PR #106, opened 2024-02-01
Install Script

sundarshankar89:
Blueprint upgrade
Install script

PR #128, opened 2024-02-08
Add serverless endpoint for validation 

nsenno-dbr:
Currently the CLI uses Databricks connect to run validation queries against an interactive Databricks cluster. This PR adds functionality to leverage a Databricks serverless SQL warehouse. Because serverless warehouses spin up almost instantly this will reduce the latency for the initial validation of SQL queries. 

TODO: Need to finish writing tests


nsenno-dbr:
I think we need to rethink how the `Validate` class is instantiated. Right now it takes some configs at the beginning, then it takes some when it runs `validate_format_result`. 

We might just want to pass in `MorphConfig` when the class is created and load up the class attributes then.

sundarshankar89:
Validate object can be instantiated at top level and passed along as part of Morph Config based on skipValidation value.

sundarshankar89 (changes requested):

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -131,7 +132,12 @@ def morph(config: MorphConfig):
     status = []
     result = MorphStatus([], 0, 0, 0, [])
 
-    ## TODO we might want to put the catalog and schema check here 
+    catalog_name = config.catalog_name
+    schema_name = config.schema_name
+
+    if catalog_name in (None, "transpiler_test") and schema_name in (None, "convertor_test"):
```
sundarshankar89, line 138: we are pushing catalog creation as part of install script, check PR #106

nsenno-dbr: @sundarshankar89 should I take this piece out then?

src/databricks/labs/remorph/config.py
```diff
@@ -11,14 +11,15 @@ class MorphConfig:
     source: str
     input_sql: str
     output_folder: str | None
-    sdk_config: Config | None
+    sdk_client: WorkspaceClient | None
     skip_validation: bool = False
+    serverless_warehouse_id: str = None
```
nfx, line 16: if we use `DATABRICKS_WAREHOUSE_ID` env variable, then it can be in the `sdk_config` automatically

labs.yml
```diff
@@ -21,6 +21,9 @@ commands:
       - name: skip-validation
         default: true
         description: Validate Transpiled Code, default True validation skipped, False validate
```
nfx, line 23: add `warehouse_types: ["SERVERLESS", "PRO"]` in the `installer` section to get warehouse prompt. see:
- https://github.com/databricks/cli/blob/36241ee55e9708fe5519c330e102174031016aea/cmd/labs/project/entrypoint.go#L49-L54
- https://github.com/databricks/cli/blob/36241ee55e9708fe5519c330e102174031016aea/cmd/labs/project/entrypoint.go#L242-L244
- https://github.com/databricks/cli/blob/36241ee55e9708fe5519c330e102174031016aea/cmd/labs/project/installer.go#L35

src/databricks/labs/remorph/config.py
nfx: i am explicitly blocking the addition of a new variable for what we already have equivalent for

src/databricks/labs/remorph/transpiler/execute.py
sundarshankar89: We will complete the install script by today, so you can remove this for now. cc: @vijaypavann-db

sundarshankar89:
covered by PR #176

PR #129, opened 2024-02-09
Fixes Merge "INTO" for delete from syntax

### LINKED ISSUE #118: Missing "INTO" for "MERGE INTO" translation
Given:
DELETE FROM
        table1
USING
        table2
WHERE
        table1.id = table2.id

We get the following error:
[PARSE_SYNTAX_ERROR] Syntax error at or near 'table1': missing 'INTO'.(line 1, pos 6)

MERGE table1 USING table2
ON
  table1.id = table2.id WHEN MATCHED THEN DELETE

Seems like the fix would just to be to add "INTO" for the "MERGE" statement.

sundarshankar89:
Closes #118


ganeshdogiparthi-db (approved):
LGTM

PR #135, opened 2024-02-12
Patch Logger function call during wheel file

### LINKED ISSUE #127: Logger shouldn't error out remorph if project_root isn't found
I tested out importing the snowflake and databricks dialect in a python script, but I get error'd out when I try to run my script 

`NotADirectoryError: Cannot find project root`

Based on how snowflake.py and databricks.py call get_logger() and how it looks for the project_root by looking for specifically "setup.py" or "pyproject.toml", it'll always fail since I have neither of those files. 

I was able to get a bandaid solution working by adding an empty "setup.py" file in remorph itself, but I think that remorph should just disable logging if project_root can't be found or set the current directory as the root if the two previous files can't be found.

sundarshankar89:
refactor logger to make only cli use `get_logger` from blueprint

closes #127

### DIFF
--- a/src/databricks/labs/remorph/__about__.py
+++ b/src/databricks/labs/remorph/__about__.py
@@ -1,1 +1,2 @@
+# DO NOT MODIFY THIS FILE
 __version__ = "0.1.2"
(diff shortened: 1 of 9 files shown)


vijaypavann-db (approved):
LGTM

PR #149, opened 2024-02-26
Add Pylint Checker

sundarshankar89:
Added Pylint


nfx (approved):
Pylint is always great

PR #172, opened 2024-03-11
`TO_NUMBER` without `format` edge case 

### LINKED ISSUE #166: TO_NUMBER should cast as number and use given precision and scale
This issue also applies to TRY_TO_NUMBER. 

When try_to_number or to_number do not specify a format but do specify a precision and format, it should be casted as a number (not a decimal) with the precision and scale given.

Input:
select * from mytable t where to_number(t.commission, 10, 2) is not null

Output
Error Parsing expression TO_DECIMAL(t.commission, 10, 2):
 * `format`: is required in Databricks [mandatory]
 * `precision` and `scale`: are considered as (38, 0) if not specified.

Expected:
select * from mytable t where CAST(t.commission AS NUMBER(10,2)) is not null

sundarshankar89:
Closes #166

### DIFF
--- a/src/databricks/labs/remorph/snow/databricks.py
+++ b/src/databricks/labs/remorph/snow/databricks.py
@@ -257,7 +257,7 @@ def _to_number(self, expression: local_expression.ToNumber):
         if precision:
             return f"CAST({func_expr} AS DECIMAL({precision}, {scale}))"
         return func_expr
-    elif not expression.expression and not precision:
+    if not expression.expression and not precision:
         exception_msg = f"""Error Parsing expression {expression}:
                          * `format`: is required in Databricks [mandatory]
                          * `precision` and `scale`: are considered as (38, 0) if not specified.
--- a/src/databricks/labs/remorph/snow/databricks.py
+++ b/src/databricks/labs/remorph/snow/databricks.py
@@ -33,6 +33,9 @@ VALID_DATABRICKS_TYPES = {
     "STRUCT",
 }
 
+PRECISION_CONST = 38
+SCALE_CONST = 0
+
 
 def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
     return self.func("DATE_TRUNC", exp.Literal.string(expression.text("unit").upper()), self.sql(expression.this))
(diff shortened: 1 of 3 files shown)


src/databricks/labs/remorph/snow/databricks.py
```diff
@@ -245,27 +245,27 @@ def _parse_json(self, expr: exp.ParseJSON):
...
+            return f"CAST({func_expr} AS DECIMAL({precision}, {scale}))"
+        return func_expr
+    if not expression.expression and not precision:
         exception_msg = f"""Error Parsing expression {expression}:
                          * `format`: is required in Databricks [mandatory]
                          * `precision` and `scale`: are considered as (38, 0) if not specified.
                       """
         raise UnsupportedError(exception_msg)
 
+    precision = 38 if not precision else precision
```
nfx, line 267: nit: better put into constant rather than a magic number

src/databricks/labs/remorph/snow/snowflake.py
```diff
@@ -197,8 +197,12 @@ def _parse_tonumber(args: list) -> local_expression.ToNumber:
 
     match len(args):
         case 1:
+            logger.warning("Precision and Scale are not specified, assuming defaults `precision`[38] and `scale`[0]")
+            logger.warning("No Format is specified, it will be inferred as simpled cast as decimal")
```
nfx, line 201:
```suggestion
            logger.warning("Precision and Scale are not specified, assuming defaults `precision`[38] and `scale`[0]. No Format is specified, it will be inferred as simpled cast as decimal")
```
don't log two log events one after another

tests/unit/snow/test_databricks.py
```diff
@@ -2980,29 +2980,29 @@ def test_to_number(dialect_context):
     validate_source_transpile, _ = dialect_context
     # Test case to validate `TO_DECIMAL` parsing with format
     validate_source_transpile(
-        """SELECT CAST(TO_NUMBER('$345', '$999.00') AS DECIMAL(38, 0)) AS col1""",
+        """SELECT TO_NUMBER('$345', '$999.00') AS col1""",
```
nfx, line 2983: practical point: you don't remove existing unit tests, you keep adding new ones.

nfx (changes requested):

src/databricks/labs/remorph/snow/snowflake.py
sundarshankar89: Done

src/databricks/labs/remorph/snow/databricks.py
sundarshankar89: Done

nfx (approved):
Lgtm

PR #184, opened 2024-03-20
Use `sqlglot` v23.0.1

sundarshankar89:


tests/unit/snow/test_databricks.py
```diff
@@ -583,14 +583,14 @@ def test_monthname(dialect_context):
     )
     # Test case to validate `monthname` parsing of date column
     validate_source_transpile(
-        """SELECT DATE_FORMAT(TO_DATE('2015-05-01'), 'MMM') AS MONTH""",
+        """SELECT DATE_FORMAT(cast('2015-05-01' as date), 'MMM') AS MONTH""",
```
nfx, line 586: nit: i wonder if we can have UPPERCASE for all SQL keywords/functions as the convention

nfx (approved):
lgtm

tests/unit/snow/test_databricks.py
sundarshankar89: conversation converts them to a single case, our validation comparison ignores them and compares

nfx (approved):
lgtm

PR #185, opened 2024-03-20
Added Databricks Source Adapter

ganeshdogiparthi-db:
Added databricks code to extract data and schema from Hive table or UC table

### DIFF
--- a/src/databricks/labs/remorph/reconcile/connectors/databricks.py
+++ b/src/databricks/labs/remorph/reconcile/connectors/databricks.py
@@ -1,0 +1,16 @@
+from pyspark.sql import DataFrame
+
+from databricks.labs.remorph.reconcile.connectors.data_source import DataSource
+from databricks.labs.remorph.reconcile.recon_config import Schema, Tables
+
+
+class DatabricksDataSource(DataSource):
+    def read_data(self, schema_name: str, catalog_name: str, query: str, table_conf: Tables) -> DataFrame:
+        # Implement Databricks-specific logic here
+        return NotImplemented
+
+    def get_schema(self, table_name: str, schema_name: str, catalog_name: str) -> list[Schema]:
+        # Implement Databricks-specific logic here
+        return NotImplemented
+
+    databricks_datatype_mapper = {}
--- a/src/databricks/labs/remorph/reconcile/connectors/databricks.py
+++ b/src/databricks/labs/remorph/reconcile/connectors/databricks.py
@@ -3,7 +3,7 @@ from pyspark.sql import DataFrame
 from pyspark.sql.functions import col
 
 from databricks.labs.remorph.reconcile.connectors.data_source import DataSource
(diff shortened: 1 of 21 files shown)


src/databricks/labs/remorph/reconcile/connectors/databricks.py
```diff
@@ -1,16 +1,33 @@
...
+            raise PySparkException(error_msg) from e
 
     def get_schema(self, table_name: str, schema_name: str, catalog_name: str) -> list[Schema]:
-        # Implement Databricks-specific logic here
-        return NotImplemented
+        try:
+            full_table_name = (
+                f"{catalog_name}.{schema_name}.{table_name}" if catalog_name else f"{schema_name}.{table_name}"
+            )
+            schema_df = self.spark.sql(f"describe table {full_table_name}").where("col_name not like '#%'").distinct()
```
sundarshankar89, line 24: we should use system tables, that we will enforce usage on UC

sundarshankar89 (changes requested):

src/databricks/labs/remorph/reconcile/connectors/databricks.py
ganeshdogiparthi-db: Yes, Using information schema for all the sources except databricks.

tests/unit/reconcile/connectors/test_databricks.py
```diff
@@ -0,0 +1,112 @@
...
+
+from databricks.labs.remorph.reconcile.connectors.databricks import DatabricksDataSource
+
+
+def initial_setup():
+    pyspark_sql_session = MagicMock()
+    spark = pyspark_sql_session.SparkSession.builder.getOrCreate()
+
+    # Define the source, workspace, and scope
+    source = "snowflake"
```
nfx, line 16: Should it be different here?

nfx (changes requested):
One copy-paste typo

tests/unit/reconcile/connectors/test_databricks.py
sundarshankar89: @ganeshdogiparthi-db: I think we need to change this variable name, In Datasource abstract class this purely indicates database engine and we use the engine name as prefix for getting secrets key.

As per our discussion on supporting HMS and UC within databricks source adapter we will use catalog_name and not source variable to differentiate.

ganeshdogiparthi-db: It is as expected. DatabricksDataSource will be intialized even for the source "snowflake" (because our target layer will always be databricks).  Also if you observe line 39 and 64 I have passed "databricks" as source and tested the output. (Old code)

Now up on a suggestion from Sundar. I have renamed the "source" param to the "engine". Now it has more context. I made appropriate changes to the code and test cases.

nfx (approved):
Lgtm

PR #206, opened 2024-04-01
Query builder is refactored to package model

ganeshdogiparthi-db:
1. Removed the query config up on a suggestion from Sundar
2. Moved most of the functions moved to data class or base query builder.
3. All other query builders extend the base query builder.


src/databricks/labs/remorph/reconcile/query_builder/hash.py
```diff
@@ -0,0 +1,58 @@
+from io import StringIO
```
nfx, line 1: This filename overlaps with hash() builtin function. Rename to avoid problems

nfx (changes requested):
Mostly good, but rename the hash file not to collide and fix formatting

src/databricks/labs/remorph/reconcile/query_builder/hash.py
ganeshdogiparthi-db: Done. Changed the file name to hash_query and test case respectively. I fixed most of the fmt issues. But still giving some weird fmt issues.

ganeshdogiparthi-db: Made structural changes up on suggestion by Sundar

nfx (approved):

sundarshankar89 (approved):
LGTM

PR #221, opened 2024-04-08
Added Data Comparison Logic

ravit-db:


src/databricks/labs/remorph/reconcile/recon_config.py
```diff
@@ -51,18 +53,18 @@ class Thresholds:
...
+    source: str = '1=1'
+    target: str = '1=1'
 
 
 @dataclass
 class Table:
     source_name: str
     target_name: str
-    join_columns: list[str] | None = None
+    joins: list[str] | None = None
```
sundarshankar89, line 64: why these names have been changed?

src/databricks/labs/remorph/reconcile/query_builder/sampling_query.py
```diff
@@ -0,0 +1,57 @@
...
+        join_condition = " and ".join(join_expr)
+        join = f" select src.* from src join recon on {join_condition}"
+
+        select_query = with_cte + join
+        return select_query
+
+    @staticmethod
+    def _get_with_clause(df: DataFrame, source: str):
+        select_clause = []
+        for row in df.take(50):
```
vil1, line 45: `50` looks like a magic number here. Also, `sample` might be better suited than `take`. 

On a more general standpoint, my gut feeling is that this function's output would be easier to format as:
`values ( x0, y0, ...), (x1, y1, ...), (xn, yn, ...) as <some alias> (column0, column1, ...)`. It would also probably avoid the need for a special case for the Oracle dialect.

vil1 (commented):
I have a few concerns about this PR. As I am not a Python developer, I may have missed some obvious warts though.

src/databricks/labs/remorph/reconcile/query_builder/sampling_query.py
```diff
@@ -0,0 +1,57 @@
...
+    @staticmethod
+    def _construct_query(
+        table: str, query_filter: str, col_expr: list[str], with_clause: str, key_cols: list[str]
+    ) -> str:
+
+        with_cte = f"{with_clause} src as ( select " + ",".join(col_expr) + f" from {table} where {query_filter} )"
+
+        join_expr = [f"recon.{col} = src.{col}" for col in key_cols]
+        join_condition = " and ".join(join_expr)
+        join = f" select src.* from src join recon on {join_condition}"
```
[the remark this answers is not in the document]
sundarshankar89: +1 TableRecon dataclass should introduce a __post_init__ method to validate mandatory attributes

src/databricks/labs/remorph/reconcile/query_builder/sampling_query.py
sundarshankar89: expressed as sqlglot expressions will as we can generate SQL in any target dialect

.github/scripts/setup_spark_remote.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -xv
+
+mkdir -p $HOME/spark
+cd $HOME/spark
+
+hatch shell
+
+version=3.5.1
```
sundarshankar89, line 10: make this version to use latest always

src/databricks/labs/remorph/reconcile/query_builder/base.py
```diff
@@ -54,10 +54,10 @@ def transform_dict(self):
 
     @property
     def select_columns(self) -> set[str]:
-        if self._table_conf.select_columns is None:
+        if self._table_conf.selects is None:
```
sundarshankar89, line 57: why is the name still changed here?

src/databricks/labs/remorph/reconcile/reconciler/comparator.py
```diff
@@ -0,0 +1,33 @@
...
+from databricks.labs.remorph.reconcile.recon_config import ReconcileOutput
+
+
+class Comparator:
+
+    def compare_data(self, source: DataFrame, target: DataFrame, join_cols: list[str], report: str) -> ReconcileOutput:
+        join = join_cols if report == "all" else "hash_value__recon"
+        df = source.alias('s').join(target.alias("t"), join, how='full')
+
+        mismatch = self._get_mismatch_data(df) if report == 'all' else None
```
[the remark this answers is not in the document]
sundarshankar89: I don't completely understand, what happens with `report = data` ?

.github/scripts/setup_spark_remote.sh
ravit-db: couldn't find latest in the apache distribution.only stable release versions are available.

src/databricks/labs/remorph/reconcile/query_builder/base.py
ravit-db: reverted the changes.

src/databricks/labs/remorph/reconcile/query_builder/sampling_query.py
ravit-db: > my gut feeling is that this funct

For sample ,we always want to get minimal sample records in smaller, medium and large volume tables say for now 50 records.If we use sample we need to keep fraction as dynamic and we need total count which will add few more extra work to worker nodes.
Table Values constructor fails in oracle with invalid table name.

src/databricks/labs/remorph/reconcile/query_builder/sampling_query.py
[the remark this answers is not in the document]
ravit-db: Added an exception

src/databricks/labs/remorph/reconcile/recon_config.py
ravit-db: reverted the names

src/databricks/labs/remorph/reconcile/query_builder/sampling_query.py
vil1: Also, `50` is still a magic number, please define it as a named constant

src/databricks/labs/remorph/reconcile/reconciler/comparator.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+        return ReconcileOutput(missing_in_src=missing_in_src, missing_in_tgt=missing_in_tgt, mismatch=mismatch)
+
+    @staticmethod
+    def _get_mismatch_data(df: DataFrame) -> DataFrame:
+        return (
+            df.filter((col("s.hash_value__recon").isNotNull()) & (col("t.hash_value__recon").isNotNull()))
+            .withColumn(
+                "hash_match",
+                when(col("s.hash_value__recon") == col("t.hash_value__recon"), lit(True)).otherwise(lit(False)),
```
vil1, line 33:
```suggestion
                col("s.hash_value__recon") == col("t.hash_value__recon")
```

sundarshankar89: @vil1 Both Threshold and Data Comparison PR needs to be re-written using sqlglot expressions, ignore till the new changes are in place.

vil1: Gotcha, although I'm not sure what it has to do with the above suggestion (that involves only the `Dataframe` API, as far as I understand)

sundarshankar89: I meant to save your time, this PR is going to significant changes, much easier for you to review once those changes are in place.

vil1: Makes sense. So I'll wait until you notify me that this PR is ready for my review. Cheers!

PR #230, opened 2024-04-09
Added grammars and intermediate representation

nfx:


Makefile
```diff
@@ -18,7 +18,7 @@ fmt-python:
 	hatch run fmt
 
 fmt-scala:
-	mvn -f parent/pom.xml scalafmt:format
+	mvn -f pom.xml scalafmt:format
```
sundarshankar89, line 21: Same fix need in test-scala

sundarshankar89 (commented):
Can we also bring in postgressql as well, 
https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/PostgreSQLLexer.g4

PR #231, opened 2024-04-10
Run `scalafmt` as part of `mvn validate`, ...

vil1:
add it as part of the `test…-core` action


sundarshankar89:
@nfx I will let you handle the merge for this whenever necessary.

PR #234, opened 2024-04-12
Feat/snow join to ir

vil1:
Add support for (some) JOIN types

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
@@ -62,8 +62,30 @@ class SnowflakeRelationBuilder extends SnowflakeParserBaseVisitor[ir.Relation] {
   }
 
   override def visitTable_source_item_joined(ctx: Table_source_item_joinedContext): ir.Relation = {
+
+    def translateJoinType(joinType: SnowflakeParser.Join_typeContext): ir.JoinType = {
+      if (joinType == null || joinType.outer_join() == null) {
+        ir.InnerJoin
+      } else if (joinType.outer_join().LEFT() != null) {
+        ir.LeftOuterJoin
+      } else if (joinType.outer_join().RIGHT() != null) {
+        ir.RightOuterJoin
+      } else if (joinType.outer_join().FULL() != null) {
+        ir.FullOuterJoin
+      } else {
+        ir.UnspecifiedJoin
+      }
+    }
+
     def buildJoin(left: ir.Relation, right: SnowflakeParser.Join_clauseContext): ir.Join = {
-      ir.Join(left, right.object_ref().accept(this), None, ir.InnerJoin, Seq(), ir.JoinDataType(false, false))
+
+      ir.Join(
+        left,
+        right.object_ref().accept(this),
+        None,
+        translateJoinType(right.join_type()),
+        Seq(),
+        ir.JoinDataType(is_left_struct = false, is_right_struct = false))
     }
     val left = ctx.object_ref().accept(this)
     ctx.join_clause().asScala.foldLeft(left)(buildJoin)
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
@@ -1,6 +1,5 @@
 package com.databricks.labs.remorph.parsers.snowflake
 
-import com.databricks.labs.remorph.parsers.intermediate.{Alias, Column, Expression, NamedTable, Project, Relation, TreeNode}
 import com.databricks.labs.remorph.parsers.snowflake.SnowflakeParser._
 import com.databricks.labs.remorph.parsers.{intermediate => ir}
 
@@ -10,10 +9,10 @@ import scala.collection.JavaConverters._
  * @see
  *   org.apache.spark.sql.catalyst.parser.AstBuilder
  */
-class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[TreeNode] {
+class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[ir.TreeNode] {
 
   // TODO investigate why this is needed
-  override protected def aggregateResult(aggregate: TreeNode, nextResult: TreeNode): TreeNode = {
+  override protected def aggregateResult(aggregate: ir.TreeNode, nextResult: ir.TreeNode): ir.TreeNode = {
     if (nextResult == null) {
       aggregate
     } else {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -56,31 +55,63 @@ class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[TreeNode] {
...
     val tableName = ctx.object_name().id_(0).getText
-    NamedTable(tableName, Map.empty, is_streaming = false)
+    ir.NamedTable(tableName, Map.empty, is_streaming = false)
   }
+
+  override def visitTable_source_item_joined(ctx: Table_source_item_joinedContext): ir.Relation = {
+
+    def translateJoinType(joinType: SnowflakeParser.Join_typeContext): ir.JoinType = {
+      if (joinType == null || joinType.outer_join() == null) {
+        ir.InnerJoin
```
sundarshankar89, line 68: can we switch to pattern matching, much easier to manage and read.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -56,31 +55,63 @@ class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[TreeNode] {
...
-class SnowflakeRelationBuilder extends SnowflakeParserBaseVisitor[Relation] {
-  override def visitObject_ref(ctx: SnowflakeParser.Object_refContext): Relation = {
+class SnowflakeRelationBuilder extends SnowflakeParserBaseVisitor[ir.Relation] {
+  override def visitObject_ref(ctx: SnowflakeParser.Object_refContext): ir.Relation = {
     val tableName = ctx.object_name().id_(0).getText
-    NamedTable(tableName, Map.empty, is_streaming = false)
+    ir.NamedTable(tableName, Map.empty, is_streaming = false)
   }
+
+  override def visitTable_source_item_joined(ctx: Table_source_item_joinedContext): ir.Relation = {
```
sundarshankar89, line 64: maybe having a from_visitor will help navigate multiple type of expressions. 
It can be a subquery, join, cte etc. Thoughts?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
vil1: I'm afraid we can't (although I totally agree it would be better expressed with pattern-matching).

The (Java) classes generated from the grammar (such as `Join_typeContext` here) do not come with a companion object providing an `unapply` method (which what Scala's pattern-matching relies upon). They aren't even data classes (they have no field, only methods). So I don't see any way around writing this kind of (arguably ugly) if-else-chains-involving-null-checks.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
sundarshankar89: I was saying something similar 
https://github.com/apache/spark/blob/daf260f74e12fc5e9fad6091f6230e71a9e6c9c1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala#L1046

sundarshankar89: I haven't yet digested the whole setup yet. so I might be not thinking this through.

vil1: Lemme give you some more context then. We are building `SnowflakeAstBuilder` as a subclass of `SnowflakeParserBaseVisitor` that ANTLR generates from the `SnowflakeParser.g4` grammar. It comes with a swath of `visit*` methods (such as the `visitTable_source_item_joined` above), basically one method for each rule in the grammar (here, the `table_source_item_joined` rule).  
So our job is to implement the `SnowflakeAstBuilder` visitor, whose structure is fully determined by the structure of the underlying grammar, in such a way that it outputs an AST that is similar to the one produced by catalyst's `AstBuilder` you mentioned above. 
So we have no choice but to cope with the fact that names and structures imposed by the Snowflake grammar (and reflected in the `SnowflakeApiBuilder`'s API) may sometimes be very different to the ones of our target AST, forcing us to produce code that isn't optimally readable/maintainable.
That being said, we should always keep that fact in mind and try to minimize the effects of this "structural discrepancy" as much as possible.
I hope this helps answering your question.

vil1:
I cannot come up with an example query for covering the 
```scala
else {
  ir.UnspecifiedJoin
}
``` 
lines codecov is (legitimately) complaining about. To the extent of my (limited) current knowledge, this branch should never get evaluated (but it has to be there for the code to compile). Should I throw an exception instead of returning `ir.UnspecifiedJoin`? Would that be enough for codecov to ignore the fact that this line isn't covered by any test? If not, what should I do?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
vil1: Another thing to keep in mind is that for the moment, we focus on (quoting @nfx) "quantity over quality". Meaning that we accept to cut some corners in order to implement as much of the Snowflake grammar as quick as possible (although some "corners" remain "uncutable", such as maintaining good test coverage, etc).
So even though reviewers' comments about code quality are (as always) very much appreciated, please be aware that addressing those concerns might get temporarily postponed for the sake of "moving fast".

nfx: @sundarshankar89 we need to strategically cut corners in the following two weeks and gather a backlog of what do we need to do to improve snowflake grammar. And then GDC can come in :)

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -56,31 +55,63 @@ class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[TreeNode] {
...
+    }
+    val left = ctx.object_ref().accept(this)
+    ctx.join_clause().asScala.foldLeft(left)(buildJoin)
+  }
+
 }
-class SnowflakeExpressionBuilder extends SnowflakeParserBaseVisitor[Expression] {
 
-  override def visitSelect_list_elem(ctx: SnowflakeParser.Select_list_elemContext): Expression = {
+class SnowflakeExpressionBuilder extends SnowflakeParserBaseVisitor[ir.Expression] {
```
nfx, line 96: @vil1 for the next PRs - split the `SnowflakeExpressionBuilder` into a separate file

nfx (approved):

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
vijaypavann-db: Does it make sense to define a function like below, and use it in the if-else conditions: 

```
def isInnerJoin(joinType): bool 
  joinType == null || joinType.outer_join() == null

```  

```
  if (isInnerJoin(joinType)):  ir.InnerJoin
  else if(isLeftJoin(joinType)): ir.LeftJoin
   ......
    else: ..... 
```

PR #235, opened 2024-04-12
Add core contribution guidelines

vil1:


sundarshankar89 (approved):

sundarshankar89:
@nfx you how are you managing unsigned commits? I don't have access to force merge.

vil1:
I cannot come up with an example query for covering the 
```scala
else {
  ir.UnspecifiedJoin
}
``` 
lines codecov is (legitimately) complaining about. To the extent of my (limited) current knowledge, this branch should never get evaluated (but it has to be there for the code to compile). Should I throw an exception instead of returning `ir.UnspecifiedJoin`? Would that be enough for codecov to ignore the fact that this line isn't covered by any test? If not, what should I do?

nfx:
@sundarshankar89 all commits are squashed once they're merged to `main` and signed, so we should be fine.

nfx (approved):
lgtm

nfx:
@sundarshankar89 
image

PR #236, opened 2024-04-12
Feat/snow where to ir

vil1:
WARNING: this branch is based on changes in #234, please wait for #234 to be merged before merging this PR

PR #237, opened 2024-04-15
Feat/snow group to ir

vil1:
Please merge after #236


vil1:
Change reformatted in #244

PR #242, opened 2024-04-15
Feat/snow pagination to ir

vil1:
to be merged after #240


vil1:
Change reformatted in #246

PR #243, opened 2024-04-16
Feat/snow pivot to ir

vil1:


vil1:
Change reformatted in #247

PR #255, opened 2024-04-17
Implement translation of Snowflake HAVING clauses

vil1:
The grammar states that `having_clause` can occur within a `group_by_clause` or straight into a `select_optional_list` (in which case, the `group_by_clause` is absent. 

That's the reason why I had to sorta duplicate the call to `buildHaving` here.

However, the Snowflake docs seems to mandate the `HAVING` clause to appear after a `GROUP BY`, so I am not sure how the latter case described above may actually occur...


nfx (approved):
lgtm

PR #256, opened 2024-04-17
Improve testing of Snowflake to IR translation

vil1:
Introduces a specific Spec for `SnowflakeExpressionBuilder` and `SnowflakeRelationBuilder`.

While adding tests in `SnowflakeExpressionBuilderSpec`, I realized that `SnowflakePredicateBuilder` and the `Predicate` trait altogether will probably have to go away:

We want `TRUE` and `FALSE` to be valid predicates. However, they are literals, which get translated to `Literal` which extends `Expression` and not `Predicate` (because other kinds of literals, such as numbers or strings cannot be considered predicates). 

In a subsequent PR, I will therefore move code in `SnowflakePredicateBuilder` inside `SnowflakeExpressionBuilder`

### DIFF
--- a/core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/ParserTestCommon.scala
+++ b/core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/ParserTestCommon.scala
@@ -1,0 +1,32 @@
+package com.databricks.labs.remorph.parsers.snowflake
+
+import com.databricks.labs.remorph.parsers.intermediate.TreeNode
+import org.antlr.v4.runtime.{CharStreams, CommonTokenStream, RuleContext}
+import org.scalatest.{Assertion, Assertions}
+
+trait ParserTestCommon { self: Assertions =>
+
+  def astBuilder: SnowflakeParserBaseVisitor[_]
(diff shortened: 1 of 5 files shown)


core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/ParserTestCommon.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  def astBuilder: SnowflakeParserBaseVisitor[_]
+  protected def parseString[R <: RuleContext](input: String, rule: SnowflakeParser => R): R = {
+    val inputString = CharStreams.fromString(input)
+    val lexer = new SnowflakeLexer(inputString)
+    val tokenStream = new CommonTokenStream(lexer)
+    val parser = new SnowflakeParser(tokenStream)
+    val tree = rule(parser)
+    // uncomment the following line if you need a peek in the Snowflake AST
+    // println(tree.toStringTree(parser))
```
nfx, line 17: not blocking, but add "if in debug from IntelliJ" check

nfx (approved):
lgtm

PR #257, opened 2024-04-18
Move SnowflakePredicateBuilder's code in SnowflakeExpressionBuilder

vil1:
And make `Predicate` a simple marker trait (which doesn't extend `Expression` anymore). This is because some expressions such as literals `True` and `False` cannot be made predicates (although they logically are)


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
```diff
@@ -84,4 +84,52 @@ class SnowflakeExpressionBuilder extends SnowflakeParserBaseVisitor[ir.Expressio
     case d if d.isDecimalDouble || d.isExactDouble => ir.Literal(double = Some(d.toDouble))
     case _ => ir.Literal(decimal = Some(ir.Decimal(decimal, None, None)))
   }
+
+  override def visitExpr(ctx: ExprContext): ir.Expression = {
+
+    def buildComparison(left: ir.Expression, right: ir.Expression, op: Comparison_operatorContext): ir.Expression = {
```
nfx, line 90: make this a private method. it's fine to have inner methods if they're just one-two lines

nfx (changes requested):

nfx (approved):
lgtm

Issue #258, opened 2024-04-19
[FEATURE]: Improve `ranking_windowed_function` rule in Snowflake grammar


vil1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

The rules for window functions, defined in `ranking_windowed_function` and its children is lacking:
- support for null ordering (`NULLS FIRST` / `NULLS LAST`) in its `ORDER BY` clause
- support for defining window frames (eg. `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`) in the `OVER` clause.

### Proposed Solution

There is some commented code in `SnowflakeParser.g4` that seems related to the missing bits described above

jimidle:
I will take this one.

jimidle:
I think it is best to do this by getting rid of all the myriad specifically specified functions in the grammar that are just standard function calls. However, that might cause quite a lot of changes in the Snowflake grammar and therefore the builders. I am not sure what the priority of Snowflake is vs TSQL? Equal? Lower? Higher?

PR #259, opened 2024-04-19
Add (partial) support for Snowflake ranking window functions

vil1:
Some bits are missing in the grammar, see #258

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
@@ -145,9 +145,13 @@ class SnowflakeExpressionBuilder extends SnowflakeParserBaseVisitor[ir.Expressio
 
   override def visitRanking_windowed_function(ctx: Ranking_windowed_functionContext): ir.Expression = {
     val windowFunction = buildWindowFunction(ctx)
-    val partitionSpec = buildPartitionSpec(ctx.over_clause())
-    val sortOrder = buildSortOrder(ctx.over_clause())
+    val overClause = Option(ctx.over_clause())
+    val partitionSpec =
+      overClause.flatMap(o => Option(o.partition_by())).map(buildPartitionSpec).getOrElse(Seq())
+    val sortOrder =
+      overClause.flatMap(o => Option(o.order_by_expr())).map(buildSortOrder).getOrElse(Seq())
     // dummy implementation because the grammar for this is missing
+    // see https://github.com/databrickslabs/remorph/issues/258
     val frameSpec = ir.WindowFrame(
       frame_type = ir.RowsFrame,
       lower = ir.FrameBoundary(current_row = false, unbounded = true, value = ir.Noop),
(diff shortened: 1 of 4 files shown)


nfx (commented):
approving to ublock, but add more unit tests and comment why + issue link

```
val commas = exprList.COMMA().asScala.map(_.getSymbol.getStopIndex) :+ exprList.getStop.getStopIndex
```

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
```diff
@@ -142,4 +143,56 @@ class SnowflakeExpressionBuilder extends SnowflakeParserBaseVisitor[ir.Expressio
...
+      upper = ir.FrameBoundary(current_row = true, unbounded = false, value = ir.Noop))
+    ir.Window(
+      window_function = windowFunction,
+      partition_spec = partitionSpec,
+      sort_order = sortOrder,
+      frame_spec = frameSpec)
+  }
+
+  private def buildWindowFunction(ctx: Ranking_windowed_functionContext): ir.Expression = {
+    if (ctx.ROW_NUMBER() != null) {
```
nfx, line 163: what do you this about this case as a special case for: https://github.com/databricks/scala-style-guide?tab=readme-ov-file#pattern-matching

image

nfx: btw, found out a statement on early returns - not relevant here, but relevant in other palces:
https://github.com/databricks/scala-style-guide?tab=readme-ov-file#return-statements

image

vil1: I'm okay with considering if/else if/else chains as pattern-matches and not put curly braces around  going forward.

In fact, as a matter of taste, I prefer to avoid curly braces whenever possible, as an incentive to write most my functions as single expressions. But guidelines are meant precisely to avoid arguing on individual preferences, so...

nfx: > In fact, as a matter of taste, I prefer to avoid curly braces whenever possible, as an incentive to write most my functions as single expressions. But guidelines are meant precisely to avoid arguing on individual preferences, so...

apparently it aligns with this agreement :)

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilderSpec.scala
```diff
@@ -30,5 +30,114 @@ class SnowflakeExpressionBuilderSpec extends AnyWordSpec with ParserTestCommon w
...
+            SortOrder(Column("b"), AscendingSortDirection, SortNullsLast),
+            SortOrder(Column("c"), DescendingSortDirection, SortNullsLast),
+            SortOrder(Column("d"), AscendingSortDirection, SortNullsLast)),
+          frame_spec = WindowFrame(
+            frame_type = RowsFrame,
+            lower = FrameBoundary(current_row = false, unbounded = true, value = Noop),
+            upper = FrameBoundary(current_row = true, unbounded = false, value = Noop))))
+    }
+
+    // see https://github.com/databrickslabs/remorph/issues/258
```
nfx, line 75: nice

nfx (approved):
lgtm

PR #261, opened 2024-04-22
[SAMPLE] BigQuery Dialect setup

vijaypavann-db:
DO NOT MERGE THIS PR


src/databricks/labs/remorph/bigquery/bigquery.py
```diff
@@ -0,0 +1,18 @@
+import logging
+from typing import ClassVar
+
+from sqlglot.dialects.bigquery import BigQuery
+
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=function-redefined
+class BigQuery(BigQuery):
```
sundarshankar89, line 10: @vijaypavann-db `config.py` also needs modification to use this class

sundarshankar89 (changes requested):

src/databricks/labs/remorph/bigquery/bigquery.py
vijaypavann-db: Done, Updated `config.py`

nfx, line 10:
```suggestion
from sqlglot.dialects import bigquery as orig

logger = logging.getLogger(__name__)


class BigQuery(orig.BigQuery):
```

nfx: don't disable linters

nfx (changes requested):

src/databricks/labs/remorph/bigquery/bigquery.py
vijaypavann-db: Done

PR #262, opened 2024-04-22
Implement translation of Snowflake WITH clauses 

vil1:


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeRelationBuilder.scala
```diff
@@ -7,6 +7,25 @@ import scala.collection.JavaConverters._
 
 class SnowflakeRelationBuilder extends SnowflakeParserBaseVisitor[ir.Relation] {
 
+  override def visitSelect_statement(ctx: Select_statementContext): ir.Relation = {
```
vil1, line 10: Moved this method here from `SnowflakeAstBuilder` (as I need it to return a `Relation` instead of a mere `TreeNode`).

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -20,23 +20,18 @@ class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[ir.TreeNode] {
...
+    buildCTE(ctx.with_expression(), select)
+  }
+
+  private def buildCTE(ctx: With_expressionContext, relation: ir.Relation): ir.Relation = {
+    if (ctx == null) {
+      relation
+    } else {
+      val ctes = ctx.common_table_expression().asScala.map(_.accept(new SnowflakeRelationBuilder))
+      ir.WithCTE(ctes, relation)
+    }
```
nfx, line 34:
```suggestion
    if (ctx == null) {
      return relation
    }
    val ctes = ctx.common_table_expression().asScala.map(_.accept(new SnowflakeRelationBuilder))
    ir.WithCTE(ctes, relation)
```
Reduce nesting :)

nfx (approved):

nfx (approved):
Lgtm

PR #263, opened 2024-04-22
Implement translation of Snowflake QUALIFY clause

vil1:
Along with a few code improvements ;).

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
@@ -6,40 +6,22 @@ import com.databricks.labs.remorph.parsers.snowflake.SnowflakeParser._
 import scala.collection.JavaConverters._
 class SnowflakeExpressionBuilder extends SnowflakeParserBaseVisitor[ir.Expression] {
 
-  override def visitSelect_list_elem(ctx: SnowflakeParser.Select_list_elemContext): ir.Expression = {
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
```diff
@@ -6,40 +6,22 @@ import com.databricks.labs.remorph.parsers.snowflake.SnowflakeParser._
...
-      }
-    } else if (ctx.expression_elem() != null) {
-      ctx.expression_elem().accept(this)
-    } else {
-      null
+  override def visitSelect_list_elem(ctx: Select_list_elemContext): ir.Expression = {
+    val rawExpression = ctx match {
+      case c if c.column_elem() != null => c.column_elem().accept(this)
+      case c if c.expression_elem() != null => c.expression_elem().accept(this)
+      case _ => null
```
nfx, line 13: Bug: "input: ir.Expression" becomes null, which it shouldn't. I guess you have to use Option here

nfx (changes requested):
One small bug

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
vil1: It's actually a not-yet-implemented case, let me just implement it properly. Stay tuned

nfx (approved):
thank you!

PR #270, opened 2024-04-24
Isort import issues and ruff warnings fixes

vijaypavann-db:
This PR aims to solve the following `isort` issues and `ruff` warnings:

### isort

* wrong-import-order : 

      *  C0411: third party import "pyspark.sql.SparkSession" should be placed before first party import "databricks.sdk.WorkspaceClient"  (wrong-import-order)
      *  C0411: third party import "sqlglot.dialects.dialect.Dialect" should be placed before first party import "databricks.sdk.WorkspaceClient"  (wrong-import-order)
      *  C0411: third party import "pyspark.sql.DataFrame" should be placed before first party import "databricks.sdk.WorkspaceClient"  (wrong-import-order)

* ungrouped-imports : 
      *  `C0412: Imports from package databricks are not grouped (ungrouped-imports)`

-------

### ruff
*  `ruff  . --fix` : 
    *  ```warning: `ruff <path>` is deprecated. Use `ruff check <path>` instead.```

PR #271, opened 2024-04-24
Refactored TableRecon, Schema, DatabaseConfig Dataclasses to config

vijaypavann-db:
### Changes
------
1. The following classes have been refactored to `config.py`: 
    * TableRecon
    * DatabaseConfig

2. The (import) changes are updated in the corresponding files.

### DIFF
--- a/src/databricks/labs/remorph/config.py
+++ b/src/databricks/labs/remorph/config.py
@@ -72,9 +72,3 @@ class DatabaseConfig:
     target_catalog: str
     target_schema: str
     source_catalog: str | None = None
-
-
-@dataclass
-class Schema:
-    column_name: str
-    data_type: str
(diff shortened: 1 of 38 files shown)


vijaypavann-db:
Depends on #270

src/databricks/labs/remorph/config.py
```diff
@@ -51,3 +52,29 @@ def get_write_dialect(self):
...
+
+@dataclass
+class DatabaseConfig:
+    source_schema: str
+    target_catalog: str
+    target_schema: str
+    source_catalog: str | None = None
+
+
+@dataclass
```
sundarshankar89, line 77: I don't think we need schema here for now.

sundarshankar89 (changes requested):

src/databricks/labs/remorph/config.py
sundarshankar89: also fix fmt errors

sundarshankar89 (approved):
LGTM

PR #274, opened 2024-04-25
Implement translation of Snowflake EXISTS, IN, BETWEEN clauses

vil1:


nfx (approved):
Lgtm

jimidle:
THe extensions scala file was not reformatted before checkin. It is only whitespace, but probably should be reformatted via mvn test or mvn compile.

Issue #275, opened 2024-04-25
[FEATURE]: Map TSQL expressions into IR


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

To fully support TSQL remorph, all TSQL expressions must be mapped to IR.

### Proposed Solution

 - Incrementally map starting with the simplest of expressions and gradually move to more complex expressions
 - Reuse any existing transformations 
 - Change grammar where necessary to simplify resolution within the parser tree (for instance have a general rule for functions)
 - In the grammar, follow these guidelines:
   - Keep the rules simple and general - don't over-specify syntax
   - Don't try and infer semantics in the grammar - accept what looks like it is syntactically correct and verify semantics later
   - Reduce ambiguities to minimal token lookahaed. ANTLR4 will take on just about any poorly specified grammar and attempt to disambiguate at parser execution time. But that does not mean it is a good strategy as parsing will become a huge time burden. The current open source grammar is exceptionally bad at this and it will be improved
   - Do not specify things like function names as individual lexer tokens when a general identifier will do. 
   - The Lexer should raise no errors, let the parser do that
   - The parser should raise errors only when it is a genuine syntax error
   -  Implement semantic checks before transformation to IR, if necessary

PR #286, opened 2024-04-26
Creates the jump off point for creating IR from the TSQL parse tree walker

jimidle:
This is the starting point for new feature issue #275


vil1 (commented):
This looks like duplicate in respect with #267

jimidle:
> This looks like duplicate in respect with #267

Yes - I did search but that did not come up as it seemed like the start of the SELECT statement, and there was no code for TSQL in main at that point. However, this issue will cover more basic elements. We won't get very far into the SELECT statement before we need all the more primitive elements of an expression.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/expressions.scala
```diff
@@ -56,7 +56,9 @@ case class Literal(
     float: Option[Float] = None,
     decimal: Option[Decimal] = None,
     double: Option[Double] = None,
+    real: Option[String] = None,
```
nfx, line 59: Changing this is not allowed - it's generated from protobuf. Use double

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -1,20 +1,53 @@
...
   }
 
   override def visitFull_column_name(ctx: Full_column_nameContext): Expression = {
     val columnName = ctx.id_.getText
     Column(columnName)
   }
+
+  override def visitPrimitive_constant(ctx: Primitive_constantContext): Expression = {
+
+    ctx match {
```
nfx, line 32: From our scala style:

For method whose entire body is a pattern match expression, put the match on the same line as the method declaration if possible to reduce one level of indentation.

nfx: Please read this (whole) document again: https://github.com/databricks/scala-style-guide?tab=readme-ov-file#pattern-matching

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -1,20 +1,53 @@
...
+      case c if c.STRING() != null => c.STRING().accept(this)
+      case c if c.INT() != null => c.INT().accept(this)
+      case c if c.FLOAT() != null => c.FLOAT() accept (this)
+      case c if c.HEX() != null => c.HEX() accept (this)
+      case c if c.REAL() != null => c.REAL() accept (this)
+    }
+  }
+
+  override def visitTerminal(node: TerminalNode): Expression = {
+    val sym = node.getSymbol
```
nfx, line 43: Do we need this constant?..

nfx (changes requested):

nfx (approved):
Lgtm

Issue #287, opened 2024-04-26
[FEATURE]: Remove all string literals from the TSqlParser.g4 gramamr


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

The lexer and parser grammars are separate files, which means you cannot use literal text to generate tokens automatically in the parser. This is poor practice anyway as then you have to name them to use them in the visitors. It is fraught with problems.

### Proposed Solution

Move any literal definitions in the parser grammar to real tokens in the lexer, so they can be referenced easily (at all), in the tree visitors.

PR #288, opened 2024-04-26
Remove literal references from the TSQL parser grammar

jimidle:
Literals in the parser grammars are discouraged as typos cause strange issues and writing visitors becomes awkward without labeling them.

Resolves: #287


nfx (approved):
woo-hoo - cleaner grammar

PR #295, opened 2024-04-29
Implement basic arithmetic and bitwise operators

jimidle:
Implement translations for the basic arithmetic and bitwise operators such as +, - * et al.

Also implements explicit  precedence via parentheses, which needs some attention in the grammar to be done perfectly, but that is for another PR.

As part of generally processing constant expressions, a NoTable AST construct is added as not all statements require a FROM clause, the SELECT being most pertinent here.  

we also clean up some of the visitor code such that unnecessary overrides are removed. The ANTLR4 visitor already calls accept() automatically so there is no need to implement an override to do exactly the same.


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilder.scala
```diff
@@ -29,10 +29,11 @@ class TSqlAstBuilder extends TSqlParserBaseVisitor[ir.TreeNode] {
...
       .map(_.accept(new TSqlExpressionBuilder))
-    val fromTable = rawExpression.table_sources().accept(new TSqlRelationBuilder)
-
-    ir.Project(fromTable, columnExpression)
-
+    if (rawExpression.table_sources() == null) {
+      // Not all SELECT statements require a table relation
+      return ir.Project(ir.NoTable(), columnExpression)
+    }
+    ir.Project(rawExpression.table_sources().accept(new TSqlRelationBuilder), columnExpression)
```
vil1, line 36:
```suggestion
   val  tableSources = Option(rawExpression.table_sources()).fold(ir.NoTable())(_.accept(new TSqlRelationBuilder))
   ir.Project(tableSources, columnExpression)
```

Merely cosmetic, but avoids some repetition.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -1,31 +1,39 @@
...
+    Column(ctx.id_.getText)
+  }
 
-  override def visitExpression(ctx: ExpressionContext): Expression = ctx match {
-    case c if c.full_column_name() != null => c.full_column_name().accept(this)
-    case c if c.primitive_expression() != null => c.primitive_expression().accept(this)
+  // This is merely a placeholder until the grammar can be fixed. It ensures that the
+  // precedence expression is correctly handled
+  override def visitBracket_expression(ctx: Bracket_expressionContext): ir.Expression = {
+    ctx.expression().accept(this)
```
vil1, line 25: What about `ctx.subquery`? Will it be handled in a future change ?

core/src/test/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilderSpec.scala
```diff
@@ -33,4 +34,78 @@ class TSqlAstBuilderSpec extends AnyWordSpec with TSqlParserTestCommon with Matc
...
+        Seq(
+          Subtract(
+            Multiply(Add(Mod(Column("a"), Literal(integer = Some(3))), Column("b")), Literal(integer = Some(2))),
+            Divide(Column("c"), Literal(integer = Some(5)))))))
+    example(query = "SELECT a || b", expectedAst = Project(NoTable(), Seq(Concat(Column("a"), Column("b")))))
+
+    example(
+      query = "SELECT a || b || c",
+      expectedAst = Project(NoTable(), Seq(Concat(Concat(Column("a"), Column("b")), Column("c")))))
+  }
```
vil1, line 110: These tests would probably be better in a `TSqlExpressionBuilderSpec` and expressed only around the expression part. Like 
```scala
example("a + b", _.expression(), Add(Column("a"), Column("b"))
```

vil1 (changes requested):
Looks good to me overall, but some details need to be addressed.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilder.scala
jimidle: OK - Just scala that I am not that familiar with yet. Will do.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: Yes - IN fact this rule is wrong and I need to fix the grammar in another PR

core/src/test/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilderSpec.scala
jimidle: Yes - that sounds good. Though I found some issues by using these tests two so I think a combination hits the spot.

jimidle: > Could you elaborate on the issues you faced? Maybe I can help.

To clarify, I mean that going through the SELECT path showed that SELECT without a Table was not currently catered for and I would have missed that if I only went for expression. But I agree that testsing the expression builder independently is the right way. I have no issues understanding the code etc.

vil1: I think we're on the same page then. We should keep a test for `SELECT 42` (or any query with no `FROM` clause) here and put tests around operator precedence and expressions in general in a `TSqlExpressionBuilderSpec`

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilder.scala
```diff
@@ -29,10 +29,8 @@ class TSqlAstBuilder extends TSqlParserBaseVisitor[ir.TreeNode] {
       .select_list_elem()
       .asScala
       .map(_.accept(new TSqlExpressionBuilder))
-    val fromTable = rawExpression.table_sources().accept(new TSqlRelationBuilder)
-
-    ir.Project(fromTable, columnExpression)
-
+    val tableSources = Option(rawExpression.table_sources())
+      .fold(ir.NoTable().asInstanceOf[ir.Relation])(_.accept(new TSqlRelationBuilder))
```
vil1, line 33:
```suggestion
      .fold[ir.Relation](ir.NoTable())(_.accept(new TSqlRelationBuilder))
```

vil1: Type inference indeed needs a little help here, but there's always a way to avoid asInstanceOf ;)

vil1 (approved):
LGTM, but please commit my suggestion before merging.

jimidle:
> LGTM, but please commit my suggestion before merging.

Note that I am not authorized to merge this.

nfx (approved):
lgtm

PR #300, opened 2024-04-30
Add IR builder for Tsql Unary expressions

alti-tude:


core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
```diff
@@ -13,6 +13,12 @@ case class And(left: Expression, right: Expression) extends Binary(left, right)
 case class Or(left: Expression, right: Expression) extends Binary(left, right) with Predicate {}
 case class Not(pred: Expression) extends Unary(pred) with Predicate {}
 
+object Minus {
```
nfx, line 16: we had a change to this in the other PRs, please rebase and resolve conflicts

nfx (approved):
@jimidle do we need this one?..

jimidle (changes requested):
Unary constructs should use names that imply they are unary. UMinus and UPlus for instance. Also, they should be case classes extending Unary in `extensions.scala`.

As i will change the grammar to deal with unary correctly, this PR will be obsoleted and I am sorry to say it is better to start again than be misguided by the current grammar.

@alti-tude Sorry mate, but would you please just close this one.

jimidle:
> @jimidle do we need this one?..

No. the PR was misguided by the state of the grammar. We can close this now as my other PRs implement this with grammar changes

jimidle:
Closed as already implemented.

PR #301, opened 2024-05-02
Implement translation of Snowflake `CREATE TABLE` command

vil1:


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -75,4 +75,63 @@ class SnowflakeDDLBuilder
...
+    val columns = ctx.collect {
+      case c if c.full_col_decl() != null => buildColumnDeclaration(c.full_col_decl())
+    }
+    val outOfLineConstraints = ctx.collect {
+      case c if c.out_of_line_constraint() != null => buildOutOfLineConstraint(c.out_of_line_constraint())
+    }.flatten
+
+    columns.map { col =>
+      val additionalConstraints = outOfLineConstraints.collect {
+        case (columnName, constraint) if columnName == col.name => constraint
```
nfx, line 101: what about other cases? shouldn't we do a filter instead?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -75,4 +75,63 @@ class SnowflakeDDLBuilder
...
+        .column_decl_item_list()
+        .column_decl_item()
+        .asScala)
+
+    ir.CreateTableCommand(tableName, columns)
+  }
+
+  private def buildColumnDeclarations(ctx: Seq[Column_decl_itemContext]): Seq[ir.ColumnDeclaration] = {
+    val columns = ctx.collect {
+      case c if c.full_col_decl() != null => buildColumnDeclaration(c.full_col_decl())
```
nfx, line 93: what about other cases? shouldn't we do a filter instead?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -75,4 +75,63 @@ class SnowflakeDDLBuilder
...
+  private def buildOutOfLineConstraint(ctx: Out_of_line_constraintContext): Seq[(String, ir.Constraint)] = {
+    val columnNames = ctx.column_list_in_parentheses(0).column_list().column_name().asScala.map(_.getText)
+    val constraints = ctx match {
+      case c if c.UNIQUE() != null => List.fill(columnNames.size)(ir.Unique)
+      case c if c.primary_key() != null => List.fill(columnNames.size)(ir.PrimaryKey)
+      case c if c.foreign_key() != null =>
+        val referencedObject = c.object_name().getText
+        val references =
+          c.column_list_in_parentheses(1).column_list().column_name().asScala.map(referencedObject + "." + _.getText)
+        references.map(ir.ForeignKey.apply)
```
nfx, line 124: can you add "unknown constraint" as a catch-all?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -75,4 +75,63 @@ class SnowflakeDDLBuilder
...
+    }
+    columnNames.zip(constraints)
+  }
+
+  private def buildInlineConstraint(ctx: Inline_constraintContext): ir.Constraint = ctx match {
+    case c if c.UNIQUE() != null => ir.Unique
+    case c if c.primary_key() != null => ir.PrimaryKey
+    case c if c.foreign_key() != null =>
+      val references = c.object_name().getText + Option(ctx.column_name()).map("." + _.getText).getOrElse("")
+      ir.ForeignKey(references)
```
nfx, line 135: can you add "unknown constraint" as a catch-all?

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
vil1: All the cases (as per the grammar) are covered here (and in the other snippet bellow). What's actually annoying is that I have absolutely no mean of proving that claim.
I could add a `case x => UnresolvedConstraint(x.getText)` but then I wouldn't be able to come up with an example query that would be successfully parsed by `SnowflakeParser` and hit that line of code, leaving said code uncovered by tests. And yet, this wouldn't be a suitable proof that the code covers all the cases.

vil1: That being said, I can add the `UnresolvedConstraint` case as a safety net, as long as we accept that it can't be covered by tests.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -75,4 +75,69 @@ class SnowflakeDDLBuilder
...
+    val columnNames = ctx.column_list_in_parentheses(0).column_list().column_name().asScala.map(_.getText)
+    val constraints = ctx match {
+      case c if c.UNIQUE() != null => List.fill(columnNames.size)(ir.Unique)
+      case c if c.primary_key() != null => List.fill(columnNames.size)(ir.PrimaryKey)
+      case c if c.foreign_key() != null =>
+        val referencedObject = c.object_name().getText
+        val references =
+          c.column_list_in_parentheses(1).column_list().column_name().asScala.map(referencedObject + "." + _.getText)
+        references.map(ir.ForeignKey.apply)
+      case c => List.fill(columnNames.size)(ir.UnresolvedConstraint(c.getText))
```
nfx, line 130:
```suggestion
      case c => 
        val filler = List.fill(columnNames.size)
        filler(ir.UnresolvedConstraint(c.getText))
```

nfx (approved):
lgtm

vil1:
@nfx: merging is blocked because of the `Unresolved*` lines are uncovered, you'll have to force merge.

nfx (approved):
lgtm

PR #307, opened 2024-05-03
TSQL - Implement unary and binary expressions with correct precedence

jimidle:
Unary operators were not correctly implemented in the ANTLR grammar file, and this is now corrected. In addition, unary operators are right associative, which affects order of operations in certain case. Precedence has also been corrected to follow the TSQL normative specification. 

Progresses: #275

PR #308, opened 2024-05-03
Correctly implement precedence in the grammar and the AST

jimidle:
Precedence was not correctly integrated into the main expression. It is now specified at the correct precedence level in the expression rule and subquery has its own place. A subquery is not an operator and as such does not have precedence, just a place in the expression rule. 

Progresses: #275

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -132,6 +132,19 @@
           <sourceDirectory>src/main/antlr4</sourceDirectory>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <version>3.6.3</version>
+        <executions>
+          <execution>
+            <id>attach-javadocs</id>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 </project>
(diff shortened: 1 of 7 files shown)


jimidle:
Awaiting merge of #307 so I can rebase this branch on main.

vil1 (commented):
I am not sure I understand why we need a `Precedence` case in the IR.

When things are represented in a tree structure, operators precedence is "embedded" in the tree structure itself.

For instance `Mult(Add(Lit(1), Lit(2)), Lit(3))` is trivially different from `Add(Lit(1), Mult(Lit(2), Lit(3)))`. What isn't trivial is whether the expression `1 + 2 * 3` should be translated into the former or the latter. However, once the translation is done, I don't think we need to carry any additional information.

If at some point we need to convert the IR tree back to a string, we can guarantee correct precedence by simply adding parenthesis around all subexpressions (`Mult(Add(Lit(1), Lit(2)), Lit(3))` would become `"1 + (2 * 3)`, etc).

jimidle:
> I am not sure I understand why we need a `Precedence` case in the IR.
> 
> When things are represented in a tree structure, operators precedence is "embedded" in the tree structure itself.
> 
> For instance `Mult(Add(Lit(1), Lit(2)), Lit(3))` is trivially different from `Add(Lit(1), Mult(Lit(2), Lit(3)))`. What isn't trivial is whether the expression `1 + 2 * 3` should be translated into the former or the latter. However, once the translation is done, I don't think we need to carry any additional information.
> 
> If at some point we need to convert the IR tree back to a string, we can guarantee correct precedence by simply adding parenthesis around all subexpressions (`Mult(Add(Lit(1), Lit(2)), Lit(3))` would become `"1 + (2 * 3)`, etc).

The original marker guaranteed that we could reproduce customer's redundant parens. It has been decided not to do this. The IR will generate correct expressions, but it will elide redundant expressions. We can wait to see if anyone complains.

jimidle:
@vil1 - Would you please review again? Thank you.

vil1 (approved):
LGTM

PR #315, opened 2024-05-06
Feature/upgrade sqlglot 23.0.1 to 23.13.7

ganeshdogiparthi-db:
* Latest SQLGLOT supports  Snowflake TIMESTAMP_NTZ
* struct_sql is an override function from the base generator class. `hasattr(e.this, "is_string") and` snippet is only added in override function.
* Added Median class in local_expressions to support median function.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ classifiers = [
 ]
 dependencies = [
   "databricks-sdk>=0.18.0",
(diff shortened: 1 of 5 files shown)


vil1 (approved):
LGTM

sundarshankar89 (approved):
LGTM

PR #316, opened 2024-05-06
Custom oracle tokenizer

ganeshdogiparthi-db:
* In the Oracle database, the long datatype can be used for character data as well. Therefore, we are mapping the long datatype to text (string) in the tokenizer. 
* We have only overridden the tokenizer because it is sufficient for handling the long datatype at present.


sundarshankar89 (approved):
LGTM

PR #322, opened 2024-05-06
Improve coverage in com.databricks.labs.remorph.parsers.snowflake

vil1:


core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
```diff
@@ -107,7 +107,7 @@ case class UnresolvedConstraint(inputText: String) extends Constraint
 case class ColumnDeclaration(
     name: String,
     dataType: DataType,
-    virtualColumnDeclaration: Option[Expression],
-    constraints: Seq[Constraint])
+    virtualColumnDeclaration: Option[Expression] = None,
+    constraints: Seq[Constraint] = Seq())
```
sundarshankar89, line 111: Why not keep constraints as `Option[Seq[Constraint]] here? else 
`Seq.empty[Constraint]`

sundarshankar89 (dismissed):
LGTM

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
vil1: I don't think it's necessary to go for `Option[Seq[Constraint]]` : what would be the difference in semantics between `None` and `Some(Seq.empty)` ?
Though you have a point with `Seq.empty` that's probably more readable than `Seq()`, I'll change that.

sundarshankar89 (approved):
LGTM

PR #325, opened 2024-05-07
Use POSEXPLODE when Snow SQL query requires index

### LINKED ISSUE #299: [FEATURE]: Use POSEXPLODE when Snowflake SQL query requires index info

### Problem statement

When given the following SQL:

SELECT 
  verticals.index AS index,    -- Accessing the index directly from the alias
  verticals.value AS value
FROM sample_data,
LATERAL FLATTEN(input => array_column, OUTER => TRUE ) AS verticals;  -- Aliasing FLATTEN as 'verticals'

remorph's translation uses to LATERAL VIEW OUTER EXPLODE. However, this does not produce index information like in Snowflake and will produce a query that fails.

### Proposed Solution

remorph should parse the query to find if ".index" is being used. If so, replace LATERAL VIEW OUTER EXPLODE with LATERAL VIEW OUTER POSEXPLODE. Use "pos" alias for the index and replace it. Use old alias for the value and replace ".value" instances.

SELECT 
  pos AS index,    -- pos represents the index of the element in the array
  verticals AS value
FROM sample_data
LATERAL VIEW POSEXPLODE(array_column) as pos, verticals;

### Additional Context

_No response_

bishwajit-db:
Use POSEXPLODE when Snow SQL query requires index.

Closes #299

### DIFF
--- a/src/databricks/labs/remorph/reconcile/query_builder/base.py
+++ b/src/databricks/labs/remorph/reconcile/query_builder/base.py
@@ -11,7 +11,6 @@ from databricks.labs.remorph.reconcile.recon_config import Schema, Table
 
 
 class QueryBuilder(ABC):
-
     def __init__(self, table_conf: Table, schema: list[Schema], layer: str, source: str):
         self._table_conf = table_conf
         self._schema = schema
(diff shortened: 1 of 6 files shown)


sundarshankar89 (approved):
LGTM

PR #330, opened 2024-05-08
Created verify_workspace_client func to handle product info

ganeshdogiparthi-db:
* For incoming workspace_client for the future user functions. We are checking and setting up proper product and product_version for telemetric data. 
* We had to use ` # pylint: disable=protected-access` since we are accessing _product and _product_version params in WorkspaceClient.


nfx (approved):
lgtm

PR #333, opened 2024-05-09
Fixed orcale missing datatypes

ganeshdogiparthi-db:
* After analysis, it is identified that for creating Tokenizer KEYWORDS mapping, we don't need Tokens for keys, So we made appropriate changes, and now we support all Oracle datatypes.


sundarshankar89 (approved):
LGTM

Issue #326, opened 2024-05-07
[FEATURE]: TSQL - Fix expressions assuming almost everything is a column


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Currently, the grammar is specified such that it uses `full_column_name` to pick up `id_`. This is correct as `id_`s are not always Column names. Whether the reference is to a column depends on context and we must resolve into `Column` references at a higher level than `expression`.

Also the extension Column and the associated visitors do not handle catalog.table.column  and are incorrect. There is a lot to fix here that cannot really be broken down as it involves grammar changes. The current grammar is broken.

### Proposed Solution

- Redo the whole column reference parser rules
- Work out what is needed for column references based on existing IR and only extend if needed
- Rejig the grammar such that full_column_name is not a part of expression (should handle `.` left recursively and not assume column name
- Generate Column using dedicated visitor

### Additional Context

This also relates to aliases, which cannot really be handled until the expression handles even basic id_ and DOT correctly. 

References: #275

PR #335, opened 2024-05-09
TSQL -Creates new context aware walkers for tracking column references

jimidle:
Distinguishes between column references and expression entities by using context aware walkers. Rejigs SELECT walking with visitor pattern. Corrects grammar for predicates. 

Closes: #326 
Progresses: #275

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
@@ -1,7 +1,7 @@
 package com.databricks.labs.remorph.parsers.tsql
 
-import com.databricks.labs.remorph.parsers.{IncompleteParser, intermediate => ir}
 import com.databricks.labs.remorph.parsers.tsql.TSqlParser._
+import com.databricks.labs.remorph.parsers.{IncompleteParser, intermediate => ir}
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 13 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -10,9 +10,44 @@ class TSqlRelationBuilder extends TSqlParserBaseVisitor[ir.Relation] with Incomp
...
+
+  /**
+   * Note that SELECT a, b, c FROM x, y, z is equivalent to SELECT a, b, c FROM x CROSS JOIN y CROSS JOIN z
+   * @param ctx
+   *   the parse tree
+   */
+  override def visitTableSources(ctx: TSqlParser.TableSourcesContext): ir.Relation = {
+    val relations = ctx.tableSource().asScala.toList.map(_.accept(this)).collect { case r: ir.Relation => r }
+    relations match {
+      case head :: Nil => head
```
vil1, line 39:
```suggestion
```
This case isn't needed as it is covered by the `foldLeft` in the case bellow

vil1 (changes requested):
I understand the intent of this PR, ie. having specialized ways of processing specific branches of the input grammar, but I disagree with the way it achieves that intent.

Specific implementations of interfaces like `TSqlParserBaseVisitor` are meant to provide  visitors that are specific _as far as the subtype of `ir.TreeNode` they produce is concerned_. `TSqlRelationBuilder` returns `ir.Relation` nodes , `TSqlExpressionBuilder` returns `ir.Expression`, and so on. We then leverage this for building specific instances of `ir.TreeNode`, in a type-safe manner, by writing `someRuleContext.accept(someSpecificParserBaseVisitorImplementation)`. 

For instance, let say we want to return a `ir.Filter` from a function whose return type is `ir.Relation`.  In order to populate the `condition` field, we need an implementation of `ParseTreeVisitor[ir.Expression]` we could pass the relevant sub-tree to using `accept`.

Note also that we cannot effectively make the output type any given sub-tree processing more specific by simply overriding the returned type of the corresponding `visit*` method in an AST builder.

When the need arises, we rather define a private `build*` method that returns the specific type we need.

We do so because we want to leverage the scala compiler as much as possible as a bug-preventing tool. Doing otherwise would ultimately lead us to write unsafe code (such as the use of `asInstanceOf` in this PR) that breaks type-safety (making errors that could have been caught at build time happen at run time).

That's why I believe that:
-  `TSqlColumnBuilder` and `TSqlSelectBuilder` do not need to be separate classes and should have their methods merged into `TSqlExpressionBuilder` and `TSqlRelationBuilder` respectively. 
- `ExpressionList` is not needed: it is only ever constructed to get access to its inner `Seq[ir.Expression]`, that should be inlined as a `...map(_.accept(...))` instead.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
jimidle: I will check that and remove it if it can indeed be removed.

PR #341, opened 2024-05-10
Creates a simple error collector for building on in the future

jimidle:
A simple error collector is created, which allows determination of errors in test code and makes that a little easier. In the future, it can be expanded to support different types of errors such as semantic errors.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -93,6 +93,16 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.typesafe.scala-logging</groupId>
+      <artifactId>scala-logging_${scala.binary.version}</artifactId>
+      <version>3.9.5</version>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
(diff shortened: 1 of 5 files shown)


core/pom.xml
```diff
@@ -93,6 +93,16 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.typesafe.scala-logging</groupId>
+      <artifactId>scala-logging_${scala.binary.version}</artifactId>
+      <version>3.9.5</version>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
```
sundarshankar89, line 102: @nfx need your input here

sundarshankar89 (changes requested):

core/pom.xml
sundarshankar89: revert to log4j.

jimidle: I didn't see that we were using log4j - at least not in the pom. I can change to use that.

jimidle:
I am going to close this and resubmit without the fork. Will resubmit using log4j if that is what we are using.

Issue #339, opened 2024-05-10
[FEATURE]: TSQL - Implement functions


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Function calls are massively over specified in the tsql grammar, unnecessarily complicating the lexer and parser by specifying most (but there are many missing) of the built-in functions directly. However, most functions have generic call patterns like this:

```antlr
 func((expression (COMMA expression)*)?)
 ```
 
 And there is no need to specifically parse them. 

### Proposed Solution

The grammar will be changed to remove directly specifying built-in function names and will generically parse them. A function builder will be used builder check argument counts for validity.
 
There are some unusually designed functions that do have to be parsed to make sense of them, and those will be retained. 

### Additional Context

References: #275

PR #342, opened 2024-05-10
Implement function calls via table lookup and remove explicit specifications

jimidle:
Replaces tons of explicit specifications of generic function calls with a a table of standard functions. We create a visitor for function call building to go with this. 

Closes: #339
Progresses: #275


core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
```diff
@@ -138,3 +138,5 @@ case class RenameConstraint(oldName: String, newName: String) extends TableAlter
 case class RenameColumn(oldName: String, newName: String) extends TableAlteration
 
 case class AlterTableCommand(tableName: String, alterations: Seq[TableAlteration]) extends Catalog {}
+
+case class ExpressionList(expressions: Seq[Expression]) extends Expression {}
```
sundarshankar89, line 142: why do we need this?

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlFunctionBuilder.scala
```diff
@@ -0,0 +1,233 @@
...
+    "PERMISSIONS" -> FunDef(0, 2),
+    "PWDENCRYPT" -> FunDef(1, 1),
+    "PWDCOMPARE" -> FunDef(2, 3),
+    "SESSIONPROPERTY" -> FunDef(1, 1),
+    "SUSER_ID" -> FunDef(0, 1),
+    "SUSER_NAME" -> FunDef(0, 1),
+    "SUSER_SID" -> FunDef(0, 2),
+    "SUSER_SNAME" -> FunDef(0, 1),
+    "USER_ID" -> FunDef(0, 1),
+    "USER_NAME" -> FunDef(0, 1))
```
sundarshankar89, line 201: should we have an IR for holding un-covered expression so we can transpile as is on target side.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
jimidle: It has gone. I changed the grammar but that is in another PR. I will rebase when that goes through

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlFunctionBuilder.scala
jimidle: > should we have an IR for holding un-covered expression so we can transpile as is on target side.

I am going to close this PR and resubmit it without the fork. 

On functions that are not covered - we may also want to collect those as lint messages (see other PR about collecting messages), but also it needs to check for user defined functions, but that is functionality for another PR I think.

jimidle:
Closing to resubmit without fork

PR #344, opened 2024-05-13
Process SELECT, table sources, and 

jimidle:
Rejigs SELECT walking with visitor pattern. 
Corrects grammar for predicates.
Simplifies SelectList elements

Closes: https://github.com/databrickslabs/remorph/issues/326
Progresses: https://github.com/databrickslabs/remorph/issues/275


jimidle:
@vil1 - rejigged to use existing visitors

sundarshankar89 (commented):
lgtm, I will let @vil1 take a pass before merging.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -1,24 +1,38 @@
...
+  override def visitFullColumnName(ctx: FullColumnNameContext): ir.Column = {
+    val columnName = ctx.id.getText
+    val fullColumnName = ctx.fullTableName() match {
+      case null => columnName
+      case ft =>
+        ft.accept(this) match {
+          case nt: ir.Literal if nt.string.isDefined => nt.string.get + "." + columnName
+          case _ => columnName
+        }
+    }
```
vil1, line 34:
```suggestion
    val fullColumnName = Option(ctx.fullTableName())
      .map(_.accept(this))
      .collect{
          case nt: ir.Literal if nt.string.isDefined => nt.string.get + "." + columnName
       }
       .getOrElse(columnName) 
    }
```

nitpicking

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -1,18 +1,74 @@
...
 
-  override def visitTableSourceItem(ctx: TableSourceItemContext): ir.Relation = {
-    val table = ctx.fullTableName().getText
-    // [TODO]: Handle Table Alias ctx.tableAlias().getText
-    ir.NamedTable(table, Map.empty, is_streaming = false)
+  override def visitSelectStatementStandalone(ctx: TSqlParser.SelectStatementStandaloneContext): ir.Relation = {
+
+    // TODO: Process ctx.WithExpression
+    ctx.selectStatement().accept(this)
+    // val tableSources = Option(rawExpression.tableSources()).fold[ir.Relation](ir.NoTable())
```
vil1, line 16: Is this commented code part of the above TODO? Otherwise we may not want to keep it.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -1,18 +1,74 @@
...
+
+    ctx.queryExpression.accept(this)
+  }
+
+  override def visitQuerySpecification(ctx: TSqlParser.QuerySpecificationContext): ir.Relation = {
+
+    // TODO: Process all the other elements of a query specification
+
+    val columns =
+      ctx.selectListElem().asScala.toList.map(_.accept(new TSqlExpressionBuilder()))
```
vil1, line 33: `toList` isn't necessary here (although its absence may result in the IDE mistakenly flagging the line as erroneous).

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -1,18 +1,74 @@
...
+    val from = Option(ctx.tableSources()).map(_.accept(new TSqlRelationBuilder)).getOrElse(ir.NoTable())
```
vil1, line 34: 👍 Going for `map ... getOrElse` helps type inference better than `fold` here.

vil1 (approved):
LGTM, despite some non-blocking comments.
It would be nice to have coverage for the two missing lines, though it could be added in a future change. If coming up with input string for the uncovered lines proves difficult/impossible, we'll have to use mocking (I'm happy to help with that).

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: I am not sure that this was my code - I just moved it. However, I have changed it.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
jimidle: It was. But looking again, I don't think it needs to be retained here.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
jimidle: IDE is fine without the toList

jimidle:
> LGTM, despite some non-blocking comments. It would be nice to have coverage for the two missing lines, though it could be added in a future change. If coming up with input string for the uncovered lines proves difficult/impossible, we'll have to use mocking (I'm happy to help with that).

Actually, coverage was helpful here. There is no possibility that there will be a Nil tableSource as we won't visit that node if it does not exist at all `(FROM tableSources)?` and the querySpecification visitor handles the case that the SELECT has no table sources -`SELECT 42`

So I removed that line. But then the compiler errors out on the  fact that the match may not be exhaustive. So it has to be there despite never being reached. I have marked it as not covered, which the docs say will stop the coverage complaining.

The other line not covered is because of IncompleteParser. The def has to be there, however we are never going to use it as far as I can see. I have removed it while we do not use it - if it does become useful, then it can be added.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -66,25 +78,83 @@ class TSqlExpressionBuilder
...
+   *
+   * For now then, they are all seen as columns.
+   *
+   * @param ctx
+   *   the parse tree
+   */
+  override def visitId(ctx: IdContext): ir.Expression = ctx match {
+    case c if c.ID() != null => ir.Column(ctx.getText)
+    case c if c.TEMP_ID() != null => ir.Column(ctx.getText)
+    case c if c.DOUBLE_QUOTE_ID() != null => ir.Column(ctx.getText)
```
vil1, line 150: we should probably remove the enclosing `"` here (as well as the `[` and `]` bellow)

jimidle: Thought about that, but there is only one isQuoted flag, so I left it for now. Just like the parens, is it important to remember which type of quote was used? I suspect not given that we are producing a different SQL dialect. I will do that in a differnt PR in the interests of getting this one over the line.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -66,25 +78,83 @@ class TSqlExpressionBuilder
...
+    case c if c.SQUARE_BRACKET_ID() != null => ir.Column(ctx.getText)
+    case c if c.RAW() != null => ir.Column(ctx.getText)
+    case _ => ir.Column(ctx.getText)
```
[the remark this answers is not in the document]
jimidle: No - unless you want to case every keyword in TSQL. The parser rule Id does not yield anything that cannot be an identifier. If it was not for the quoting, we could just ignore the token type altogether.  So that RAW is only there because it is specifically mentioned. Every other id token, we just need the raw text.

PR #345, opened 2024-05-13
Rework functions to remove overly explicit definitions in the grammar

jimidle:
Replaces tons of explicit specifications of generic function calls with a a table of standard functions. We create a visitor for function call building to go with this.

Closes: #342 339
Progresses: #275


jimidle:
@sundarshankar89 Functions marked as non-convertible will generate an ir.UnresolvedFunction though will will need to work on this with user defined functions

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -1,22 +1,214 @@
...
 
+import java.util.Locale
 import scala.collection.JavaConverters._
 
 class TSqlExpressionBuilder
     extends TSqlParserBaseVisitor[ir.Expression]
     with ParserCommon
     with IncompleteParser[ir.Expression] {
 
+  case class FunDef(argMin: Int, argMax: Int = Int.MaxValue, convertible: Boolean = true)
```
vil1, line 17: `Arity` or `FunctionArity` may be more informative names than `FunDef`. Looking at the usage bellow, it may also benefit from being defined as an ADT:

```scala
sealed trait FunctionArity
case class FixedArity(nbParameters: Int) extends FunctionArity
case class VariableArity(argMin: Int, argMax: Int) extends FunctionArity
```

The `convertible` parameter seems to always be set to its default value, so I ignored it for now.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -1,22 +1,214 @@
...
+
   protected override def wrapUnresolvedInput(unparsedInput: String): ir.UnresolvedExpression =
     ir.UnresolvedExpression(unparsedInput)
 
+  // TODO: Add more functions as we find them
+  private val functions: Map[String, FunDef] = Map(
```
vil1, line 23: Yay tabulated function! Brings up memories of game programming in C++ ❤️ 

One minor drawback is that it trumps code coverage somehow (all the lines will be considered as covered as long as the `function` member is evaluated in a test, which happens whenever a `TSqlExpressionBuilder` object is instantiated). 

Good news is, the changes required to fix that are fairly mechanical:
```suggestion
  def functionArities(functionName: String): FunDef = functionName match {
```
and then rewrite every following `"FUNC_NAME" -> FuncDef(x, y)` line as `case "FUNC_NAME" => FuncDef(x, y)`.

Because doing so would make a lot of lines marked as uncovered, we should move the definition to a separate file (for example as a member of a `FunctionArities` object) and decide that having uncovered lines in that file is acceptable for now. 

It would also make it easier to later adapt and reuse this information in other dialects.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -133,4 +327,34 @@ class TSqlExpressionBuilder
...
+
+  override def visitStandardFunction(ctx: StandardFunctionContext): ir.Expression = {
+    val name = ctx.funcId.getText
+    val uName = name.toUpperCase(Locale.getDefault())
+    val defnOption = functions.get(uName)
+
+    val args = ctx.expression() match {
+      case null => Seq.empty
+      case _ => ctx.expression().asScala.map(_.accept(this))
+    }
```
vil1, line 339:
```suggestion
    val args = Option(ctx.expression()).map(_.asScala.map(_.accept(this))).getOrElse(Seq.empty)
```

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -133,4 +327,34 @@ class TSqlExpressionBuilder
...
+        if (args.size < defn.argMin || args.size > defn.argMax) {
+          // TODO: This should be a semantic error
+          throw new IllegalArgumentException(
+            s"Function $name expects between ${defn.argMin} and ${defn.argMax} arguments")
+        }
+        if (!defn.convertible) {
+          // This function was flagged in the table as something that cannot be converted with current functionality
+          return ir.UnresolvedFunction(name, args, is_distinct = false, is_user_defined_function = false)
+        }
+        ir.CallFunction(name, args)
```
vil1, line 357:
```suggestion
      case Some(defn) if defn.argMin <= args.size && defn.argMax >= args.size =>
        ir.CallFunction(name, args)
      case _ => ir.UnresolvedFunction(name, args, is_distinct = false, is_user_defined_function = false)
```

vil1 (changes requested):
Much valuable work, requesting a few adjustments though.

jimidle:
> Much valuable work, requesting a few adjustments though.

The one lint not covered is rewritten in another PR, so I did not bother with it. To cover the other, I will mark a funciotn that we cannot convert.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: > `Arity` or `FunctionArity` may be more informative names than `FunDef`. Looking at the usage bellow, it may also benefit from being defined as an ADT:
> 
> ```
> sealed trait FunctionArity
> case class FixedArity(nbParameters: Int) extends FunctionArity
> case class VariableArity(argMin: Int, argMax: Int) extends FunctionArity
> ```
> 
> The `convertible` parameter seems to always be set to its default value, so I ignored it for now.

I don't mind changing the name but I think it is likely that this table may expand to have information other than arity. It does have a parameter that indicates whether we can convert it for instance. I guess we can change it again later  so I will change to use this trait pattern.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: Yeah - did not consider it important to have every entry in the table checked. I think then, that moving the function lookup and table to a common file and over time annotating common functions vs TSQL specific vs Sonwflake vs X, is a good idea so I will do that in this PR.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: OK. I have done that elsewhere as you may have seen. Was not sure which way of doing it is considered `best`

vil1 (approved):
LGTM. As discussed, having lines in `FunctionBuilder` that aren't covered by tests is OK for now.

vil1:
@sundarshankar89 could you use your super-power and merge this one (despite codecov being unhappy)?

sundarshankar89:
> @sundarshankar89 could you use your super-power and merge this one (despite codecov being unhappy)?

Nope. Can't merge without all required checks are met.

vil1:
Now you should be able to merge it @sundarshankar89

Issue #340, opened 2024-05-10
[FEATURE]: Build a rudimentary error listener that collects parser and semantic errors


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

The standard output for ANTLR4 recognizers is just to print a basic message to standard err. It would be useful to collect these, and provide rudimentary additional information such as number of errors. This will enhance testing and also allow the builders and any other walkers to provide semantic errors or errors about inability to convert code (linting down the line, perhaps).

This will make it easier for test specifications to stop if parsing errors occur, amongst other things.

### Proposed Solution

A quick and dirty collector will be written that collects errors without doing anything too fancy. Maybe go as far as printing the source line in error with a carat, but such things can always be added later.

PR #346, opened 2024-05-13
Provide a rudimentary error collector for syntax errors and future semantic and linting errors

jimidle:
A simple error collector is created, which allows determination of errors in test code and makes that a little easier. In the future, it can be expanded to support different types of errors such as semantic errors.

Closes: #340

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/ErrorCollector.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/ErrorCollector.scala
@@ -1,16 +1,44 @@
 package com.databricks.labs.remorph.parsers
 
+import org.antlr.v4.runtime._
 import org.apache.logging.log4j.{LogManager, Logger}
-import org.antlr.v4.runtime.{BaseErrorListener, RecognitionException, Recognizer, Token}
 import org.json4s._
 import org.json4s.jackson.Serialization
 import org.json4s.jackson.Serialization.write
 
 import scala.collection.mutable.ListBuffer
 
+sealed trait ErrorCollector extends BaseErrorListener {
+  def logErrors(): Unit = {}
+  def errorsAsJson: String = "{}"
+  def errorCount: Int = 0
+  private[parsers] def formatErrors: Seq[String] = Seq()
+  def reset(): Unit = {}
+}
+
 case class ErrorDetail(line: Int, charPositionInLine: Int, msg: String, offendingToken: Token)
 
-class ErrorCollector(sourceCode: String, fileName: String) extends BaseErrorListener {
+class DefaultErrorCollector extends ErrorCollector {
+
+  var count: Int = 0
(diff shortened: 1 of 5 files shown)


jimidle:
@sundarshankar89 - Reworked to use Log4j as supplied by scala-core

core/src/main/scala/com/databricks/labs/remorph/parsers/ErrorCollector.scala
```diff
@@ -0,0 +1,59 @@
...
+      recognizer: Recognizer[_, _],
+      offendingSymbol: Any,
+      line: Int,
+      charPositionInLine: Int,
+      msg: String,
+      e: RecognitionException): Unit = {
+    errors += ErrorDetail(line, charPositionInLine, msg, offendingSymbol.asInstanceOf[Token])
+  }
+
+  def formatErrors(): Seq[String] = {
```
vil1, line 29:
```suggestion
  private [parsers] def formatErrors: Seq[String] = {
```

core/src/main/scala/com/databricks/labs/remorph/parsers/ErrorCollector.scala
```diff
@@ -0,0 +1,59 @@
...
+        if (start > 0) error.offendingToken.getStartIndex - start + 3 else error.offendingToken.getStartIndex - start
+      val marker =
+        " " * markerStart + "^" *
+          (error.offendingToken.getStopIndex - error.offendingToken.getStartIndex + 1)
+      s"File: $fileName, Line: ${error.line}, Token: ${error.offendingToken.getText}\n$windowedLine\n$marker"
+    }
+  }
+
+  def logErrors(): Unit = {
+    val formattedErrors = formatErrors()
```
vil1, line 46:
```suggestion
    val formattedErrors = formatErrors
```

vil1 (changes requested):
A few comments and suggestions.

core/src/test/scala/com/databricks/labs/remorph/parsers/ParserTestCommon.scala
```diff
@@ -2,28 +2,40 @@ package com.databricks.labs.remorph.parsers
...
-import org.antlr.v4.runtime.{CharStream, CharStreams, CommonTokenStream, Parser, RuleContext, TokenSource, TokenStream}
+import org.antlr.v4.runtime._
 import org.scalatest.{Assertion, Assertions}
 
 trait ParserTestCommon[P <: Parser] { self: Assertions =>
 
   protected def makeLexer(chars: CharStream): TokenSource
   protected def makeParser(tokens: TokenStream): P
-
+  protected def makeErrHandler(chars: String): ErrorCollector = null
```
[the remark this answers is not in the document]
jimidle: Default handler created for sealed trait

jimidle:
> A few comments and suggestions.

All addressed now I think? Seems I cannot mark as resolved.

Issue #165, opened 2024-03-07
Common Snowflake syntax conversion errors


dougspadotto:
I believe these 2 Snowflake syntax detection and conversion should be covered by remorph and they're still returning errors:

1. When the Snowflake code had setting options like “!set exit_on_error=true”, remorph returns an error. We should just warn, comment out, or remove the line and progress.

2. When passing a parameter, for example, MERGE INTO &SCHEMA_NAME.ATABLE T … remorph also gives an error with the unexpected ‘&’. We should convert to the Databricks SQL way of passing parameters, not only for schema/table names but also on values like DATE = &INPUT_DATE.

sundarshankar89:
@dougspadotto variables are represented with `$` in snowflake not `&' 
https://docs.snowflake.com/en/sql-reference/session-variables
snow the query works when `$` is used

dougspadotto:
@sundarshankar89 , when using variable substitution, the & syntax is utilized. Please check this out: https://docs.snowflake.com/en/user-guide/snowsql-use#enabling-variable-substitution

 I'm also adding an example of a working script for reference.
<img width="756" alt="image" src="https://github.com/databrickslabs/remorph/assets/5835211/32ae71a3-6d8a-479e-9257-66c3c2e7de64">

dougspadotto:
@sundarshankar89 , is there any plan to support variable substitution? It seems to be a matter of adding & as a  PARAMETER_TOKEN at remorph/src/databricks/labs/remorph/snow/databricks.py, but I don't want to do a PR because I haven't investigated deeply any other consequences of this change.

sundarshankar89:
we will pick this up in the coming week as below output.
https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/sql-ref-variables

PR #350, opened 2024-05-14
Remove SET options

vijaypavann-db:
closes 1st issue in #165  

Current behaviour:
```
ParserError(file_name='hive1.sql', exception='Invalid expression / Unexpected token. Line 1, Col: 19.\n  !set exit_on_error--> = <--true\nselect * from catalog.information_schema.tables limit 10;\n!set header=false\nselect * from m')
```

This PR removes the SET options and prints the Warning message

### DIFF
--- a/src/databricks/labs/remorph/helpers/file_utils.py
+++ b/src/databricks/labs/remorph/helpers/file_utils.py
@@ -1,5 +1,9 @@
 import codecs
+import logging
+import re
 from pathlib import Path
+
+logger = logging.getLogger(__name__)
 
 
 # Optionally check to see if a string begins with a Byte Order Mark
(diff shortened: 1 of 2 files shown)


src/databricks/labs/remorph/helpers/file_utils.py
```diff
@@ -101,3 +105,25 @@ def refactor_hexadecimal_chars(input_string: str) -> str:
...
+    lines = []
+    output_string = input_string
+
+    for line in input_string.splitlines():
+        if line.startswith('!set'):
+            lines.append(line)
+            output_string = re.sub(f"{line}", "", output_string)
+
+    set_options = "\n ".join(lines)
+    logger.warning(f"WARNING: `{source.upper()}` SET Options are not supported: \n {set_options}")
```
sundarshankar89, line 127: @vijaypavann-db: we need to handle this https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-aux-set-variable.html

sundarshankar89 (changes requested):

sundarshankar89:
Closing this and will be revisited

PR #351, opened 2024-05-15
Expand translation of Snowflake `expr`

vil1:
Some bit of the grammar are still not covered, some seem to be plain wrong (such as `subquery` being considered a valid `expr`), other are "duplicated" from `predicate`.


sundarshankar89 (approved):
This is very common use case, wonderful to see this coming along.

jimidle:
The Snowflake grammar is almost but not quite completely broken on it s definition of expressions and predicates. Those rules need to be reworked to be left recursive with precedence, like the TSQL grammar. I will add a new grammar option today that is a full TSQL grammar. However, I hink we can add to that to make it Snowflake compatible and potentially a universal parser. 

So let's approve this PR and connect on how far to take the Snowflake grammar.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3681,7 +3681,7 @@ expr_list_sorted
 expr
     : object_name DOT NEXTVAL
     | expr LSB expr RSB //array access
-    | expr COLON expr   //json access
+    | expr COLON json_path   //json access
     | expr DOT (VALUE | expr)
     | expr COLLATE string
     | case_expression
```
jimidle, line 3687: every time we change this rule, it will like break. We need to rework it to be truly left recursive with precedence corrected.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3699,34 +3699,43 @@ expr
...
+    ;
+
+json_path
+    : json_path_elem (DOT json_path_elem)*
+    ;
+
+json_path_elem
+    : ID | DOUBLE_QUOTE_ID
+    ;
+
```
jimidle, line 3726: We should not need partial rules to avoid self recursion. The rule should look something like this:

```antlr
searchCondition
    : LPAREN searchCondition RPAREN         #scPrec
    | NOT searchCondition                   #scNot
    | searchCondition AND searchCondition   #scAnd
    | searchCondition OR searchCondition    #scOr
    | predicate                             #scPred
    ;

predicate
    : EXISTS LPAREN subquery RPAREN
    | freetextPredicate
    | expression comparisonOperator expression
    | expression ME expression ////SQL-82 syntax for left outer joins; PE. See https://stackoverflow.com/questions/40665/in-sybase-sql
    | expression comparisonOperator (ALL | SOME | ANY) LPAREN subquery RPAREN
    | expression NOT* BETWEEN expression AND expression
    | expression NOT* IN LPAREN (subquery | expressionList) RPAREN
    | expression NOT* LIKE expression (ESCAPE expression)?
    | expression IS nullNotnull
    ;
```

Which is what TSQL looks like right now (though it will likely change)

jimidle (approved):
I think we accept this as is and then revisit the whole expression and predicate tree in the Snowflake grammar.  It can be done without throwing away the current visitor code.

PR #354, opened 2024-05-16
Better TSQL parser, with intent of creating a universal parser for all dialects

jimidle:
A new open source parser is added, which is complete and tested up to TSQL 2000R2.

While this is currently intended to replace the poorly formed grammar currently in use, the is a possibility that we could adapt it to Snowflake and other dialects and use it as a universal parser (hence the current name). 

This PR adds the grammar, with configurations and unit test for parse only. The unit tests can be expanded to check AST in another PR.

Progresses: #275 as a better grammar is needed to complete this


sundarshankar89:
I love the work being done here, but flagging as `do-not-merge` as I want to discuss how we are going to use common grammars

jimidle:
> I love the work being done here, but flagging as `do-not-merge` as I want to discuss how we are going to use common grammars and extensions

Sure - no problem. That's exactly why I pushed it. @vil1 and I discussed this briefly. I suggest that we use this for TSQL right now, as it is much better, then worry about making it Snowflake compatible down the line. For TSQL, the current visitors will not change very much as both the current grammar and this one more or less use the same names etc as the TSQL spec. But for Snowflake, there might be quite a bit of rejigging the current visitors; probably just mechanical, but would require changes. Whether or not we could use a common visitor is another thing to talk about.

vil1 (approved):
LGTM

sundarshankar89:
@nfx This is the start of universal grammar, tagging you for your input on the design.

jimidle:
@sundarshankar89 - would you add this to the merge queue please?

core/src/main/antlr4/com/databricks/labs/remorph/parsers/library/USQlCursors.g4
```diff
@@ -0,0 +1,158 @@
+// MIT License
```
nfx, line 1: the code in this repo is under https://github.com/databrickslabs/remorph/blob/main/LICENSE, where this license is coming from? what is the original repository?

core/src/main/antlr4/com/databricks/labs/remorph/parsers/library/USQlCursors.g4
```diff
@@ -0,0 +1,158 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/**
+ * This grammar contains all statements that deal with cursors other than
+ * CREATE cursor
+ */
+parser grammar USqlCursors;
```
nfx, line 27:
```suggestion
parser grammar Cursors;
```
`USql` is a bit confusing. if we do get something common, it has to be a bit more carefully named.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/library/USQlCursors.g4
```diff
@@ -0,0 +1,158 @@
...
+
+///////////////////////////////////////////////////////////
+// DDL statements that control cursors
+//
+ cursorDdl
```
nfx, line 32: we need to get the lineage comments based on which SQL standards we got these rules named after.

if we do get any common parser bits, all of them have to stem from Spark's grammar, which is derived from HIVE's:

https://github.com/apache/spark/blob/7e8b60b5ae7d6453bc1ce51b5112c975f9aa8757/sql/api/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBaseParser.g4

nfx (changes requested):
it is too early to introduce a common layer without a proper research across dialects we need to support. also, the licensing part is triggering couple of questions.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/library/USQlCursors.g4
jimidle: Sure.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/library/USQlCursors.g4
jimidle: https://github.com/jimidle/tsqlparser

PR #355, opened 2024-05-16
Simplifies alias grammar and implements TableAlias

jimidle:
Implements the new extension TableAlias and handles the simple case of NamedTable aliases.


sundarshankar89 (approved):
LGTM

PR #360, opened 2024-05-20
Integrate threshold reconciliation with core 

ravit-db:
- Added threshold reconciliation to the core
- Added ThresholdOutput dataclass to the ReconcileOutput
- Added end-to-end reconciliation core to integrate all components(data,schema,threshold)

### DIFF
--- a/src/databricks/labs/remorph/reconcile/recon_config.py
+++ b/src/databricks/labs/remorph/reconcile/recon_config.py
@@ -172,6 +172,13 @@ class ReconcileOutput:
     mismatch: MismatchOutput | None = None
     missing_in_src: DataFrame | None = None
     missing_in_tgt: DataFrame | None = None
+    threshold_output: ThresholdOutput | None = None
+
+
+@dataclass
+class ThresholdOutput:
+    threshold_df: DataFrame | None
+    threshold_mismatch_count: int = 0
 
 
 @dataclass
--- a/tests/unit/reconcile/query_builder/test_threshold_query.py
+++ b/tests/unit/reconcile/query_builder/test_threshold_query.py
@@ -76,7 +76,7 @@ def test_build_threshold_query_with_single_threshold(table_conf_with_opts, table
     table_conf = table_conf_with_opts
     table_conf.jdbc_reader_options = None
     table_conf.transformations = [
-        Transformation(column_name="s_acctbal", source="trim(s_acctbal)", target="trim(s_acctbal)")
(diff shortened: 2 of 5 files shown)


tests/unit/reconcile/query_builder/test_threshold_query.py
```diff
@@ -19,7 +19,7 @@ def test_threshold_comparison_query_with_one_threshold(table_conf_with_opts, tab
     table_schema, _ = table_schema
     table_schema.append(Schema("s_suppdate", "timestamp"))
     comparison_query = ThresholdQueryBuilder(
-        table_conf, table_schema, "source", get_dialect("oracle")
+        table_conf, table_schema, "target", get_dialect("databricks")
```
ganeshdogiparthi-db, line 22: Why change? I wanted to test that It works the same irrespective of layer and source. 
test_threshold_comparison_query_with_dual_threshold test case covers for target and databricks.

src/databricks/labs/remorph/reconcile/recon_config.py
```diff
@@ -172,6 +172,13 @@ class ReconcileOutput:
     mismatch: MismatchOutput | None = None
     missing_in_src: DataFrame | None = None
     missing_in_tgt: DataFrame | None = None
+    threshold_output: ThresholdOutput | None = None
+
+
+@dataclass
+class ThresholdOutput:
+    threshold_df: DataFrame | None
+    threshold_mismatch_count: int = 0
```
ganeshdogiparthi-db, line 181: We can add threshold_mismatch_col also to this. @sundarshankar89 call to implement it now or not.
Cavait: Since threshold comparison is not cached .. we are already hitting the source and target two times (sample and count). If we implement this it will be one more iteration.

tests/unit/reconcile/query_builder/test_threshold_query.py
ravit-db: Yes.But as per logic data resides in Databricks views all the data types are related to Databricks.So we wanted to restrict to Databricks and Target.

ganeshdogiparthi-db: No need .. It works as a negative case also. I suggest revert it back.

ravit-db: done.

sundarshankar89 (approved):
LGTM

Issue #356, opened 2024-05-16
[FEATURE]: TSQL Implement full visit chain from start rule down to statements


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

The visitors to create Batch all the way down need to be implemented to get rid of the need for aggregating rules with multiple children. The default generated visitors will call aggregate when a rule generates multiple children, which is there as a catch-all really. 

### Proposed Solution

Write the visitors for the levels up from SelectStatementStandalone up to the tsqlfile.

### Additional Context

#275

PR #363, opened 2024-05-20
TSQL: Implement full visit chain from highest grammar rule to SELECT

jimidle:
Leaving the default ANTLR generated implementations of the visit chain in between tSqlFile and select statements mean that we had to override the aggregateResult visitor, which happens because ANTLR will not be able to know which alts were actually visited and so expects you to decide wha to do with them.

We eliminate the problem by implement the visitors in the chain, though they only deal with SELECT at the moment.

Closes: #356 
Progresses: #275


sundarshankar89 (approved):
LGTM

PR #364, opened 2024-05-20
Implement column alias processing for TSQL

jimidle:
Column aliases are now fully supported within SELECT statements and some other statements. Note that column alias is required in other places as the visitor is expanded.

Progresses: #275


sundarshankar89 (approved):
LGTM

PR #365, opened 2024-05-20
Implement COLLATEd expressions/columns

jimidle:
Support COLLATEd expressions, which yieds COLLATE support for SELECT columns and WHERE columns.

Progresses: #275


sundarshankar89 (approved):
LGTM

PR #366, opened 2024-05-20
Implement the CASE expression

jimidle:
Implements the CASE expression in both expression and searchCondition form. Note that there is no big difference in the two cases:

```sql
CASE WHEN a <4 AND f > 0 ....
CASE e WHEN  G>5 ...
```

It is just a matter of either having the first expression or not. There is also little point in trying to enforce the difference syntactically, so the grammar is changed to accept searchCondition for all WHEN clauses.


sundarshankar89 (approved):
LGTM

PR #367, opened 2024-05-20
Implement timezone and provide test case

jimidle:
Timezone specification is very simple and is part of the expression tree:

```sql
a AT TIME ZONE 'UTC'
```

Visitor implemented and test provided.

Progresses: #275


vil1 (approved):
LGTM

PR #368, opened 2024-05-20
Implement hierarchical functions

jimidle:
These functions are not special in any way and have just been moved to the standard function table.

Progresses: #275


sundarshankar89 (approved):
LGTM

PR #370, opened 2024-05-21
Refactor Sowflake `expr` rule using labels

vil1:
That allows for simpler implementation of the translation relying on method dispatch rather than pattern-matching.


jimidle (commented):
I think that the precedence in this stack of rules is not correct. The expression rule was broken like the TSQL one and as well as changing to labels, which is good, the rule order needs changing and bracket_expression needs to be absorbed into it. 

Given that it was already broken, I am hesitant to request that to be fixed in this PR. It coudl be fixed in a separate PR without breaking the changes to the visitors.

PR #373, opened 2024-05-21
Implement XML functions such as query, and local id assignment

jimidle:
Implements two more elements of select lists, being 

`@localid = column`

and the various XML column related functions.

Both are additions to the  selectListElem visitor.

New IR is created that is used when the visitor recognizes XML functions (and can expand to other types of functions). The IR indicates which function is to be applied to which column expression.


jimidle:
Note that coverage percentage seems to be affected by looking at the number of lines that changed, but not accounting for lines that we don't cover such as grammar changes. If they are, then the .g4 files should not be counted in the coverage % as here I have one line that cannot be reached and it means the patch coverage is not hitting the required target despite all executed vist code being tested.

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
```diff
@@ -206,33 +221,45 @@ object FunctionBuilder {
...
     case _ => None
   }
   // GCOVR_EXCL_STOP
 
+  def functionType(name: String): FunctionType = {
+    val uName = name.toUpperCase(Locale.getDefault())
+    val defnOption = functionArity(uName)
+    defnOption match {
+      case Some(fixedArity: FixedArity) => fixedArity.functionType
+      case Some(variableArity: VariableArity) => variableArity.functionType
```
vil1, line 236:
```suggestion
      case Some(FixedArity(_ , functionType, _)) => functionType
      case Some(VariableArity(_, _, functionType, _)) => functionType
```

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
```diff
@@ -206,33 +221,45 @@ object FunctionBuilder {
...
     val defnOption = functionArity(uName)
 
     defnOption match {
-      case Some(FixedArity(_, false)) | Some(VariableArity(_, _, false)) =>
-        // Should raise a warning/lint error here
+      case Some(fixedArity: FixedArity) if !fixedArity.isConvertible =>
         ir.UnresolvedFunction(name, args, is_distinct = false, is_user_defined_function = false)
 
-      case Some(FixedArity(arity, true)) if args.length == arity =>
+      case Some(fixedArity: FixedArity) if args.length == fixedArity.arity =>
```
vil1, line 249:
```suggestion
      case Some(FixedArity(arity, _, _)) if args.length == arity =>
```

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
```diff
@@ -206,33 +221,45 @@ object FunctionBuilder {
...
         ir.CallFunction(name, args)
 
-      case Some(VariableArity(argMin, argMax, true)) if args.length >= argMin && args.length <= argMax =>
+      case Some(variableArity: VariableArity) if !variableArity.isConvertible =>
```
vil1, line 252:
```suggestion
      case Some(VariableArity(_, _, _, true)) =>
```

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
```diff
@@ -206,33 +221,45 @@ object FunctionBuilder {
...
+        ir.UnresolvedFunction(name, args, is_distinct = false, is_user_defined_function = false)
+
+      case Some(variableArity: VariableArity)
+          if args.length >= variableArity.argMin && args.length <= variableArity.argMax =>
```
vil1, line 256:
```suggestion
      case Some(VariableArity(argMin, argMax, _, _)) if args.length >= argMin && args.length <= argMax =>
```

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
```diff
@@ -206,33 +221,45 @@ object FunctionBuilder {
...
         ir.CallFunction(name, args)
 
       case Some(_) =>
-        // Should raise a warning/lint error here about mismatched argument count
         ir.UnresolvedFunction(name, args, is_distinct = false, is_user_defined_function = false)
 
       case None =>
-        // Should raise a warning/lint error here about unknown function
         ir.UnresolvedFunction(name, args, is_distinct = false, is_user_defined_function = false)
```
vil1, line 263: These two cases could be merged as a single `case _ =>`

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -10,9 +10,39 @@ import scala.collection.JavaConverters.{asScalaBufferConverter, collectionAsScal
...
-    ctx.expressionElem.accept(this)
+  override def visitSelectListElem(ctx: TSqlParser.SelectListElemContext): ir.Expression = {
+    val handlers = Seq(
+      // TODO: asterisk not fully handled
+      Option(ctx.asterisk()).map(_ => ctx.asterisk().accept(this)),
+      // TODO: UDT elements seem broken in the grammar
+      Option(ctx.udtElem()).map(_ => ctx.udtElem().accept(this)),
+      Option(ctx.LOCAL_ID()).map(_ => buildLocalAssign(ctx)),
+      Option(ctx.expressionElem()).map(_ => ctx.expressionElem().accept(this)))
+    handlers.flatten.headOption.getOrElse(ir.UnresolvedExpression("Unsupported SelectListElem"))
```
vil1, line 21:
```suggestion
ctx match {
  case c if c.asterisk() != null => c.asterisk().accept(this)
  case c if c.udtElem() != null => c.udtElem().accept(this)
  case c if c.LOCAL_ID() != null => buildLocalAssign(ctx)
  case c if c.expressionElem() != null => ctx.expressionElem().accept(this)
  case _ => ir.UnresolvedExpression("Unsupported SelectListElem")
}
```

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -10,9 +10,39 @@ import scala.collection.JavaConverters.{asScalaBufferConverter, collectionAsScal
...
+      case ME => ir.Assign(localId, ir.Subtract(localId, expression))
+      case SE => ir.Assign(localId, ir.Multiply(localId, expression))
+      case DE => ir.Assign(localId, ir.Divide(localId, expression))
+      case MEA => ir.Assign(localId, ir.Mod(localId, expression))
+      case AND_ASSIGN => ir.Assign(localId, ir.BitwiseAnd(localId, expression))
+      case OR_ASSIGN => ir.Assign(localId, ir.BitwiseOr(localId, expression))
+      case XOR_ASSIGN => ir.Assign(localId, ir.BitwiseXor(localId, expression))
+      case _ => ir.UnresolvedExpression(ctx.getText) // Handle unexpected operation types
+    }
+  }
```
vil1, line 45: Same remark as above: this could be fully covered using mockito

vil1 (changes requested):

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
jimidle: I cannot do this as it is forbidden in the style guide. It was actually the way I first implemented it

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
jimidle: Same comment:

https://github.com/databricks/scala-style-guide?tab=readme-ov-file#pattern-matching

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
jimidle: Except I think that down the line, we will weant to distinguish between them if we are linting, so I wanted to leave that there. If we are not ever going to lint, then they can be merged of course.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: Ah - I see what you mean. Force line 43 to be executed. However, it can't really be reached, so I am a bit loathe to spend time on mockito until it definitely gives some benefit.

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
jimidle: I woudl prefer to err in favor of the style guide here.

core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
vil1: TIL

vil1 (approved):
LGTM

sundarshankar89:
@jimidle Adding it to Queue, will be merged once conflicts are resolved.

PR #378, opened 2024-05-22
TSQL: Adds full support for Money, negative constants

jimidle:
Moves negative constants to the lexer, where they are more easily processed.  Performs numeric conversion as per Snowflake, whereby the smallest numeric representation is used but no smaller than integer. We also support Money directly as it is a specific type and is represented as such in the IR.


vil1 (approved):
LGTM

PR #379, opened 2024-05-22
Implement scalar sub queries for expressions

jimidle:
Scalar subqueries return exactly one row with one column and can be used as expressions in SELECT lists and WHERE clauses amongst other things.  

Note that the parser and AST builder do not check to see if the sub-query is actually a scalar.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
@@ -152,4 +152,6 @@ case class JsonAccess(json: Expression, path: Seq[String]) extends Expression {}
 case class Collate(string: Expression, specification: String) extends Expression {}
 case class Iff(condition: Expression, thenBranch: Expression, elseBranch: Expression) extends Expression {}
 
+case class ScalarSubquery(relation: Relation) extends Expression {}
+
 case class Timezone(expression: Expression, timeZone: Expression) extends Expression {}
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
@@ -135,6 +135,10 @@ class TSqlExpressionBuilder extends TSqlParserBaseVisitor[ir.Expression] with Pa
     buildConstant(ctx.con)
   }
 
+  override def visitExprSubquery(ctx: ExprSubqueryContext): ir.Expression = {
+    ir.ScalarSubquery(ctx.subquery().accept(new TSqlRelationBuilder))
+  }
+
   override def visitExprTz(ctx: ExprTzContext): ir.Expression = {
     val expression = ctx.expression().accept(this)
     val timezone = ctx.timeZone.expression().accept(this)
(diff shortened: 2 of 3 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
```diff
@@ -151,3 +151,5 @@ case class ArrayAccess(array: Expression, index: Expression) extends Expression
 case class JsonAccess(json: Expression, path: Seq[String]) extends Expression {}
 case class Collate(string: Expression, specification: String) extends Expression {}
 case class Iff(condition: Expression, thenBranch: Expression, elseBranch: Expression) extends Expression {}
+
+case class ScalarSubquery(relation: Relation) extends Expression {}
```
sundarshankar89, line 155: @jimidle why this is an expression? and also I would rather prefer to have this as type of `relation`

``` select * from table``` here table can be identifier or a subquery or join type.

Comment (jimidle) [reply core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala:155]: It is the "exception (or perhaps expression ;) ) that proves the rule is wrong."

You are referring to derived tables, but this is a scalar subquery, which is an expression over a relation.

Scalar subqueries are an expression yielding a single column and a single row. and here are used as columns. Hence it is found in expression as a select list item, though it is built with the relation builder. 

Subqueries are relations and are built as such if they are, for instance table sources (which is what you are referring to above). Here though, they are select list expressions as in the test. Hence they can be aliased, assigned from and so on and must be expressions to be columns:


```sql
SELECT
     EmployeeID,
     Name,
     (SELECT AvgSalary FROM Employees) AS AverageSalary. -- expression
FROM
      Employees; 
```

```scala
expectedAst = 
        Batch(
          Seq(Project(
            NamedTable("Employees", Map(), is_streaming = false),
            Seq(
              Column("EmployeeID"),
              Column("Name"),
              Alias(
                ScalarSubquery(Project(NamedTable("Employees", Map(), is_streaming = false), Seq(Column("AvgSalary")))),
                Seq("AverageSalary"),
                None))))))
```

When used as a relation in a FROM statement, subqueries generate a derived table:

```sql
SELECT derived.AverageSalary
FROM (
    SELECT AVG(Salary) AS AverageSalary
    FROM Employees
) AS derived;
```

I cannot show the ir for this here because that is waiting on my other PRs all coming together, one of which addresses tableSource.  This branch is waiting for other PRs, which are in turn waiting on others :). However, visitTableSource produces a relation for a derived table, as you would expect.

In short, the ir that applies to columns must also apply to scalar subqueries and they must be expressions here as `Expression` ir will generally not accept `Relation` as an argument (see `Alias` in the ir above for instance).

In short, while TSQL is a relatively simple language, it gets messy in parts.

sundarshankar89: got it

sundarshankar89 (approved):
LGTM

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
vil1: 🤯 that explains why I was wrong to add [this comment](https://github.com/databrickslabs/remorph/blob/03457ef1206d1bb0161cbb523221554d1f7250b5/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4#L3705)

vil1 (approved):
LGTM, thanks for the explanation.

PR #382, opened 2024-05-23
Introduce coverage tests

vil1:
Parses the queries present under `tests/resources/functional/snowflake` and turn them into coverage (acceptance) tests by verifying that `SnowflakeAstBuilder` processes them without exception and without producing trees containing any `Unresolved*` nodes.

Running these tests is made part of a subsequent, non-blocking (ie that doesn't fail the build upon test failure) step of the overall build workflow.

A summary of tests results is published as a comment in the PR, and as a JSON file (produced as an artifact in the build for the moment). Individual failures can be explored in the `checks` section of a PR.


acceptance/src/test/scala/com/databricks/labs/remorph/acceptance/AcceptanceTestRunner.scala
```diff
@@ -0,0 +1,115 @@
...
+  config.testFileSource.listTests.foreach { test =>
+    registerTest(test.testName) {
+      runAcceptanceTest(test)
+    }
+
+  }
+
+}
+
+class SnowflakeAcceptanceSuite
```
nfx, line 110: can you add a reporter for 

https://github.com/databrickslabs/remorph/blob/main/src/databricks/labs/remorph/coverage/commons.py#L21-L35

acceptance/src/test/scala/com/databricks/labs/remorph/acceptance/AcceptanceTestRunner.scala
```diff
@@ -0,0 +1,115 @@
+package com.databricks.labs.remorph.acceptance
```
nfx, line 1: can we rename the module to `coverage`

nfx (approved):

nfx (approved):
Lgtm

PR #383, opened 2024-05-23
TSQL: Add support for *, INSERTED.*, DELETED.* in SELECT <list>

jimidle:
Adds support for select list elements involving `*`:
  
  ```sql
SELECT *
SELECT t.*
SELECT server..t.x.*

DELETE FROM TestTable
   OUTPUT deleted.*
   WHERE ID = 2; 
 ``` 
  
 Also simplifies tableName and removes fullTableName
 
 Relies on: #373

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,2 @@
 ignore:
-  - "**/*.g4"
   - core/src/main/scala/com/databricks/labs/remorph/parsers/FunctionBuilder.scala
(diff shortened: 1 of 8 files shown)


vil1 (changes requested):
I'd prefer if we would avoid using `// $COVERAGE-OFF$` altogether. For this kind of cases that we-know-but-cannot-actually-prove are unreachable, we can write test cases using mockito (see [this example](https://github.com/databrickslabs/remorph/blob/5f988ec7a4476c5e429a014bbabce79850299607/core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilderSpec.scala#L217))

vil1 (approved):
LGTM

PR #384, opened 2024-05-23
Create full coverage for the FunctionBuilder

jimidle:
A cleanup task so that we can remove the file from the ignored files in the coverage configuration and ensure that the function lookup and build code is fully tested.


vil1 (approved):

jimidle:
@sundarshankar89 This should automerge now Sir!

PR #386, opened 2024-05-24
Feature/fix integration issues

ravit-db:
- Fixed the HashQueryBuilder to handle the column order in case of column mapping
- Fixed the Literal conversion in threshold query builder
- Fixed the schema retriever with right param


sundarshankar89 (approved):
LGTM

Issue #387, opened 2024-05-24
[FEATURE]: TSQL support ALL and DISTINCT as modifiers for certain functions


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Some SQL functions all specification of using ALL values of an expression or just the DISTINCT values. This needs to be supported and the function call ir should be modified accordingly

### Proposed Solution

Add ALL and DISTINCT into the expression tree as modifiers with new #labels
Implement visitor function for new labels
Modify right hand expression 

### Additional Context

#275

PR #400, opened 2024-05-28
TSQL: Support DISTINCT in SELECT list and aggregate functions

jimidle:
There are two use cases for DISTINCT, one is in a `SELECT` _list_ such as:

```sql
SELECT DISTINCT a, b, c ....
```

The other is in aggregate functions, which is usually `COUNT` such as:

```sql
SELECT COUNT(DISTINCT x), y, z ...
```

Here we support `DISTINCT <expression>` and `DISTINCT` arguments to aggregate functions.

Note that `ALL` is the default behavior in SQL and therefore ALL causes no changes in the ir from how it was produced prior to this PR.

Closes: #387


vil1 (approved):
LGTM

PR #402, opened 2024-05-28
Reduce builtIn functions to minimum possible use of explicit syntax

jimidle:
We remove explicit specification of the builtIn functions, leaving only those that definitely require a special syntax.


vil1 (approved):
LGTM, but maybe the PR title (and hence the release message that will ultimately get generated out of it) could be improved.

jimidle:
> LGTM, but maybe the PR title (and hence the release message that will ultimately get generated out of it) could be improved.

Actually, I will mark this as internal as how we process functions isn't important to the outside world I think?

PR #405, opened 2024-05-29
Implement Lakeview Dashboard Publisher

bishwajit-db:
Implement Lakeview dashboard Publisher


src/databricks/labs/remorph/helpers/dashboard_publisher.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, ws: WorkspaceClient, installation: Installation):
+        self._ws = ws
+        self._installation = installation
+
+    def create(self, dashboard_file: Path, parameters: dict[str, Any] | None = None) -> ObjectInfo:
+        if parameters is None:
+            parameters = {}
+
+        dashboard_data = self._substitute_params(dashboard_file, parameters)
+        self._installation.upload(dashboard_file.name, dashboard_data)
```
vijaypavann-db, line 20: I think we should add `try` and `except` blocks since we're interacting with a remote DB workspace.

tests/unit/helpers/test_dashboard_publisher.py
```diff
@@ -0,0 +1,45 @@
...
+from databricks.sdk import WorkspaceClient
+from databricks.sdk.service.workspace import ObjectInfo, ObjectType
+
+from databricks.labs.remorph.helpers.dashboard_publisher import DashboardPublisher
+
+
+def test_create_dashboard():
+    installation = MockInstallation(is_global=False)
+    workspace_client = create_autospec(WorkspaceClient)
+    dashboard_file = Path(__file__).parent / Path('../../resources/Remorph-Reconciliation.lvdash.json')
```
vijaypavann-db, line 14: As per this [link](https://stackoverflow.com/questions/6028000/how-to-read-a-static-file-from-inside-a-python-package), we shouldn't use `__file__`. 

```
Bad ways to avoid:
Bad way #1: using relative paths from a source file
This was previously described in the accepted answer. At best, it looks something like this:

from pathlib import Path

resource_path = Path(__file__).parent / "templates"
data = resource_path.joinpath("temp_file").read_bytes()
What's wrong with that? The assumption that you have files and subdirectories available is not correct. This approach doesn't work if executing code which is packed in a zip or a wheel, and it may be entirely out of the user's control whether or not your package gets extracted to a filesystem at all.

```

sundarshankar89: This is ok for test cases. all are valid if this is being used inside our core lib.

src/databricks/labs/remorph/helpers/dashboard_publisher.py
sundarshankar89: @bishwajit-db can you fix this.

bishwajit-db: IMO, we can add retry logic based on databricks.sdk.retries.retried, but we should keep the exception handling to the calling code. Handling the exception here will either swallow it or duplicate handling in both here and the caller. I plan to add that to the installer. What do you say, @sundarshankar89 @vijaypavann-db?

tests/unit/helpers/test_dashboard_publisher.py
bishwajit-db: Done. Used importlib.

sundarshankar89:
LGTM

vijaypavann-db (approved):

Issue #406, opened 2024-05-29
[FEATURE]: TSQL Implement coverage for TSQL as per Snowflake


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Transpile

### Problem statement

Th coverage module is currently Snowflake centered and needs to be either adjusted to accommodate TSQL, or indeed any future SQL variants. 

### Proposed Solution

Expand the existing code to include a TSQL test suite, and adjust the instantiation of parser/translator source  accordingly.

PR #408, opened 2024-05-29
Create coverage suite for TSQL

jimidle:
Creates coverage tests for TSQL, in the same manner as the coverage for Snowflake.

Coverage is a non-blocking test suite that allos us to measure how far along the path of syntax coverage we are. It captures parsing errors and checks the IR to see if contains any UnresolvedXXX, which indicates something that we do not currently cover. 

Closes: #406


jimidle:
@vil1 I am not sure how I managed to create this branch from my other PR branch and therefore have it include those changes too. Resubmitting.

PR #409, opened 2024-05-29
feat: Implement coverage tests for TSQL

jimidle:
Creates coverage tests for TSQL, in the same manner as the coverage for Snowflake.

Coverage is a non-blocking test suite that allows us to measure how far along the path of syntax coverage we are, and how much of the syntax we are able to convert to DBSQL IR. 

We capture parsing errors and check the IR to see if contains any UnresolvedXXX, which flags constructs that we do not currently cover.

Closes: #406


vil1 (approved):
Jolly good! 👍 

We'll also want to add a step to the `coverage-tests` job to run coverage with `-d tsql` (in `push.yml`) but that can come later.

PR #413, opened 2024-05-30
Added validation for join columns for all query builders and limiting rows for reports

ganeshdogiparthi-db:


src/databricks/labs/remorph/reconcile/query_builder/base.py
```diff
@@ -122,3 +126,10 @@ def _get_transform(datatype: str):
                 transform = _get_transform(schema_dict.get(column_name, column_name))
                 return transform_expression(node, transform)
         return node
+
+    def _validate(self, field: set[str] | list[str] | None, message: str):
+        if field is None:
+            logger.error(f"Exception {message}")
```
sundarshankar89, line 132:
```suggestion
            logger.error(f"Exception for {self.table_conf.target_name} target table in {self.layer} layer --> {message}")
```

sundarshankar89 (changes requested):

sundarshankar89 (approved):
LGTM

PR #415, opened 2024-05-30
TSQL: Create coverage tests for TSQL -> Databricks functions 

jimidle:
As we now have translation capability for them, we start out by examining the TSQL functions and providing the equivalent Databricks SQL for them.

  - Some functions, mainly the common ANSI defined SQL functions are directly equivalent. 
  - Some have the same name but mean different things.
  - Some functions have equivalent ways to achieve the same thing.
  - Some functions will likely be unsupportable without changes to Databricks SQL.
  - Some functions may be unsupportable due to fundamental differences between TSQL and Databricks SQL engines/models.

### DIFF
--- a/tests/resources/functional/tsql/test_abs/test_abs_1.sql
+++ b/tests/resources/functional/tsql/test_abs/test_abs_1.sql
@@ -1,0 +1,6 @@
+
+-- tsql sql:
+SELECT abs(col1) AS abs_col1 FROM tabl;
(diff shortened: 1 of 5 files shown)


jimidle:
Rather than adding possibly hundreds of commits to this one PR, I will open new ones to add more coverage

vil1 (approved):
LGTM

PR #420, opened 2024-05-31
Adds more coverage tests for functions to TSQL coverage

jimidle:
More functions are added to the set of coverage tests for TSQL conversions

### DIFF
--- a/tests/resources/functional/tsql/functions/test_checksum_agg/test_checksum_agg_1.sql
+++ b/tests/resources/functional/tsql/functions/test_checksum_agg/test_checksum_agg_1.sql
@@ -1,0 +1,10 @@
+## CHECKSUM_AGG
+
+There is no direct equivalent of CHECKSUM_AGG in Databricks SQL. The following
+conversion is a suggestion and may not be perfectly functional.
+
+-- tsql sql:
+SELECT CHECKSUM_AGG(col1) FROM t1;
+
+-- databricks sql:
+SELECT MD5(CONCAT_WS(',', COLLECT_LIST(col1))) FROM t1;
--- a/tests/resources/functional/tsql/functions/test_count/test_count_1.sql
+++ b/tests/resources/functional/tsql/functions/test_count/test_count_1.sql
@@ -1,0 +1,9 @@
+## COUNT
+
+The TSQl COUNT function and the DataBricks COUNT function are equivalent.
+
+-- tsql sql:
+SELECT COUNT(*) FROM t1;
+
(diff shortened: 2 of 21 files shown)


tests/resources/functional/tsql/functions/test_checksum_agg/test_checksum_agg_1.sql
```diff
@@ -0,0 +1,10 @@
+## CHECKSUM_AGG
+
+There is no direct equivalent of CHECKSUM_AGG in Databricks SQL. The following
```
nfx, line 3: @vil1 do we allow free-form text here?.. 

@jimidle could these files be "mostly valid SQL"?

tests/resources/functional/tsql/functions/test_checksum_agg/test_checksum_agg_1.sql
```diff
@@ -0,0 +1,10 @@
+## CHECKSUM_AGG
```
nfx, line 1: we shouldn't have one file per folder, as it'll get very difficult to support with the amount of verifications we want:
- tests/resources/functional/tsql/functions/test_checksum_agg/test_checksum_agg_1.sql
- tests/resources/functional/tsql/functions/test_count/test_count_1.sql
- ...

can we semantically group these files into folders? e.g.
- tests/resources/functional/tsql/functions/checksum_agg_1.sql
- tests/resources/functional/tsql/functions/count_1.sql
- ...

nfx (approved):

tests/resources/functional/tsql/functions/test_checksum_agg/test_checksum_agg_1.sql
jimidle: In the prelude I have started putting markdown with intent of auto producing a mini guide as we add functions, but otherwise it is free-form. That could of course use SQL comments, which would then be removed before rendering markdown. Either way is fine. Using SQl comments would mean that IDES etc can parser it as pure SQL - however, the IDE requires a dialect, so when I select TSQL, it flags a lot of Datarbicks SQL and vice-versa

tests/resources/functional/tsql/functions/test_checksum_agg/test_checksum_agg_1.sql
jimidle: I would also prefer not to have so many subdirectories. I merely copied the existing Snowflake layout to keep it the same.

PR #430, opened 2024-06-04
TSQL: Implement functions with specialized syntax

jimidle:
Implement translation of the following functions and pseudo-functions:

```tsql
NEXT VALUE FOR sequence
```

WIP

Progresses: #275


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlFunctionBuilder.scala
```diff
@@ -8,8 +8,10 @@ class TSqlFunctionBuilder extends FunctionBuilder {
     case "@@CURSOR_STATUS" => FunctionDefinition.notConvertible(0)
     case "@@FETCH_STATUS" => FunctionDefinition.notConvertible(0)
     // The ConversionStrategy is used to rename ISNULL to IFNULL
-    case "ISNULL" => FunctionDefinition.standard(2).withConversionStrategy(TSqlFunctionConverters.FunctionRename)
+    case "ISNULL" => FunctionDefinition.standard(2).withConversionStrategy(TSqlFunctionConverters.Rename)
     case "MODIFY" => FunctionDefinition.xml(1)
+    case "NEXTVALUEFOR" =>
+      FunctionDefinition.standard(1).withConversionStrategy(TSqlFunctionConverters.NextValueFor)
   }
```
nfx, line 16:
```suggestion
    case "NEXTVALUEFOR" => FunctionDefinition.standard(1).withConversionStrategy(nextValueFor)
  }
  
  private def nextValueFor(irName: String, args: Seq[ir.Expression]): ir.Expression = {
      // Note that this conversion assumes that the CREATE SEQUENCE it references was an increment in ascending order.
      // We may run across instances where this is not the case, and will have to handle that as a special case, perhaps
      // with external procedures or functions in Java/Scala, or even python.
      // For instance a SequenceHandler supplied by the user.
      //
      // Given this, then we use this converter rather than just the simple Rename converter.
      // TODO: Implement external SequenceHandler?
      ir.CallFunction("MONOTONICALLY_INCREASING_ID", List.empty)
    }

```

more about the pattern: conversion strategies are mostly Single Method Interfaces (SMIs), that can be represented as `(String, Seq[ir.Expression]) => ir.Expression` type, so it would improve readability if we don't add extra static final singletons (`object`) just to have one method. In Java8+ it would have been called an anti-pattern. It's definitely anti-pattern in Scala.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlFunctionBuilder.scala
vil1: Moreover, if a trait/abstract class has a single abstract method like this:
```scala
trait Foo{
  def bar(x: X, y: Y): Z
}
``` 

It can be instantiated with simply:
```scala
val foo: Foo = (x, y) => computeSomeZ(x, y)
```

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
```diff
@@ -170,3 +170,10 @@ case class Timezone(expression: Expression, timeZone: Expression) extends Expres
 case class Money(value: Literal) extends Expression {}
 
 case class WithinGroup(expression: Expression, order: Seq[SortOrder]) extends Expression {}
+
+// Note that Databricks SQL supports FILTER() used as an expression.
+case class FilterExpr(input: Seq[Expression], lambdaFunction: LambdaFunction) extends Expression {}
+case class ValueArray(expresssions: Seq[Expression]) extends Expression {}
```
vil1, line 176:
```suggestion
case class ValueArray(expressions: Seq[Expression]) extends Expression {}
```

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -405,4 +405,90 @@ class TSqlExpressionBuilder(functionBuilder: FunctionBuilder)
...
+    val (keys, values) = ctx
+      .getOrElse(List.empty)
+      .map { keyValueContext =>
+        val expressions = keyValueContext.expression().asScala.toList
+        (expressions.head.accept(this), expressions(1).accept(this))
+      }
+      .unzip
+
+    ir.NamedStruct(keys, values)
+  }
```
vil1, line 439:
```suggestion
  override def visitJsonObject(ctx: JsonObjectContext): ir.Expression = {
    val jsonKeyValues = Option(ctx.jsonKeyValue()).map(_.asScala).getOrElse(Nil)
    val namedStruct = buildNamedStruct(jsonKeyValues)
    val absentOnNull = checkAbsentNull(ctx.jsonNullClause())
    buildJsonObject(namedStruct, absentOnNull)
  }

  private def checkAbsentNull(ctx: JsonNullClauseContext): Boolean = {
    ! Option(ctx).exists(_.ABSENT() != null)
  }

  private def buildNamedStruct(ctx: Seq[JsonKeyValueContext]): ir.NamedStruct = {
    val (keys, values) = ctx
      .map { keyValueContext =>
        val expressions = keyValueContext.expression().asScala.toList
        (expressions.head.accept(this), expressions(1).accept(this))
      }
      .unzip

    ir.NamedStruct(keys, values)
  }
```

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -405,4 +405,90 @@ class TSqlExpressionBuilder(functionBuilder: FunctionBuilder)
...
+   * @param args
+   *   the list of expressions yield JSON values
+   * @param absentOnNull
+   *   whether we should remove NULL values from the JSON array
+   * @return
+   *   IR for the JSON_ARRAY function
+   */
+  private[tsql] def buildJsonArray(args: Seq[ir.Expression], absentOnNull: Boolean): ir.Expression = {
+    if (absentOnNull) {
+      val lambdaVariable = ir.UnresolvedNamedLambdaVariable(Seq("x"))
```
vil1, line 459: Currently, acceptance tests verify that the output tree doesn't contain any node with a class name starting with "Unresolved". Going forward in improving acceptance tests success rate, we'll have to either rename `UnresolvedNamedLambdaVariable` to, say, `NotYetResolvedNamedLambdaVariable` or, better, make acceptance tests more specific about the `Unresolved*` classes they don't allow in output trees.

jimidle: Yes - I also spotted that. But this is pre-declared IR, so I could not rename it.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: While a good idea, the code:

```scala
  private def checkAbsentNull(ctx: JsonNullClauseContext): Boolean = {
    ctx match {
      case null => true
      case clause if clause.ABSENT() == null => false
      case _ => true
    }
  }
  ```
  
  Is required because:
  
    - If the clause does not exist, the ABSENT ON NULL is assumed - so true (case null prevents null access)
    - If the clause exists and ABSENT == null then the clause was NULL ON NULL - so false
    - If the clause exists and ABSENT != null then the clause was ABSENT ON NULL - so true
    
    Your code:
    
 ```scala
    private def checkAbsentNull(ctx: JsonNullClauseContext): Boolean = {
    ! Option(ctx).exists(_.ABSENT() != null)
  }
```
 
 Does not implement the same logic and it took me a minute to see why. I will add the above as a comment.

vil1 (approved):

jimidle:
@nfx - just needs you to approve your requested changes to merge this.

nfx (approved):
Lgtm

PR #431, opened 2024-06-04
Use default_factory in recon_config.py

derekyidd:
With Python version 3.11, when trying to initialize the config classes, we are getting ValueError: mutable default <class 'databricks.labs.remorph.reconcile.recon_config.StatusOutput'> for field status is not allowed: use default_factory

This change addresses this error by using default_factory.


nfx (approved):
Lgtm

sundarshankar89 (approved):

PR #433, opened 2024-06-05
Snowflake add missing function definitions

vil1:
Add entries in `SnowflakeFunctionBuilder` for functions that were used in the acceptance tests but not "defined" yet.

Note that the grammar around `function_call` has been reworked, this work is still in progress. Future work will include:
- completing the definition of `builtin_function` (functions that have special syntax)
- improving implementation around `aggregate_function` and `ranking_windowed_function`


jimidle:
I would like to merge #430 before we merge this so we don't start repeating the same edits. That PR finishes all functions except for the more exotic stuff like freeText. You can then use the TSQL grammar changes as a template without worrying I will change them half way through your own re-work on Snowflake.

vil1:
> I would like to merge #430 before we merge this so we don't start repeating the same edits. That PR finishes all functions except for the more exotic stuff like freeText. You can then use the TSQL grammar changes as a template without worrying I will change them half way through your own re-work on Snowflake.

What edits do you refer to? the one in `FunctionBuilder`?

nfx (approved):
Lgtm

PR #437, opened 2024-06-06
Grammar clean up and keyword expansion

jimidle:
An internal cleanup of grammar and keywords  

  - Removes superfluous comments
  - Corrects hacks with "" strings
  - Expands keywords - for instance SELECT * FROM TABLE now works
  - Removes unused accumulated rules from TSQL


vil1 (approved):
👍

PR #443, opened 2024-06-07
TSQL: Implement WITH CTE

jimidle:
We implement CTE support for TSQL as in the following example:

```tsql
WITH cteTable (col1, col2, col3count)
AS
    SELECT col1, fred, COUNT(OrderDate) AS counter
    FROM Table1
    WHERE fred IS NOT NULL
)
SELECT col2, col1, col3
FROM cteTable
```


nfx (approved):
Lgtm

PR #445, opened 2024-06-07
Lose superflous alt in selectListElem

jimidle:
Removes superfluous alts defined in select list element rule in TSQL grammar. Also removes the accepts on it which were rightly not used.

Note that the rule for parsing udt column declarations may still be relevant for `CREATE TYPE`, hence it is not removed.


nfx (approved):
lgtm

PR #449, opened 2024-06-10
Snowflake:: Snowflake grammar clean up

jimidle:
Change rule names and Context names to camelCase from snake_case.

While this appears to be a huge PR, it is strictly confined to name changes so a review shoudl not take long.


vil1 (approved):
YES!

PR #455, opened 2024-06-11
TSQL: Improve function coverage

jimidle:
Improves TSQL coverage by adding missing functions to the relevant tables and sharing common functions between dialects.


core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
```diff
@@ -476,11 +494,31 @@ class FunctionBuilderSpec extends AnyFlatSpec with Matchers with TableDrivenProp
   "buildFunction" should "Apply known TSql conversion strategies" in {
     val functionBuilder = new TSqlFunctionBuilder
 
-    val result1 = functionBuilder.buildFunction("ISNULL", Seq(ir.Column("x"), ir.Literal(integer = Some(0))))
-    result1 match {
+    var result = functionBuilder.buildFunction("ISNULL", Seq(ir.Column("x"), ir.Literal(integer = Some(0))))
```
vil1, line 497: Please avoid using `var`.

A more practical approach would be to define a table property check:
```scala
val functions = Table(
  ("inFunctionName", "outFunctionName"),
  ("GET_BIT", "GETBIT"),
  // etc.
)
forAll(functions){(inFn, outFn) =>
  functionBuilder.buildFunction(inFn) match {
    case f:ir.CallFunction => f.function_name shouldBe outFn
    case _ => fail(s"$inFn TSql conversion failed")
  }
}
```

Also, there's a typo on line 520 (><)'

vil1 (changes requested):
please see my comment about `var`

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
jimidle: OK - yeah I guess that is better. In tests I am usually just looking to get it going but example is not implemented in that Spec I think

jimidle: Change - please approve changes.

vil1 (approved):
lgtm

PR #456, opened 2024-06-11
Better coverage reports

vil1:
Previously, tsql coverage report wasn't included in the `coverage-report.json` artifact produced by the build (though they were correctly included in the `Coverage tests results` comment), this PR fixes it.


tests/resources/functional/tsql/functions/test_get_bit_1.sql
```diff
@@ -5,5 +5,5 @@
 -- tsql sql:
 SELECT GET_BIT(42, 7);
 
--- GET_BIT sql:
+-- databricks sql:
```
jimidle, line 8: This is already fixed in my own PR. But I guess it will merge OK.

PR #460, opened 2024-06-12
Add lexer tests for TSql and Snowflake lexers

jimidle:
Adds lexer test Specs for both TSql and Snowflake lexers.

Here we clean up a lot of things in both lexers:

  - Quoted ID does not span multiple lines
  - Quoted ID correctly matches embedded "", though it is inadvisable.
     *Note* that the lexer preserves the text as it is in the source including the ""
  - STRING accepts escaped characters and not just ''
  - A single ' as a token is removed and the grammar rules that tried to use
    them have been removed or corrected
  - Attempts to lex URIs and enforce the normative spec lexicaly and syntactically
     have been removed.
  - Incorrect attempts to distinguish (incorrectly) storage options have been removed
     *Note* that we will soon remove all the keywords that are over-specified in the lexer and parser
     in favor of a keyword checking helper used when building the IR.
  - Configuration for antlr-maven-plugin is changed to not generate Listeners we do not use
  - Configuration for antlr-maven-plugin is changed to treat warnings as errors so that we do not check
    in parser grammar that cannot parse.

### DIFF
--- a/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
+++ b/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
@@ -1119,25 +1119,19 @@ LISTAGG: 'LISTAGG';
 
 DUMMY: 'DUMMY'; //Dummy is not a keyword but rules reference it. As to be cleaned.
 
-SPACE: [ \t\r\n]+ -> channel(HIDDEN);
+SPACE: [ \t\r\n]+ -> skip;
 
 SQL_COMMENT    : '/*' (SQL_COMMENT | .)*? '*/' -> channel(HIDDEN);
 LINE_COMMENT   : '--' ~[\r\n]*                 -> channel(HIDDEN);
 LINE_COMMENT_2 : '//' ~[\r\n]*                 -> channel(HIDDEN);
 
 // TODO: ID can be not only Latin.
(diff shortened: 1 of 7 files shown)


core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
```diff
@@ -1117,31 +1117,24 @@ VARIANT          : 'VARIANT';
...
-SPACE: [ \t\r\n]+ -> channel(HIDDEN);
+SPACE: [ \t\r\n]+ -> skip;
 
 SQL_COMMENT    : '/*' (SQL_COMMENT | .)*? '*/' -> channel(HIDDEN);
 LINE_COMMENT   : '--' ~[\r\n]*                 -> channel(HIDDEN);
 LINE_COMMENT_2 : '//' ~[\r\n]*                 -> channel(HIDDEN);
 
 // TODO: ID can be not only Latin.
-DOUBLE_QUOTE_ID    : '"' ~'"'+ '"';
+DOUBLE_QUOTE_ID    : '"' ~[\r\n"]+ '"';
```
vil1, line 1129: This rule needs to be LL(2) in order to properly capture escaped double quotes (`""`).

vil1 (changes requested):
I added a check in the lexer specs which shows that `DOUBLE_QUOTE_ID` doesn't properly handle `""`.

vil1 (approved):
... and fixed the tests

PR #461, opened 2024-06-12
Simplify the lexers and parsers by removing all unnecessary lexer tokens 

jimidle:
All the overdone keyword tokens and over-specified option grammar rules are being gradually removed to be replaced with simple rules that look for XXX (= expression|ON|OFF|etc) and just pass all of them to the parse tree walkers. 

Grammar is easier to maintain and will also result in smaller and faster parsers.


vil1 (approved):
LGTM

PR #467, opened 2024-06-13
fix: Change order of search for "" in quoted lexer rule.

jimidle:
As the lexer uses an NFA, it will match the first alt when either could match. I will leave the " in the alt explicit.


vil1 (approved):
LGTM

PR #249, opened 2024-04-16
Added `Generate Recon Config` support to `databricks labs remorph generate-recon-config` cli command

vijaypavann-db:
## Changes

Added CLI command `databricks labs remorph generate-recon-config`  to generate Recon Config for Data Reconciliation.

Logic:
-  This command prompts user for the following: 
      * Source
      * Secret Scope
      * Source, Target Catalog and Schema
      * include/exclude a set of tables
- Based on the provided information it will generate `recon-config-file`
- This feature is intended to enhance our tool's capabilities in generating the `recon-config-file` dynamically which will eventually minimise the unintended manual errors and streamline the process of prompting and saving configuration details. 

- Sample `recon-config-file`

```
{
  "source_catalog": "catalog",
  "source_schema": "schema",
  "tables": [
    {
      "column_mapping": None,
      "drop_columns": None,
      "filters": None,
      "jdbc_reader_options": None,
      "join_columns": None,
      "select_columns": None,
      "source_name": "table1",
      "target_name": "table1",
      "thresholds": None,
      "transformations": None
    },
    {
      "column_mapping": None,
      "drop_columns": None,
      "filters": None,
      "jdbc_reader_options": None,
      "join_columns": None,
      "select_columns": None,
      "source_name": "table2",
      "target_name": "table2",
      "thresholds": None,
      "transformations": None
    },
    {
      "column_mapping": None,
      "drop_columns": None,
      "filters": None,
      "jdbc_reader_options": None,
      "join_columns": None,
      "select_columns": None,
      "source_name": "table3",
      "target_name": "table3",
      "thresholds": None,
      "transformations": None
    }
  ],
  "target_catalog": "tgt_catalog",
  "target_schema": "tgt_schema"
}
```

### Linked issues
Resolves 
https://github.com/databrickslabs/remorph/issues/209

### Functionality
- [x]  added new CLI command

### Tests
- [x]  manually tested
- [x]  added unit tests

### TODO

- [ ]  add integration tests
- [ ]  verify on staging environment (attach screenshot )

PR #469, opened 2024-06-17
Generate recon config for oracle data source

vijaypavann-db:
Merge priority order: 
* #249 
*  and then this.

### DIFF
--- a/src/databricks/labs/remorph/helpers/reconcile_utils.py
+++ b/src/databricks/labs/remorph/helpers/reconcile_utils.py
@@ -263,7 +263,7 @@ class ReconcileConfigUtils:
             if filter_type != "all":
                 proceed_run_prompt = (
                     f"Would you like to run reconciliation `{filter_type[:-1]}ing` "
-                    f" {','.join(subset_tables_list)} tables? "
+                    f" ({', '.join(subset_tables_list)}) tables? "
                 )
 
             proceed_run = self._prompts.confirm(proceed_run_prompt)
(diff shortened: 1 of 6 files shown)


src/databricks/labs/remorph/helpers/reconcile_utils.py
```diff
@@ -91,3 +100,366 @@ def run(self):
         logger.info(f"\nReconcile job started. Please check the job_url `{job_run_url}` for the current status.")
 
         return True
+
+
+class ReconcileConfigUtils:
```
nfx, line 105: revert the move of this class, as it's not clear what you've changed and it's not possible to review. don't mix refactoring with feature addition

src/databricks/labs/remorph/reconcile/connectors/oracle.py
```diff
@@ -91,3 +95,31 @@ def _get_timestamp_options() -> dict[str, str]:
...
+            logger.info(f" Executing query: {tables_query}")
+            tables_df = self.reader(tables_query).load()
+
+            tables_list = [
+                Table(source_name=row.table_name.lower(), target_name=row.table_name.lower())
+                for row in tables_df.collect()
+            ]
+            return build_table_recon(catalog, schema, tables_list)
+        except (RuntimeError, PySparkException) as e:
+            return self.log_and_throw_exception(e, "list_tables", tables_query)
```
nfx, line 125: make this method private

labs.yml
```diff
@@ -51,3 +51,5 @@ commands:
         description: Directory to store the generated lineage file
   - name: configure-secrets
     description: Utility to setup Scope and Secrets on Databricks Workspace
+  - name: generate-recon-config
```
nfx, line 54: why not calling it `configure-recon`?

nfx (changes requested):
don't mix refactoring and feature addition in the same PR, it's impossible to review what the increment is.

PR #471, opened 2024-06-17
Improve id definition and translations

vil1:
This somewhat involved PR improves the definition, usage and translation of the idea of id (pun intended).

An id may be quoted as in `SELECT * FROM "My Table"`, in which case it is to be resolved in a case sensitive manner. In some dialects (eg. TSQL), several delimiters are allowed (making `"My Table"` and `[My Table]` refer to the same object).

These two reasons makes it incorrect to represent ids with mere strings. This PR introduces two new AST constructs: 
- `Id` that represents a single id along with its case sensitivity (and with delimiters striped),
- `ObjectReference` that consists in a non empty list of ids. 

Other definitions that previously relied on ids being represented as mere strings have been modified accordingly, except for `NamedTable`'s `unparsed_identifier` (due to the intent the field name was indicating). @nfx, is it OK to change this field to an `ObjectReference`? (that can be done in a subsequent PR as well).


jimidle (changes requested):
Changing  to id (COMMA id)* liekly needs some tweaks to the visitors

jimidle (approved):
lgtm

PR #477, opened 2024-06-20
Transpile Presto `cast as JSON` to `cast as String`

ganeshdogiparthi-db:


src/databricks/labs/remorph/snow/presto.py
```diff
@@ -38,6 +39,13 @@ class Parser(presto.Parser):
...
         FUNCTIONS = {
+            **presto.Parser.FUNCTIONS,
             "APPROX_PERCENTILE": _build_approx_percentile,
             "STRPOS": locate_to_strposition,
         }
+
+    class Tokenizer(presto.Tokenizer):
+        KEYWORDS = {
+            **presto.Tokenizer.KEYWORDS,
+            "JSON": TokenType.TEXT,
```
vijaypavann-db, line 50: what would happen in the below cases: 
- SELECT CAST(ARRAY[1, 23, 456] AS JSON); -- JSON '[1,23,456]'
- SELECT CAST(ARRAY[1, NULL, 456] AS JSON); -- JSON '[1,null,456]'
- SELECT CAST(ARRAY[ARRAY[1, 23], ARRAY[456]] AS JSON); -- JSON '[[1,23],[456]]'
- SELECT CAST(MAP_FROM_ENTRIES(ARRAY[('k1', 1), ('k2', 23), ('k3', 456)]) AS JSON); -- JSON '{"k1":1,"k2":23,"k3":456}'
- SELECT CAST(CAST(ROW(123, 'abc', true) AS ROW(v1 BIGINT, v2 VARCHAR, v3 BOOLEAN)) AS JSON); -- JSON '[123,"abc",true]'

As per the above change, all will be converted to `STRING`, which I don't think is correct. 
I believe it's specific to only your project usecase and the change shouldn't impact the default `presto` behavior.

vijaypavann-db (commented):
Need more discussion

src/databricks/labs/remorph/snow/presto.py
ganeshdogiparthi-db: Everything works in Databricks. The idea here is that the cast( as JSON) counterpart in Databricks will be TO_JSON, but TO_JSON expects a struct col in Databricks, which is not the case in presto. So, ideally, I believe this should be valid.
image

vijaypavann-db (approved):

Issue #357, opened 2024-05-17
[FEATURE]: TSQL Implement remaining non-standard functions


jimidle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

The TSQL remorph now handles all the standard format functions `func LPAREN (e (COMMA e)*)` but there are a number of functions that are reasonably common that have non-standard syntax. they are not really functions in the normal sense of the word, but are expressed syntactically as if they were. 

### Proposed Solution

 - Implement visitors and possibly a specialized function builder.
 - Extended standard function builder with new entries for functionarity when possible
 - Implement over a number of PRs for the different function types

### Additional Context

#275

PR #487, opened 2024-06-24
feat: Implement freetext function visitor, simplify grammar

jimidle:
Freetext functions, $PARTITION and HIERARCHYID functions are unsupported in Databricks SQL, but we must still support generation of IR to indicate this. We also  therefore prevent errors when visiting functionCall.

Here we complete handling of all TSQL functions from a parsing and IR generating point of view. There are likely to be many conversions of functions and flagging of unsupported functions as the coverage tests are expanded to support them.

Progresses: #275 
Closes: #357

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
@@ -433,6 +433,12 @@ class TSqlExpressionBuilder() extends TSqlParserBaseVisitor[ir.Expression] with 
     functionBuilder.buildFunction(ctx.f.getText, List.empty)
   }
 
+  override def visitHierarchyidStaticMethod(ctx: HierarchyidStaticMethodContext): ir.Expression = {
+    // Databricks SQL does not support HIERARCHYID functions, so there is no point in trying to convert these
+    // functions. We do need to generate IR that indicates that this is a function that is not supported.
+    functionBuilder.buildFunction("HIERARCHYID", List.empty)
+  }
+
   // format: off
   /**
    * Check if the ABSENT ON NULL clause is present in the JSON clause. The behavior is as follows:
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
```diff
@@ -392,6 +392,11 @@ class TSqlExpressionBuilder() extends TSqlParserBaseVisitor[ir.Expression] with
     ctx.expression().accept(this)
   }
 
+  override def visitPartitionFunction(ctx: PartitionFunctionContext): ir.Expression = {
+    // $$PARTITION is not supported in Databricks SQL, so we will report it is not supported
+    functionBuilder.buildFunction(s"$$PARTITION", List.empty)
```
vil1, line 397:
```suggestion
    functionBuilder.buildFunction("$PARTITION", List.empty)
```

core/src/test/scala/com/databricks/labs/remorph/parsers/tsql/TSqlFunctionBuilderSpec.scala
```diff
@@ -18,6 +17,7 @@ class TSqlFunctionBuilderSpec extends AnyFlatSpec with Matchers with TableDriven
       ("functionName", "expectedArity"), // Header
 
       // TSql specific
+      (s"$$PARTITION", Some(FunctionDefinition.notConvertible(0))),
```
vil1, line 20:
```suggestion
      ("$PARTITION", Some(FunctionDefinition.notConvertible(0))),
```

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala
jimidle: I cannot do this as the Scala compiler incorrectly complains that "$PARTITION" is a an unresolved interpolation. It is apparently a known issue and I had to do quite a bit of searching to work out why it was saying that. It only seems to happen when the string is the first parameter of a function.

core/src/test/scala/com/databricks/labs/remorph/parsers/tsql/TSqlFunctionBuilderSpec.scala
jimidle: This does not seem to give the error, but given it gives it elsewhere, I thought it better to be consistent.

vil1 (approved):

PR #488, opened 2024-06-25
Snowflake driver and user transformation to partition & join col

ganeshdogiparthi-db:
* Reverted the code to apply only user transformations to key cols and join cols in `Hash query` and `Threshold source query`
* In the compare class we use `null safe equality join` for `hash comparison`.
* Created a map entry for snowflake driver in `jdbc_reader.py`


sundarshankar89 (approved):
LGTM

vijaypavann-db:
Is `SnowflakeDriver`: `net.snowflake.client.jdbc.SnowflakeDriver`  available on all clusters by default, or do we need to install something similar to the `ojdbc8` jar?

ganeshdogiparthi-db:
@vijaypavann-db 
It is available already in our cluster. I have never installed it manually.

ganeshdogiparthi-db:
@sundarshankar89  Can we close this please?

vijaypavann-db (approved):
LGTM

PR #489, opened 2024-06-25
Snowflake, improve coverage test success rate

vil1:


core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3227,7 +3229,9 @@ functionCall
     ;
 
 builtinFunction
-    : trim = (TRIM | LTRIM | RTRIM) L_PAREN expr (COMMA string)? R_PAREN # builtinTrim
+    : trim = (TRIM | LTRIM | RTRIM) L_PAREN expr (COMMA string)? R_PAREN                 # builtinTrim
+    | EXTRACT L_PAREN part = (STRING | ID) FROM expr R_PAREN                             # builtinExtract
+    | ARRAY_AGG L_PAREN DISTINCT? expr R_PAREN (WITHIN GROUP orderByClause)? overClause? # builtinArrayAgg
```
jimidle, line 3234: I folded the WITHIN GROUP modifier into [expression in TSQL](https://github.com/databrickslabs/remorph/blob/main/core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4#L2909), then the visitor deals with the IR for it. Then ARRAY_AGG can be removed as a token and it becomes a standardFunction. That means that WITHIN GROUP can be applied to anything, but we are expecting verified syntax for the conversion. If we ever do not, then we can add a basic semantic verification pass to reject impossible constructs.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
```diff
@@ -348,6 +351,20 @@ class SnowflakeExpressionBuilder()
...
     functionBuilder.buildFunction(ctx.trim.getText, expression :: characters)
   }
+
+  override def visitBuiltinExtract(ctx: BuiltinExtractContext): ir.Expression = {
+    val part = ir.Id(removeQuotes(ctx.part.getText))
+    val date = ctx.expr().accept(this)
+    functionBuilder.buildFunction(ctx.EXTRACT().getText, Seq(part, date))
+  }
+
+  override def visitBuiltinArrayAgg(ctx: BuiltinArrayAggContext): ir.Expression = {
```
jimidle, line 361: By folding WITHIN GROUP in to [the expression](https://github.com/databrickslabs/remorph/blob/main/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlExpressionBuilder.scala#L377), then this visitor can be deleted along with the rule.

jimidle (changes requested):
A few tweaks suggested to simplify the grammar and visitors.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
jimidle: That said, maybe check that the syntax is the same as in TSQL.

jimidle (approved):
Nice

PR #496, opened 2024-06-26
Implement TSQL FOR and OPTION clauses as part of SELECT and other statements

jimidle:
TSQL has clauses FOR and OPTION for SELECT, and later other statements such as INSERT. 

The FOR clause has no equivalent in Databricks SQL, but we must visit the clause anyway, even though we build no IR for it. Later, it could be a good idea to report this via the Error handler as unsupported (not Unresolved), along with functions and other things that we are unable to convert to anything sensible. 

The OPTION clause provides TSQL with optimization hints for various queries. We support parsing it and do build the options as they may be useful to add as comments to the generated converted query as useful hints in the migration process. In general, it is better to leave optimization to the Catalyst optimizer. However, as options can be set on the optimizer, knowing what hints were given to TSQL could be useful.


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/OptionBuilder.scala
```diff
@@ -0,0 +1,47 @@
...
+      case _ if ctx.OFF() != null => OptionOff(id)
+      case _ if ctx.AUTO() != null => OptionAuto(id)
+      case _ if ctx.STRING() != null => OptionString(id, ctx.STRING().getText)
+      case _ if ctx.expression() != null =>
+        val supplement = if (ctx.id(1) != null) Some(ctx.id(1).getText) else None
+        OptionExpression(id, ctx.expression().accept(expressionBuilder), supplement)
+
+      // All other cases being OptionOn as it is a single keyword representing true
+      case _ => OptionOn(id)
+    }
```
vil1, line 45:
```suggestion
    ctx match {
      case c if c.DEFAULT() != null => OptionDefault(id)
      case c if c.ON() != null => OptionOn(id)
      case c if c.OFF() != null => OptionOff(id)
      case c if c.AUTO() != null => OptionAuto(id)
      case c if c.STRING() != null => OptionString(id, c.STRING().getText)
      case c if c.expression() != null =>
        val supplement = if (c.id(1) != null) Some(ctx.id(1).getText) else None
        OptionExpression(id, c.expression().accept(expressionBuilder), supplement)
      case _ if id == "DEFAULT" => OptionDefault(id)
      case _ if id == "ON" => OptionOn(id)
      case _ if id == "OFF" => OptionOff(id)
      case _ if id == "AUTO" => OptionAuto(id)
      // All other cases being OptionOn as it is a single keyword representing true
      case _ => OptionOn(id)
    }
```

This would translate (arguably weird) cases like `DEFAULT = AUTO` to `OptionAuto("DEFAULT")` rather than to `OptionDefault("DEFAULT")`, which I believe is more "correct".

jimidle: Yeah - we rely on correct syntax here of course and in general, any IR generation that relie on options would have to look for them anyway (I think, but maybe not in some cases). But your suggestion seems fine. I will check it locally and push the change.

nfx (approved):
lgtm

PR #499, opened 2024-06-27
TSQL:  Implement TOP, FETCH and OFFSET clauses for TSQL SELECT

jimidle:
TSQL select implements the LIMIT clause via the TOP clause, which also allows phrases such as:

```tsql
SELECT TOP 10 PERCENT
```

TSQL limiting clauses such as FETCH and NEXT also allow expressions rather than just NUMERIC, as well as returning more than the TOP N in an ORDER BY clause if more rows than the last row match the ORDER BY list. Hence the IR has changed to accoomodate expression, `WITH TIES` and whether the expression represents a PERCENT or a straight scalar.


vil1 (approved):

nfx (approved):
lgtm

PR #506, opened 2024-06-28
TSQL: Implement remaining join pivot and apply operations

jimidle:
Here we implement:

 - CROSS JOIN (explicit)
 - CROSS APPLY
 - OUTER APPLY
 - PIVOT
 - UNPIVOT
 
The APPLY joins are peculiar to TSQL, but should be translatable to LEFT LATERAL or LEFT LATERAL VIEW with an explode function. This however is not going to be a simple translation in the manner of function to function translation, and so we will defer this to Databricks SQL generation, at least for now.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
@@ -131,7 +131,7 @@ class TSqlRelationBuilder extends TSqlParserBaseVisitor[ir.Relation] {
   override def visitTableSource(ctx: TableSourceContext): ir.Relation = {
     val left = ctx.tableSourceItem().accept(this)
     ctx match {
-      case c if c.joinPart() != null => c.joinPart().asScala.foldLeft(left)(buildJoin)
+      case c if c.joinPart() != null => c.joinPart().asScala.foldLeft(left)(buildJoinPart)
     }
   }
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -141,25 +141,104 @@ class TSqlRelationBuilder extends TSqlParserBaseVisitor[ir.Relation] {
...
-    case jt if jt.outerJoin().RIGHT() != null => ir.RightOuterJoin
-    case jt if jt.outerJoin().FULL() != null => ir.FullOuterJoin
-    case _ => ir.UnspecifiedJoin
+  private def buildJoinPart(left: ir.Relation, ctx: JoinPartContext): ir.Relation = {
+    ctx match {
+      case c if c.joinOn() != null => buildJoinOn(left, c.joinOn())
+      case c if c.crossJoin() != null => buildCrossJoin(left, c.crossJoin())
+      case c if c.apply() != null => buildApply(left, c.apply())
+      case c if c.pivot() != null => buildPivot(left, c.pivot())
+      case _ => buildUnpivot(left, ctx.unpivot()) // Only case left
```
vil1, line 150:
```suggestion
      case c if c.unpivot() != null => buildUnpivot(left, ctx.unpivot())
      case _ => ir.UnspecifiedJoin
```

jimidle: At the point of the `case _`, the only possibility is `unpivot()` and the `case _` cannot be reached. Is there a general scala convention to avoid using `case _` as I have done here?

nfx (approved):

PR #512, opened 2024-07-01
TSQL: Simplifies named table tableSource, implements columnAlias list

jimidle:
The ANTLR grammar for tableSource is simplified and rationalized such that named tables are dealt with in only one place. 

Additionally, column alias list is correctly parsed and implemented in IR.

Table hints are collected and processed, and a new Relation is created, TableWithHints. Some hints may have direct counterparts in the Catalyst optimizer, some we may use to comment the translated code as an aid to migration. 

In general, query optimizations are best left to the plan optimizer, but it is possible that some interventions in TSQL were useful and may also be useful within Databricks SQL.


jimidle:
Waiting on other PRs before this can be completed.

vil1 (approved):

vil1 (approved):

PR #515, opened 2024-07-01
TSQL: Implement translation of INSERT statement

jimidle:
Here we fully implement the TSQL INSERT statement, including all target options and optional clauses as well as CTEs.


jimidle:
Now blocked by #512

nfx (approved):
lgtm

PR #525, opened 2024-07-03
TSQL: Support generic FOR options

jimidle:
Some TSQL option clauses use the keyword FOR, which is not allowed to be used as an identifier without escaping it using the standard syntax `[FOR]` as it would be seen as a table alias in a `SELECT` statement. Hence it must be catered for explicitly in the generic option syntax. 

For example:

```tsql
SELECT * FROM t FOR XML RAW
            OPTION (
                OPTIMIZE FOR UNKNOWN
            )
```

In the example, the first `FOR` would be seen as an alias for table `t` if it were allowed to be a keyword usable directly as an Identifier.

Here, we expand the ANTLR rule for parsing generically formed options, such that options containing `FOR` are correctly parsed as options.

### DIFF
--- a/core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
+++ b/core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
@@ -3610,7 +3610,6 @@ keyword
     | FIRST
     | FMTONLY
     | FOLLOWING
-    | FOR
     | FORCE
     | FORCE_FAILOVER_ALLOW_DATA_LOSS
     | FORCE_SERVICE_ALLOW_DATA_LOSS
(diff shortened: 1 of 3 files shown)


core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
```diff
@@ -4136,6 +4135,7 @@ optionList: genericOption (COMMA genericOption)*
  * KEYWORD = VALUE KB        - Some sort of size value, where KB can be various things so is parsed as any id()
  * KEYWORD (=)? DEFAULT      - A fairly redundant option, but sometimes people want to be explicit
  * KEYWORD (=)? AUTO         - The option is set to AUTO, which occurs in a few places
+ * somthing FOR something    - The option is a special case such as OPTIMIZE FOR UNKNOWN
```
vil1, line 4138:
```suggestion
 * something FOR something    - The option is a special case such as OPTIMIZE FOR UNKNOWN
```

vil1 (approved):

vil1 (approved):

PR #529, opened 2024-07-04
Implement a new error strategy for simplifying error messages

jimidle:
Here we implement a new error strategy for constructing error messages such that they are actually useful instead of listing out 780 possible tokens that were expected.


core/src/main/scala/com/databricks/labs/remorph/parsers/SqlErrorStrategy.scala
```diff
@@ -0,0 +1,86 @@
...
+  }
+  // $COVERAGE-ON$
+
+  override protected def reportInputMismatch(recognizer: Parser, e: InputMismatchException): Unit = {
+    val msg = new StringBuilder()
+    msg.append(getTokenErrorDisplay(e.getOffendingToken))
+    msg.append(" was unexpected ")
+    msg.append(generateMessage(recognizer, e))
+    msg.append("\nexpecting one of: ")
+    msg.append(buildExpectedMessage(recognizer, e.getExpectedTokens))
```
nfx, line 41: use interpolated stings: `val msg = s"${getTokenErrorDisplay(e.getOffendingToken)}  was unexpected ..."` - more readable

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlErrorStrategy.scala
```diff
@@ -0,0 +1,963 @@
...
+
+    // We build the messages by looking at the stack trace of the exception, but if the
+    // rule translation is not found, or it is the same as the previous message, we skip it,
+    // to avoid repeating the same message multiple times. This is because a recognition error
+    // could be found in a parent rule or a child rule but there is no extra information
+    // provided by being more specific about the rule location. ALos, in some productions
+    // we may be embedded very deeply in the stack trace, so we want to avoid too many contexts
+    // in a message.
+    val stack = e.getStackTrace
+    stack.foreach { traceElement =>
```
nfx, line 47:
```suggestion
    val messages = e.getStackTrace.flatMap { traceElement => ...
```
flatMap would be more idiomatic here.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlErrorStrategy.scala
```diff
@@ -0,0 +1,963 @@
...
+  /**
+   * When building the list of expected tokens, we do some custom manipulation so that we do not produce a list of 750
+   * possible tokens because there are so many keywords that can be used as id/column names. If ID is a valid expected
+   * token, then we remove all the keywords that are there because they can be an ID.
+   * @param expected
+   *   the set of valid tokens at this point in the parse, where the error was found
+   * @return
+   *   the expected string with tokens renamed in more human friendly form
+   */
+  import scala.collection.mutable
```
nfx, line 82: can you move the import to the top?

nfx (approved):
this definitely makes errors more digestible. few nits to make it a bit more idiomatic.

core/src/main/scala/com/databricks/labs/remorph/parsers/SqlErrorStrategy.scala
jimidle: Will. do - @vil1 also suggested the same

PR #532, opened 2024-07-08
Use Oracle library only if the recon source is Oracle

bishwajit-db:
Use Oracle library for recon job only if the recon source is Oracle


vijaypavann-db (approved):

nfx (approved):
lgtm

PR #536, opened 2024-07-08
Add support for Snowflake's date and time literals

vil1:
It seems undocumented (maybe it is a deprecated feature), but Snowflake seems to accept date and timestamp literals expressed like `DATE'2024-07-08'` or `TIMESTAMP'2024-07-08 15:50:00'` (this has been confirmed by @vijaypavann-db on a real snowflake instance). This PR introduces support for this syntax.


nfx (approved):
lgtm

PR #540, opened 2024-07-09
TSQL: Implement TSQL UPDATE/DELETE statements

jimidle:
Here we add support for the TSQL update statement in all syntactical forms, including support for UDF column transformations. The DELETE statement is also implemented here as they share many common clauses.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
@@ -244,6 +244,40 @@ class TSqlRelationBuilder extends TSqlParserBaseVisitor[ir.LogicalPlan] {
     ctx.expressionList().expression().asScala.map(_.accept(expressionBuilder))
   }
 
+  override def visitMergeStatement(ctx: MergeStatementContext): ir.LogicalPlan = {
+    val merge = ctx.merge().accept(this)
+    Option(ctx.withExpression())
+      .map { withExpression =>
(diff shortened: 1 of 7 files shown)


jimidle:
MERGE is only partially implemented as I ran out of time. Hence I will pick it up in my next PR

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -245,6 +245,109 @@ class TSqlRelationBuilder extends TSqlParserBaseVisitor[ir.Relation] {
...
+    }
+
+    val output = Option(ctx.outputClause()).map(_.accept(this))
+    val tableSourcesOption = Option(ctx.tableSources()).map(_.tableSource().asScala.map(_.accept(this)))
+    val sourceRelation = tableSourcesOption.map {
+      case Seq(tableSource) => tableSource
+      case sources =>
+        sources.reduce(
+          ir.Join(_, _, None, ir.CrossJoin, Seq(), ir.JoinDataType(is_left_struct = false, is_right_struct = false)))
+    }
```
vil1, line 344:
```suggestion
    val sourceRelation = tableSourcesOption.map { tableSources =>
     tableSources.tail.foldLeft(tableSources.head)(
          ir.Join(_, _, None, ir.CrossJoin, Seq(), ir.JoinDataType(is_left_struct = false, is_right_struct = false))))
    }
```

vil1 (approved):
lgtm

PR #541, opened 2024-07-10
plsql snowflake

sundarshankar89:


sundarshankar89:
covered as part of #548  so closing this.

PR #545, opened 2024-07-11
Support CTAS in TSQL Grammar and add more comparison operators.

bishwajit-db:
Support CTAS in TSQL Grammar.
Support for !=, !<, !> operators


core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
```diff
@@ -1860,6 +1860,7 @@ createTable
     : CREATE TABLE tableName LPAREN columnDefTableConstraints (COMMA? tableIndices)* COMMA? RPAREN (
         LOCK simpleId
     )? tableOptions* (ON id | DEFAULT | onPartitionOrFilegroup)? (TEXTIMAGE_ON id | DEFAULT)? SEMI?
+    | CREATE TABLE tableName (LPAREN columnNameList RPAREN)? tableOptions? AS selectStatementStandalone SEMI?
```
nfx, line 1863: create two rules - `createTableDefault` and `createTableAs`. this will simplify the parser. add unit and functional tests

nfx (changes requested):

bishwajit-db:
@jimidle, could you please take over the PR so that the changes can be included along with DDLBulider for TSQL?

nfx:
@bishwajit-db Jim will pick it up after vacation

jimidle:
> @bishwajit-db Jim will pick it up after vacation

The CREATE TABLE grammar will need quite a lot of changes, as it is still in the style of the original grammar we had, but the operator stuff looks fine in the current way they are being handled - those operators were obviously missed originally. That grammar form is possibly going to change, but I will make sure to include these operators if I change it.

jimidle:
@nfx This PR can be merged as-is, as the operators were just missing from the original grammar and the changes to CREATE TABLE are innocuous. There will be significant changes to CREATE once I start on it as there is more missing than CTAS and as CREATE xxx is expressed it will cause ANTLR to do a lot of lookahead to predict the correct alts. Changing the whole of create is a series of additional PRs.

jimidle (approved):
lgtm

PR #547, opened 2024-07-11
TSQL: Remove the SIGN fragment from numerical tokens

### LINKED ISSUE #546: [BUG]: TSQL arithmetic operators clash with numeric signs

### Category of Bug / Issue

TranspileParserError

### Current Behavior

The following input `x +1` fails to be parsed as an expression whereas `x + 1` succeeds. 

This is due to the fact that the `SIGN` fragment is embedded in numerical tokens (`INT`, `REAL`, `FLOAT`, etc). Therefore, `x +1` results in two tokens (`x` and `+1`), which doesn't match any `expression` rule, while `x + 1` results in three tokens (`x`, `+` and `1`).

### Expected Behavior

Both `x +1` and `x + 1` should be parsable as expressions.

### Steps To Reproduce

_No response_

_No response_

### Sample Query

_No response_

### Operating System

macOS

### Version

latest via Databricks CLI

vil1:
Fixes #546 

The proposed solution isn't completely satisfying though, as it makes it impossible to parse negative literals out of TSQL input: the input string `-2` is now parsed as `UMinus(Literal(2))` (instead of simply `Literal(-2)`). While being a mere inconvenience (it just delegates the handling of `UPlus` and `UMinus` to downstream processes), the fact that Snowflake grammar is able to parse `-2` as `Literal(-2)` indicates that there might be a better solution to this problem. So I suggest we keep #546 open even if we merge this PR, so that @jimidle can later come back to it and find a better solution.

### DIFF
--- a/core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlLexer.g4
+++ b/core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlLexer.g4
@@ -952,13 +952,11 @@ STRING options {
     caseInsensitive = false;
 }: 'N'? '\'' ('\\' . | '\'\'' | ~['])* '\'';
 
-fragment SIGN: [+-];
-
-INT   : SIGN? DEC_DIGIT+;
-HEX   : SIGN? '0' 'X' HEX_DIGIT*;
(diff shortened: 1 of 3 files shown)


nfx (approved):
lgtm

PR #548, opened 2024-07-11
Add `LET` and `DECLARE` statements parsing in Snowflake PL/SQL procedures

sundarshankar89:
Added Grammar for 
`DECLARE` and `LET` statements
`RETURN` Statement
Parameterized Expression
Refactor using ScalaSubquery

Added IR visitor for `DECLARE` and `LET` with Variable Assignment and ResultSet Assignment
Not covered `CURSOR`


core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -151,6 +153,9 @@ copyIntoTable
     ) R_PAREN files? pattern? fileFormat? copyOptions*
     ;
 
+declare: DECLARE (id dataType (DEFAULT expr | (L_PAREN selectStatement R_PAREN))? SEMI)+
+    ;
```
vil1, line 156: Split the rule in smaller components to makes visitor implementation easier:
```suggestion
declareClause: DECLARE declarationStatement+
    ;
 
 declarationStatement
    : id dataType                                                                #declarationSimple
    | id dataType DEFAULT expr                                       #declarationDefault
    | id dataType L_PAREN selectStatement R_PAREN #declarationSelect
    ;
```

This is just a sketch (you may want to factorize the `id dataType` part out at some point) but it would make the visitor implementation more straightforward:

In the method that uses the content of `declareClause` (and other stuff) to build a `ir.CreateProcedure`, you'd just have to do something like `ctx.declareClause().declareStatement().asScala.map(_.accept(statementBuilder))`. 

In `SnowflakeStatementBuilder` you would just have to implement three small methods for each case of `declareStatement`, namely `visitDeclareSimple(ctx: DeclareSimpleContext)`, `visitDeclareDefault(ctx: DeclareDefaultContext)` and `visitDeclareSelect(ctx: DeclareSelectContext)`.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -188,6 +193,12 @@ formatName: FORMAT_NAME EQ string
 formatType: TYPE EQ typeFileformat formatTypeOptions*
     ;
 
+let: LET id dataType? ASSIGN expr SEMI | LET id dataType? DEFAULT expr SEMI | id ASSIGN expr SEMI
+    ;
```
vil1, line 196: You can simply write:
```suggestion
let: LET id dataType? ((ASSIGN | DEFAULT) expr)? SEMI 
    ;
```
Or use labels (as described above), whatever makes the visitor implementation easier.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -219,4 +218,38 @@ class SnowflakeDDLBuilder
...
+    val ids = ctx.id().asScala.map(_.getText)
+    val dataTypes = ctx.dataType().asScala.map(dt => Some(DataTypeBuilder.buildDataType(dt)))
+    val expressions = ctx.expr().asScala.map(_.accept(expressionBuilder)).map(Some(_))
+
+    // Padding the lists to ensure they all have the same size
+    // [TODO]: This is a temporary fix. We need to handle this in a better way
+    val paddedDataTypes = dataTypes ++ Seq.fill(ids.size - dataTypes.size)(None)
+    val paddedExpressions = (expressions ++ Seq.fill(ids.size - expressions.size)(None))
+
+    ids.zip(paddedDataTypes).zip(paddedExpressions).map { case ((name, dataType), expr) =>
```
vil1, line 233: Whenever you find yourself needing to write this kind of complicated logic it is a sign that the corresponding grammar could be improved to make your life easier (see the comments above in `SnowflakeParser.g4`.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -219,4 +218,38 @@ class SnowflakeDDLBuilder
...
+      ir.SetVariable(name, dataType, expr)
+    }
+
+  }
+
+  private def buildReturn(ctx: ReturnStatementContext): ir.Expression = {
+    // [TODO]: implement return statement for TABLE(res)
+    ctx.accept(expressionBuilder)
+  }
```
vil1, line 242: There's nothing wrong with this method _per se_, but we usually use this pattern of defining a `private def buildXYZ(...): XYZ` only when `XYZ` isn't a type returned by an existing `SnowflakeXYZBuilder`. Otherwise, we simply delegate to said `SnowflakeXYZBuilder`.

vil1 (commented):
First, I think we should define a `trait Statement extends TreeNode` and use it for everything related to scripting/PLSQL (including variable declaration and assignment, conditionals, loops, etc). 

We would then define a `SnowflakeStatementBuilder` for those, which would help keep `SnowflakeExpressionBuilder` relatively small and manageable.

It will be OK for statements to contain expressions like in `case class SetVariable(name: Id, value: Expression) extends Statement`, but not the other way around.

Then I have a few remarks about how the grammar could be defined in a way that makes the implementation of visitors easier to express, I'll try to add these as comments in the relevant files.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -1601,10 +1600,10 @@ callerOwner: CALLER | OWNER
 executaAs: EXECUTE AS callerOwner
     ;
 
-procedureBody: string
+procedureBody: (let | call | executeImmediate | string)*
```
vil1, line 1603: We should start grouping things like `let`, `call`, etc under a `statement` rule (similar to what is done with `expr`.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
```diff
@@ -236,3 +236,11 @@ case class ColumnAliases(input: Relation, aliases: Seq[Id]) extends RelationComm
 
 case class TableFunction(functionCall: Expression) extends Relation {}
 case class Lateral(expr: Relation) extends Relation {}
+
+case class CreateProcedure(
+    name: String,
+    parameters: Seq[FunctionParameter],
+    variables: Option[Seq[SetVariable]],
```
vil1, line 243: It's merely a detail but `Option[Seq[X]]` is redundant in most cases (it appears here and there in the codebase but we'll have to clean that up eventually). There is rarely any meaningful difference between `None` and `Some(Seq())`. So we should prefer going for simply `Seq[X]` in general.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/extensions.scala
```diff
@@ -236,3 +236,11 @@ case class ColumnAliases(input: Relation, aliases: Seq[Id]) extends RelationComm
...
+    body: Option[Plan],
```
vil1, line 244: This will have to become a `Seq[Statement]` instead.

nfx:
@sundarshankar89 please add some functional examples of PL/SQL procedures with LET/DECLARE

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/commands.scala
```diff
@@ -87,3 +87,14 @@ case class WriteStreamOperationStart(
...
     with Command
+
+
+
+case class CreateVariable(name: LogicalPlan, defaultExpr: Expression, replace: Boolean) extends UnaryNode with Command {
+  override def child: LogicalPlan = name
+}
+
+case class SetVariable(name: String, value: LogicalPlan) extends UnaryNode with Command {
+
```
nfx, line 98:
```suggestion
```

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/commands.scala
```diff
@@ -87,3 +87,14 @@ case class WriteStreamOperationStart(
...
     extends LeafNode
     with Command
+
+
+
+case class CreateVariable(name: LogicalPlan, defaultExpr: Expression, replace: Boolean) extends UnaryNode with Command {
+  override def child: LogicalPlan = name
+}
+
+case class SetVariable(name: String, value: LogicalPlan) extends UnaryNode with Command {
```
nfx, line 97:
```suggestion
case class SetVariable(name: String, value: Expression) extends UnaryNode with Command {
```
variables are results of expressions, not plans.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/commands.scala
```diff
@@ -87,3 +87,14 @@ case class WriteStreamOperationStart(
     foreach_batch: Option[StreamingForeachFunction])
     extends LeafNode
     with Command
+
+
+
+case class CreateVariable(name: LogicalPlan, defaultExpr: Expression, replace: Boolean) extends UnaryNode with Command {
```
nfx, line 93:
```suggestion
case class CreateVariable(name: String, defaultExpr: Expression, replace: Boolean) extends UnaryNode with Command {
```
why name is the `LogicalPlan`? it makes no sense.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/expressions.scala
```diff
@@ -4,6 +4,10 @@ import java.util.{Locale, UUID}
 
 // Expression used to refer to fields, functions and similar. This can be used everywhere
 // expressions in SQL appear.
+case class ExpressionLogicalPlan(expression: Expression) extends LeafNode {
+  override def output: Seq[Attribute] = Seq.empty
+}
+
```
nfx, line 10:
```suggestion
```
no need to wrap expression, it's unnecessary.

nfx (changes requested):

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
sundarshankar89: Taken care

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
sundarshankar89: Done

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/commands.scala
sundarshankar89: Fixed

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/expressions.scala
sundarshankar89: Fixed

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/commands.scala
sundarshankar89: I was making them as a command trying to map it to how Spark maps them.

sundarshankar89: Either I keep them as command defined them as `leafnode` or create a new Branch called statements which raps `expressions`

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/expressions.scala
nfx: we have `ScalarSubquery` for a think exactly opposite to this, just in case.

sundarshankar89: That's a good direction, let me see if that helps.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -142,8 +142,10 @@ otherCommand
     | truncateMaterializedView
     | truncateTable
     | unset
-    | call
-    | beginTxn
+    | scriptingStatement
```
nfx, line 145:
```suggestion
    | call
    | beginTxn
    | scriptingStatement
```
otherwise this breaks the bc-compat

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -142,8 +142,10 @@ otherCommand
...
+    ;
+
+scriptingStatement: call | declareCommand | let | returnStatement | beginTxn
```
nfx, line 148:
```suggestion
procStatement
    : call 
    | declareCommand 
    | let 
    | returnStatement 
    | beginTxn
```

scripting statement will grow bigger.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -162,6 +164,15 @@ copyIntoTable
     ) R_PAREN files? pattern? fileFormat? copyOptions*
     ;
 
+declareCommand: DECLARE declareStatement+
```
nfx, line 167: would be good to leave a pointer to official documentation

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -162,6 +164,15 @@ copyIntoTable
...
+    ;
+
+declareStatement
+    : id dataType SEMI
+    | id dataType DEFAULT L_PAREN selectStatement R_PAREN SEMI
+    | id dataType DEFAULT expr SEMI
```
nfx, line 173:
```suggestion
    : id dataType SEMI                                                                                   # declareSimple 
    | id dataType DEFAULT L_PAREN selectStatement R_PAREN SEMI.  # declareSelect
    | id dataType DEFAULT expr SEMI                                                          # declareExpr
```

use rule specifiers. see `exprBitNot` as the example.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -199,6 +210,14 @@ formatName: FORMAT_NAME EQ string
 formatType: TYPE EQ typeFileformat formatTypeOptions*
     ;
 
+let
+    : LET key = id dataType? (ASSIGN | DEFAULT) value = expr SEMI
+    | key = id dataType? (ASSIGN | DEFAULT) value = expr SEMI
```
nfx, line 215:
```suggestion
    : LET? key = id dataType? (ASSIGN | DEFAULT) value = expr SEMI
```
won't this be simpler?

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -1601,16 +1620,22 @@ callerOwner: CALLER | OWNER
 executaAs: EXECUTE AS callerOwner
     ;
 
-procedureDefinition: string | DBL_DOLLAR
+procedureBody: (let | call | executeImmediate | string)*
```
nfx, line 1623:
```suggestion
procedureBody: procStatement+
```
didn't you mean this?

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -1601,16 +1620,22 @@ callerOwner: CALLER | OWNER
...
+    ;
+
+procedureDefinition: DBL_DOLLAR | declareCommand? BEGIN procedureBody returnStatement END SEMI
```
nfx, line 1626:
```suggestion
procedureDefinition: DBL_DOLLAR | declareCommand? BEGIN procStatement+ END SEMI
```
there might be multiple `returnStatement` in the body, it's not necessary to have it always the last one. can we just embed the `procStatement+`?

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -2228,7 +2253,7 @@ taskOverlap: ALLOW_OVERLAPPING_EXECUTION EQ trueFalse
 sql: EXECUTE IMMEDIATE DBL_DOLLAR | sqlCommand | call
     ;
 
-call: CALL objectName L_PAREN exprList? R_PAREN
+call: CALL objectName L_PAREN exprList? R_PAREN SEMI
```
nfx, line 2256:
```suggestion
call: CALL objectName L_PAREN exprList? R_PAREN
```
don't do that. it'll break BC compat
image

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3210,8 +3237,11 @@ primitiveExpression
     | BOTH_Q            # primExprBoth
     | ARRAY_Q           # primExprArray
     | OBJECT_Q          # primExprObject
+    | COLON id          # primVariable
     ;
 
+parameterExpression: COLON id
+    ;
```
nfx, line 3244:
```suggestion
    ;
    
```
nit

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/commands.scala
```diff
@@ -87,3 +87,10 @@ case class WriteStreamOperationStart(
     foreach_batch: Option[StreamingForeachFunction])
     extends LeafNode
     with Command
+
+case class CreateVariable(name: String, dataType: DataType, defaultExpr: Option[Expression], replace: Boolean)
```
nfx, line 91:
```suggestion
case class CreateVariable(name: String, dataType: DataType, defaultExpr: Option[Expression] = None, replace: Boolean = false)
```
1. use defaults
2. is it shared between snow & TSQL?

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/commands.scala
```diff
@@ -87,3 +87,10 @@ case class WriteStreamOperationStart(
...
+    extends LeafNode
+    with Command
+
+case class SetVariable(name: String, dataType: Option[DataType], value: Expression) extends LeafNode with Command
```
nfx, line 95:
```suggestion
case class SetVariable(name: String, value: Expression, dataType: Option[DataType] = None) extends LeafNode with Command
```
1. optional args always have to be the last
2. use defaults

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeCommandBuilderSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+    with Matchers
+    with MockitoSugar
+    with IRHelpers {
+
+  override protected def astBuilder: SnowflakeCommandBuilder =
+    new SnowflakeCommandBuilder
+
+  "translate Declare to CreateVariable Expression" in {
+    example(
+      "x number default 0;",
```
nfx, line 20: can you use UPPERCASE for keywords everywhere in tests, so that it's easier for reviewers to understand

nfx (changes requested):

sundarshankar89:
@vil1  and @nfx  Ready for Review.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -144,6 +144,10 @@ otherCommand
     | unset
     | call
     | beginTxn
+    | procStatement
```
nfx, line 147:
```suggestion
    | declareCommand
```
otherwise we have a recursion of `call` and `executeImmediate` on the `otherCommand`.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -162,6 +166,18 @@ copyIntoTable
     ) R_PAREN files? pattern? fileFormat? copyOptions*
     ;
 
+//https://docs.snowflake.com/en/sql-reference/snowflake-scripting/declare
```
nfx, line 169:
```suggestion
// see https://docs.snowflake.com/en/sql-reference/snowflake-scripting/declare
```
nit: space after comment tokens.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -199,6 +215,15 @@ formatName: FORMAT_NAME EQ string
 formatType: TYPE EQ typeFileformat formatTypeOptions*
     ;
 
+let
+    //variable and resultset are covered under the same visitor since expr is common
```
nfx, line 219:
```suggestion
    // variable and resultset are covered under the same visitor since expr is common
```
nit: space after comment tokens

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -144,6 +144,10 @@ otherCommand
     | unset
     | call
     | beginTxn
+    | procStatement
+    ;
+
+procStatement: call | declareCommand | let | executeImmediate | returnStatement
```
nfx, line 150: can we add a catch-all `UnresolvedStatement` for those that we don't parse yet? otherwise it's a regression from `procedureDefinition: string | DBL_DOLLAR`.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3109,13 +3137,14 @@ exprList: expr (COMMA expr)*
...
-    | iffExpr                                   # exprIff
-    | bracketExpression                         # exprBracket
+    : objectName DOT NEXTVAL # exprNextval
+    | expr DOT expr          # exprDot
+    | expr COLON expr        # exprColon
+    | expr COLLATE string    # exprCollate
+    | caseExpression         # exprCase
+    | iffExpr                # exprIff
+    //Removed it so expression precedance takes over.
+    //| bracketExpression                         # exprBracket
```
nfx, line 3147: why do we need this?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeCommandBuilder.scala
```diff
@@ -0,0 +1,55 @@
...
+    extends SnowflakeParserBaseVisitor[ir.Command]
+    with ParserCommon[ir.Command]
+    with IncompleteParser[ir.Command] {
+
+  private val expressionBuilder = new SnowflakeExpressionBuilder
+
+  protected override def wrapUnresolvedInput(unparsedInput: String): ir.UnresolvedCommand =
+    ir.UnresolvedCommand(unparsedInput)
+
+  // [TODO]: Implement Cursor and Exception for Declare Statements.
```
nfx, line 16:
```suggestion
  // TODO: Implement Cursor and Exception for Declare Statements.
```
let's start all todo's with `// TODO` consistently - it'll be easier to autocreate issues from those.

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeCommandBuilderSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+      "price NUMBER(13,2) DEFAULT 111.50;",
+      _.declareStatement(),
+      CreateVariable(
+        name = Id("price"),
+        dataType = DecimalType(Some(13), Some(2)),
+        defaultExpr = Some(Literal(float = Some(111.5f))),
+        replace = false))
+
+    example(
+      "query_statement RESULTSET := (select col1 from some_table);",
```
nfx, line 43:
```suggestion
      "query_statement RESULTSET := (SELECT col1 FROM some_table);",
```
can we consistently use UPPERCASE for keywords everywhere?

nfx (changes requested):

PR #549, opened 2024-07-12
remorph metrics logging for usage for experimental functions

ganeshdogiparthi-db:


src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -229,6 +229,8 @@ def verify_workspace_client(workspace_client: WorkspaceClient) -> WorkspaceClien
         workspace_client.config._product = "remorph"
     if workspace_client.config._product_version != __version__:
         workspace_client.config._product_version = __version__
+
+    logger.info(f"user: {workspace_client.current_user.me().user_name}")
```
sundarshankar89, line 233: can you remove that from CLI so we don't log twice.

sundarshankar89 (changes requested):

src/databricks/labs/remorph/transpiler/execute.py
ganeshdogiparthi-db: I have checked @sundarshankar89 
Transpile:
In cli.py, we don't call for `verify_workspace_client`, so this is logged only once.
We call `verify_workspace_client` in experimental functions, which are not related to CLI, so this shouldn't be a problem.

Reconcile:
In reconcile we log the `user` in CLI and trigger the job. 
In the workflow job, while running the execute.py/recon function, we log it in the workflow job logs, so it does not log twice.

Please let me know if you still want me to remove it from reconcile CLI.

ganeshdogiparthi-db: To follow up on your suggestion, I have removed the logger from cli.py/transpile and called `verify_workspace_client` to execute.py/morph function. This should cover the CLI and any cx wrapper around the morph function.

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -155,6 +155,7 @@ def morph(workspace_client: WorkspaceClient, config: MorphConfig):
     :param config: The configuration for the morph operation.
     :param workspace_client: The WorkspaceClient object.
     """
+    ws_client: WorkspaceClient = verify_workspace_client(workspace_client)
```
sundarshankar89, line 158: nit: Add a comment why we are reassigning the wsclient.

sundarshankar89 (approved):
LGTM small nit

src/databricks/labs/remorph/transpiler/execute.py
ganeshdogiparthi-db: Done

.github/scripts/setup_spark_remote.sh
```diff
@@ -34,4 +34,4 @@ if [ $? -ne 0 ]; then
     #sleep time to start the server
     sleep 2m
     echo "Server Already Running"
-fi
+fi
```
sundarshankar89, line 37: remove this space so there is no change here.

sundarshankar89 (changes requested):

.github/scripts/setup_spark_remote.sh
ganeshdogiparthi-db: done

pyproject.toml
```diff
@@ -49,7 +49,7 @@ dependencies = [
   "pytest-cov>=4.0.0,<5.0.0",
   "black>=23.1.0",
   "ruff>=0.0.243",
-  "databricks-connect",
+  "databricks-connect==15.1",
```
nfx, line 52: don't attach specific databricks connect version. there

PR #559, opened 2024-07-17
Changes in Final reconcile output for column category 

ganeshdogiparthi-db:
Changes in the Final reconcile output for column category give priority to status over absolute mismatch since the table threshold for the mismatch is introduced


sundarshankar89 (approved):
LGTM

sundarshankar89 (approved):
LGTM

PR #560, opened 2024-07-17
Snowflake implement translation of UPDATE command

vil1:
Includes a more general (but small) refactoring around visiting many subtrees with a same rule.


nfx (approved):
lgtm

PR #563, opened 2024-07-19
Improve remorph-core coverage report, add a dry-run github workflow

vil1:
Set `transpiled` to 0 are we do only parsing ATM, shorten errors and display relative paths to make the report smaller


nfx (approved):
lgtm

PR #703, opened 2024-07-23
Define (almost all) the remaining Snowflake functions

vil1:
A significant portion of errors in coverage tests are due to functions that were not yet defined.

To improve the situation, I scraped Snowflake's documentation with a combination of bash, curl and xpath to extract the name and arities of the 700+ functions that are documented there. Because the documentation isn't really uniform (in terms of HTML structure), some of the arities thus extracted are still wrong and will have to be manually verified and fixed one by one. This will be a long and tedious process, which is why I decided to issue a PR even though there is still work to be done on that front (for the sake of making the coverage dashboard look better).


nfx (approved):
lgtm

PR #705, opened 2024-07-23
[chore] Make singletons for the relevant `DataType` instances

nfx:
This PR makes case objects for things like `NullType`, `StringType`, etc.


sundarshankar89 (approved):
LGTM

PR #710, opened 2024-07-24
Ensure that date and time formatters always use UTC in tests

jimidle:
Ensures that the tests suite works even if the machine running the tests is not set to UTC.


jimidle:
This PR is also solved in  #715 - so I will close this.

PR #715, opened 2024-07-25
Feature: introduce core transpiler

vil1:
A `SnowflakeToDatabricksTranspiler` is introduced that converts Snowflake queries into Databricks SQL. It is plugged in the coverage tests suites so that going forward, said test actually test transpilation

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
@@ -1,14 +1,19 @@
 package com.databricks.labs.remorph.generators.sql
 
-import com.databricks.labs.remorph.generators.GeneratorContext
+import com.databricks.labs.remorph.generators.{Generator, GeneratorContext}
 import com.databricks.labs.remorph.parsers.{intermediate => ir}
+import com.databricks.labs.remorph.transpilers.TranspileException
 
-import java.text.SimpleDateFormat
+import java.time.format.DateTimeFormatter
+import java.time.{Instant, LocalDate, LocalDateTime, ZoneId, ZonedDateTime}
 import java.util.Locale
 
-class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper()) {
-  private val dateFormat = new SimpleDateFormat("yyyy-MM-dd")
-  private val timeFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS")
+class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper())
+    extends Generator[ir.Expression, String] {
+  private val dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd")
+  private val timeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.of("UTC"))
+
+  override def generate(ctx: GeneratorContext, tree: ir.Expression): String = expression(ctx, tree)
 
   def expression(ctx: GeneratorContext, expr: ir.Expression): String = {
     expr match {
@@ -19,7 +24,9 @@ class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper()) {
       case l: ir.Literal => literal(ctx, l)
       case fn: ir.Fn => callFunction(ctx, fn)
       case ir.UnresolvedAttribute(name, _, _) => name
-      case _ => throw new IllegalArgumentException(s"Unsupported expression: $expr")
+      case i: ir.Id => id(ctx, i)
+      case a: ir.Alias => alias(ctx, a)
+      case x => throw TranspileException(s"Unsupported expression: $x")
     }
   }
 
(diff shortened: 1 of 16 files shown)


core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -1,20 +1,29 @@
...
+import java.time.format.DateTimeFormatter
+import java.time.{Instant, LocalDate, LocalDateTime, ZoneId, ZonedDateTime}
 import java.util.Locale
 
-class ExpressionGenerator {
-  private val dateFormat = new SimpleDateFormat("yyyy-MM-dd")
-  private val timeFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS")
+class ExpressionGenerator extends Generator[ir.Expression, String] {
+  private val dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd")
+  private val timeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS")
```
jimidle, line 14: Won't this change still suffer from returning the default TZ and not UTC? The tests fail on my system because it is on Belize/Central America. 

```scala
private val timeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.of("UTC"))
```

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -31,12 +40,18 @@ class ExpressionGenerator {
       case ir.StringType => orNull(l.string.map(doubleQuote))
       case ir.DateType =>
         l.date match {
-          case Some(date) => doubleQuote(dateFormat.format(date))
+          case Some(date) =>
+            doubleQuote(
+              LocalDate.from(ZonedDateTime.ofInstant(Instant.ofEpochMilli(date), ZoneId.of("UTC"))).format(dateFormat))
```
jimidle, line 45: Or does this change force the conversion into UTC? Perhaps we need to know if the incoming queries are assuming a default timezone other than UTC. SqlServer can do this, though it is not a good practice of course.

jimidle (changes requested):

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -1,20 +1,29 @@
 package com.databricks.labs.remorph.generators.sql
 
-import com.databricks.labs.remorph.generators.GeneratorContext
-import com.databricks.labs.remorph.parsers.intermediate.Literal
+import com.databricks.labs.remorph.generators.{Generator, GeneratorContext}
+import com.databricks.labs.remorph.parsers.intermediate.{Expression, Literal}
```
nfx, line 4: this import is redundant because of the following line - `intermediate => ir`

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -31,12 +40,18 @@ class ExpressionGenerator {
       case ir.StringType => orNull(l.string.map(doubleQuote))
       case ir.DateType =>
         l.date match {
```
nfx, line 42: nit: perhaps let's refactor it into a separate method

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
```diff
@@ -0,0 +1,42 @@
...
+      case ir.InnerJoin => "INNER JOIN"
+      case ir.FullOuterJoin => "FULL OUTER JOIN"
+      case ir.LeftOuterJoin => "LEFT OUTER JOIN"
+      case ir.LeftSemiJoin => "LEFT SEMI JOIN"
+      case ir.LeftAntiJoin => "LEFT ANTI JOIN"
+      case ir.RightOuterJoin => "RIGHT OUTER JOIN"
+      case ir.CrossJoin => "JOIN"
+    }
+
+    val condition = join.join_condition.map(expressionGenerator.generate(ctx, _)).getOrElse("")
```
nfx, line 38: we're missing the `USING (col1, col2)` syntax here. join conditions can be empty as well, but then there should be no space before it

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
vil1: good catch, thanks!

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
vil1: roger that

jimidle (approved):

core/src/test/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGeneratorTest.scala
```diff
@@ -0,0 +1,57 @@
...
+class LogicalPlanGeneratorTest extends AnyWordSpec with GeneratorTestCommon[ir.LogicalPlan] with ir.IRHelpers {
+
+  override protected val generator = new LogicalPlanGenerator
+
+  "Project" should {
+    "transpile to SELECT" in {
+      ir.Project(namedTable("t1"), Seq(ir.Id("c1"))) generates "SELECT c1 FROM t1"
+      ir.Project(namedTable("t1"), Seq(ir.Star(None))).doesNotTranspile
+    }
+  }
```
nfx, line 16:
```suggestion
  ir.Project(namedTable("t1"), Seq(ir.Id("c1"))) shouldGenerate "SELECT c1 FROM t1"
  ir.Project(namedTable("t1"), Seq(ir.Star(None))) shouldFail
```

@vil1 i wonder if we can have this kind of test DSL?... this way we can run all examples in parallel and re-run only those that failed... though it might be too much

nfx (approved):
lgtm

PR #742, opened 2024-07-30
Snowflake: introduce a specific call mapper

vil1:


core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -12,299 +12,329 @@ case class CallFunction(function_name: String, arguments: Seq[Expression]) exten
...
+    case CaseInsensitiveCallFunction("XPATH_SHORT", args) => XPathShort(args.head, args(1))
+    case CaseInsensitiveCallFunction("XPATH_STRING", args) => XPathString(args.head, args(1))
+    case CaseInsensitiveCallFunction("XXHASH64", args) => XxHash64(args)
+    case CaseInsensitiveCallFunction("YEAR", args) => Year(args.head)
+    case CaseInsensitiveCallFunction("ZIP_WITH", args) => ZipWith(args.head, args(1), args(2))
     case _ => call // fallback
   }
 }
 
+class SnowflakeCallMapper extends CallMapper {
```
nfx, line 325: does not belong to this file. add it to `rules` package under `showflake` and create a dedicated optimizer rule

core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
```diff
@@ -1,6 +1,18 @@
 package com.databricks.labs.remorph.transpilers
 
+import com.databricks.labs.remorph.parsers.{intermediate => ir}
+
 trait Transpiler {
 
-  def transpile(input: String): String
```
nfx, line 5: keep the public API of this trait only to this one single method. no other methods allowed

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -12,299 +12,329 @@ case class CallFunction(function_name: String, arguments: Seq[Expression]) exten
...
+object CaseInsensitiveCallFunction {
+  def unapply(call: CallFunction): Option[(String, Seq[Expression])] = {
+    Some((call.function_name.toUpperCase(), call.arguments))
+  }
+}
+
 class CallMapper {
 
   /** This function is supposed to be overridden by dialects */
   def convert(call: Fn): Fn = call match {
```
nfx, line 31:
```suggestion
  def convert(call: Fn): Fn = withNormalizedName(call) match {
```
this diff is too big and unnecessary. just add a private method to fix the name.

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -69,7 +75,12 @@ class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper())
 
   private def callFunction(ctx: GeneratorContext, fn: ir.Fn): String = {
     val call = callMapper.convert(fn)
-    s"${call.prettyName}(${call.children.map(expression(ctx, _)).mkString(", ")})"
+    call match {
+      case r: RLike => rlike(ctx, r)
```
nfx, line 79: remove RLike from Fn trait and move it to expressions.scala, next to Like

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -155,4 +166,6 @@ class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper())
   private def orNull(option: Option[String]): String = option.getOrElse("NULL")
 
   private def doubleQuote(s: String): String = s""""$s""""
+
+  private def singleQuote(s: String): String = s"'$s'"
```
nfx, line 170: let's revert this for now.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
vil1: why?

vil1: From the [documentation](https://docs.databricks.com/en/sql/language-manual/data-types/string-type.html)

> While you can use double quotes (") instead of single quotes (') to delimit a string literal, this is discouraged as it is not standard SQL.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
vil1: It's only big because it fixes a big chunk of code that was introduced without tests and bypassing code review.

core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
vil1: Good catch, I should have made these methods protected

nfx (approved):
lgtm

PR #751, opened 2024-08-01
Generate CAST (and datatypes)

vil1:


jimidle (approved):
lgtm

core/src/main/scala/com/databricks/labs/remorph/generators/sql/DataTypeGenerator.scala
```diff
@@ -0,0 +1,34 @@
+package com.databricks.labs.remorph.generators.sql
+
+import com.databricks.labs.remorph.generators.GeneratorContext
+import com.databricks.labs.remorph.parsers.{intermediate => ir}
+
+object DataTypeGenerator {
+
+  def generateDataType(ctx: GeneratorContext, dt: ir.DataType): String = dt match {
+    case ir.NullType => "VOID"
```
nfx, line 9: we don't really have that one

nfx (approved):
g

PR #752, opened 2024-08-01
Generate CTEDefinition

vil1:


jimidle (approved):
lgtm

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
```diff
@@ -140,4 +141,14 @@ class LogicalPlanGenerator(val expr: ExpressionGenerator, val explicitDistinct:
     s"${optionComments}" +
       s"${plan}"
   }
+
+  private def cte(ctx: GeneratorContext, cteDefinition: ir.CTEDefinition): String = {
```
nfx, line 145: let's use SubqueryAlias instead of CTEDefinition - https://github.com/databrickslabs/remorph/blob/main/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/rules/TopPercentToLimitSubquery.scala#L47C9-L58C72

nfx (changes requested):

nfx (approved):
lgtm

PR #754, opened 2024-08-01
Generate IN

vil1:


core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -246,6 +253,15 @@ class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper())
     (objRef.head +: objRef.tail).map(id(ctx, _)).mkString(".")
   }
 
+  private def isInCollection(ctx: GeneratorContext, isInCollection: snowflake.IsInCollection): String = {
+    val values = isInCollection.collection.map(expression(ctx, _)).mkString(", ")
+    s"${expression(ctx, isInCollection.expression)} IN ($values)"
+  }
+
+  private def isInRelation(ctx: GeneratorContext, isInRelation: snowflake.IsInRelation): String = {
+    s"${expression(ctx, isInRelation.expression)} IN (${logicalPlanGenerator.generate(ctx, isInRelation.relation)})"
```
nfx, line 262: From snowflake you have to return `ir.In(Seq(ir.ScalarSubquery(...)))`

nfx (changes requested):

PR #755, opened 2024-08-01
TSQL: Adds tests and support for SELECT OPTION(...) generation

jimidle:
Here, we add code generation support for TSQL `SELECT ... OPTION(...)` clause. 

Any query hints supplied with a SELECT statement are now generated as comments in the output code, in case they may be useful in assessing query performance after remorph transpilation. As in the following sample:

```tsql
SELECT * FROM t
            OPTION (
            MAXRECURSION 10,
            SOMETHING ON,
            SOMETHINGELSE OFF,
            SOMEOTHER AUTO,
            SOMEstrOpt = 'STRINGOPTION')
```

Transpiles to:

```sql
/*
   The following statement was originally given the following OPTIONS:

    Expression options:

     MAXRECURSION = 10

    String options:

     SOMESTROPT = 'STRINGOPTION'

    Boolean options:

     SOMETHING ON
     SOMETHINGELSE OFF

    Auto options:

     SOMEOTHER AUTO


 */
SELECT * FROM t
```

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
@@ -64,7 +64,7 @@ class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper())
     }
 
     val boolOptions = opts.boolFlags.map { case (key, value) =>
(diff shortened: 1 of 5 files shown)


vil1 (approved):
lgtm

nfx (approved):
lgtm

PR #760, opened 2024-08-01
Implement error strategy for Snowflake parsing and use error strategy for all parser instances

jimidle:
Here we implement the equivalent of `TSqlErrorStrategy.scala` for Snowflake.

An implementation of SqlErrorStrategy influences how error messages are reported to users of the transpiler and other parsing tools. It translates raw token names and  parser rules into something that makes more sense for SQL authors.

In addition we ensure that the relevant error strategy is used everywhere that an instance of a dialect parser is instantiated.


vil1 (changes requested):
I proposed some [changes](#761 ) in the implementation  of this PR to make more "scala-idiomatic". It would be nice to add some unit tests as well.

nfx (approved):
lgtm

jimidle:
> I proposed some [changes](#761) in the implementation of this PR to make more "scala-idiomatic". It would be nice to add some unit tests as well.

There are in fact tests for the TSQL one, I would essentially just duplicate them. I will do it.

PR #765, opened 2024-08-02
Implement TSQL specific function call mapper

jimidle:
Here, we implement the function call mapper (Rule) that allows us to make complex transformations from TSQL specific functions into Databricks compatible constructs. Transformations are sometimes argument reordering, and sometimes replacement with a different expression. 

In this PR we use the transformation of TSQL `DATEADD` to Databricks SQL `DATE_ADD`, `ADD_MONTHS` and `xxx + INTERVAL n {days|months|etc}`

For example the TSQL:

```tsql
SELECT DATEADD(hh, 7, col1) AS add_hours_col1 FROM tabl;
```

Translates to:

```sql
SELECT col1 + INTERVAL 7 HOUR AS add_hours_col1 FROM tabl;
```


nfx (approved):
lgtm

PR #766, opened 2024-08-02
TSQL: Improve transpilation coverage

jimidle:
Here we modify some existing tests to improve coverage of TSQL and fix some bugs in code generation or add additional tests.


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/rules/TSqlCallMapper.scala
```diff
@@ -8,10 +8,42 @@ class TSqlCallMapper extends CallMapper {
...
     }
   }
 
+  /**
+   * Converts a CHECKSUM_ARG call to a MD5 function call sequence
+   * @param args
+   * @return
+   */
+  private def checksumAgg(args: Seq[Expression]): Expression = {
+    CallFunction("MD5", Seq(CallFunction("CONCAT_WS", Seq(Literal(","), CallFunction("COLLECT_LIST", args)))))
```
nfx, line 24:
```suggestion
    Md5(ConcatWs(Literal(","), CollectList(args)))
```
Can we use concrete IR nodes when remapping, instead of generic CallFunction? same for genBitSet. otherwise it's inconsistent.

nfx (changes requested):
small nits to prevent problems early in the future copy-pasted code.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/rules/TSqlCallMapper.scala
jimidle: Let me look again. I tried that first  but because of some call sequence it ended up not translating and now I do not remember why. So I switched back to CallFunction, which then translates as it goes back through the mapper. I agree with you though - let me find out why.

jimidle: Mappers fixed to pass through already resolved conrete functions and CallFunction IR replaced with concrete classes.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/rules/TSqlCallMapper.scala
```diff
@@ -8,7 +8,43 @@ class TSqlCallMapper extends CallMapper {
...
+    }
+  }
+
+  /**
+   * Converts a CHECKSUM_ARG call to a MD5 function call sequence
+   * @param args
+   * @return
+   */
+  private def checksumAgg(args: Seq[Expression]): Expression = {
+    Md5(ConcatWs(Seq(Literal(","), CollectList(args.head))))
```
nfx, line 30: beautiful

nfx (approved):
lgtm

PR #771, opened 2024-08-05
Added Translation Support for `!` as `commands` and `&` for `Parameters`

### LINKED ISSUE #165: Common Snowflake syntax conversion errors

sriram251-code:
closes #165 

- In the Snowflake code, the setting option `!set exit_on_error=true` caused remorph to return an error. We have now modified remorph to comment out this setting, treating it as a command. As a result, instead of producing an error, the output will now append  '--' before the command .
ex:
Snowflake input sql : `!set exit_on_error=true`
Remorph output: `-- !set exit_on_error=true`
 
- In Snowflake, passing a parameter such as `select emp_id from abc.emp where emp_id = &ids;` caused remorph to give an error due to the unexpected `&` symbol. We have now implemented a change to convert `&` to its Databricks equivalent `$` which represents parameters, resolving this issue.
ex: 
Snowflake input SQL: `select emp_id from abc.emp where emp_id = &ids;`
Remorph output: `select emp_id from abc.emp where emp_id = $ids;`


jimidle (changes requested):
Please expand on the PR description with better grammar etc, if this is a public facing PR.

Also, the CI build fails. Please check that the build works before changing PRs from draft.

sriram251-code:
As per comment necessary changes done

sundarshankar89 (approved):
LGTM

nfx (changes requested):
also add coverage in ANTLR grammar - https://github.com/databrickslabs/remorph/tree/main/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake and map it into UnresolvedCommand IR

sriram251-code:
I have added the antlr grammar to parse the commands with  "!" BANG prefix. This should ensure that this will get captured in unresolved CommandIR. Please le me know if you want me to add explecit test case for it.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -146,6 +146,10 @@ otherCommand
     | beginTxn
     | declareCommand
     | let
+    | scriptingCommand
+    ;
+
+scriptingCommand: BANG set | STRING
```
nfx, line 152: this rule is incorrect. it has to be something like `BANG STRING`. The name `scriptingCommand` is also incorrect, it has to be  snowsqlCliCommand - https://docs.snowflake.com/en/user-guide/snowsql-use. Rename it in python code as well.

Add `SnowsqlCliCommand` into `extensions.scala` and code generate it as the UnresolvedCommand. Add 100% unit test coverage.

Also add `&` to the lexer and create SnowsqlCliVariable. Make sure that it's covered in `SnowflakeToDatabricksTranspilerTest`.

tests/resources/functional/snowflake/test_tokens_parameter/test_tokens_parameter_4.sql
```diff
@@ -0,0 +1,5 @@
+-- snowflake sql:
+select &col_name from TEST_USER.EMP_TBL;
+
+-- databricks sql:
+select $col_name from TEST_USER.EMP_TBL;
```
nfx, line 5: where do we use this syntax? please add references to documentation.

tests/resources/functional/snowflake/test_command/test_command_1.sql
```diff
@@ -0,0 +1,6 @@
+
+-- snowflake sql:
+!set exit_on_error = true;
+
+-- databricks sql:
+-- !'set exit_on_error = true';
```
nfx, line 6: generate into: `-- snowsql command: !set exit_on_error = true;`

src/databricks/labs/remorph/snow/snowflake.py
```diff
@@ -228,6 +228,12 @@ class Tokenizer(Snowflake.Tokenizer):
             r"(?i)var\s+\w+\s+=\s+\w+?": TokenType.VAR,
         }
 
+        SINGLE_TOKENS = {
+            **Snowflake.Tokenizer.SINGLE_TOKENS,
+            "&": TokenType.PARAMETER,  # https://docs.snowflake.com/en/user-guide/snowsql-use#substituting-variables-in-a-session
```
nfx, line 233: This is incorrect. This breaks existing snowflake tokens, as sqlglot already maps parameter to `$` - `"$": TokenType.PARAMETER`.

This is purely `snowcli` variable substitution concept. *We need a new token type here.*

nfx (changes requested):
Core impl is missing. make sure that both token types are also covered in `SnowflakeToDatabricksTranspilerTest`.

legacy impl: you must not override the `!` (NOT) token, that already has existing semantics and will break core queries. Add new token type for CLI variable substitutions, because you're using an incorrect generator for them.

src/databricks/labs/remorph/snow/snowflake.py
```diff
@@ -228,6 +228,12 @@ class Tokenizer(Snowflake.Tokenizer):
...
+            "!": TokenType.COMMAND,
```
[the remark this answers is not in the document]
sriram251-code: The program behaves as follows:
1)If the SnowSQL command starts with an exclamation mark !, it will generate the output as commented with --.
Example:

Input: !(2 = 2)
Output: -- !'(2 = 2)'

synatx tree: Command(
  this=!,
  expression=Literal(this=(2 = 2), is_string=True))

2)If the exclamation mark ! appears within the query but not at the starting position, it will generate the equivalent DBSQL output.
Ex:
input: select * from emp where !(2 = 2)
output: SELECT * FROM emp WHERE  !(2 = 2) 
syntax tree: Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Identifier(this=emp, quoted=False))),
  where=Where(
    this=Anonymous(
      this=!,
      expressions=[
        EQ(
          this=Literal(this=2, is_string=False),
          expression=Literal(this=2, is_string=False))])))

tests/resources/functional/snowflake/test_tokens_parameter/test_tokens_parameter_4.sql
sriram251-code: making necessary changes

tests/resources/functional/snowflake/test_command/test_command_1.sql
sriram251-code: done

src/databricks/labs/remorph/snow/snowflake.py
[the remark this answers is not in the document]
nfx: add a unit test to prove it with automation.

```sql
!whatever
SELECT !(2 = 2) AS always_false
```

and make sure it translates.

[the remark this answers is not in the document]
sriram251-code: added the test case

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
sundarshankar89: @nfx  just saw this comment, you want the ir to defined for `snowsqlclicommand` or to resolved as `unresolvedcommand` ?

nfx: unresolved is probably better

src/databricks/labs/remorph/snow/snowflake.py
sriram251-code: for snowflake both '$' and '&' will be converted to "$" in databricks.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -148,6 +155,9 @@ otherCommand
     | let
     ;
 
+snowSqlCommand: BANG batch
+    ;
```
jimidle, line 159: This will parse `! <any valid sql batch>`, which includes an entire set of SQL statements, but does not include many of the actual commands listed in the Snowflake docs at: https://docs.snowflake.com/en/user-guide/snowsql-use

Here is an explanation of why I think we cannot implement this in the ANTLR grammar:

Looking through the actual available commands, they are all pretty standard simple command syntax, so you might first think, given that almost all keywords are allowed as identifiers, what you actually want is akin to:

```antlr
snowSqlCommand: BANG id exprList ;
```

But because some commands accept things that are not actually an expression, you need an expansion on expression for commands like this:

```
user#> !abort 77589bd1-bcbf-4ec8-9ebc-6c949b00614d;
```

Because the UUID is not a part of the standard syntax and we cannot make a token for this as it would eat 4-5. Here you only want to consume it without a syntax error, so specify the sequence and because the ! acts as a gate, it will not affect the rest of the grammar. So you will get to something like this:

```antlr
snowSqlCommand: BANG commandExpr (COMMA? commandExpr)* ;

commandExpr
    : (id | INTEGER)+ DOT  (id | INTEGER)+ DOT  (id | INTEGER)+ DOT  (id | INTEGER)+  // UUID 
    | expr
    ;
```

But this sequence is of indeterminate length as it relies on a newline to terminate it and newlines are not passed thorough to the Snowflake parser. So as written above, it would eat anything that looks like an expression, even on the next line as we humans see it (the parser just sees a sequence of tokens). The list is of indeterminate length because of `!print` and `!system`.  

Next, can we use a simple pre-processor? I think so. Convert it to `SQLCOMMAND "!system df;lkfglkjfklgj kl. "`

The BANG operator isn't a thing in Snowflake SQL, it is always NOT. However, It would be better avoid that. So I think that we need to use either lexer modes for this, or exact matches on tokens.

The problem  is that Snowflake already uses ! as a special character in system stored procedure calls, so if there is a clash there, then a simple lexer rule would not work. You can investigate that CALL:

```antlr
// Snowfllake allows calls to special internal stored procedures, named x.y!entrypoint
call: CALL objectName (BANG id)? L_PAREN exprList? R_PAREN
    ;
```
To see if there are any command clashes. If not then you can do this in the LEXER:

```antlr
SQLCOMMAND: '!' WS* ('abort' | 'system' | etc....)  ~[\r\n]*;
```

Note that you need to spell out all the tokens as otherwise the rule would eat the CALL special !name.

Add a test for each command and make sure to add a test that ensures CALL x.y!FRED() works.

If there are any clashes, then it is probably best if I take a deeper look but: `WS* ~'('` could be added before `~[\n\r]`

sriram251-code: implemented accordingly

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
```diff
@@ -914,4 +914,28 @@ fragment DEC_DIGIT   : [0-9];
...
+        | 'queries'
+        | 'quit'
+        | 'rehash'
+        | 'result'
+        | 'set'
+        | 'source'
+        | 'spool'
+        | 'system'
+        | 'variables'
+    ) ~[\r\n]*
```
jimidle, line 940: I think that this will fail if the input does not have a trailing \n, which can happen on Windows, or tests. This should work though:

```antlr
(~[\r\n] | EOF)*
```

Write a test where the input is just a single `!` with no `\n`, a test where it is `!set` and a test that is `! badcommand` (and checks that a syntax error is thrown'. See `TSqlErrorStategySpec` for testing for syntax errors.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -63,4 +63,14 @@ class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[ir.LogicalPlan] wit
     case c if c.queryStatement() != null => c.queryStatement().accept(this)
     case c => c.accept(dmlBuilder)
   }
+
+  override def visitSnowSqlCommand(ctx: SnowSqlCommandContext): ir.UnresolvedCommand = {
+    val input = ctx.getText.trim
+    ir.UnresolvedCommand(if (input.endsWith(";")) { /* Removing ; as it will get appended in generator */
```
jimidle, line 69: Don't trim or check for ';'. Then in the generator do not add anything to the text of the command - preserve as it was in the input.

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilderSpec.scala
```diff
@@ -429,6 +429,10 @@ class SnowflakeAstBuilderSpec extends AnyWordSpec with SnowflakeParserTestCommon
         UpdateTable(namedTable("t1"), None, Seq(Assign(Id("c1"), Literal(short = Some(42)))), None, None, None))
 
     }
+
+    "translate BANG to Unresolved Expression" in {
+      example("!set error_flag = true;", _.snowSqlCommand(), UnresolvedCommand("!set error_flag = true"))
+    }
```
jimidle, line 435: Add extra tests as above:

```
"!"
"!set dfsdfds"
"!badcommand". // Check for a syntax error
```

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
```diff
@@ -10,6 +10,15 @@ class SnowflakeToDatabricksTranspilerTest extends AnyWordSpec with TranspilerTes
 
     "transpile queries" in {
 
+      /* Test BANG ! command */
+      "!set error_flag = true;" transpilesTo "--!set error_flag = true;"
+      "!define tablename=CENUSTRACKONE;" transpilesTo "--!define tablename=CENUSTRACKONE;"
+      "!print Include This Text" transpilesTo "--!print Include This Text;"
+      "!abort 77589bd1" transpilesTo "--!abort 77589bd1;"
```
jimidle, line 17: Plus the tests as above. It is possible that you may not need the EOF modification to the LEXER token.

jimidle (changes requested):
Please see comments about expended tests and not changing the text of the input command.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
sriram251-code: @jimidle as disscussed this will work fine and no further change needed.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
sriram251-code: done

jimidle:
Just requires the extra simple tests and to resolve the transpiler test merge conflict and we are good.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
sriram251-code: Hi @jimidle, in the generate method within LogicalPlanGenerator.scala, there's a case statement:
case b: ir.Batch => b.children.map(generate(ctx, _)).mkString("", ";\n", ";")

This appends a ";" at the end of the batch.
If I don't remove the ";" in visitSnowSqlCommand, we'll end up with two ";" at the end of the generate method.

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilderSpec.scala
sriram251-code: done

sriram251-code:
@jimidle as disscussed please review :) thanks.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
jimidle: I guess it can be left as is.

jimidle (approved):
lgtm @nfx

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
```diff
@@ -33,6 +33,7 @@ class LogicalPlanGenerator(val expr: ExpressionGenerator, val explicitDistinct:
     case i: ir.InsertIntoTable => insert(ctx, i)
     case ir.DeleteFromTable(target, None, where, None, None) => delete(ctx, target, where)
     case ir.NoopNode => ""
+    case unresolved: ir.UnresolvedCommand => s"--${unresolved.inputText}"
```
nfx, line 36:
```suggestion
    case ir.UnresolvedCommand(inputText) => s"-- ${inputText}"
```
be a bit more scala idiomatic

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
```diff
@@ -10,6 +10,14 @@ class SnowflakeToDatabricksTranspilerTest extends AnyWordSpec with TranspilerTes
 
     "transpile queries" in {
 
+      /* Test BANG ! command */
```
nfx, line 13: create new `should` block right above `transpile queries` in approx the following way:

```
"snowsql commands" should {
   "transpile BANG" in {
      // ... move assertions here
   }
}
```

remove redundant `/* Test BANG ! command */` comments

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
sriram251-code: done

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
sriram251-code: done

nfx (approved):

PR #778, opened 2024-08-08
Improve correctness of snowflake transpilation by improving SnowflakeCallMapper

vil1:


core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -1616,9 +1627,11 @@ case class BoolAnd(left: Expression) extends Unary(left) with Fn {
 }
 
 /** collect_list(expr) - Collects and returns a list of non-unique elements. */
-case class CollectList(left: Expression) extends Unary(left) with Fn {
-  override def prettyName: String = "COLLECT_LIST"
+case class CollectList(expr: Expression, cond: Option[Expression]) extends Expression with Fn {
```
nfx, line 1630: Spark has no `cond` parameter. convert it in the optimizer rule.
```
@ExpressionDescription(
  usage = "_FUNC_(expr) - Collects and returns a list of non-unique elements.",
  examples = """
    Examples:
      > SELECT _FUNC_(col) FROM VALUES (1), (2), (1) AS tab(col);
       [1,2,1]
  """,
  note = """
    The function is non-deterministic because the order of collected results depends
    on the order of the rows which may be non-deterministic after a shuffle.
  """,
  group = "agg_funcs",
  since = "2.0.0")
case class CollectList(
    child: Expression,
    mutableAggBufferOffset: Int = 0,
    inputAggBufferOffset: Int = 0) extends Collect[mutable.ArrayBuffer[Any]] {
```

also - can we change test examples to use COLLECT_LIST, which is more common?

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -252,7 +249,10 @@ class CallMapper {
     case CallFunction("SOUNDEX", args) => SoundEx(args.head)
     case CallFunction("SPACE", args) => StringSpace(args.head)
     case CallFunction("SPARK_PARTITION_ID", _) => SparkPartitionID()
-    case CallFunction("SPLIT", args) => StringSplit(args.head, args(1), args(2))
+    case CallFunction("SPLIT", args) =>
+      val delim = if (args.size >= 3) Some(args(2)) else None
+      StringSplit(args.head, args(1), delim)
+    case CallFunction("SPLIT_PART", args) => StringSplitPart(args.head, args(1), args(2))
```
nfx, line 255: we don't have `StringSplitPart` in Spark. Convert the call in the optimizer rule.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -270,6 +270,7 @@ class CallMapper {
     case CallFunction("TO_CSV", args) => StructsToCsv(args.head, args(1))
     case CallFunction("TO_DATE", args) => ParseToDate(args.head, args(1))
     case CallFunction("TO_JSON", args) => StructsToJson(args.head, args(1))
+    case CallFunction("TO_NUMBER", args) => ToNumber(args.head, args(1))
```
nfx, line 273: we don't have `ToNumber` in Spark. Make the conversion in optimizer rule

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -279,6 +280,7 @@ class CallMapper {
     case CallFunction("TRANSLATE", args) => StringTranslate(args.head, args(1), args(2))
     case CallFunction("TRIM", args) => StringTrim(args.head, args(1))
     case CallFunction("TRUNC", args) => TruncDate(args.head, args(1))
+    case CallFunction("TRY_TO_NUMBER", args) => TryToNumber(args.head, args(1))
```
nfx, line 283: we don't have `TryToNumber ` in Spark. Make the conversion in optimizer rule

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -1292,12 +1294,21 @@ case class SparkPartitionID() extends LeafExpression with Fn {
  * split(str, regex, limit) - Splits `str` around occurrences that match `regex` and returns an array with a length of
  * at most `limit`
  */
-case class StringSplit(left: Expression, right: Expression, c: Expression) extends Expression with Fn {
+case class StringSplit(left: Expression, right: Expression, c: Option[Expression]) extends Expression with Fn {
```
nfx, line 1297: signature cannot change, as we want to keep it as close to Spark as possible.

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -1829,9 +1842,9 @@ case class ArrayIntersect(left: Expression, right: Expression) extends Binary(le
  * array_join(array, delimiter[, nullReplacement]) - Concatenates the elements of the given array using the delimiter
  * and an optional string to replace nulls. If no value is set for nullReplacement, any null value is filtered.
  */
-case class ArrayJoin(left: Expression, right: Expression, c: Expression) extends Expression with Fn {
+case class ArrayJoin(left: Expression, right: Expression, c: Option[Expression]) extends Expression with Fn {
```
nfx, line 1845: signature cannot change

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
```diff
@@ -2422,3 +2435,20 @@ case class RowNumber() extends LeafExpression with Fn {
...
+}
+
+/**
+ * try_to_number(expr, fmt) - Returns expr cast to DECIMAL using formatting fmt, or NULL if expr does not match the
+ * format.
+ */
+case class TryToNumber(expr: Expression, fmt: Expression) extends Binary(expr, fmt) with Fn {
+  override def prettyName: String = "TRY_TO_NUMBER"
+  override def dataType: DataType = UnresolvedType
+}
```
nfx, line 2454: belongs to snowflake package

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -1,16 +1,111 @@
...
+      val fmt = getArg(1).collect { case f @ StringLiteral(_) =>
+        f
+      }
+      val precPos = fmt.fold(1)(_ => 2)
+      val prec = getArg(precPos).collect { case IntLiteral(p) =>
+        p
+      }
+      val scale = getArg(precPos + 1).collect { case IntLiteral(s) =>
+        s
+      }
```
nfx, line 68:
```suggestion
      val fmt = getArg(1).collect { 
        case f @ StringLiteral(_) => f
      }
      val precPos = fmt.fold(1)(_ => 2)
      val prec = getArg(precPos).collect { 
        case IntLiteral(p) => p
      }
      val scale = getArg(precPos + 1).collect { 
        case IntLiteral(s) => s
      }
```
didn't it want to format it like this, btw? having one-letter variables on one line is jarring.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -1,16 +1,111 @@
...
+
+  private def toNumber(args: Seq[ir.Expression]): ir.Expression = {
+    val getArg: Int => Option[ir.Expression] = args.lift
+    if (args.size < 2) {
+      throw TranspileException("not enough arguments to TO_NUMBER")
+    } else if (args.size == 2) {
+      ir.ToNumber(args.head, args(1))
+    } else {
+      val fmt = getArg(1).collect { case f @ StringLiteral(_) =>
+        f
```
nfx, line 60: what would happen if the argument is not the string literal? panic? computer reboot?

nfx (changes requested):
most of the changes to `intermediate` package have to be done in the `snowflake` package instead

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
vil1: In fact we do https://docs.databricks.com/en/sql/language-manual/functions/split_part.html

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
vil1: https://docs.databricks.com/en/sql/language-manual/functions/to_number.html

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
vil1: https://docs.databricks.com/en/sql/language-manual/functions/try_to_number.html

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
vil1: That's the result of auto format, CI won't pass if I commit your suggestion

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
vil1: `fmt` will be `None`

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/functions.scala
vil1: https://docs.databricks.com/en/sql/language-manual/functions/collect_list.html
Screenshot 2024-08-13 at 09 56 45

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
nfx: kk

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
nfx: make a default value for the format.

nfx (approved):
rebase

PR #784, opened 2024-08-12
Aggregates Reconcile Updates

vijaypavann-db:
* Updated `Table Deployment` to accommodate `Aggregate Tables` deployment
* Updated Tables persist logic
* Updated documentation


src/databricks/labs/remorph/resources/reconcile/queries/installation/aggregate_metrics.sql
```diff
@@ -1,11 +1,11 @@
 CREATE TABLE IF NOT EXISTS aggregate_metrics (
     recon_table_id BIGINT NOT NULL,
-    rule_id BIGINT NOT NULL,
+    rule_id BIGINT,
```
sundarshankar89, line 3: When can rule_id be null?

vijaypavann-db: There could be cases where provided column names are not present in the table, In those situations we are capturing the exception in the Delta Tables.

Example below:  
`Runtime exception occurred while fetching data using SELECT min(p_discount) AS source_min_p_discount, max(p_price11) AS source_max_p_price11, count(p_id) AS source_count_p_id FROM hive_metastore.default.product_ctgry : [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter with name **p_price11** cannot be resolved. Did you mean one of the following? `

vijaypavann-db: reverted `rule_id` to NOT NULL, as per the discussion.

src/databricks/labs/remorph/reconcile/connectors/databricks.py
```diff
@@ -57,7 +57,7 @@ def read_data(
         if catalog:
             table_with_namespace = f"{catalog}.{schema}.{table}"
         else:
-            table_with_namespace = f"{schema}.{table}"
+            table_with_namespace = f"hive_metastore.{schema}.{table}"
```
nfx, line 60: simplify with 

```
if not catalog:
  catalog = "hive_metastore"
table_with_namespace = f"{catalog}.{schema}.{table}"
```

vijaypavann-db: Done

sundarshankar89 (approved):
LGTM

PR #786, opened 2024-08-12
Corrections to Snowflake grammar for explicit precedence

jimidle:
PR #548 unfortunately incorrectly removed support for parenthesized expressions. 

This was not caught because we had no coverage tests for expression precedence so an expression such as:

```sql
(a + b) * c
```

Generates a syntax error.

The PR also:
  - removed `bracketExpression` from the `expr` rule, but left it in the grammar despite being unused (removing it would have pointed out its real use in the IR generator)
  - Incorrectly duplicated `COLON id` 

Here we reinstate expression precedence correctly, add test coverage for the grammar changes that were made, and correct the duplication of `COLON id`


nfx (approved):
lgtm

PR #788, opened 2024-08-12
TSQL: Implement grammar for CREATE DATABASE and CREATE DATABASE SCOPED OPTION

jimidle:
The TSQL grammar for `CREATE DATABASE` and `CREATE DATABASE SCOPED OPTION` was not consistent with the TSQL documentation.  Here we correct the grammar errors.

While it was originally my intention to implement these all the way through to code gen, the DDL needs some serious work, and I would rather implement the grammar for it all in a series of grammar only PRs, then come back through and implement all the way to codegen in a subsequent set of PRs. The reason being that as each DDL statement is fixed at the grammar level, all the other DDL statements will gradually simplify at the same time and I do not wish to keep changing grammar then have to tweak previously defined code gen.


nfx (approved):
lgtm

PR #792, opened 2024-08-13
Make non-standard default for Snowflake rank-related functions' window frame explicit

vil1:


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
```diff
@@ -309,6 +310,33 @@ class SnowflakeExpressionBuilder()
       frame_spec = frameSpec)
   }
 
+  // see: https://docs.snowflake.com/en/sql-reference/functions-analytic#list-of-window-functions
+  private val SnowflakeRankRelatedWindowFunctions = Set(
```
nfx, line 314:
```suggestion
  private val rankRelatedWindowFunctions = Set(
```
is it CONSTANT_CASE or a ClassCase? i'm confused what this is. let's remove `snowflake` prefix, as we're already in the snowflake package.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeExpressionBuilder.scala
```diff
@@ -309,6 +310,33 @@ class SnowflakeExpressionBuilder()
...
+
+  /**
+   * For rank-related window functions, snowflake's default frame deviate from ANSI standard. So in such case, we must
+   * make the frame specification explicit. see:
+   * https://docs.snowflake.com/en/sql-reference/functions-analytic#usage-notes-for-window-frames
+   */
+  private def snowflakeDefaultFrameSpec(windowFunction: ir.Expression): Option[ir.WindowFrame] = {
+    val rankRelatedDefaultFrameSpec = ir.WindowFrame(ir.RowsFrame, ir.UnboundedPreceding, ir.UnboundedFollowing)
+    windowFunction match {
+      case fn: ir.Fn if rankRelatedWindowFunctions.contains(fn.prettyName) => Some(rankRelatedDefaultFrameSpec)
```
nfx, line 335:
```suggestion
    windowFunction match {
      case fn: ir.Fn if rankRelatedWindowFunctions.contains(fn.prettyName) => 
        val rankRelatedDefaultFrameSpec = ir.WindowFrame(ir.RowsFrame, ir.UnboundedPreceding, ir.UnboundedFollowing)
        Some(rankRelatedDefaultFrameSpec)
```
forgotten to nit this on performance.

nfx (approved):

PR #796, opened 2024-08-13
TSQL: Implement correct grammar for CREATE TABLE in all forms

jimidle:
Here we correct the existing grammar for CREATE TABLE in all its forms such as graph nodes, CTAS and FILETABLE.


jimidle:
Note grammar only until the whole CREATE xxxx set is cleaned up and made orthogonal.

nfx (approved):
lgtm

PR #801, opened 2024-08-15
Root Table Identifier

### LINKED ISSUE #23: Based on a set of SQL queries, determine a set of root tables
Root table is a dataset that is not created within the provided folder of queries. It's availability is the blocker for migration.

Root table has no upstream dependencies.

Image

sundarshankar89:
Closes #23 

Depends on #812


core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
```diff
@@ -19,4 +20,14 @@ abstract class BaseTranspiler extends Transpiler {
...
+
+object TranspilerFactory {
+  def getTranspiler(engine: String): Transpiler = {
+    engine.toLowerCase() match {
+      case "tsql" => new TSqlToDatabricksTranspiler()
+      case "snowflake" => new SnowflakeToDatabricksTranspiler()
+      case _ => throw new IllegalArgumentException(s"Unsupported source database: $engine")
+    }
+  }
+}
```
nfx, line 33:
```suggestion
```
leads to anti-patterns.

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,51 @@
+package com.databricks.labs.remorph
+
+import com.databricks.labs.remorph.parsers.{intermediate => ir}
+import com.databricks.labs.remorph.transpilers.TranspilerFactory
+
+class RootTableIdentifier(engine: String) {
+
+  private val transpiler = TranspilerFactory.getTranspiler(engine)
```
nfx, line 8:
```suggestion
class RootTableIdentifier(Transpiler transpiler) {
```
it doesn't seem that you fully understood my previous comments about dependency injection. let's prohibit any use of Factory pattern other than ContextFactory, because i don't see the pattern used in an effective way.

this code design is bad, as one is coupled to invariants defined by a factory class and nothing else than that. what if we want to test with transpiler that always fails only on second query?

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,51 @@
...
+    var child = ""
+    var parent = Set.empty[String]
+
+    def collectTables(node: ir.LogicalPlan): Unit = {
+      node match {
+        case n: ir.InsertIntoTable =>
+          child = n.target
+            .find(_.isInstanceOf[ir.NamedTable])
+            .map(_.asInstanceOf[ir.NamedTable].unparsed_identifier)
+            .getOrElse("unknown") // Unknown is when query is a Select Query. May be file name becomes the table name.
```
nfx, line 30: make it an argument to the function, because we may get more than one unknown, which in fact point to many tables.

core/src/main/scala/com/databricks/labs/remorph/Dag.scala
```diff
@@ -0,0 +1,30 @@
+package com.databricks.labs.remorph
+import scala.collection.mutable
+
+case class Node(name: String)
+
+case class Edge(from: Node, to: Node)
+
+class DAG {
```
nfx, line 8: DAG is too generic name not to have generic parameters.

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,51 @@
...
+    p.find(_.isInstanceOf[ir.NamedTable]).map(_.asInstanceOf[ir.NamedTable].unparsed_identifier).toSet
+  }
+
+  private def findTables(dag: DAG, plan: ir.LogicalPlan): DAG = {
+    var child = ""
+    var parent = Set.empty[String]
+
+    def collectTables(node: ir.LogicalPlan): Unit = {
+      node match {
+        case n: ir.InsertIntoTable =>
```
nfx, line 26:
```suggestion
        case ir.NamedTable(name, _, _) =>
```
why do you need so many unnecessary branches?..

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,51 @@
...
+      node.children.foreach(collectTables)
+    }
+
+    collectTables(plan)
+    val childNode = Node(child)
+    dag.addNode(childNode)
+    parent.foreach { parentTable =>
+      val parentNode = Node(parentTable)
+      dag.addNode(parentNode)
+      dag.addEdge(parentNode, childNode)
```
nfx, line 46: we need more infromation on the edge - e.g. read and write. one query plan will always result in zero or more triplets. 

For example: `SELECT * FROM a LEFT JOIN b RIGHT JOIN c` will result in:
* `a -> read -> b`
* `a -> read -> c`

And `CREATE TABLE a AS SELECT FROM b` will result in:
* `b -> write -> a`

core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
```diff
@@ -4,11 +4,12 @@ import com.databricks.labs.remorph.parsers.{intermediate => ir}
 
 trait Transpiler {
   def transpile(input: String): String
+  def parse(input: String): ir.LogicalPlan
```
nfx, line 7: i am not sure we have to add `parse` method to this public interface.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: Made changes.

sundarshankar89: Took care.

core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
sundarshankar89: I need `parse` output to create DAG? should I create a `parse` object and refactor all transpiler with that?

core/src/main/scala/com/databricks/labs/remorph/Dag.scala
sundarshankar89: how about `TableLineage` ?

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,48 @@
...
+
+class RootTableIdentifier(transpiler: Transpiler) {
+
+  def processQuery(sql: String, dag: Lineage): Lineage = {
+    val parsedSql = transpiler.parse(sql)
+    findTables(dag, parsedSql)
+  }
+
+  private def getTablName(p: ir.LogicalPlan): Set[String] = {
+    p.find(_.isInstanceOf[ir.NamedTable]).map(_.asInstanceOf[ir.NamedTable].unparsed_identifier).toSet
```
vil1, line 14:
```suggestion
    p.collect{
      case nt: ir.NamedTable => nt.unparsed_identifier
    }.toSet
```

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: in the above example?

For example: SELECT * FROM a LEFT JOIN b RIGHT JOIN c will result in:

`a -> read -> b`
`a -> read -> c`

how are b and c related should there `b -> read -> c` as well
And `CREATE TABLE a AS SELECT FROM b` will result in:

`b -> write -> a`

and in below example  `CREATE TABLE d AS SELECT FROM a LEFT JOIN b RIGHT JOIN c`
Should it be 

`d -> read -> a` or  'a -> write -> d`

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: Tackled

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: resolved

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: Done

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  private def getTableFromExpression(p: ir.Expression): Set[String] = {
+    // TODO Tackle Unresolved Expressions
+    p.collect { case e: ir.ScalarSubquery =>
+      fetchTableName(e.relation)
+    }.toSet
+  }
+
+  private def getTableList(p: ir.LogicalPlan): Set[String] = {
```
nfx, line 24: why not a full name? add
```
case class FullName(catalog: String, schema: String, table: String)
```
and emit those instead

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,79 @@
...
+  private def findTables(graph: Lineage, plan: ir.LogicalPlan): Lineage = {
+    var child: ChildActionMap = null
+    var parent = Set.empty[String]
+
+    def updateChild(target: ir.LogicalPlan, action: Action.Operation): ChildActionMap = {
+      ChildActionMap(fetchTableName(target), action)
+    }
+
+    def collectTables(node: ir.LogicalPlan): Unit = {
+
```
nfx, line 45:
```suggestion
```

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,79 @@
+package com.databricks.labs.remorph
+
+import com.databricks.labs.remorph.parsers.{intermediate => ir}
+import com.databricks.labs.remorph.transpilers.Transpiler
+
+private case class ChildActionMap(name: String, action: Action.Operation)
+
+class RootTableIdentifier(transpiler: Transpiler) {
```
nfx, line 8:
```suggestion
class RootTableIdentifier(transpiler: Transpiler, queries: Seq[String]) {
```
1. let's make this interface more robust and immutable by shifting queries to constructor arg
2. introduce `Parser` interface with `parse(String): LogicalPlan` as a single public method.

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  def processQuery(sql: String, graph: Lineage): Lineage = {
```
nfx, line 10: hide this from public api

core/src/test/scala/com/databricks/labs/remorph/RootTableIdentifierTest.scala
```diff
@@ -0,0 +1,254 @@
...
+    val resultDAG = rootTableIdentifier.processQuery(sql, dag)
+    val expectedNodes = Set(Node("output_table"), Node("input_table_1"), Node("input_table_2"))
+    val expectedEdges =
+      Set(
+        Edge(Node("input_table_1"), Action.Write, Node("output_table")),
+        Edge(Node("input_table_2"), Action.Write, Node("output_table")))
+
+    assert(resultDAG.getNodes == expectedNodes)
+    assert(resultDAG.getEdges == expectedEdges)
+  }
```
nfx, line 35: this class design is brittle, as it leaks graph to the consumer of the instance. Public API has to be:

```
rootDiscovery = RootTableIdentifier(parser, Seq(
  "SELECT ... query1",
  "CREATE TABLE query2 AS SELECT ...",
"CREATE TABLE query3 AS SELECT ...",
"CREATE TABLE query4 AS SELECT ...",
  "SELECT ... query5",
))

assert(rootDiscovery.roots == Seq("query2", "query87"))
assert(rootDiscovery.sorted == Seq("query2", "query87", "query89", "query2", "query1", "query243", ...))
```

current test cases are not good enough, as they focus on the wrong thing and leak too much implementaiton detail.

nfx (changes requested):
change interfaces of root identifier not to leak implementation detail

core/src/main/scala/com/databricks/labs/remorph/Lineage.scala
```diff
@@ -0,0 +1,35 @@
+package com.databricks.labs.remorph
+import scala.collection.mutable
+
+case object Action extends Enumeration {
+  type Operation = Value
+  val Read, Write, Update, Delete, Merge = Value
+}
+
+case class Node(name: String)
```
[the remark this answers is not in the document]
sundarshankar89: Done

core/src/main/scala/com/databricks/labs/remorph/Lineage.scala
```diff
@@ -0,0 +1,35 @@
+package com.databricks.labs.remorph
```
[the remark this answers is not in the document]
sundarshankar89: Done

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: Done

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: We shouldn't move the queries to Contructor, we can have `queries`, `etl` or `workflows` 
I didn't understand the need parser interface.

core/src/test/scala/com/databricks/labs/remorph/RootTableIdentifierTest.scala
sundarshankar89: Done

core/src/main/scala/com/databricks/labs/remorph/RootTableIdentifier.scala
sundarshankar89: I have created a parser interface added them in `transpiler` let me know if you want me to implement parser separately.

core/src/main/scala/com/databricks/labs/remorph/graph/lineage.scala
```diff
@@ -0,0 +1,41 @@
...
+
+case object Action extends Enumeration {
+  type Operation = Value
+  val Read, Write, Update, Delete, Merge = Value
+}
+
+case class Node(catalog: Option[String], schema: String, table: String)
+
+case class Edge(from: Node, operator: Action.Operation, to: Node)
+
```
sriram251-code, line 13: why edge needed? We can have Node and from: Node, to: Node can and  operator: Action.Operation can be the properties of the Node.

core/src/main/scala/com/databricks/labs/remorph/graph/RootTableIdentifier.scala
```diff
@@ -0,0 +1,117 @@
...
+import com.databricks.labs.remorph.graph.{Action, Lineage, Node}
+import com.databricks.labs.remorph.transpilers.Parser
+
+private case class FullTableIdentifier(catalog: Option[String], schema: String, table: String)
+private case class ChildParentMap(
+    child: FullTableIdentifier,
+    action: Action.Operation,
+    parent: Set[FullTableIdentifier])
+
+class RootTableIdentifier(parser: Parser, queries: Seq[String]) {
```
nfx, line 13: use `PlanParser` interface

nfx (changes requested):

sundarshankar89:
too divergent, will close and open a new branch

PR #808, opened 2024-08-19
Fix translation of Snowflake's join types

vil1:
Previously, `NATURAL` and `CROSS` joins were not properly handled, this change fixes it.


nfx (approved):
lgtm

PR #810, opened 2024-08-19
[snowflake] fix data type mapping and explain `UnparsedType`

nfx:
This PR increases coverage of snowflake data types.


coverage/src/main/scala/com/databricks/labs/remorph/coverage/AcceptanceTestSource.scala
```diff
@@ -10,6 +11,22 @@ trait AcceptanceTestSource {
...
+    case p => p.toString
+  }
+
+  @tailrec private def checkProjectRoot(current: Path): Path = {
+    // check if labs.yml exists in the current folder
+    if (Files.exists(current.resolve("labs.yml"))) {
+      current
+    } else {
+      checkProjectRoot(current.getParent)
+    }
```
vil1, line 26: Suggested version would work even in the (arguably very unlikely) case where `labs.yml` is located directly under `/`.

```suggestion
  def projectRoot: String = checkProjectRoot(Paths.get(".")).toString

  @tailrec private def checkProjectRoot(current: Path): Path = {
    // check if labs.yml exists in the current folder
    if (Files.exists(current.resolve("labs.yml"))) {
      current
    } else {
      if (current.getParent != null) {
        checkProjectRoot(current.getParent)
      } else {
        throw new RuntimeException("Could not find project root")
      }
    }
```

vil1 (approved):
lgtm (despite a nit-picking comment about some unlikely corner-case)

PR #812, opened 2024-08-19
TSQL: Implement CREATE TABLE

jimidle:
Here, we implement transpilation of the TSQL `CREATE TABLE` command with its many options and forms, including CTAS and graph node syntax, as well as covering syntactical differences for the analytics variants of SQL Server.

We also move the DDL and DML visitors from the AST and Relation visitors, to give better separation of responsibility.


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlDDLBuilder.scala
```diff
@@ -0,0 +1,42 @@
+package com.databricks.labs.remorph.parsers.tsql
+
+import com.databricks.labs.remorph.parsers.{OptionAuto, OptionExpression, OptionOff, OptionOn, OptionString, ParserCommon, intermediate => ir}
+
+import scala.collection.JavaConverters.asScalaBufferConverter
+
+class TSqlDDLBuilder extends TSqlParserBaseVisitor[ir.Catalog] with ParserCommon[ir.Catalog] {
+
+  private val optionBuilder = new OptionBuilder(new TSqlExpressionBuilder)
```
nfx, line 9: inject via constructor instead of always initialising. this will be useful once we parametrise all those AST builders to support different sub versions of a dialect

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlDMLBuilder.scala
```diff
@@ -0,0 +1,194 @@
...
+import com.databricks.labs.remorph.parsers.tsql.TSqlParser._
+import com.databricks.labs.remorph.parsers.tsql.rules.InsertDefaultsAction
+import com.databricks.labs.remorph.parsers.{intermediate => ir}
+
+import scala.collection.JavaConverters.asScalaBufferConverter
+
+class TSqlDMLBuilder extends TSqlParserBaseVisitor[ir.Modification] {
+
+  private val expressionBuilder = new TSqlExpressionBuilder
+  private val relationBuilder = new TSqlRelationBuilder
```
nfx, line 12: inject via constructor

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlDDLBuilder.scala
jimidle: No problem.

jimidle:
Replaced with a better one

PR #817, opened 2024-08-19
[snowflake] `UPDATE` as `MERGE INTO`

nfx:


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/UpdateToMerge.scala
```diff
@@ -0,0 +1,11 @@
+package com.databricks.labs.remorph.parsers.snowflake.rules
+
+import com.databricks.labs.remorph.parsers.intermediate.{Assign, LogicalPlan, MergeIntoTable, Rule, UpdateAction, UpdateTable}
+
+class UpdateToMerge extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = plan transform {
+    case UpdateTable(target, Some(source), assigns, Some(where), _, _) =>
+      val set = assigns.filter(_.isInstanceOf[Assign]).map(_.asInstanceOf[Assign])
```
vil1, line 8: nit:
```suggestion
      val set = assigns.collect{ case a: Assign => a }
```

nfx: thanks!

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
```diff
@@ -97,38 +100,37 @@ class LogicalPlanGenerator(val expr: ExpressionGenerator, val explicitDistinct:
     s"(${generate(ctx, setOp.left)}) $op$duplicates (${generate(ctx, setOp.right)})"
   }
 
-  private def generateMerge(ctx: GeneratorContext, mergeIntoTable: ir.MergeIntoTable) = {
+  private def update(ctx: GeneratorContext, update: UpdateTable): String = {
+    val target = generate(ctx, update.target)
+    val source = update.source.map(generate(ctx, _)).getOrElse("")
+    val set = update.set.map(expr.generate(ctx, _)).mkString(", ")
+    val where = update.where.map(expr.generate(ctx, _)).getOrElse("")
+    s"UPDATE $target SET $set $source WHERE $where"
```
vil1, line 108: Is this correct? [the documentation](https://docs.databricks.com/en/sql/language-manual/delta-update.html) doesn't seem to specify anything appearing between the assignments (introduced by `SET`) and the predicates (introduced by `WHERE`). 

Also, this code would ouptut `WHERE` even if no predicate is specified.

nfx: thanks!

PR #818, opened 2024-08-20
Validation Strategy Approach for Transpiled Queries

sundarshankar89:
Validation Strategy Approach for Transpiled Queries using Catalyst Parser

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -81,7 +81,6 @@
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-catalyst_${scala.binary.version}</artifactId>
       <version>${spark.version}</version>
-      <classifier>tests</classifier>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
@@ -3,9 +3,23 @@ package com.databricks.labs.remorph.transpilers
 import com.databricks.labs.remorph.parsers.{intermediate => ir}
 import com.github.vertical_blank.sqlformatter.SqlFormatter
 import com.github.vertical_blank.sqlformatter.core.FormatConfig
+import org.apache.spark.sql.catalyst.parser.CatalystSqlParser
 
 trait Transpiler {
-  def transpile(input: String): String
+  def transpile(input: String, skipValiation: Boolean = false): String
+}
+
+trait Validator {
+  def validate(transpiledCode: String): String = {
+    try {
+      CatalystSqlParser.parsePlan(transpiledCode)
+      transpiledCode
+    } catch {
+      // TODO Enhance custom messaging to support complex errors we encounter
+      case e: Exception =>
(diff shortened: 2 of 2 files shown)


sundarshankar89:
@nfx  or @vil1  Thoughts?

core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
```diff
@@ -3,9 +3,23 @@ package com.databricks.labs.remorph.transpilers
 import com.databricks.labs.remorph.parsers.{intermediate => ir}
 import com.github.vertical_blank.sqlformatter.SqlFormatter
 import com.github.vertical_blank.sqlformatter.core.FormatConfig
+import org.apache.spark.sql.catalyst.parser.CatalystSqlParser
 
 trait Transpiler {
-  def transpile(input: String): String
+  def transpile(input: String, skipValiation: Boolean = false): String
```
nfx, line 9: don't pollute existing interface. validation is a separate stage and has to be a separate interface, if we need it

core/pom.xml
```diff
@@ -81,7 +81,6 @@
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-catalyst_${scala.binary.version}</artifactId>
       <version>${spark.version}</version>
-      <classifier>tests</classifier>
```
nfx, line 84: this will increase the size of the remorph jar by hundreds of megabytes - do we want it now?

nfx (changes requested):

core/pom.xml
sundarshankar89: This is a proposal, or a idea to validate transpiled queries. 
We can discuss I would really like to have it and add block comments on which query potentially fail during validation.

core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
sundarshankar89: We can align of interface I wanted to come up with small though process on validation strategy for transpiled queries.

vil1:
Is transpiled query validation a feature we want to provide to our users, or just something that we want to use internally to make sure we produce syntactically-correct DB SQL?

In the latter case, I agree with @nfx: we should put that in a separate module so that we don't ship spark's dependencies (aka. half the Internet) along with remorph.

In the former case, we should still keep it in a separate module (unless we figure a way to perform validation without pulling spark) so that users can still use remorph without validation (or opt-in for validation at the cost of a bigger download). An alternate solution would be to provide validation as a web-service, but that would require hosting said web-service (which shouldn't be that big of an issue) but wouldn't work for users whose queries are too sensitive to be sent over a public network.

About the `Transpiler` interface, I also agree with @nfx that we shouldn't change it. A simple approach would be to define it as a subclass of `Transpiler`:

```scala
class ValidatingTranspiler(transpiler: Transpiler) extends Transpiler {
  override def transpile(input: String): String = {
    val transpiledQuery = transpiler.transpile(input)
    validate(transpiledQuery)
  }
}
```

sundarshankar89:
Will open up design document for this,.

PR #821, opened 2024-08-20
[chore] cleanup snowflake functional tests

nfx:
Having 300+ folders in a single folder is unmaintainable. This PR moves files around in a bit more nested structure, as well as reformatting complex cases.


sundarshankar89 (approved):
LGTM

PR #841, opened 2024-08-22
Improve coverage test success rate around snowflake's conversion functions

vil1:
Two tests are still failing, those related to the translation of Snowflake's  `TO_TIMESTAMP` function. 

To fix these I figured that we first need to agree on:
- how we convert the provided time format. For instance, remorph-py converts `yyyy-MM-dd` to `yyyy-MM-d`. Both format have the same effect (in the sense that in DB SQL, `TO_TIME(x, 'yyyy-MM-dd')` and `TO_TIME(x, 'yyyy-MM-d')` yield the same result), but are syntactically different. Are we sure this conversion is necessary?
- how do we infer the format string (that is required by `TO_TIME`) when none is provided to `TO_TIMESTAMP` ?


nfx (approved):
lgtm

PR #845, opened 2024-08-23
Changing String data type to Varchar for validation dataset

aman-db:
Changing the String datatype to varchar for validation dataset. Using varchar is better as we have this datatype across multiple databases.


nfx (approved):
lgtm

PR #856, opened 2024-08-26
Fix CommentBasedQueryExtractor so that it works as intended even when…

vil1:
… there are more than 2 dialects in the file


nfx (approved):
lgtm

PR #864, opened 2024-08-28
unresolved commands `alter session | stream...` `create stream` `create task`

sundarshankar89:
`alter session | stream...` `create stream` `create task` and `execute task`


core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/unresolved.scala
```diff
@@ -37,7 +37,7 @@ case class UnresolvedExtractValue(child: Expression, extraction: Expression) ext
   override def dataType: DataType = UnresolvedType
 }
 
-case class UnresolvedCommand(inputText: String) extends LeafNode with Command {
+case class UnresolvedCommand(inputText: String) extends Catalog with Command {
```
sundarshankar89, line 40: This is because All Commands are a Catalog Operations.

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
```diff
@@ -36,7 +36,6 @@ class LogicalPlanGenerator(val expr: ExpressionGenerator, val explicitDistinct:
     case c: ir.CreateTableCommand => createTable(ctx, c)
     case t: ir.TableSample => tableSample(ctx, t)
     case ir.NoopNode => ""
-    case ir.UnresolvedCommand(inputText) => s"--${inputText}"
```
nfx, line 39: this looks like a regression, rebase.

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
```diff
@@ -37,6 +37,7 @@ class LogicalPlanGenerator(val expr: ExpressionGenerator, val explicitDistinct:
     case t: ir.TableSample => tableSample(ctx, t)
     case a: ir.AlterTableCommand => alterTable(ctx, a)
     case ir.NoopNode => ""
+    case ir.UnresolvedCommand(inputText) => s"--${inputText}"
```
sundarshankar89, line 40: Added this back, as unresolvedCommand Node can be surfaced in between any node traversal?

Only Immediate Children is tackled inside `batch` method.

sriram251-code: We are handlelling it inside the batch(ctx, b),
`  private def batch(ctx: GeneratorContext, b: Batch): String = {
    val seqSql = b.children
      .map {
        case ir.UnresolvedCommand(text) =>
          s"-- ${text.stripSuffix(";")}"
        case query => s"${generate(ctx, query)}"
      }
      .filter(_.nonEmpty)
    if (seqSql.nonEmpty) {
      seqSql.mkString("", ";\n", ";")
    } else {
      ""
    }`

 As "generate" is a recursive function. We can remove the  `case ir.UnresolvedCommand(inputText) => s"--${inputText}".`

core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
sundarshankar89: @nfx  see the comment below.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -154,9 +155,23 @@ class SnowflakeDDLBuilder
...
 
-  override def visitAlterSession(ctx: AlterSessionContext): ir.Catalog = {
+  override def visitAlterSession(ctx: AlterSessionContext): ir.UnresolvedCommand = {
     // Added replace formatting the incoming text a=b to a = b
-    ir.UnresolvedCatalog("ALTER SESSION SET " + ctx.sessionParams().getText.replace("=", " = "))
+    val alterSession = formatContextString(ctx)
+    ir.UnresolvedCommand(alterSession)
+  }
+
+  private def formatContextString(ctx: ParserRuleContext) = {
```
sriram251-code, line 164: Can we include it in Utils so that it can be reused?

sundarshankar89: Done

sriram251-code (approved):
lgmt

core/src/main/scala/com/databricks/labs/remorph/utils/Strings.scala
```diff
@@ -60,4 +62,25 @@ object Strings {
...
+
+  /**
+   * Formats the text of a given `ParserRuleContext` by concatenating the text of all its children nodes,
+   * separated by spaces.
+   *
+   * @param ctx the `ParserRuleContext` to format.
+   * @return a `String` representing the concatenated text of all children nodes, separated by spaces.
+   *         If the context has no children, returns an empty string.
+   */
+  def formatContextString(ctx: ParserRuleContext): String = {
```
nfx, line 74: this method belongs better to com.databricks.labs.remorph.parsers.{IncompleteParser or ParserCommon}. I also do think that we should return not a String, but rather `UnresolvedCommand`.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/utils/Strings.scala
sundarshankar89: ok will reorganise.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
```diff
@@ -941,6 +941,8 @@ SQLCOMMAND:
     ) ~[\r\n]*
 ;
 
+ANY_BANG: BANG ID ~[\r\n]*;
```
nfx, line 944: remove from this PR, it's irrelevant to the description of PR.

@jimidle , please take a look at this separately. is there a better way of doing `SQLCOMMAND`, so that `ANY_BANG` won't interfere with it?

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -149,10 +149,13 @@ otherCommand
     | call
     | beginTxn
     | declareCommand
-    | let
+    | let // to wide rule need to look at the let rule precedence.
+    ;
+
+snowSqlCommand: SQLCOMMAND | unresolvedBangCommnad
```
nfx, line 155: irrelevant to this PR

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -64,7 +65,14 @@ class SnowflakeAstBuilder extends SnowflakeParserBaseVisitor[ir.LogicalPlan] wit
...
 
+  override def visitOtherCommand(ctx: OtherCommandContext): ir.LogicalPlan = {
+    ctx.accept(commandBuilder)
+  }
+
   override def visitSnowSqlCommand(ctx: SnowSqlCommandContext): ir.UnresolvedCommand = {
-    ir.UnresolvedCommand(ctx.getText)
+    ctx match {
+      case c if ctx.SQLCOMMAND() != null => ir.UnresolvedCommand(ctx.getText)
+      case _ => null
```
nfx, line 75: why can't it be unresolved command in all cases?

tests/resources/functional/snowflake/test_skip_unsupported_operations/test_skip_unsupported_operations_9.sql
```diff
@@ -2,12 +2,12 @@
 -- snowflake sql:
 
                 CREATE TASK t1
```
nfx, line 4: these we should parse into IR for SQL notebook + job creation later on.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
sundarshankar89: Reverted

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
sundarshankar89: Reverted

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
sundarshankar89: Reverted will raise a seperate PR

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
sundarshankar89: Will raise a seperate PR

core/src/main/scala/com/databricks/labs/remorph/parsers/IncompleteParser.scala
```diff
@@ -13,4 +15,40 @@ trait IncompleteParser[T] extends ParseTreeVisitor[T] {
...
+          if (child.isInstanceOf[RuleContext]) {
+            builder.append(formatContextRecursive(child.asInstanceOf[RuleContext]))
+          } else {
+            builder.append(child.getText)
+          }
+          builder.append(" ")
+        }
+        // remove any trailing spaces and some regular expression
+        // TODO Fix this to not include regex rather check based on Generic String Context Expression.
+        builder.toString
```
jimidle, line 44: I have already implemented retrieval of the original input text given a context - however it is not checked in yet:

```scala
package com.databricks.labs.remorph.utils

import org.antlr.v4.runtime.misc.Interval

object ParsingUtils {

  /**
   * Creates a string representation of the text represented by the given ANTLR ParserRuleContext.
   *
   * Note that this should exactly reflect the original input text as bounded by the source interval
   * recorded by the pareser.
   */
  def getTextFromParserRuleContext(ctx: org.antlr.v4.runtime.ParserRuleContext): String =
    ctx.start.getInputStream.getText(new Interval(ctx.start.getStartIndex, ctx.stop.getStopIndex))
}
```

Implementing the same for say two tokens is also trivial. This way you get back the exact input text. There are more nuances to this such as child terminal nodes and error nodes. It is best to retrieve from the input.

I will split that utility into its own PR and submit it.

jimidle (changes requested):
Do not use hand crafted text builder for Context text.

sundarshankar89:
@nfx  ready for review

nfx (approved):
lgtm

PR #875, opened 2024-09-02
Updated Spark setup script

vijaypavann-db:
1. Updated Spark setup script to check whether `spark gzip file` exists

1. Refactored the script to remove the warnings: 
    * `Double quote to prevent globbing and word splitting`
    * `Use 'cd ... || exit' or 'cd ... || return' in case cd fails.`
    * `Consider using 'grep -c' instead of 'grep|wc -l'.`
    * `Usage: sleep seconds`, converted 2m to 120 seconds

1. Tested the following Scenarios:
    1. Scenario: Extracted Spark folder is already present
        Outcome: Directly starts the spark server using `sbin/start-connect-server.sh`
    1. Scenario: Extracted Spark folder **is not present**, and Zip file (`spark-<VERSION>.tgz`) is present
        Outcome:  Extract the zip file and start the spark server
    1. Scenario: Extracted Spark folder **is not present**, and Zip file **is not present**
        Outcome:  Download, Extract and start the spark server

### DIFF
--- a/.github/scripts/setup_spark_remote.sh
+++ b/.github/scripts/setup_spark_remote.sh
@@ -2,31 +2,36 @@
 
 set -xv
 
-mkdir -p $HOME/spark
-cd $HOME/spark
+mkdir -p "$HOME"/spark
+cd "$HOME"/spark || exit
 
 version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark- | tr -d / | sort -r --version-sort | head -1)
 spark=spark-${version}-bin-hadoop3
-spark_connect=spark-connect_2.12
+spark_connect="spark-connect_2.12"
 
-mkdir ${spark}
-cd {spark}
+mkdir "${spark}"
+
 
 SERVER_SCRIPT=$HOME/spark/${spark}/sbin/start-connect-server.sh
 
 ## check the spark version already exist ,if not download the respective version
-if [ -f ${SERVER_SCRIPT} ];then
+if [ -f "${SERVER_SCRIPT}" ];then
   echo "Spark Version already exists"
 else
-  wget https://dlcdn.apache.org/spark/spark-${version}/${spark}.tgz
-  tar -xvf ${spark}.tgz
+  if [ -f "${spark}.tgz" ];then
+    echo "${spark}.tgz already exists"
+  else
+    wget "https://dlcdn.apache.org/spark/spark-${version}/${spark}.tgz"
+  fi
+  tar -xvf "${spark}.tgz"
 fi
 
+cd "${spark}" || exit
 ## check spark remote is running,if not start the spark remote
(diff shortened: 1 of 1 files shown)


.github/scripts/setup_spark_remote.sh
```diff
@@ -2,31 +2,36 @@
 
 set -xv
```
bishwajit-db, line 3: Consider also adding -e to exit on any command failure

.github/scripts/setup_spark_remote.sh
```diff
@@ -2,31 +2,36 @@
...
 
-mkdir -p $HOME/spark
-cd $HOME/spark
+mkdir -p "$HOME"/spark
+cd "$HOME"/spark || exit
 
 version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark- | tr -d / | sort -r --version-sort | head -1)
```
bishwajit-db, line 8: Add error checking after the version assignment. For, example,
if [ -z "$version" ]; then echo "Failed to extract Spark version"; exit 1; fi

.github/scripts/setup_spark_remote.sh
```diff
@@ -2,31 +2,36 @@
...
 else
-  wget https://dlcdn.apache.org/spark/spark-${version}/${spark}.tgz
-  tar -xvf ${spark}.tgz
+  if [ -f "${spark}.tgz" ];then
+    echo "${spark}.tgz already exists"
+  else
+    wget "https://dlcdn.apache.org/spark/spark-${version}/${spark}.tgz"
+  fi
+  tar -xvf "${spark}.tgz"
 fi
```
bishwajit-db, line 27: Should the script clean up the downloaded .tgz file after extraction?

.github/scripts/setup_spark_remote.sh
```diff
@@ -2,31 +2,36 @@
 
 set -xv
 
-mkdir -p $HOME/spark
-cd $HOME/spark
+mkdir -p "$HOME"/spark
+cd "$HOME"/spark || exit
```
bishwajit-db, line 6: Exit with non-zero status if cd fails. Like exit 1. Same for other places.

bishwajit-db (changes requested):

.github/scripts/setup_spark_remote.sh
vijaypavann-db: Keeping it to avoid downloading every time

.github/scripts/setup_spark_remote.sh
vijaypavann-db: Done

.github/scripts/setup_spark_remote.sh
vijaypavann-db: done

.github/scripts/setup_spark_remote.sh
vijaypavann-db: done

sundarshankar89 (approved):
LGTM

bishwajit-db (approved):
LGTM

PR #876, opened 2024-09-02
Add T-SQL verification queries

### LINKED ISSUE #636: Verification examples for TSQL

bishwajit-db:
Add T-SQL verification queries.  Resolves #636 

Currently, the tests are ignored and will be enabled later when the parser starts supporting these queries.


sundarshankar89 (changes requested):
I would prefer to have meaningful filenames

bishwajit-db:
> I would prefer to have meaningful filenames

Yes, that would be better. Right now, it just uses the example file name. However, it will make the process more complex and require another LLM inference step to generate the filenames based on the content. Can we skip it unless absolutely required?

jimidle:
> > I would prefer to have meaningful filenames
> 
> Yes, that would be better. Right now, it just uses the example file name. However, it will make the process more complex and require another LLM inference step to generate the filenames based on the content. Can we skip it unless absolutely required?

I think it is OK to leave this - it is already a massive PR because of the huge number of examples.

jimidle (approved):
It's impossible to check all examples in a review but that is just how it has to be unfortunately.

gueniai:
@bishwajit-db I've made a note on the issue so that we have a placeholder to add the Databricks version of all of these, so that we can go ahead and merge this and then work on the other part on a separate PR.

jimidle (approved):
LGTM - mechanical change

vil1 (approved):
lgtm

sundarshankar89:
@bishwajit-db closing this PR and will delete the branch.

PR #880, opened 2024-09-03
Unresolved Bang

sundarshankar89:
This adds a lexer rule to catch any or all unresolved bang into a bucket `unresolvedSnowSqlCommand` if this is not implemented the grammar for `let` creates them as `unresolvedCommand` which is incorrect assumption.

### DIFF
--- a/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
+++ b/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeLexer.g4
@@ -941,6 +941,8 @@ SQLCOMMAND:
     ) ~[\r\n]*
 ;
 
+ANY_BANG: BANG ID ~[\r\n]*;
+
 STRING_START: '\'' -> pushMode(stringMode);
 
 // This lexer rule is needed so that any unknown character in the lexicon does not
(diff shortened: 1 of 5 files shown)


jimidle (commented):
I do not believe that this will work sundar. I am also of the opinion that we are expecting valid input to the parser such that:

```
!badcommand something something
```

Is OK to reject with a syntax error on the `!`

I am working on a tool chain that will stop when syntax errors are detected and we would not try to generate IR using an incorrect ParseTree, which is what we are doing in places now.

I don't think that we need this change. The CALL issue is a big problem actually - they implemented that without regard to anything else.

sundarshankar89:
> I do not believe that this will work sundar. I am also of the opinion that we are expecting valid input to the parser such that:
> 
> ```
> !badcommand something something
> ```
> 
> Is OK to reject with a syntax error on the `!`
> 
> I am working on a tool chain that will stop when syntax errors are detected and we would not try to generate IR using an incorrect ParseTree, which is what we are doing in places now.
> 
> I don't think that we need this change. The CALL issue is a big problem actually - they implemented that without regard to anything else.

@nfx how do you want to tackle this current framework resolves them unresolvedCommand this rule is added to circumvent that, I agree with the approach but will require separate PR, Can we take this in as Tech Debt and tackle once we have our functional tests running as expected.

sundarshankar89:
@nfx  as discussed I will close this and remove the invalid `bang SQL command`

PR #882, opened 2024-09-03
Add basis of ANTLR utilities, starting with text extractor

jimidle:
Adds a start point for useful ANTLR utilities.

  - Given an ANTLR ParserRuleContext : retrieve the original text from the input source


core/src/main/scala/com/databricks/labs/remorph/utils/ParsingUtils.scala
```diff
@@ -0,0 +1,15 @@
+package com.databricks.labs.remorph.utils
+
+import org.antlr.v4.runtime.misc.Interval
+
+object ParsingUtils {
```
nfx, line 5: shouldn't it be in `ParserCommon`?

core/src/main/scala/com/databricks/labs/remorph/utils/ParsingUtils.scala
```diff
@@ -0,0 +1,15 @@
...
+
+  /**
+   * Creates a string representation of the text represented by the given ANTLR ParserRuleContext.
+   *
+   * Note that this should exactly reflect the original input text as bounded by the source interval
+   * recorded by the pareser.
+   */
+  def getTextFromParserRuleContext(ctx: org.antlr.v4.runtime.ParserRuleContext): String =
```
nfx, line 13:
```suggestion
  def getTextFromParserRuleContext(ctx: ParserRuleContext): String =
```
it might be better like this

core/src/main/scala/com/databricks/labs/remorph/utils/ParsingUtils.scala
jimidle: Can be. I assumed ParserCommon was for IR production stuff, but it can go there.

core/src/main/scala/com/databricks/labs/remorph/utils/ParsingUtils.scala
jimidle: Ther was some reason I spelled it out, but I no longer remember - can do though.

nfx (approved):
lgtm

PR #777, opened 2024-08-07
Upgrade script Implementation

vijaypavann-db:
closes #769

PR #884, opened 2024-09-03
updated remorph reconcile workflow to use wheels instead of pypi

vijaypavann-db:
Currently, the `[REMORPH] Reconciliation Runner` Workflow refers to [remorph pypi](https://pypi.org/project/databricks-labs-remorph/) as a dependent library.

We have come across a few scenarios where: 
* Customer doesn't have public internet access, so they're not able to install the dependent library.
* Not able to use the new features until the release happens

This PR updates the Workflow to use the uploaded wheel in the project root wheels directory `.remorph\wheels` instead of referring to PyPI.


vijaypavann-db:
before: Workflow referring to pypi dependent library
before
--

after: Workflow referring to workspace user project wheels directory `.remorph\wheels` 
after

vijaypavann-db:
merge after #777

src/databricks/labs/remorph/deployment/job.py
```diff
@@ -108,7 +108,9 @@ def _recon_job_settings(
 
     def _job_recon_task(self, jobs_task: Task, recon_config: ReconcileConfig) -> Task:
         libraries = [
-            compute.Library(pypi=compute.PythonPyPiLibrary("databricks-labs-remorph")),
+            compute.Library(
+                whl=f"/Workspace/{self._installation.install_folder()}/wheels/databricks_labs_remorph-{self._product_info.version()}-py3-none-any.whl"
```
nfx, line 112: use the result of `self._wheels.upload_to_wsfs()`, otherwise we won't be able to write integration tests over current branch. e.g. `self._product_info.version()` is `0.5.1+920240404234234`, where `9` is the number of commits from the last tagged release and the rest is a timestamp. see https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/installer/workflows.py#L609

nfx (changes requested):
slightly better implementation advised

src/databricks/labs/remorph/deployment/job.py
vijaypavann-db: Updated the code, please review @nfx

src/databricks/labs/remorph/deployment/installation.py
```diff
@@ -25,8 +25,9 @@ def __init__(
         self._recon_deployment = recon_deployment
 
     def install(self, config: RemorphConfigs):
+        wheel_paths: list[str] = []  # self._upload_wheel()
```
vijaypavann-db, line 28: PR #777 consists the `upload wheel` and `Upgrades` code. This is stacked PR after this.

vijaypavann-db:
PR #777 consists the `upload wheel` and `Upgrades` code. This is stacked PR after #777.

vijaypavann-db:
@nfx Please check this PR

src/databricks/labs/remorph/deployment/job.py
```diff
@@ -100,15 +101,16 @@ def _recon_job_settings(
...
             ],
             "max_concurrent_runs": 2,
             "parameters": [JobParameterDefinition(name="operation_name", default="reconcile")],
         }
 
-    def _job_recon_task(self, jobs_task: Task, recon_config: ReconcileConfig) -> Task:
+    def _job_recon_task(self, jobs_task: Task, recon_config: ReconcileConfig, remorph_wheel_path: str) -> Task:
         libraries = [
-            compute.Library(pypi=compute.PythonPyPiLibrary("databricks-labs-remorph")),
+            compute.Library(whl=f"{remorph_wheel_path}"),
```
nfx, line 113:
```suggestion
            compute.Library(whl=remorph_wheel_path),
```
why do you need to wrap one variable into a string?

tests/unit/deployment/test_job.py
```diff
@@ -60,7 +60,7 @@ def test_deploy_new_job(oracle_recon_config):
     product_info = ProductInfo.from_class(RemorphConfigs)
     name = "Recon Job"
     job_deployer = JobDeployment(workspace_client, installation, install_state, product_info)
-    job_deployer.deploy_recon_job(name, oracle_recon_config)
+    job_deployer.deploy_recon_job(name, oracle_recon_config, "remorph-0.5.0-py3-none-any.whl")
```
nfx, line 63: this will break when we release new versions.

nfx (changes requested):

tests/unit/deployment/test_job.py
vijaypavann-db: I have tested with the latest release, it's not breaking. 
Here we are creating mocked objects and asserting `jobs.create.assert_called_once()`. 
So, I don't think it will fail in new versions. Please advise.

src/databricks/labs/remorph/deployment/job.py
vijaypavann-db: Done

nfx (approved):
lgtm

PR #892, opened 2024-09-06
Rephrase implementation using a monadic `Result`

vil1:
Incidentally, this helped fixing a bug where tests `transpile BANG with negative scenario unknown command` and `transpile BANG with negative scenario unknown command2` were mistakenly expected to return an empty string. 

The changes introduced by the original PR should have detected that as wrong (these tests should actually raise an error), but failed to do so because of a combination of `Option` fields in `Result` and `.getOrElse("")` calls here and there.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
@@ -16,19 +16,12 @@ object WorkflowStage {
   case object GENERATE extends WorkflowStage
 }
 
-case class Result(
-    stage: WorkflowStage = WorkflowStage.PARSE,
-    errorsJson: String = "",
-    tree: Option[ParserRuleContext] = None,
-    plan: Option[ir.LogicalPlan] = None,
-    optimizedPlan: Option[ir.LogicalPlan] = None,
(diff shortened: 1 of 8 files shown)


core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
```diff
@@ -65,45 +54,15 @@ trait Formatter {
...
-    val parsed = parse(input)
-    val visited = processStage(parsed, r => visit(parsed))
-    val optimized = processStage(visited, r => optimize(visited))
-    val generated = processStage(optimized, r => generate(optimized))
+  protected def generate(optimizedLogicalPlan: ir.LogicalPlan): Result[String]
 
-    if (generated.inError()) generated
-    else Result(stage = WorkflowStage.GENERATE, output = Some(format(generated.output.get)))
+  override def transpile(input: SourceCode): Result[String] = {
+    (parse _ >> visit >> optimize >> generate)(input)
```
vil1, line 66:
```suggestion
   for {
     parsed <- parse(input)
     visited <- visit(parsed)
     optimized <- optimize(visited)
     output <- generate(optimized)
   } yield output
```

nfx (approved):
lgtm

PR #897, opened 2024-09-12
Simplify lateral_struct tests to use PARSE_JSON instead of FROM_JSON

vil1:
I've also manually checked that all the `test_lateral_struct_*` tests work on Snowflake and DB and yield the same results


nfx (changes requested):
`ci` is failing

PR #899, opened 2024-09-13
Make CallMapper a Rule[LogicalPlan]

### LINKED ISSUE #898: [FEATURE]: `CallMapper` needs to become a `Rule[LogicalPlan]`

Transpile

### Problem statement

Currently, the (dialect-specific) `CallMapper` is called during the generation phase (by the `ExpressionGenerator`), thus after every other optimizer `Rule` has been applied.

However some optimizer rules may depend on the CallMapper's result. For instance, rules that rely on expressions `dataType` need to be applied after the CallMapper, since it transforms `CallFunction` (that have an `UnresolvedType`) into specific `Fn` nodes (that may have a more precise data type).

We therefore have to pull the CallMapper out the ExpressionGenerator and make it into a Rule so that it can be applied before the rules that depend on it.

### Proposed Solution

The `CallMapper` class could easily become a subclass of `Rule[LogicalPlan]` by simply wrapping its already defined `convert` method (care should me taken as to catch the possible `IndexOutOfBoundException`).

Unit tests that rely on `ExpressionGenerator` being responsible of calling the `CallMapper` (basically every test in `[DIALECT]ExpressionGeneratorTest` that imply a `CallFunction`) would need to be rephrased and relocated in a proper `[DIALEC]CallMapperTest`).

### Additional Context

_No response_

vil1:
fixes #898

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
@@ -8,8 +8,7 @@ import java.time._
 import java.time.format.DateTimeFormatter
 import java.util.Locale
 
-class ExpressionGenerator(val callMapper: ir.CallMapper = new ir.CallMapper())
-    extends Generator[ir.Expression, String] {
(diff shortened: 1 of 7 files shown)


nfx (approved):
lgtm

PR #901, opened 2024-09-13
Patch/reconcile support temp views

ganeshdogiparthi-db:
Right now remorph-reconcile supports only physical tables. However, some of the clients use temporary views for reconciliation. To support temporary views we had to make changes to support `global_temp`.


src/databricks/labs/remorph/reconcile/connectors/databricks.py
```diff
@@ -54,9 +56,12 @@ def read_data(
         query: str,
         options: JdbcReaderOptions | None,
     ) -> DataFrame:
-        if not catalog:
-            catalog = "hive_metastore"
-        table_with_namespace = f"{catalog}.{schema}.{table}"
+        namespace_catalog = "hive_metastore" if not catalog else catalog
```
nfx, line 59: add the relevant changes in `core` and a functional test to make we keep this feature in proper implementation.

nfx (changes requested):

nfx (approved):

PR #907, opened 2024-09-17
Patch Function without Brackets

sundarshankar89:
Introduced lexer and parser rules for tackling functions without parenthesis.

see example documentation here: https://docs.snowflake.com/en/sql-reference/functions/current_date

Impacted Functions
- CURRENT_DATE
- CURRENT_TIME
- CURRENT_TIMESTAMP
- LOCALTIME
- LOCALTIMESTAMP

this fixes the functional tests as well.

TODO: Fix/Add More Test cases


core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3245,7 +3245,9 @@ functionCall: builtinFunction | standardFunction | rankingWindowedFunction | agg
 builtinFunction: EXTRACT L_PAREN (string | ID) FROM expr R_PAREN # builtinExtract
     ;
 
-standardFunction: functionName L_PAREN (exprList | paramAssocList)? R_PAREN
+standardFunction
+    : functionName L_PAREN (exprList | paramAssocList)? R_PAREN
+    | functionOptinalBrackets (L_PAREN R_PAREN)?
```
vil1, line 3250:
```suggestion
    | functionOptionalBrackets (L_PAREN R_PAREN)?
```

sundarshankar89:
> I think changes in SnowflakeExpressionBuilder (at least most of them) should go in a separate `Rule[LogicalPlan]` that would be executed before `SnowflakeCallMapper`.
> 
> In the ExpressionBuilder, we would simply produce `CallFunction("CURRENT_DATE", Seq())`, etc. And in that new Rule, we would catch these and change them into `DateFormat` accordingly.

Done we will refactor.

core/src/main/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspiler.scala
```diff
@@ -15,6 +19,7 @@ class SnowflakeToDatabricksTranspiler extends BaseTranspiler {
   private val generator = new LogicalPlanGenerator(exprGenerator, optionGenerator)
   private val optimizer =
     ir.Rules(
+      new SnowflakeFractionalSecond,
```
nfx, line 22:
```suggestion
      new ConvertFractionalSecond,
```
better name

nfx (approved):

core/src/main/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspiler.scala
sundarshankar89: Done

PR #911, opened 2024-09-17
TSQL: Implement CREATE TABLE

jimidle:
Here, we implement transpilation of the TSQL CREATE TABLE command with its many options and forms, including CTAS and graph node syntax, as well as covering syntactical differences for the analytics variants of SQL Server.

We also move the DDL and DML visitors from the AST and Relation visitors, to give better separation of responsibility.


nfx (approved):
okay to merge, needs follow-up fixes, though

PR #912, opened 2024-09-18
Move coverage tests into the core module

vil1:


core/pom.xml
```diff
@@ -45,21 +47,35 @@
       <version>${antlr.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.json4s</groupId>
-      <artifactId>json4s-jackson_${scala.binary.version}</artifactId>
-      <version>4.0.7</version>
+      <groupId>com.fasterxml.jackson.module</groupId>
+      <artifactId>jackson-module-scala_${scala.binary.version}</artifactId>
+      <version>2.15.0</version>
```
nfx, line 52: extract property

core/pom.xml
```diff
@@ -99,12 +115,22 @@
       <version>0.8.1</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.lihaoyi</groupId>
+      <artifactId>upickle_2.12</artifactId>
```
nfx, line 120: replace scala binary version

core/pom.xml
```diff
@@ -99,12 +115,22 @@
...
     <dependency>
       <groupId>com.lihaoyi</groupId>
       <artifactId>ujson_${scala.binary.version}</artifactId>
       <version>${ujson.version}</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.lihaoyi</groupId>
+      <artifactId>os-lib_${scala.binary.version}</artifactId>
+      <version>0.10.1</version>
```
nfx, line 132: extract property

core/pom.xml
```diff
@@ -115,6 +141,11 @@
       <artifactId>snowflake-jdbc</artifactId>
       <version>${snowflake-jdbc.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.github.tototoshi</groupId>
+      <artifactId>scala-csv_${scala.binary.version}</artifactId>
+      <version>2.0.0</version>
```
nfx, line 147: extract property

core/pom.xml
```diff
@@ -215,40 +246,71 @@
...
-                    <argument>--save-dev</argument>
-                    <argument>antlr-format</argument>
-                    <argument>antlr-format-cli</argument>
-                  </arguments>
-                </configuration>
               </execution>
-              <!-- antlr-fmt -->
+            </executions>
+            <configuration>
+              <mainClass>"com.databricks.labs.remorph.Main"</mainClass>
```
nfx, line 291: why do we need quotes here?

core/src/main/scala/com/databricks/labs/remorph/Main.scala
```diff
@@ -9,6 +9,8 @@ object Main extends App with ApplicationContext {
       exampleDebugger.debugExample(args("name"), args.get("dialect"))
     case Payload("debug-me", _) =>
       prettyPrinter(workspaceClient.currentUser().me())
+    case Payload("coverage", args) =>
```
nfx, line 12: rename command to `debug-coverage` - that way customers won't be tempted to call it.

core/src/main/scala/com/databricks/labs/remorph/coverage/CoverageTest.scala
```diff
@@ -67,18 +51,13 @@ object Main {
...
+          target_dialect = "databricks",
           file = os.Path(test.inputFile).relativeTo(sourceDir).toString)
         val report = runner.runQuery(exampleQuery)
         val reportEntryJson = ReportEntry(header, report).asJson
         os.write.append(outputFilePath, ujson.write(reportEntryJson, indent = -1) + "\n")
       }
     }
-    // scalastyle:off
-    println(s"Successfully produced coverage report in $outputFilePath")
-    // scalastyle:on
```
nfx, line 80: can you also print the coverage %? either in this PR or the next one. then https://github.com/databrickslabs/remorph/blob/main/src/databricks/labs/remorph/coverage/local_report.py and https://github.com/databrickslabs/remorph/blob/main/src/databricks/labs/remorph/coverage/commons.py#L62-L78 could be removed

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/coverage/CoverageTest.scala
vil1: I will reintroduce that `println` (in an improved version with %) in the next PR

nfx (approved):
lgtm

PR #915, opened 2024-09-18
TSQL: Reorg of DML IR and generation production after CREATE TABLE

jimidle:
Follows Snowflake reorg of separation DDL and DML and cleanly separates CREATE TABLE etc from DML. Also changes the IR Builders for TSQL such that they are all injected via a visitor coordination class rather than created within each builder, thus making each visitor independent of any other, which solves cyclic dependencies.  

An implementation of the VisitorCoordinator can also supply any other common resources in the same manner.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/VisitorCoordinator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/VisitorCoordinator.scala
@@ -1,0 +1,30 @@
+package com.databricks.labs.remorph.parsers
+import com.databricks.labs.remorph.parsers.tsql.DataTypeBuilder
+import org.antlr.v4.runtime.tree.ParseTreeVisitor
+
+/**
+ * <p>
+ *   An implementation of this class should provide an instance of each of the ParseTreeVisitors
+ *   required to build IR and each visitor should be initialized with a reference to the implementation
+ *   of this class so that the visitors can call each other without needing to know the specific implementation
+ *   of any visitor or creating circular dependencies between them.
+ * </p>
+ * <p>
+ *   Implementations can also supply other shared resources, builders, etc. via the same mechanism
+ * </p>
+ */
+abstract class VisitorCoordinator {
+
+  // Parse tree visitors
+  val astBuilder: ParseTreeVisitor[_]
+  val relationBuilder: ParseTreeVisitor[_]
+  val expressionBuilder: ParseTreeVisitor[_]
+  val dmlBuilder: ParseTreeVisitor[_]
+  val ddlBuilder: ParseTreeVisitor[_]
+
+  // A function builder that can be used to build function calls for a particular dialect
+  val functionBuilder: FunctionBuilder
+
+  // Common builders that are used across all parsers, but can still be overridden
+  val dataTypeBuilder = new DataTypeBuilder
+}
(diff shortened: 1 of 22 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilder.scala
```diff
@@ -12,8 +11,10 @@ import scala.collection.JavaConverters.asScalaBufferConverter
  */
 class TSqlAstBuilder extends TSqlParserBaseVisitor[ir.LogicalPlan] {
 
-  private val relationBuilder = new TSqlRelationBuilder
-  private val expressionBuilder = new TSqlExpressionBuilder
+  private val expressionBuilder = new TSqlExpressionBuilder(null)
+  private val relationBuilder = new TSqlRelationBuilder(expressionBuilder)
+  expressionBuilder.relationBuilder = relationBuilder
```
nfx, line 16: if expression builder and relation builder need each other - they have to be in the same class. I guess `TSqlExpressionBuilder` :)

nfx (changes requested):
if expression builder and relation builder need each other - they have to be in the same class. I guess TSqlExpressionBuilder :)

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilder.scala
jimidle: Might be that the visitors that cause this interplay need moving to be builders. Moving to injection here only points out what the relationship already was of course.

jimidle: I implemented a visitor coordinator such that each visitor has access to the other visitors if it needs them, without being dependent on any particular implementation of the other visitors. The VisitorCoordinator is then injected into each visitor. The coordinator can also supply any other shared resources using the same technique.

nfx (approved):
lgtm

core/src/main/scala/com/databricks/labs/remorph/parsers/VisitorCoordinator.scala
```diff
@@ -0,0 +1,30 @@
...
+ *   An implementation of this class should provide an instance of each of the ParseTreeVisitors
+ *   required to build IR and each visitor should be initialized with a reference to the implementation
+ *   of this class so that the visitors can call each other without needing to know the specific implementation
+ *   of any visitor or creating circular dependencies between them.
+ * </p>
+ * <p>
+ *   Implementations can also supply other shared resources, builders, etc. via the same mechanism
+ * </p>
+ */
+abstract class VisitorCoordinator {
```
nfx, line 16: keep it in the tsql package, for now - it doesn't make sense to introduce an abstraction in common package before it's mature enough and adopted by both parsers

nfx (changes requested):
fix conflicts

core/src/main/scala/com/databricks/labs/remorph/parsers/VisitorCoordinator.scala
jimidle: Fair point - I thought about that too. It is fairly clear that TSQL and Snowflake are following the same builder set and it seems that new dialects should follow the same pattern but neither are completely finished with IR building. So I will leave the abstract class but not apply the same to Snowflake at the moment. I can remove the abstract class if you prefer.

PR #917, opened 2024-09-20
Added private key authentication for sf

ganeshdogiparthi-db:
* Right now, we only support password-based authentication for Snowflake.
* This PR includes support for private key authentication to Snowflake.
* First priority is given to the `pem_private_key` secret. If not found, then, we check for `sfPassword` in the secret scope. If both are not found, we raise an exception.
* Since Spark JDBC does not properly support privatekey auth, we have to use password-based authentication when the user provides JDBCOptions for now.
* Added `cryptography` dependency for processing the user provided private key.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -74,3 +74,8 @@ https://github.com/databrickslabs/blueprint/blob/main/LICENSE
 
 Databricks Connect - https://pypi.org/project/databricks-connect/
 Copyright (2019) Databricks Inc.
+
+cryptography - https://github.com/pyca/cryptography
+Copyright 2013-2023 The Python Cryptographic Authority and individual contributors.
+License - https://github.com/pyca/cryptography/blob/main/LICENSE
+This library requires compiled dependencies, including a C compiler and OpenSSL.
--- a/NOTICE
+++ b/NOTICE
@@ -60,6 +60,10 @@ scala-csv - https://github.com/tototoshi/scala-csv
 Copyright 2013-2015 Toshiyuki Takahashi
 License - https://github.com/tototoshi/scala-csv/blob/master/LICENSE.txt
 
+cryptography - https://github.com/pyca/cryptography
+Copyright 2013-2023 The Python Cryptographic Authority and individual contributors.
+License - https://github.com/pyca/cryptography/blob/main/LICENSE
+
 This software contains code from the following open source projects, licensed under the BSD license:
 
 ANTLR v4 - https://github.com/antlr/antlr4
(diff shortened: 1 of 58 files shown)


pyproject.toml
```diff
@@ -55,6 +55,7 @@ dependencies = [
   "databricks-labs-pylint~=0.4.0",
   "mypy~=1.10.0",
   "numpy==1.26.4",
+  "cryptography>=41.0.3"
```
nfx, line 58: this is test-time dependencies and it's installed only on CI and developer machines, you probably intend this library to be executed on customer machines - move it to line 19. also what license is this library? does it require any compiled dependencies? we also need to reflect it in `NOTICE` file.

nfx (changes requested):

pyproject.toml
ganeshdogiparthi-db: Thanks, corrected the code.
It is apache2.0 license https://github.com/pyca/cryptography?tab=Apache-2.0-2-ov-file
and According to the internet, it requires the below-complied dependencies.

```
C Compiler: A tool that translates C code into machine code. Examples include gcc (GNU Compiler Collection) or clang.
OpenSSL: A software library that provides cryptographic functions. It must be installed on your system because the cryptography library relies on it for certain operations.
```

ganeshdogiparthi-db:
* There is a recent release in pylint 3.3.0. I had to add `too-many-positional-arguments` to disable list. I see this setting `max-args = 9` is not working as expected.

pyproject.toml
```diff
@@ -556,7 +557,8 @@ disable = [
   "fixme",
   "consider-using-assignment-expr",
   "logging-fstring-interpolation",
-  "consider-using-any-or-all"
+  "consider-using-any-or-all",
```
sundarshankar89, line 560: Rebase from main Fixed pylint to specific version

sundarshankar89 (changes requested):

NOTICE
```diff
@@ -73,4 +73,9 @@ Copyright (2023) Databricks, Inc.
...
 
 Databricks Connect - https://pypi.org/project/databricks-connect/
-Copyright (2019) Databricks Inc.
\ No newline at end of file
+Copyright (2019) Databricks Inc.
+
+cryptography - https://github.com/pyca/cryptography
+Copyright 2013-2023 The Python Cryptographic Authority and individual contributors.
+License - https://github.com/pyca/cryptography/blob/main/LICENSE
+This library requires compiled dependencies, including a C compiler and OpenSSL.
```
nfx, line 81:
```suggestion
```

irrelevant

NOTICE
```diff
@@ -73,4 +73,9 @@ Copyright (2023) Databricks, Inc.
 https://github.com/databrickslabs/blueprint/blob/main/LICENSE
 
 Databricks Connect - https://pypi.org/project/databricks-connect/
-Copyright (2019) Databricks Inc.
\ No newline at end of file
+Copyright (2019) Databricks Inc.
+
+cryptography - https://github.com/pyca/cryptography
```
nfx, line 78: move this to a proper section - under Apache 2.0. this section is for Databricks License

nfx (changes requested):

NOTICE
ganeshdogiparthi-db: Sorry may bad. I though everything below is Apache 2.0
Done.

nfx (approved):
lgtm

PR #920, opened 2024-09-20
Ensure that successful parsing is recorded even if transpilation fails and record reconciled generations

jimidle:
A small bug meant that if the transpile failed, then the fact that the parse was successful was not recorded. Here we correct that oversight.

We also add an extra field to the report entry to record what percentage of code generations resulted in the expected translation as currently we record that there was a generation difference but do not report how many resulted in the expected translation according to the coverage test.

### DIFF
--- a/coverage/src/main/scala/com/databricks/labs/remorph/coverage/QueryRunner.scala
+++ b/coverage/src/main/scala/com/databricks/labs/remorph/coverage/QueryRunner.scala
@@ -2,7 +2,7 @@ package com.databricks.labs.remorph.coverage
 
 import com.databricks.labs.remorph.queries.ExampleQuery
 import com.databricks.labs.remorph.transpilers.WorkflowStage.PARSE
-import com.databricks.labs.remorph.transpilers.{Formatter, Result, SnowflakeToDatabricksTranspiler, SourceCode, TSqlToDatabricksTranspiler, Transpiler}
+import com.databricks.labs.remorph.transpilers._
 import com.databricks.labs.remorph.utils.Strings
 
 trait QueryRunner extends Formatter {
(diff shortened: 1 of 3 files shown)


coverage/src/main/scala/com/databricks/labs/remorph/coverage/QueryRunner.scala
```diff
@@ -21,17 +21,25 @@ abstract class BaseQueryRunner(transpiler: Transpiler) extends QueryRunner {
...
+        ReportEntryReport(parsed = 1, statements = 1, transpilation_error = Some(errorJson))
       case Result.Success(output) =>
         if (exampleQuery.expectedTranslation.map(format).exists(_ != format(output))) {
           val expected = exampleQuery.expectedTranslation.getOrElse("")
           ReportEntryReport(
             parsed = 1,
             transpiled = 1,
             statements = 1,
-            transpilation_error = Some(compareQueries(expected, output)))
+            reconciliation_error = Some(compareQueries(expected, output)))
```
nfx, line 34: our downstream ETL pipelines rely on `transpilation_error` for this purpose. why do we need this? this is still a transpilation error if it's not converting things correctly.

coverage/src/main/scala/com/databricks/labs/remorph/coverage/ReportEntry.scala
```diff
@@ -14,7 +14,10 @@ case class ReportEntryReport(
     parsing_error: Option[String] = None,
     transpiled: Int = 0, // 1 for success, 0 for failure
     transpiled_statements: Int = 0, // number of statements transpiled
-    transpilation_error: Option[String] = None) {
+    transpilation_error: Option[String] = None,
+    reconciled: Int = 0,
+    reconciled_statements: Int = 0,
+    reconciliation_error: Option[String] = None) {
```
nfx, line 20:
```suggestion
    transpilation_error: Option[String] = None) {
```
remove `reconciled` - we don't have cycles to implement it in downstream pipelines

nfx (changes requested):

coverage/src/main/scala/com/databricks/labs/remorph/coverage/QueryRunner.scala
jimidle: Valentin and I wanted to separated the cases where the generation was successful but it does not match the expected output as at the moment we don't get to be told that. However, we can still count this as a transpilation error if you think that downstream tools might need to see it that way.

coverage/src/main/scala/com/databricks/labs/remorph/coverage/ReportEntry.scala
jimidle: OK

nfx (approved):
lgtm

PR #922, opened 2024-09-20
feat: Implement all of TSQL predicates except for SOME ALL ANY

jimidle:
Thee IR generation for TSQL predicates was not complete. Here, we implement all of the IR for predicates such as IN, IS, BETWEEN, LIKE....

Additionally the LIKE IR was incorrectly assuming that the ESCAPE character had to be a single character at parse time, but in fact it accepts a full expression which must evaluate to a single character at runtime. We correct that here.


nfx (approved):
lgtm

PR #923, opened 2024-09-23
Fetch table definitions for TSQL

### LINKED ISSUE #918: [FEATURE]: Fetch TableDefinition entries from all catalogs/databases/tables in TSQL

Transpile

### Problem statement

 Fetch TableDefinition entries from all catalogs/databases/tables in TSQL.

### Proposed Solution

Implementation is done for Snowflake https://github.com/databrickslabs/remorph/issues/902. 
The same thing will be implemented in TSQL.

### Additional Context

_No response_

sriram251-code:
closes #918 

This pull request retrieves table definitions from TSQL and structures them using a case class named TableDefinition.

The TableDefinition class encapsulates the following properties for each table:

catalog: The catalog name (String)
schema: The schema name (String)
table: The table name (String)
location: An optional field for table location (Option[String])
tableFormat: An optional field for the table format (Option[String])
viewText: An optional field for view definition, if applicable (Option[String])
columns: A sequence of ColumnDetail objects representing the table's columns (Seq[ColumnDetail])
sizeGb: The table size in gigabytes (Int)
comments: The comment given on the table(Option[String])

This structured approach allows for efficient querying and manipulation of Snowflake table metadata within the application.


core/src/main/scala/com/databricks/labs/remorph/discovery/queries.scala
```diff
@@ -19,7 +19,8 @@ case class TableDefinition(
     tableFormat: Option[String] = None,
     viewText: Option[String] = None,
     columns: Seq[StructField] = Seq.empty,
-    sizeGb: Int = 0)
+    sizeGb: Int = 0,
+    comments: Option[String] = None)
```
sundarshankar89, line 23: Just checking is the ask is to capture only Table Comments, how about Column Comments?

sriram251-code: @nfx @gueniai let us know if we need to capture the column comments? Now we are only capturing table comments.

gueniai: We discussed this during standup, but just confirming here that we do need column comments as well

core/src/main/scala/com/databricks/labs/remorph/parsers/intermediate/ddl.scala
```diff
@@ -59,6 +59,7 @@ case class StructField(name: String, dataType: DataType, nullable: Boolean = tru
 case class StructType(fields: Seq[StructField]) extends DataType
 case class MapType(keyType: DataType, valueType: DataType) extends DataType
 case object VariantType extends DataType
+case class ColumnDetail(name: String, dataType: DataType, nullable: Boolean = true, comment: Option[String] = None)
```
sundarshankar89, line 62: Can you add a comment on what or why this IR is needed

sriram251-code: As we can't use StructField to store the comments for each column

core/src/main/scala/com/databricks/labs/remorph/discovery/queries.scala
sriram251-code: done

PR #937, opened 2024-09-26
Translation Complexity Analyzer

jimidle:
Here, we create a query analyzer to extend the current analysis of a translation source query history. 

The complexity analysis is intended as a guide to members of a translation project as to how much work is involved in converting the source system's SQL to Databricks SQL.

This version of the analyzer establishes the framework for assessing and reporting query complexity, but it does not yet attempt to ascribe rules to the analysis of the plan in order to accumulate an overall complexity score. As such, the output from this version of the analyzer is not useful as complexity measure, but will become so when we add properly though t out rules about what scores each element of a query should contribute to the overall score. 

For instance:

 - what score does being unable to parse the query give vs parsing but falling to generate? 
 - what score does a simple AND condition contribute vs a chain of CASEs?
 - and so on...

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
@@ -18,7 +18,15 @@ object SqlComplexity {
     macroRW[SqlComplexity.VERY_COMPLEX.type])
 }
 
+case class SourceTextComplexity(lineCount: Int, textLength: Int)
+
 class EstimationAnalyzer extends LazyLogging {
   // TODO: We will do this and the rest of the analysis in the next PR
   def countStatements(plan: ir.LogicalPlan): Int = 1
+
+  def estimateComplexity(plan: ir.LogicalPlan): SqlComplexity = SqlComplexity.LOW
+
+  def sourceTextComplexity(query: String): SourceTextComplexity = {
+    SourceTextComplexity(query.split("\n").length, query.length)
+  }
 }
--- a/core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
@@ -107,26 +107,22 @@ class EstimationAnalyzer extends LazyLogging {
 
   private def logicalPlanEvaluator: PartialFunction[LogicalPlan, Int] = { case lp: LogicalPlan =>
     try {
-      // scalastyle:off println
-      println(lp)
-      // scalastyle:on println
       1
     } catch {
       case _: Throwable => 5
     }
   }
 
-  def expressionEvaluator: PartialFunction[Expression, Int] = { case expr: Expression =>
+  private def expressionEvaluator: PartialFunction[Expression, Int] = { case expr: Expression =>
     try {
-      // scalastyle:off println
-      println(expr)
-      // scalastyle:on println
       1
     } catch {
       case _: Throwable => 5
     }
   }
 
+  // TODO: Verify these calculations and decide if they are all needed or not. May not harm to keep them around anyway
+  // TODO: calculate complexity using above stats not just the median score
   def summarizeComplexity(reportEntries: Seq[EstimationReportRecord]): EstimationStatistics = {
     val scores = reportEntries.map(_.analysisReport.score)
 
@@ -152,7 +148,7 @@ class EstimationAnalyzer extends LazyLogging {
 
     // Percentiles
     def percentile(p: Double): Double = {
-      val k = (p * (sortedScores.size - 1) + 1).toInt - 1
+      val k = (p * (sortedScores.size - 1)).toInt
       sortedScores(k)
     }
     val percentile25 = percentile(0.25)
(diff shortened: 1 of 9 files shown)


core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
```diff
@@ -11,14 +14,159 @@ object SqlComplexity {
...
+  def evaluateTree(
+      node: TreeNode[_],
+      logicalPlanVisitor: PartialFunction[LogicalPlan, Int],
+      expressionVisitor: PartialFunction[Expression, Int]): Int = {
+    if (node == null) return 0 // Return 0 if the node is null
+
+    node match {
+      case lp: LogicalPlan =>
+        if (lp.children == null) return 0
+        val currentValue = if (logicalPlanVisitor.isDefinedAt(lp)) logicalPlanVisitor(lp) else 0
```
vil1, line 72:
```suggestion
        val currentValue = logicalPlanVisitor.applyOrElse(lp, 0)
```

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
```diff
@@ -11,14 +14,159 @@ object SqlComplexity {
...
+        val childrenValue = lp.children.map(child => evaluateTree(child, logicalPlanVisitor, expressionVisitor)).sum
+        val expressionsValue = lp.expressions.map(expr => evaluateTree(expr, logicalPlanVisitor, expressionVisitor)).sum
+        currentValue + childrenValue + expressionsValue
+
+      case expr: Expression =>
+        if (expr.children == null) return 0
+        val currentValue = if (expressionVisitor.isDefinedAt(expr)) expressionVisitor(expr) else 0
```
vil1, line 79:
```suggestion
        val currentValue = expressionVisitor.applyOrElse(expr, 0)
```

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
```diff
@@ -11,14 +14,159 @@ object SqlComplexity {
...
+    SourceTextComplexity(query.split("\n").length, query.length)
+  }
+
+  private def logicalPlanEvaluator: PartialFunction[LogicalPlan, Int] = { case lp: LogicalPlan =>
+    try {
+      1
+    } catch {
+      case _: Throwable => 5
+    }
+  }
```
vil1, line 114: I don't see what exception could be thrown while evaluating `1`. Maybe this is just some placeholder code that will be filled later?

Also, catching `Throwable` is advised against. Instead, you should `catch NonFatal(e) =>`

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
```diff
@@ -11,14 +14,159 @@ object SqlComplexity {
...
+  def estimateComplexity(query: ExecutedQuery, plan: ir.LogicalPlan): ComplexityEstimate = {
+    val s = sourceTextComplexity(query.source)
+    ComplexityEstimate(SqlComplexity.LOW, countStatements(plan), s.textLength, s.lineCount)
+  }
+
+  def evaluateTree(node: TreeNode[_]): Int = {
+    evaluateTree(node, logicalPlanEvaluator, expressionEvaluator)
+  }
+
+  def evaluateTree(
```
vil1, line 63: The various null-checks in this method look a bit "un-scala-ish". Did you actually stumble upon a case where `.children` returns `null`? If so, that's definitely a bug. A proper solution would be to define a default implementation for `TreeNode#children` that returns `Seq.empty`.

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
jimidle: The 1 is indeed just a placeholder for the next PR. I see that some of the nodes come back as null (which likely means there are missing bits in IR building). I will change to catch NonFatal.

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
jimidle: I did not try and pin it down yet. I thought about what we might do here, but I think we want to report (and assign complexity to), malformed IR. Of course, we should be ensuring that anything that we do not support creates an UnsupportedXXXX and that perhaps we can ensure that in the Incomplete stuff in visitors - but that is complicated by sometimes visitChildren  being the correct thing to call instead of it being default implementation for an unimplemented visitor.

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationReporter.scala
```diff
@@ -17,18 +17,45 @@ class ConsoleEstimationReporter(estimate: EstimationReport) extends EstimationRe
...
+            os.write(outputFilepath, output)
+            record.withQueries(queryFilePath.toString, Some(outputFilepath.toString))
+          case None =>
+            record.withQueries(queryFilePath.toString, None)
+        }
+      } else {
+        record.withQueries("<redacted>", None)
+      }
+    }
+    val newEstimate = estimate.withRecords(newRecords)
```
vil1, line 57: I think we could do better here, but we can leave it as technical debt for now

vil1 (approved):
lgtm (with a non-blocking comment)

nfx (approved):
lgtm

PR #941, opened 2024-09-27
Establish rules infrastructure for complexity analysis

jimidle:
Here we finalize the metrics used for complexity calculations and establish the way rules contribute to scores, ready for concrete implementation of scoring rules and text complexity contribution, in a follow up and final PR.


pyproject.toml
```diff
@@ -628,8 +628,8 @@ evaluation = "max(0, 0 if fatal else 10.0 - ((float(5 * error + warning + refact
 # Tells whether to display a full report or only the messages.
 # reports =
 
-# Activate the evaluation score.
-score = true
+# Activate the evaluation ruleScore.
+ruleScore = true
```
nfx, line 632: irrelevant change, revert

CHANGELOG.md
```diff
@@ -56,7 +56,7 @@ Dependency updates:
...
+* Bump sqlglot from 25.5.1 to 25.6.1 ([#585](https://github.com/databrickslabs/remorph/issues/585)). In this release, the `sqlglot` dependency is updated from version 25.5.1 to 25.6.1 in the 'pyproject.toml' file. This update includes bug fixes, breaking changes, new features, and improvements. Breaking changes consist of updates to the QUALIFY clause in queries and the canonicalization of struct and array inline constructor. New features include support for ORDER BY ALL, FROM ROWS FROM (...), RPAD & LPAD functions, and exp.TimestampAdd. Bug fixes address issues related to the QUALIFY clause in queries, expansion of SELECT * REPLACE, RENAME, transpiling UDFs from Databricks, and more. The pull request also includes a detailed changelog, commit history, instructions for triggering Dependabot actions, and commands for reference, with the exception of the compatibility ruleScore for the new version, which is not taken into account in this pull request.
```
nfx, line 59: irrelevant, revert.

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
```diff
@@ -30,10 +30,16 @@ object SqlComplexity {
     macroRW[SqlComplexity.VERY_COMPLEX.type])
 }
 
-case class ComplexityEstimate(complexity: SqlComplexity, statementCount: Int, charCount: Int, lineCount: Int)
+// TODO: case class ComplexityEstimate(complexity: SqlComplexity, statementCount: Int, charCount: Int, lineCount: Int)
```
nfx, line 33: don't leave commented code

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/RuleDefinitions.scala
```diff
@@ -0,0 +1,77 @@
...
+ * </p>
+ */
+class RuleDefinitions {
+
+  // TODO: Maybe we should use a match like we do in FunctionBuilder?
+
+  val rules: Map[String, Int] = Map(
+    // We were unable to parse the query at all. This adds a significant amount of work to the conversion, but it is
+    // work for the core team, not the user, so are able to filter these out of calculations if desired.
+    "PARSE_FAILURE" -> 100,
```
nfx, line 29: can we instead create the concrete classes, because string constants may get mistyped, causing harder-to-diagnose bugs. e.g.

```
case object ParseFailureRule extends RuleScore(100)
```

also, `case class RuleScore(rule: String, score: Int, from: Seq[RuleScore])` can simply become `Seq[RuleScore]`, right?

core/src/main/scala/com/databricks/labs/remorph/discovery/Anonymizer.scala
```diff
@@ -178,4 +178,30 @@ class Anonymizer(parser: PlanParser[_]) extends LazyLogging {
...
+   * </p>
+   * <p>
+   *   We could improve this hash by removing comments and normalizing whitespace perhaps,
+   *   but whether we would get any gains from that is debatable
+   * </p>
+   *
+   * @param query The text of the query to parse
+   * @return A fingerprint representing the query text
+   */
+  def fingerprint(query: String): String = {
```
nfx, line 201: what if this method is private and called in these branches 

```
      case Result.Failure(PARSE, errorJson) => ...
      case Result.Failure(_, errorJson) => ...
```

to replace `unknown` placeholder?

core/src/main/scala/com/databricks/labs/remorph/discovery/Anonymizer.scala
```diff
@@ -178,4 +178,30 @@ class Anonymizer(parser: PlanParser[_]) extends LazyLogging {
...
+    val masked = query.replaceAll("\\b\\d+\\b", "42").replaceAll("'[^']*'", "?")
```
nfx, line 202: what about double quotes (`"`) and backticks (`)?

nfx (changes requested):

CHANGELOG.md
jimidle: I presume this came in from rebase on main - I will take a look

core/src/main/scala/com/databricks/labs/remorph/discovery/Anonymizer.scala
jimidle: I can add them.

core/src/main/scala/com/databricks/labs/remorph/discovery/Anonymizer.scala
jimidle: Cool - I can do that and add an apply for the processQuery()

pyproject.toml
jimidle: No idea why this happened unless the IDE rename has made this change in a refactor error.

core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/RuleDefinitions.scala
vil1: What we really need here (and in many other places, including most of the foreseeable future developments) is something akin to [this](https://en.wikipedia.org/wiki/Attribute_grammar). 

TLDR; we need the ability to attach arbitrary pieces of information (here it is complexity estimates, but it could as well be some schema information we fetch from an actual database, etc) to the nodes of our precious IR tree, while retaining its structure (because that structure is what we actually care about). 

Once you have this ability, folding such a "labelled-tree" into whatever you need (be it an overall cost estimate, a string containing some SQL code, etc) becomes almost trivial. The best part being that you could label the same tree with different such additional-pieces-of-information at no added cost.

I know how to perform such tree-labelling in a Functional-Progamming-with-recursion-schemes setting, as I applied these techniques with great success on similar problems in the past (namely, transpiling SPARQL queries to Spark-SQL code), though I understand why we won't be considering it as a viable option (it involves writing code that can rapidly become quite difficult to wrap one's head around). However, it would certainly be worth trying to translate these techniques into our more-or-less-imperative setting (even though I cannot guarantee it would be feasible).

nfx (approved):
lgtm

PR #952, opened 2024-10-03
Miscellaneous improvements to Snowflake parser

vil1:
Improves Snowflake parse success rate by about 4% (and incidentally bump transpilation success rate by 2%)


core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3321,17 +3334,17 @@ withExpression: WITH commonTableExpression (COMMA commonTableExpression)*
...
 commonTableExpression
-    : tableName = id (L_PAREN columns += id (COMMA columns += id)* R_PAREN)? AS L_PAREN selectStatement setOperators* R_PAREN
+    : tableName = id (L_PAREN columns += id (COMMA columns += id)* R_PAREN)? AS L_PAREN ((selectStatement setOperators*) | expr) R_PAREN
     ;
 
 selectStatement
     : selectClause selectOptionalClauses limitClause?
     | selectTopClause selectOptionalClauses //TOP and LIMIT are not allowed together
+    | L_PAREN selectStatement R_PAREN
     ;
```
jimidle, line 3344: What is the reason to add the `L_PAREN` `R_PAREN` here? It means that it can be `(((( SELECT ... ))))` etc.

It is probably too much work now, but the parser should not concern itself with disallowing TOP and LIMIT together as that will cause a somewhat meaningless syntax error, where we would want (in a parser rather than a a translator perhaps), something like `"TOP and LIMIT are mutually exclusive in a SELECT statement"`.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3321,17 +3334,17 @@ withExpression: WITH commonTableExpression (COMMA commonTableExpression)*
...
 
 setOperators
-    : (UNION ALL? | EXCEPT | MINUS_ | INTERSECT) selectStatement //EXCEPT and MINUS have same SQL meaning
-    | L_PAREN selectStatement R_PAREN
+    : (UNION ALL? | EXCEPT | MINUS_ | INTERSECT) selectStatement  //EXCEPT and MINUS have same SQL meaning
```
jimidle, line 3347: If we are claning up we should probably rename tokens that have a trailing _ as they were meant to avoid symbol clashes in various generated targets (where keywords like NULL would clash with target language builtins). ANTLR does that automatically now - this is not enough to request changes though.

nfx (approved):
created comments as tech debt

PR #963, opened 2024-10-04
Transpile Snow MONTHS_BETWEEN function correctly

bishwajit-db:
Transpile Snow MONTHS_BETWEEN function correctly


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -51,6 +51,7 @@ class SnowflakeCallMapper extends ir.CallMapper with ir.IRHelpers {
       case ir.CallFunction("LISTAGG", args) =>
         ir.ArrayJoin(ir.CollectList(args.head, None), args.lift(1).getOrElse(ir.Literal("")), None)
       case ir.CallFunction("MONTHNAME", args) => ir.DateFormatClass(args.head, ir.Literal("MMM"))
+      case ir.CallFunction("MONTHS_BETWEEN", args) => ir.MonthsBetween(args.head, args(1), ir.Literal(true))
```
sundarshankar89, line 54: @bishwajit-db Can you also add a test inside `SnowflakeToDatabricksTranspilerTest` as well.

sundarshankar89 (changes requested):

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
bishwajit-db: Done

nfx (approved):
lgtm

PR #966, opened 2024-10-07
Establish incomplete parser and aggregate functions for all Visitors

jimidle:
While we had implemented an IncompleteParser trait, it was somewhat incompletely implemented. Here, we ensure that all visitors implement the `ParserCommon` trait and supply the method `unresolved(String)`, which ensure that null is never inserted in the IR. 

We now provide logger warnings output to indicate that there are missing visitors that are relying on default implementations and the text held by an UnresolvedXXX node indicates which visitor was missing and what source text the visitor is supposed to handle.

The main purpose of this change is to ensure that the IR is never populated with null nodes or children so that IR visitors do not need to check for these errors. All unimplemented visitors will be reported as such and can then be implemented as we do not wish to rely on default behavior.

The default implementation of every visitor is just a visit of all children of the node. The visit then aggregates the results of all the visits and returns the last thing visited, which happens to work in some cases. This is the job of the `aggregateResult` method. 

However, this is not the way to implement the visitors in a production system as we should cover all elements of the parse tree with an actual visitor that explicitly shows the intent of the visitor.

`wrapUnresovedInput` is still implemented and it can be used by concrete implementations of visitors to show that we know we have not implemented support for a particular ParseTree.

We also implement some of the required visitors that were missing to date, although not all of them. we can gradually (or in one hit) implement the reported missing visitors from here, but this is a big enough change as it is.


core/src/main/scala/com/databricks/labs/remorph/parsers/IncompleteParser.scala
```diff
@@ -1,16 +1,25 @@
 package com.databricks.labs.remorph.parsers
 
+import com.typesafe.scalalogging.LazyLogging
 import org.antlr.v4.runtime.tree.{ParseTreeVisitor, RuleNode}
 
-trait IncompleteParser[T] extends ParseTreeVisitor[T] {
+trait IncompleteParser[T] extends ParseTreeVisitor[T] with LazyLogging {
 
-  protected def wrapUnresolvedInput(unparsedInput: String): T
+  // Note that this is neveer called from here, but may be useful in implementing visitors
```
vil1, line 8:
```suggestion
  // Note that this is never called from here, but may be useful in implementing visitors
```

vil1 (changes requested):
This a huge improvement! 

However, concrete implementations of `wrapUnresolvedInput`, `defaultResult` and `aggregateResult` look very repetitive. They seem to differ only in the type of `Unresolved*` they return (while `aggregateResult` is the same everywhere).

The idea behind `wrapUnresolvedInput` (which doesn't seem to be called anymore now BTW) was to avoid this repetition by abstracting over "the type of `UnresolvedStuff` we return" for each Builder. We should use a similar idea (or refurbish `wrapUnresolvedInput`) so that `IncompleteParser` can provide the implementation for `defaultResult` (and `aggregateResult`) and avoid repetition. The trait would look like:

```scala
trait IncompleteParser[T] extends ParseTreeVisitor[T] {
  def unresolved(msg: String): T
  
  def wrapUnresolvedInput(unparsedInput: RuleNode): T = unresolved(getTextFromParserRuleContext(unparsedInput.getRuleContext)))

  def defaultResult: T = unresolved("Unimplemented visitor returns defaultResult!")

  def aggregateResult(agg: T, next:T): T = Option(next).getOrElse(agg)

  abstract override def visitChildren(node: RuleNode): T = {
    val stackTrace = Thread.currentThread().getStackTrace
    val callingMethod = stackTrace(4).getMethodName
    val implementingClass = this.getClass.getSimpleName
    logger.warn(s"Unimplemented visitor for method: $callingMethod in class: $implementingClass")
    super.visitChildren(node)
  }
}
```

jimidle:
> This a huge improvement!
> 
> However, concrete implementations of `wrapUnresolvedInput`, `defaultResult` and `aggregateResult` look very repetitive. They seem to differ only in the type of `Unresolved*` they return (while `aggregateResult` is the same everywhere).
> 
> The idea behind `wrapUnresolvedInput` (which doesn't seem to be called anymore now BTW) was to avoid this repetition by abstracting over "the type of `UnresolvedStuff` we return" for each Builder. We should use a similar idea (or refurbish `wrapUnresolvedInput`) so that `IncompleteParser` can provide the implementation for `defaultResult` (and `aggregateResult`) and avoid repetition. The trait would look like:
> 
> ```scala
> trait IncompleteParser[T] extends ParseTreeVisitor[T] {
>   def unresolved(msg: String): T
>   
>   def wrapUnresolvedInput(unparsedInput: RuleNode): T = unresolved(getTextFromParserRuleContext(unparsedInput.getRuleContext)))
> 
>   def defaultResult: T = unresolved("Unimplemented visitor returns defaultResult!")
> 
>   def aggregateResult(agg: T, next:T): T = Option(next).getOrElse(agg)
> 
>   abstract override def visitChildren(node: RuleNode): T = {
>     val stackTrace = Thread.currentThread().getStackTrace
>     val callingMethod = stackTrace(4).getMethodName
>     val implementingClass = this.getClass.getSimpleName
>     logger.warn(s"Unimplemented visitor for method: $callingMethod in class: $implementingClass")
>     super.visitChildren(node)
>   }
> }
> ```

I thought about doing that, but decided to check this in, then resolve how to do this with generics later. I was trying to enforce implementation of defaultResult but because this is already supplied by the ANTLR library, the compiler just thought it was already implemented.

I totally agree though that it is all repetitive, so let me look at the code you are suggesting here. Maybe I can/should do that in this PR and not wait.

wrapUnresolvedInput will now be called deliberately from visitors (for instance visitPredicate in snowflake will be changed to call this once this PR is merged). It turns out that knowing which visitors are not implemented is much better than knowing what bit of text isn't dealt with (which I could probably still do now that I think about it, but getTextFromParseRule isn't available to this trait I think as it is in ParserCommon?). Let me take another look at this...

core/src/main/scala/com/databricks/labs/remorph/parsers/ParserCommon.scala
```diff
@@ -1,23 +1,100 @@
...
   protected def occursBefore(a: ParseTree, b: ParseTree): Boolean = {
     a != null && b != null && a.getSourceInterval.startsBeforeDisjoint(b.getSourceInterval)
   }
 
   def visitMany[R <: RuleContext](contexts: java.lang.Iterable[R]): Seq[A] = contexts.asScala.map(_.accept(self)).toSeq
 
   /**
-   * Creates a string representation of the text represented by the given ANTLR ParserRuleContext.
+   * <p>
+   *   An implementation sof this should return some type of ir.UnresolvedXYZ object that represents the
```
vil1, line 19:
```suggestion
   *   An implementation of this should return some type of ir.UnresolvedXYZ object that represents the
```

core/src/main/scala/com/databricks/labs/remorph/parsers/ParserCommon.scala
```diff
@@ -1,23 +1,100 @@
...
+  /**
+   * <p>
+   *   Creates a string representation of the text represented by the given ANTLR ParserRuleContext.
+   * </p>
+   * <p>
+   *   Note that this should exactly reflect the original input text as bounded by the source interval
+   *   recorded by the parser.
+   * </p>
+   */
+  def getTextFromParserRuleContext(ctx: RuleContext): String = ctx match {
```
vil1, line 43: nitpicking:

```suggestion
  protected def getTextFromParserRuleContext(ctx: RuleContext): String = ctx match {
```

vil1 (approved):
lgtm (remaining comments are non-blocking)

core/src/main/scala/com/databricks/labs/remorph/parsers/ParserCommon.scala
jimidle: It is accessed from outside the visitor in tests

core/src/main/scala/com/databricks/labs/remorph/parsers/ParserCommon.scala
```diff
@@ -1,23 +1,100 @@
...
+  /**
+   * Overrides the default visitChildren to report that there is an unimplemented
+   * visitor def for the given RuleNode, then call the ANTLR default visitor
+   * @param node the RuleNode that was not visited
+   * @return T an instance of the type returned by the implementing visitor
+   */
+  abstract override def visitChildren(node: RuleNode): A = {
+    caller = Thread.currentThread().getStackTrace()(4).getMethodName
+    implementor = this.getClass.getSimpleName
+    logger.warn(
```
nfx, line 94: we better need to collect this information in our reports

nfx (approved):
lgtm

PR #972, opened 2024-10-08
Move intermediate package out of parsers

vil1:
A long-overdue refactoring that moves the `intermediate` package out of the `parsers` one. As parsers are meant to depend on the intermediate representation and not the other way around.


nfx (approved):
lgtm

PR #974, opened 2024-10-08
Implement missing visitor functions for Snowflake and TSQl

jimidle:
A number of required visitor methods were missing from both Snowflake and TSQL builder classes. When a visitor is not implemented, the default visitor is called, which visits all the children of the ParseTree. In some cases this happens to work as it results in just one IR node being returned. However, this is not the way to traverse the ParseTree as a small change in the ANTLR grammar will accidentally break the visitor as the default result is no longer valid. 

Here we implement all the missing visitors not covered by #966


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
```diff
@@ -21,6 +22,66 @@ class SnowflakeDDLBuilder
...
+  override def visitDdlCommand(ctx: DdlCommandContext): ir.Catalog = ctx match {
+    case a if a.alterCommand() != null => a.alterCommand().accept(this)
+    case c if c.createCommand() != null => c.createCommand().accept(this)
+    case d if d.dropCommand() != null => d.dropCommand().accept(this)
+    case u if u.undropCommand() != null => u.undropCommand().accept(this)
+  }
+
+  override def visitCreateCommand(ctx: CreateCommandContext): ir.Catalog = {
+    val methods: Seq[() => ParserRuleContext] = Seq(
+      ctx.createAccount _,
```
nfx, line 34: i'm struggling to understand this syntax, other people would as well

nfx (approved):
looks good for how, needs more improvement

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeDDLBuilder.scala
jimidle: i changed to this because it is smaller. It is just a list of functions basically and the first one that returns non null is used. I can:

  - Change to just List( ... )
  - Go back to a big switch (I just found that a bit ugly)
  - Change the grammar such that this rule is split into a few sub-collections that are generally related

PR #975, opened 2024-10-08
Implement more missing visitor functions for Snowflake and TSQL

jimidle:
A number of required visitor methods were missing from both Snowflake and TSQL builder classes. When a visitor is not implemented, the default visitor is called, which visits all the children of the ParseTree. In some cases this happens to work as it results in just one IR node being returned. However, this is not the way to traverse the ParseTree as a small change in the ANTLR grammar will accidentally break the visitor as the default result is no longer valid. 

Here we implement a number of the missing visitors not covered by #966


nfx (approved):
lgtm

PR #982, opened 2024-10-09
Functional Test Ignore Folder

sundarshankar89:
Added core-only folder to be ignored inside our parameterized test; it will ignore the core_engine folder in all dialects.


nfx (approved):
lgtm

Issue #978, opened 2024-10-09
[BUG]: Unable to parse function arguments


bishwajit-db:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of Bug / Issue

TranspileParserError

### Current Behavior

When trying to translate the Snow query  `SELECT ARRAY_REMOVE([2, 3, 4.00::DOUBLE, 4, NULL], 4);` it fails to parse the function arguments correctly.

### Expected Behavior

It should be able to correctly parse the query.

### Relevant log output or Exception details

```shell
the following error is thrown:

[{"line":1,"charPositionInLine":31,"msg":"could not process 'SELECT ARRAY_REMOVE([2, 3, 4.00::'","offendingTokenWidth":2,"offendingTokenText":"::"},{"line":1,"charPositionInLine":20,"msg":"'[' was unexpected while parsing a SELECT statement in a Snowflake batch\nexpecting one of: Select Statement, '('","offendingTokenWidth":1,"offendingTokenText":"["}]
```

### Sample Query

```shell
SELECT ARRAY_REMOVE([2, 3, 4.00::DOUBLE, 4, NULL], 4);
```

### Operating System

macOS

### Version

latest via Databricks CLI

PR #984, opened 2024-10-09
Snowflake: Correct parsing of Array literals

jimidle:
Here we correct the ANTLR grammar and IR generation for Snowflake arrayliterals of the form:

```sql
[2, 3, 4.00::DOUBLE, 4, NULL]
```
The existing grammar only accepted literal values for the array, but in fact, because we must allow casts, the elements of the array need to be parsed as expressions.

The current type determination in visitArrayLiteral may be too naive but does not affect code generation as the generator does not use it. It only determines the type of the array by looking at the first expression in the array literal, but when an array literal contains a double (as above) and the first value appears to be an integer, then the array literal type should probably be typed as DoubleType and not IntegerType. We will need a function that determines the highest base type that
covers all the values, if this ever becomes relevant. The code is commented so that we do not forget this.

closes: #978

### DIFF
--- a/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
+++ b/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
@@ -3190,7 +3190,7 @@ jsonLiteral: LCB kvPair (COMMA kvPair)* RCB | LCB RCB
 kvPair: key = string COLON literal
     ;
 
-arrayLiteral: LSB literal (COMMA literal)* RSB | LSB RSB
+arrayLiteral: LSB expr (COMMA expr)* RSB | LSB RSB
     ;
 
 dataTypeSize: L_PAREN num R_PAREN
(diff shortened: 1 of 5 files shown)


core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4
```diff
@@ -3190,7 +3190,7 @@ jsonLiteral: LCB kvPair (COMMA kvPair)* RCB | LCB RCB
 kvPair: key = string COLON literal
     ;
 
-arrayLiteral: LSB literal (COMMA literal)* RSB | LSB RSB
+arrayLiteral: LSB expr (COMMA expr)* RSB | LSB RSB
```
sundarshankar89, line 3193: Instead doing this can we make Literal Schema Aware?

sundarshankar89: Rough idea

```
literal
    : DATE string
    | TIMESTAMP string
    | string
    | sign? DECIMAL
    | sign? (REAL | FLOAT)
    | trueFalse
    | jsonLiteral
    | arrayLiteral
    | NULL_
    | string :: cast
    | PARAM // A question mark can be used as a placeholder for a prepared statement that will use binding.
    ;
```

nfx: @sundarshankar89 separate PR, also - `ARRAY(CURENT_USER(), 'foo')` is a valid literal of `ARRAY<STRING>`.

nfx (approved):
lgtm

PR #988, opened 2024-10-10
Revert "Table comments captured as part of Snowflake Table Definition"

sundarshankar89:
Reverts databrickslabs/remorph#949


sundarshankar89:
@sriram251-code I reverted the code.

PR #989, opened 2024-10-10
Table comments captured as part of Snowflake Table Definition

sriram251-code:
closes https://github.com/databrickslabs/remorph/issues/945

This pull request involve adding support for table comments, modifying the TableDefinition case class, and introducing a new integration test for Snowflake table definitions.


sundarshankar89 (approved):
LGTM

core/src/main/scala/com/databricks/labs/remorph/discovery/SnowflakeTableDefinitions.scala
```diff
@@ -91,12 +94,14 @@ class SnowflakeTableDefinitions(conn: Connection) {
...
           val columns = rs
             .getString("DERIVED_SCHEMA")
-            .split("~")
+            .split("‡")
             .map(x => {
-              val data = x.split(":")
+              val data = x.split("§")
               val name = data(0)
               val dataType = getDataType(data(1))
-              ColumnDetail(name, dataType, data(2).toBoolean)
```
nfx, line 99: don't use `ColumnDetail`, add `metadata` field to `StructField` and add comment there, the same way we do it in Spark.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/discovery/SnowflakeTableDefinitions.scala
sriram251-code: changed the StructField to.
case class StructField(name: String, dataType: DataType, nullable: Boolean = true, metadata: Option[Metadata] = None)
The Metadata class i have used is belongs to spark.(org.apache.spark.sql.types.Metadata)
I have made it as Optional field because we dont have to change the current implementation of StructField which has been used in multiple places.

core/src/main/scala/com/databricks/labs/remorph/discovery/SnowflakeTableDefinitions.scala
```diff
@@ -3,6 +3,7 @@ package com.databricks.labs.remorph.discovery
 import com.databricks.labs.remorph.intermediate.{DataType, StructField}
 import com.databricks.labs.remorph.parsers.snowflake.{SnowflakeLexer, SnowflakeParser, SnowflakeTypeBuilder}
 import org.antlr.v4.runtime.{CharStreams, CommonTokenStream}
+import org.apache.spark.sql.types.MetadataBuilder
```
nfx, line 6: don't depend on `org.apache.spark` from this package, create a class inside of `intermediate` package and depend on that one.

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/discovery/SnowflakeTableDefinitions.scala
sriram251-code: ok

nfx (approved):
lgtm

nfx (commented):
lgtm

PR #990, opened 2024-10-10
Added upgrade script and modified metrics sql

ganeshdogiparthi-db:
- Modified the insert statement for the metrics table to cast to Bigint
- updated metrics.sql
- added upgrade script
- modified setup_spark_remote.sh to check URL instead of sleep.


nfx (approved):
looks good for now

PR #991, opened 2024-10-10
Query History From Folder

sundarshankar89:
Enhanced Query History Interface to hold metadata from reading from Folder

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/discovery/FileQueryHistory.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/discovery/FileQueryHistory.scala
@@ -1,0 +1,10 @@
+package com.databricks.labs.remorph.discovery
+
+class FileQueryHistory extends QueryHistoryProvider {
+
+  override def history(): QueryHistory = {
+    val queries = Seq.empty[ExecutedQuery]
+    QueryHistory(queries)
+  }
+
+}
--- a/core/src/main/scala/com/databricks/labs/remorph/discovery/queries.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/discovery/queries.scala
@@ -8,8 +8,8 @@ import java.time.Duration
 case class QuerySpec(
     timestamp: Timestamp = new Timestamp(System.currentTimeMillis()),
     duration: Duration = Duration.ofMillis(0),
-    user: Option[String],
-    filename: Option[String])
+    user: Option[String] = None,
+    filename: Option[String] = None)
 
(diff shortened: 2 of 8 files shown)


jimidle (approved):
lgtm

core/src/main/scala/com/databricks/labs/remorph/discovery/FileQueryHistory.scala
```diff
@@ -0,0 +1,34 @@
...
+import java.nio.file.{Files, Path}
+import scala.collection.JavaConverters._
+import scala.io.Source
+
+class FileQueryHistory(path: Path) extends QueryHistoryProvider {
+
+  private def extractQueriesFromFile(file: File): Seq[ExecutedQuery] = {
+    val fileContent = Source.fromFile(file)
+    val queries = fileContent.getLines().mkString("\n").split(";").map(_.trim).filter(_.nonEmpty)
+    queries.map(ExecutedQuery("10", _, QuerySpec(filename = Some(file.getName))))
```
nfx, line 13: here you loose the query id. should be `{$filename}#{$numberInFile}`

core/src/main/scala/com/databricks/labs/remorph/discovery/queries.scala
```diff
@@ -5,7 +5,13 @@ import com.databricks.labs.remorph.intermediate.StructField
 import java.sql.Timestamp
 import java.time.Duration
 
-case class ExecutedQuery(id: String, timestamp: Timestamp, source: String, duration: Duration, user: String)
+case class QuerySpec(
```
nfx, line 8: why do we need queryspec? don't change interfaces without a need

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/discovery/queries.scala
sundarshankar89: how do we capture future metadata, today we add local file, tomorrow we might be getting other metadata trigger from ETL Job, etc, I feel we meed to have QuerySpec like outline or a Map[String,String] thoughts?

nfx: move those fields to `ExecutedQuery` and we're good to merge

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/discovery/queries.scala
sundarshankar89: Done

PR #993, opened 2024-10-10
Adorn all UnresolvedXXX Ir with ruleName and token name

jimidle:
Here we add additional information to Unresolved nodes so that we can count how many times a particular parser rule was unable to handle certain syntaxes. 

NB: This appears to b e a huge PR, but a lot of the changes are repetitive changes for many tests and many error returns, so it is a lot smaller than it looks.


core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
```diff
@@ -130,7 +130,7 @@ class EstimationAnalyzer extends LazyLogging {
   private def logicalPlanEvaluator: PartialFunction[ir.LogicalPlan, RuleScore] = { case lp: ir.LogicalPlan =>
     try {
       lp match {
-        case ir.UnresolvedCommand(_) =>
+        case ir.UnresolvedCommand(_, _, _, _) =>
```
vil1, line 133:
```suggestion
        case _:ir.UnresolvedCommand =>
```

nfx (approved):
lgtm

Issue #959, opened 2024-10-03
[BUG]: Ranking window functions have an invalid window frame


pamons-databricks:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of Bug / Issue

TranspileParserError

### Current Behavior

Ranking window functions such as row_number, rank, dense_rank, etc. are being transpiled into using an invalid window frame

### Expected Behavior

Ranking window functions should a valid window frame
It looks like the only valid one is `rows between unbounded preceding and current row`

### Steps To Reproduce

Run transpile on the sample query

### Relevant log output or Exception details

```shell
Window Frame specifiedwindowframe(RowFrame, unboundedpreceding$(), unboundedfollowing$()) must match the required frame specifiedwindowframe(RowFrame, unboundedpreceding$(), currentrow$()).
```

### Sample Query

```shell
select 
    row_number() over (partition by id order by report_month)
    , rank() over (partition by id order by report_month)
    , dense_rank() over (partition by id order by report_month)
    , percent_rank() over (partition by id order by report_month)
from test_tbl
```

### Operating System

macOS

### Version

latest via Databricks CLI

sriram251-code:
Investigation : remorph generates ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING by default for all the window functions. But for the below functions ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING is not supported.
row_number(),rank(), dense_rank(), percent_rank(), lead(), lag().

PR #999, opened 2024-10-14
Some window functions does not support window frame conditions

### LINKED ISSUE #958: [BUG]: lag and lead functions are having window frames added

### Category of Bug / Issue

TranspileParserError

### Current Behavior

When transpiling a query with a lag or lead, a window frame is being added.
Databricks does not support window frames in a lead or lag

### Expected Behavior

The window frame should not be included

### Steps To Reproduce

Run transpile on the sample query

```shell
Cannot specify window frame for lag function.
```

### Sample Query

```shell
select 
    id
    , type AS ch_type
    , LEAD(type) OVER (PARTITION BY id ORDER BY day) AS gn_type
from test_tbl
```

### Operating System

macOS

### Version

latest via Databricks CLI

sriram251-code:
closes #958  #959 
Investigation:

Issue 1:
Remorph generates **ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING** by default for all window functions. However, this behavior is not supported for the following functions:

**row_number()
rank()
dense_rank()
percent_rank()
lead()
lag()**

Issue 2:
By default, Remorph adds NULLS LAST in the ORDER BY clause of window functions, without considering whether the sorting is ASC or DESC. This behavior needs to be corrected to align with the documented behavior in the following references:

[Snowflake documentation on ORDER BY usage](https://docs.snowflake.com/en/sql-reference/constructs/order-by#usage-notes)
[Databricks documentation on ORDER BY syntax](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-orderby.html)

ex:
SELECT
  rank() OVER (
    PARTITION BY col1
    ORDER BY col2 DESC
  ) AS rank_col1
FROM tabl;

is transpiled to:
SELECT
  RANK() OVER (PARTITION BY col1 ORDER BY col2 DESC NULLS LAST) AS rank_col1
FROM tabl;


sundarshankar89 (approved):
LGTM

vil1 (commented):
If I understand this correctly, when no window frame is explicitly defined in snowflake for any of the LAG, DENSE_RANK, LEAD, PERCENT_RANK, RANK, ROW_NUMBER, we are not able to emulate snowflake's default. I think we should warn the user about this somehow.

PR #1000, opened 2024-10-14
With Recursive

sundarshankar89:
Added grammar enhancements for With Recursive statements.


jimidle (changes requested):
The change back to ColumnList is fine but we need to ensure that an error in the list does not cause the visitor to ignore valid columns because there was an error such as a missing comma or extra comma - see #1023 for how to do this.

sundarshankar89:
> The change back to ColumnList is fine but we need to ensure that an error in the list does not cause the visitor to ignore valid columns because there was an error such as a missing comma or extra comma - see #1023 for how to do this.

Let me check on that.

sundarshankar89:
> The change back to ColumnList is fine but we need to ensure that an error in the list does not cause the visitor to ignore valid columns because there was an error such as a missing comma or extra comma - see #1023 for how to do this.

SnowflakeSelectStatement doesn't work like TSQL one, I will create a Seperate PR and link it here.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -90,9 +90,19 @@ class SnowflakeAstBuilder(override val vc: SnowflakeVisitorCoordinator)
...
       case None =>
-        val ctes = vc.relationBuilder.visitMany(ctx.commonTableExpression())
-        ir.WithCTE(ctes, relation)
+        if (ctx.RECURSIVE() == null) {
+          val ctes = vc.relationBuilder.visitMany(ctx.commonTableExpression())
+          ir.WithCTE(ctes, relation)
+        } else {
+          ir.UnresolvedCommand(
+            ruleText = contextText(ctx),
+            message = "*** WITH RECURSIVE IS NOT SUPPORTED ***",
```
nfx, line 99: there are reports that `WITH RECURSIVE` could be implemented with various slow-performance hacks in PySpark - see https://stackoverflow.com/questions/52562607/recursive-cte-in-spark-sql

given that, we should emit new `WithRecursive` node and try to translate it in PySpark later

nfx (changes requested):

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
sundarshankar89: Sure

sundarshankar89: Ready for review

nfx (approved):
lgtm

PR #1003, opened 2024-10-15
Add rule name and token name to syntax error reporting

jimidle:
In order that our various coverage tests and conversion metrics are able to report on which rules and tokens are most commonly found to yield errors, we must adorn the syntax error class with those items. 

With this improvement, we can provide coverage and estimation reports that indicate where the most common errors are being found. From there we can indicate solving which errors might be likely to progress a conversion the quickest and in terms of coverage, addressing which errors would increase coverage the most.


nfx (approved):
lgtm

PR #1004, opened 2024-10-15
Rework error return system and use it to generate parsing error and coverage stats

jimidle:
Here we stop converting error messages from the parse, IR builders and source generators to strings and JSON strings prematurely. Errors are now returned as a `RemorphError` and then that is converted to JSON where it is appropriate.

The estimation system now preserves the error structure, including the ruleName and `TOKEN` when a parsing error occurs and similarly for errors produced when building IR and finding missing visitors etc.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
@@ -1,7 +1,7 @@
 package com.databricks.labs.remorph.coverage.estimation
 
 import com.databricks.labs.remorph.coverage.EstimationReportRecord
-import com.databricks.labs.remorph.intermediate.ParsingError
+import com.databricks.labs.remorph.intermediate.{ParsingErrors}
 import com.databricks.labs.remorph.{intermediate => ir}
 import com.typesafe.scalalogging.LazyLogging
 import upickle.default._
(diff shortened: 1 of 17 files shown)


core/src/main/scala/com/databricks/labs/remorph/coverage/estimation/EstimationAnalyzer.scala
```diff
@@ -31,7 +32,15 @@ object SqlComplexity {
 
 case class SourceTextComplexity(lineCount: Int, textLength: Int)
 
-case class EstimationStatistics(allStats: EstimationStatisticsEntry, successStats: EstimationStatisticsEntry)
+case class ParseFailStats(ruleNameCounts: Map[String, Int], tokenNameCounts: Map[String, Int])
```
sundarshankar89, line 35: Love this.

sundarshankar89: My 2 cents, we should expose this outside of Estimation Reporter also. 
If some folks directly run the transpile without running estimator. (When we actually build it)

jimidle: Yes - that can be possible. This is the second of a few PRs that are going to improve all of this. So, we will put this here fist, then, having all the mechanisms, it can be propagated to wherever it is useful.

nfx (approved):
lgtm

PR #1006, opened 2024-10-16
Return `Result` out of generators, instead of throwing exceptions

vil1:
This is a rather important change although the most part (changes in the `Generator`s) is quite mechanical.

The signatures in the `Generator` trait have been changed so that the `generate` method returns a `Result[Out]` (instead of just `Out` before).

This introduces a difficulty, as the result of recursive calls can no longer be used directly, but must be accounted for potential failures by the caller. To make it easier to work with `Result[String]` (which have been aliased to `SQL`), an `SQLInterpolater` has been added (see `SQLInterpolatorSpec` for examples of how it can be used), as well as `mkSql` extension methods for `Seq[SQL]` (similar to the way `mkString` works on `Seq[String]`).

 Finally, we change `ReportEntry` produced by coverage tests to contain `RemorphError`s instead of JSON strings, to open the way for more structured and more informative such reports (which was the initial trigger for this whole change).

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/coverage/ReportEntry.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/coverage/ReportEntry.scala
@@ -1,5 +1,7 @@
 package com.databricks.labs.remorph.coverage
 
+import com.databricks.labs.remorph.intermediate.RemorphError
+
 case class ReportEntryHeader(
     project: String,
     commit_hash: Option[String],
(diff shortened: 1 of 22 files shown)


core/src/main/scala/com/databricks/labs/remorph/coverage/ReportEntry.scala
```diff
@@ -11,16 +13,24 @@ case class ReportEntryHeader(
 case class ReportEntryReport(
     parsed: Int = 0, // 1 for success, 0 for failure
     statements: Int = 0, // number of statements parsed
-    parsing_error: Option[String] = None,
+    parsing_error: Option[RemorphError] = None,
```
jimidle, line 16: In my changes, I had to make this a List[RemorphError] as the parser can return more than one error message. Unless you have used some magic here?

Unfortunately we have both decided to change this and error capture in general,  at the same time. Fortunately we have arrived at more or less the same solution.

jimidle (approved):
LGTM - I will need to rework my own PR on top of this when it is merged. we already identified some more work that we can do on top of this.

core/src/main/scala/com/databricks/labs/remorph/coverage/ReportEntry.scala
vil1: Yeah, going for `List[RemorphError]` here is something I considered as well. Not only here, but in `Failure` as well, since we ultimately want the ability to accumulate multiple errors. But I deemed that to a future PR.

Another way would be to have something of the form 
```scala
case class RemorphErrors(errors: Seq[RemorphError]) extends RemorphError
```
To implement such accumulation while minimizing the amount of changes.

nfx (approved):
lgtm

PR #1022, opened 2024-10-25
Improve Snowflake grammar to improve syntax error recovery

jimidle:
Here we improve the grammar specifications for both Snowflake and TSQL (less was needed) so that they both can recover equally when parsing more than one statement in a batch, when one or more statements are found to be in error.

The AST generator for Snowflake no longer eats good statements and assumes that they were all bad (TSQL already did this).

We now see that both parsers recover using the shortest number of tokens.

A minor modification to the grammar makes the `LET` keyword in an assignment non-optional, as per the Snowflake documentation. If the documentation was wrong, we will merely detect a missing `LET` keyword and recover sensibly by inserting it.


nfx (approved):
lgtm

PR #1023, opened 2024-10-25
TSQL: prevent selectListElem from ignoring valid columns when errorNodes are accumulated

jimidle:
In code such as:

```tsql
SELECT col1, , col2 ...
```

Where the parser recovers nicely from the extraneous ',', the visit method would ignore the recovery and only return the errorNode output.

Here we use a builder method rather than a visit method and return a Seq[] rather than just a single IR node. The generator then receives all the valid column definitions.

We also modify the code generator such that it does not add ',' after the generated comments representing error nodes.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
@@ -239,12 +239,12 @@ class LogicalPlanGenerator(
     }
 
     // Don't put commas after unresolved expressions as they are error comments only
-    val generatedExpressions = proj.expressions.map { exp: ir.Expression => expr.generate(ctx, exp) -> exp }
-    val sqlParts = generatedExpressions
+    val sqlParts = proj.expressions
       .map {
-        case (OkResult(sqlStr), exp) if !exp.isInstanceOf[ir.Unresolved[_]] => s"$sqlStr, "
-        case (OkResult(sqlStr), _) => sqlStr
+        case u: ir.Unresolved[_] => expr.generate(ctx, u)
+        case exp: ir.Expression => expr.generate(ctx, exp).map(_ + ", ")
       }
+      .collect { case OkResult(sqlStr) => sqlStr }
       .mkString
       .stripSuffix(", ")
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
```diff
@@ -237,7 +237,18 @@ class LogicalPlanGenerator(
...
+    // Don't put commas after unresolved expressions as they are error comments only
+    val generatedExpressions = proj.expressions.map { exp: ir.Expression => expr.generate(ctx, exp) -> exp }
+    val sqlParts = generatedExpressions
+      .map {
+        case (OkResult(sqlStr), exp) if !exp.isInstanceOf[ir.Unresolved[_]] => s"$sqlStr, "
+        case (OkResult(sqlStr), _) => sqlStr
+      }
+      .mkString
+      .stripSuffix(", ")
+
```
vil1, line 250:
```suggestion
    val sqlParts = proj.expressions.map { 
      case u: ir.Unresolved[_] => expr.generate(ctx, u)
      case exp: ir.Expression => expr.generate(ctx, exp).map(_ + ", ")
    }
    .mkSql

```

vil1 (changes requested):
I suggested a change to make things more idiomatic in `project`, but I wonder if expecting an `OkResult` when generating an `Unresolved` is a good idea. My guts say that we shoudl rather return a `PartialResult` in that case (though it might be the point of a future PR). 
In any case, the suggested code will keep working when we make that change.

jimidle:
> I suggested a change to make things more idiomatic in `project`, but I wonder if expecting an `OkResult` when generating an `Unresolved` is a good idea. My guts say that we shoudl rather return a `PartialResult` in that case (though it might be the point of a future PR). In any case, the suggested code will keep working when we make that change.

Yes - I want to talk about this in the meeting - what is success and what is failure

jimidle:
> I suggested a change to make things more idiomatic in `project`, but I wonder if expecting an `OkResult` when generating an `Unresolved` is a good idea. My guts say that we shoudl rather return a `PartialResult` in that case (though it might be the point of a future PR). In any case, the suggested code will keep working when we make that change.

Actually, we habve to do a collect after the generate because we want the sqlStr not the OKResult or PartialResult, but I will add that when we change the return type

PR #1027, opened 2024-10-28
Complete evaluate statements scores and make evaluation statement based rather than file based

jimidle:
PLaceholder: Move to statement based evaluations so we know how much of a batch we correctly converted


jimidle:
This PR was affected by many other changes and it will be easier to cancel this one and re work the changes on top of the current state of the code.

PR #1047, opened 2024-11-01
Generic stored procedure parsing

jimidle:
Here we create an ANTLR grammar library mechanism for including grammars, homogonize certain high level rule names, and implement a shared grammar for stored procedures.

The shared grammar caters for TSQL and Snowflake stored procedures, but is intended to be expanded for all future SQL dialects as the standard SQL stored procedures are all very similar and dialect specific extensions such as LANGUAGE xyz in Snowflake do not create unresolvable ambiguities.

The lexers for Snowflake and TSQL have been redesigned to share many common tokens, leaving only a few for the specific lexers, where they would take too much time to merge at the moment. But the intention is to eventually have a common lexer for all dialects, once the over-specified tokens (such as option names) are replaced with `id` or `genericOption`.

This PR is essentially partially finished as the complete stored procedure grammar is not yet defined and the rules for TSQL and Snowflake are yet to be combined. This is left for a future PR.

NB: The ANTLR grammar linter will have to be upgraded to cater for library grammars and to process includes as it is currently only able to process a single grammar. For now, the linting build step is changed to continue on error.


.gitignore
```diff
@@ -17,4 +17,5 @@ spark-warehouse/
 remorph_transpile/
 /linter/gen/
 /linter/src/main/antlr4/library/gen/
-.databricks-login.json
\ No newline at end of file
+.databricks-login.json
+/core/src/main/antlr4/com/databricks/labs/remorph/parsers/lib/gen/
```
nfx, line 21:
```suggestion
/core/src/main/antlr4/com/databricks/labs/remorph/parsers/*/gen/
```

core/src/main/antlr4/com/databricks/labs/remorph/parsers/lib/basesnowflake.g4
```diff
@@ -0,0 +1,204 @@
...
+// $antlr-format allowShortBlocksOnASingleLine true
+// $antlr-format minEmptyLines 0
+// $antlr-format alignSemicolons ownLine
+// $antlr-format alignColons trailing
+// $antlr-format singleLineOverrulesHangingColon true
+// $antlr-format alignLexerCommands true
+// $antlr-format alignLabels true
+// $antlr-format alignTrailers true
+// =================================================================================
+lexer grammar basesnowflake;
```
nfx, line 20: move this into `snowflake` directory. same for `basetsql`.

core/src/main/antlr4/com/databricks/labs/remorph/parsers/lib/procedure.g4
```diff
@@ -0,0 +1,121 @@
...
+
+declareElement:
+    id dataType SEMI                                   # declareSimple
+    | id dataType (DEFAULT | EQ) expr SEMI             # declareWithDefault
+    | id CURSOR FOR expr SEMI                          # declareCursorElement
+    | id RESULTSET ((ASSIGN | DEFAULT) expr)? SEMI     # declareResultSet
+    | id EXCEPTION LPAREN INT COMMA string RPAREN SEMI # declareException
+;
+
+// TODO: Complete definition of SQL/PSM rules
```
nfx, line 121: can't we just include https://github.com/apache/spark/blob/master/sql/api/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBaseParser.g4#L45-L311 here?

nfx (changes requested):

core/src/main/antlr4/com/databricks/labs/remorph/parsers/lib/basesnowflake.g4
jimidle: I went one better and removed the need for it by merging more of the lexers into a single unified lexer

core/src/main/antlr4/com/databricks/labs/remorph/parsers/lib/procedure.g4
jimidle: Possibly, but not in this PR I suggest ;). There would likely be token renaming etc, but it beats writing the rules again from scratch, for sure!

nfx (approved):
this is a big one, but okay :)

PR #1048, opened 2024-11-01
Handling any_keys_match from presto

ganeshdogiparthi-db:
`any_keys_match` is a presto function to do a lambda function on map keys and returns a boolean. We don't have such a direct function in Databricks. We are handling it using existing functions and like the below.

Presto: select any_keys_match(data, x -> x = 'enable_bwob')
Databricks: select exists(map_keys(data), x -> x = 'enable_bwob')


sundarshankar89 (approved):
LGTM

nfx (approved):
where's the core implementation for this?

PR #1163, opened 2024-11-05
Rework serialization using circe+jackson

vil1:


core/src/main/scala/com/databricks/labs/remorph/Main.scala
```diff
@@ -43,13 +49,10 @@ object Main extends App with ApplicationContext {
...
 
   // placeholder for global CLI flags
-  private var cliFlags: Map[String, String] = Map.empty
+  private val cliFlags: Map[String, String] = Map.empty
 
   // parse json from the last CLI argument
   private def route: Payload = {
-    val payload = ujson.read(args.last).obj
-    val command = payload("command").str
-    cliFlags = payload("flags").obj.mapValues(_.str).toMap
```
nfx, line 52: you have to set `cliFlags`

core/src/main/scala/com/databricks/labs/remorph/coverage/EstimationReport.scala
```diff
@@ -21,11 +19,6 @@ case class EstimationReport(
   }
 }
 
-object EstimationReport {
```
nfx, line 24: nice

core/src/main/scala/com/databricks/labs/remorph/coverage/encoders.scala
```diff
@@ -0,0 +1,76 @@
...
+import io.circe.syntax._
+
+import java.io.{PrintWriter, StringWriter}
+import java.sql.Timestamp
+import java.time.Duration
+
+trait ErrorEncoders {
+
+  implicit val codecConfiguration: Configuration =
+    Configuration.default.withDiscriminator("$type").withSnakeCaseMemberNames
```
nfx, line 18: correct me if i'm wrong, but all of this is required because we have `RemorphError` trait that we serialize?

better option is:

- not to include `RemorphError` in json serialization
- change the interface of it to always convert into `problems: Seq[Problem]`, even if it's a single error.
- it's fine to have `RemorphError` trait, but we always have to post-process it into `problems: Seq[Problem]` before we do serialization.
- we have less risks of nested structs leaking into JSON repr and we're closer to tabular form of this
- `Problem.code` will essentially be a simple name of class that you're putting into `$type` discriminator
- we may not need `circle` library then :) 

```
case class Problem(code: String, message: String)
sealed trait RemorphError {
  def problems: Seq[Problem]
}
```

nfx (changes requested):

NOTICE
```diff
@@ -77,4 +77,14 @@ Copyright (2023) Databricks, Inc.
 https://github.com/databrickslabs/blueprint/blob/main/LICENSE
 
 Databricks Connect - https://pypi.org/project/databricks-connect/
-Copyright (2019) Databricks Inc.
\ No newline at end of file
+Copyright (2019) Databricks Inc.
+
+circe - https://github.com/circe/circe
+Copyright (c) 2015, Ephox Pty Ltd, Mark Hibberd, Sean Parsons, Travis Brown, and other contributors. All rights reserved.
+https://github.com/circe/circe/blob/series/0.14.x/LICENSE
```
nfx, line 84: this is Apache2 license, move it away from Databricks License section to a respective Apache2 section, right after `cryptography`, for example.

core/pom.xml
```diff
@@ -123,15 +122,24 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>com.lihaoyi</groupId>
-      <artifactId>upickle_2.12</artifactId>
-      <version>${upickle.version}</version>
+      <groupId>io.circe</groupId>
+      <artifactId>circe-core_2.12</artifactId>
```
nfx, line 126:
```suggestion
      <artifactId>circe-core_${scala.binary.version}</artifactId>
```

nfx (changes requested):
please fix `NOTICE` file

nfx (approved):
lgtm

PR #1167, opened 2024-11-06
Avoid processing escapes in interpolator arguments

vil1:
Calling `treatEscapes` on the interpolator's arguments was a mistake (calling it on string parts wasn't).

The reason is that string parts come from our scala program, like the `\n` in `code"$foo\n$bar"`, while arguments (may) come from the input query (and as such should be left as-is).


nfx (approved):
lgtm

PR #1168, opened 2024-11-06
Create preprocessor and workflow for Jinja template processing

jimidle:
A preprocessor and workflow to allow conversion of embedded SQL in Jinja templates while preserving the templates and, as much as possible, the original text format.

See the preprocessors/README.md for full details of the implementation.

Note that this PR implements about 95% of the use cases for templates, but there are likely other places to support them, such as `if` templates in expression lists. But we will now address these on a case by case basis throuhg testing.

NOTE: This PR only supports templates in TSQL, but it is already too big, so Snowflake support will be added in a separate PR, which is now trivial to do, but would add another 70+ lines of code changes to this PR.


core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+ - The lexer and parser are not burdened with the template syntax
+ - The lexer and parser can be used with a few modifications
+
+#### Disadvantages
+
+ - Templates can be placed anywhere, so the grammar will need to be strategically adorned with valid
+   places that templates can occur. This is not a trivial task, however it is doable if we accept that
+   we will not be able to process some esoteric uses of templates, which are generally abuses of the concept.
+   For instance `SELECT A {{ , }} B {{ , }} C` is not a reasonable construct. It would need to be dealt
```
nfx, line 93: leave it as the `UnresolvedNode` and we'll add a support for it if it's used often

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+## Approaches to handling templates
+
+We should first note that our template handler/preprocessor is not expanding the templates, but locating them and
+translating embedded SQL. The templates themselves will be left in place, and the translated SQL will be adorned
+with them. We are writing a XXX SQL plus DBT to Databricks SQL plus DBT, such that users will then maintain the
+Databricks SQL alongside their templates.
+
+We therefore have the following approaches:
+
+### Lexing and Placeholding
```
nfx, line 74: please add a diagram for this approach

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+#### Disadvantages
+
+ - Templates can be placed anywhere, so the grammar will need to be strategically adorned with valid
+   places that templates can occur. This is not a trivial task, however it is doable if we accept that
+   we will not be able to process some esoteric uses of templates, which are generally abuses of the concept.
+   For instance `SELECT A {{ , }} B {{ , }} C` is not a reasonable construct. It would need to be dealt
+   with manually.
+
+### In-place parsing
```
nfx, line 96: please add a diagram for this approach

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+#### Advantages
+
+ - No need for a pre-processor as such
+
+#### Disadvantages
+
+ - We end up with code dealing with templates intermingled with the dialect parsing code
+ - We still need to track the templates and their locations, so we are not saving any coding effort
+
+### Translate on the fly
```
nfx, line 111: please add a diagram for this approach. i struggle to understand how this will work

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+## The preprocessor
+
+We therefore conclude that the best approach is to use a preprocessor that will locate and track the templates
+and replace them in the pre-processed stream with a placeholder.
+
+As the templates are just text with the macro types above sprinkled in, we can create a preprocessor
+that will always run against the given input text, even if, with un-templated SQL, the preprocessor
+will merely pass through what it sees. However, there may well be other functionality for the preprocessor
+to provide in the future such as perhaps parameter tracking/processing, or perhaps daisy-chaining of
```
nfx, line 134: > perhaps daisy-chaining of

we have `ir.Rules`, so that's a common practice already

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+pre-processors that perform specific tasks.
+
+Hence the preprocessor will be a simple text processor that will:
+
+ - find and replace all `{{ }}`, `{# #}` and `{% %}` blocks with a placeholder
```
nfx, line 139: > blocks with a placeholder

text placeholder or IR node already? how the placeholder will look like?

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Workflow
+```mermaid
+---
+title: Preprocessor Workflow
+config:
+  direction: RL
+  theme: dark
+---
+flowchart TD
```
nfx, line 157: can you adjust this diagram to use our existing concepts - `SnowflakeParser` / `TSqlParser`, `PlanParser`, `LogicalPlanGenerator`, `ExpressionGenerator`, `LogicalPlan`, `Expression` and `ParserRuleContext` + `Parser` from antlr? i struggle to see what components are existing and what components are added here.

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+    User->>DBTProcessor: Load DBT Project
+    DBTProcessor->>ConfigLoader: Load Configuration
+    ConfigLoader-->>DBTProcessor: Configuration Loaded
+    DBTProcessor->>DBTProcessor: Adjust Configuration
+    DBTProcessor->>Preprocessor: Process Template
+    Preprocessor-->>DBTProcessor: Translated Template
+    DBTProcessor->>Template Directory: Store Translated Template
+    DBTProcessor->>OutputWriter: Write Translated Project
+    OutputWriter-->>User: Translated DBT Project
+```
```
nfx, line 205:
```mermaid
sequenceDiagram
    participant User
    participant DBTProcessor
    participant ConfigLoader
    participant Preprocessor
    participant Template Directory
    participant OutputWriter

    User->>+DBTProcessor: Load DBT Project
    DBTProcessor->>+ConfigLoader: Load Configuration
    ConfigLoader-->>-DBTProcessor: Configuration Loaded
    DBTProcessor->>DBTProcessor: Adjust Configuration
    DBTProcessor->>+Preprocessor: Process Template
    Preprocessor-->>-DBTProcessor: Translated Template
    DBTProcessor->>Template Directory: Store Translated Template
    DBTProcessor->>OutputWriter: Write Translated Project
    OutputWriter-->>User: Translated DBT Project
```

add closing activity blocks to make it more readable

core/src/main/scala/com/databricks/labs/remorph/preprocessor/README.md
```diff
@@ -0,0 +1,205 @@
...
+A simplified sequence diagram of the process is shown below:
+
+```mermaid
+sequenceDiagram
+    participant User
+    participant DBTProcessor
+    participant ConfigLoader
+    participant Preprocessor
+    participant Template Directory
+    participant OutputWriter
```
nfx, line 194: make references to the existing components we have.

I also think we should create a copy of a dbt project instead of doing in-place translation... though, it might be a configurable option.

jimidle:
Comments noted. I agree with all and see that I can change the spec to point out these things. I am currently doing a PoC of placeholding and replace and will then update the spec to address the above.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -347,6 +348,18 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
...
+    case Some(errorResult) => errorResult
+    case None =>
+      ctx.jinjaTemplate().accept(this)
+  }
+
+  override def visitJinjaTemplate(ctx: TSqlParser.JinjaTemplateContext): ir.LogicalPlan = errorCheck(ctx) match {
+    case Some(errorResult) => errorResult
+    case None =>
+      ir.JinjaAsStatement(ctx.getText)
+  }
```
vil1, line 361:
```suggestion
  override def visitJinjaTemplate(ctx: TSqlParser.JinjaTemplateContext): ir.LogicalPlan = errorCheck(ctx).getOrElse(ir.JinjaAsStatement(ctx.getText))
```

core/src/main/scala/com/databricks/labs/remorph/preprocessors/Processor.scala
```diff
@@ -0,0 +1,10 @@
+package com.databricks.labs.remorph.preprocessors
+
+import com.databricks.labs.remorph._
+import org.antlr.v4.runtime.{CharStream, Lexer}
+
+trait Processor extends TransformationConstructors {
```
vil1, line 6:
```suggestion
trait Processor {
```
Let the descendants of `Processor` decide whether they need to mix `TransformationConstructors`.

core/src/main/scala/com/databricks/labs/remorph/preprocessors/Processor.scala
```diff
@@ -0,0 +1,10 @@
...
+  protected def createLexer(input: CharStream): Lexer
+  def pre(input: PreProcessing): Transformation[Parsing]
```
vil1, line 8: It feels a bit odd for this method to return (a `Transformation` whose output is) a `Parsing` phase.

It is likely that what we want to express here is that `pre` it a `Transformation` that takes a `String` and returns a `String`, while also changing the `Phase` from `PreProcessing` to `Parsing`.

That would be better expressed as (inside `Processor` trait, which would then have a good reason for mixing in `TransformationConstructors`):

```scala
final def pre: Transformation[String] = {
  get
  .flatMap{
    case PerProcessing(input) => 
       val preprocessed  = preprocess(input)
      update{
         case _ => Parsing(preprocessed)
       }
    case _ => lift(KoResult(// pre shouldn't be called outside PerProcessing phase)
}

def preprocess(input: String): String
```

core/src/main/scala/com/databricks/labs/remorph/preprocessors/Processor.scala
jimidle: I see. Now I have to go back and see why I thought that was needed ;)

core/src/main/scala/com/databricks/labs/remorph/preprocessors/Processor.scala
jimidle: I also wondered about this. But then I wondered if it was just that I got used to Parsing as being the start point. Let's have a chat aboiut this.

vil1: Well, we should have changed `PlanParser#parse` the same way I suggested above for `Processor#pre` (getting rid of the `input` parameter and instead pull it from the state) long ago. It just hasn't happened yet, unfortunately (my bad, probably).

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -347,6 +348,18 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
       result
   }
 
+  override def visitTsiJinja(ctx: TsiJinjaContext): ir.LogicalPlan = errorCheck(ctx) match {
+    case Some(errorResult) => errorResult
+    case None =>
+      ctx.jinjaTemplate().accept(this)
+  }
```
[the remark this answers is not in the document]
jimidle: I will change this. There are other visitors that can probably also be changed for the same reason - but I wanted to preserve that pattern really as they all do that errorCheck(ctx) match sequence

core/src/main/scala/com/databricks/labs/remorph/coverage/QueryRunner.scala
```diff
@@ -17,13 +17,16 @@ abstract class BaseQueryRunner(transpiler: Transpiler) extends QueryRunner {
       exampleQuery: ExampleQuery,
       output: String,
       error: Option[RemorphError] = None): ReportEntryReport = {
-    val expected = exampleQuery.expectedTranslation.getOrElse("")
+    val expected = exampleQuery.expectedTranslation
     val parsed = if (error.isEmpty) 1 else 0
-    if (exampleQuery.expectedTranslation.map(format).exists(_ != format(output))) {
+    val formattedOutput = if (exampleQuery.shouldFormat) format(output) else output
+    val formattedExpected = expected.map(e => if (exampleQuery.shouldFormat) format(e) else e)
```
asnare, line 23: I think this can be tidied up a bit:
```suggestion
    val formattedOutput = if (exampleQuery.shouldFormat) format(output) else output
    val formattedExpected = if (exampleQuery.shouldFormat) expected.map(format))
```
There's a variant also possible, but I don't think it's really an improvement:
```suggestion
    val (formattedOutput, formattedExpected) = {
      if (exampleQuery.shouldFormat) {
        (format(output), expected.map(format))
      } else {
        (output, expected)
      }
    }
```

core/src/main/scala/com/databricks/labs/remorph/preprocessors/jinja/TemplateManager.scala
```diff
@@ -0,0 +1,51 @@
+package com.databricks.labs.remorph.preprocessors.jinja
+
+import com.databricks.labs.remorph.utils.Sed
+
+import scala.collection.mutable
+
+class TemplateManager {
+  private val templates = mutable.Map[String, TemplateElement]()
+  private var counter = 0
```
asnare, line 9: A minor nit: I tend to prefer `private[this]` for fields: by inspection it means I don't have to worry about looking for possible cross-instance access.

This is covered in the Style Guide, along with another reason:
 - https://github.com/databricks/scala-style-guide?tab=readme-ov-file#private-fields
 - https://github.com/databricks/scala-style-guide?tab=readme-ov-file#privatethis

core/src/main/scala/com/databricks/labs/remorph/utils/Strings.scala
```diff
@@ -61,3 +62,16 @@ object Strings {
     }
   }
 }
+
+class Sed(rules: (String, String)*) {
+  private val compiledRules: Seq[(Regex, String)] = rules.map { case (regex, replace) =>
```
asnare, line 67: If you decide to keep the current `(String, String)` API:
```suggestion
  private[this] val compiledRules: Seq[(Regex, String)] = rules.map { case (regex, replace) =>
```

asnare (commented):
I've mostly looked at the Scala code, not the big picture, and have some comments/suggestions.

sundarshankar89:
@jimidle, how do the macros get tackled? Macros are expressions or functions wrapped in jina. 

for eg: https://github.com/rlsalcido24/lakehouse_utils/blob/main/macros/snowflake/array_agg.sql

jimidle:
> @jimidle, how do the macros get tackled? Macros are expressions or functions wrapped in jina.
> 
> for eg: https://github.com/rlsalcido24/lakehouse_utils/blob/main/macros/snowflake/array_agg.sql

They don't need to be as far as I can tell as they are constrained to JInja syntax. But they will be handles where they are defined.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -590,6 +603,7 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
     // correct source code to be given to remorph.
     val aggregateFunction = ctx.pivotClause().expression().accept(vc.expressionBuilder)
     val column = ctx.pivotClause().fullColumnName().accept(vc.expressionBuilder)
+    // TODO: All other aliases are handled as ir.Id, but here we use ir.Literal - should it change?
```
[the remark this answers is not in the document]
jimidle: IT's just an observation for. someone to take a deeper look later really. No need fore an issue particularly..

core/src/main/scala/com/databricks/labs/remorph/utils/Strings.scala
jimidle: I'd have to change the transpiler as well, so I will change this to `private[this]`

core/src/main/scala/com/databricks/labs/remorph/coverage/QueryRunner.scala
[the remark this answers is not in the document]
asnare: @jimidle: Hmm. Maybe I'm missing something?

If I enumerate the situations:

 * `Some(x)`, `shouldFormat=true` ⇒ `Some(formatted(x))`
 * `Some(x)`, `shouldFormat=false` ⇒ `Some(x)`
 * `None`, `shouldFormat=true` ⇒ `None`
 * `None`, `shouldFormat=false` ⇒ `None`

Under these, hoisting the `shouldFormat`-branch outside of `.map()` is equivalent?

(To be clear, it wasn't a question of correctness.)

asnare (approved):
lgtm

gueniai (approved):
LGTM

PR #1171, opened 2024-11-07
make hatch a pre-requisite

ericvergnaud:
Our current `make` script installs `hatch` which can be very flaky depending on the OS and how `python/pip` were installed.
This PR assumes `hatch` is already installed


nfx (approved):
lgtm

PR #1172, opened 2024-11-07
update setup docs

ericvergnaud:
there are prerequisites for developing remorph locally, which are not yet documented
also, setting up IntelliJ IDEA blindly may lead to an unusable environment
This PR documents these steps


nfx (approved):
lgtm

Issue #869, opened 2024-08-29
[FEATURE]: IR to Listen and Generate Code Comments


sundarshankar89:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Transpile

### Problem statement

Currently the ANTLR Parser moves all comments into Hidden Channel. Need to map them to concrete IR and introduce them at appropriate place during code generation.

### Proposed Solution

Resolving comments back to output in a sensible manner is not a trivial task and first we must establish some principles in the source language and target language. 

#### Goal - Generate comments in the translated code and preserve their context wherever possible. 

In our case the source and target languages can be considered equivalent as the comments have common syntax, so there are no complications caused by the need to resolve comment structure; for instance if the source allows multi-line in-syntax comments (`/* xxx */`), but the target only allows line comments (`-- xxx`).

Additionally, as we will pretty print the resulting output, we can ignore trying to preserve the number of spaces and so on that precede or follow a particular token. So long as the correct comment style is placed in the correct place, then the formatter will make it all look good.

#### Comment style

Using the Databricks documentation for some examples, and adding our own, we can create a sample input that challenges our system such that if the test input is translated correctly, then all input should be translated correctly. The starting point for such a test is in the Additional context section. Note that this is a starting point and has not been thought through exhaustively.

If we study the test we can establish some principles around comment association, knowing that as we are not actually reading the comments, we make guess the associativity incorrectly, but by using these principles, we _should_ get the intended output anyway.

The principles are therefore:

1. If a comment follows some token on the same line, then it likely belongs to that token (though we know that people's intent might be the opposite).
2. A comment that follows some token on the next line is likely associated with the following token, even if that token is EOF, therefore -
3. Any comment preceding a token but not on the same line, belongs to that token, even if it is EOF, except
4. Multiline comments ('/* */')  may be intended to address the previous token but we will choose to associate them with the following token.
5. Single line comments should always be rendered with a trailing new line
6. Multi-line comments should be rendered with a trailing new line if they do not end on the same line as the following token with which they are associated.
7. Ambiguity is always solved with the nearest following token (we could modify this, if say the comment starts on the same line as the previous token and does not end on the same line as the following token. 

These principles will need to be tested in practice - these kinds of things always have strange edge-cases, where the source author thought they were being mighty clever.

#### Procedure

[...]

PR #1182, opened 2024-11-11
Attach SnowFlake line comments to SELECT statements

ericvergnaud:
There is a need to transpile SQL comments
This PR:
 - sends `LINE_COMMENT`s to a dedicated ANLTR4 channel, such that comments do not interfere with SQL constructs
 - adds a `comments` array field to the `LogicalPlan` case class
 - moves the `origin` field of `TreeNode` to a 2nd parameter list such that it is ignored during comparisons, and refactors accordingly
 - refactors the `LogicalPlan` pipeline such that it's possible to use the original `TokenStream` to fetch the comment tokens and apply them appropriately to the plan items. The abstract plan is also no loger dependent on ANTLR4.
 - adds a test for checking transpilation

Progresses #869 

Future PRs will expand this to:
 - TSQL parser
 - more constructs
 - inline comments


core/src/main/scala/com/databricks/labs/remorph/intermediate/plans.scala
```diff
@@ -6,7 +6,7 @@ package com.databricks.labs.remorph.intermediate
  * the [[Command]] type that is used to execute commands on the server. [[Plan]] is a union of Spark's LogicalPlan and
  * QueryPlan.
  */
-abstract class Plan[PlanType <: Plan[PlanType]] extends TreeNode[PlanType] {
+abstract class Plan[PlanType <: Plan[PlanType]]()(origin: Origin) extends TreeNode[PlanType]()(origin) {
```
ericvergnaud, line 9: origin in 2nd parameter list is required to only compare relevant fields

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakePlanParser.scala
```diff
@@ -1,20 +1,68 @@
...
+class SnowflakePlanParser extends PlanParser {
 
   private val vc = new SnowflakeVisitorCoordinator(SnowflakeParser.VOCABULARY, SnowflakeParser.ruleNames)
 
-  override protected def createLexer(input: CharStream): Lexer = new SnowflakeLexer(input)
-  override protected def createParser(stream: TokenStream): SnowflakeParser = new SnowflakeParser(stream)
-  override protected def createTree(parser: SnowflakeParser): ParserRuleContext = parser.snowflakeFile()
-  override protected def createPlan(tree: ParserRuleContext): ir.LogicalPlan = vc.astBuilder.visit(tree)
-  override protected def addErrorStrategy(parser: SnowflakeParser): Unit =
+  override def parseLogicalPlan(parsing: Parsing): Transformation[LogicalPlan] = {
```
ericvergnaud, line 16: we need both the plan and the token stream to attach comments so had to move away from the original design

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlPlanParser.scala
```diff
@@ -1,19 +1,51 @@
...
+class TSqlPlanParser extends PlanParser {
 
   val vc = new TSqlVisitorCoordinator(TSqlParser.VOCABULARY, TSqlParser.ruleNames)
 
-  override protected def createLexer(input: CharStream): Lexer = new TSqlLexer(input)
-  override protected def createParser(stream: TokenStream): TSqlParser = new TSqlParser(stream)
-  override protected def createTree(parser: TSqlParser): ParserRuleContext = parser.tSqlFile()
-  override protected def createPlan(tree: ParserRuleContext): ir.LogicalPlan = vc.astBuilder.visit(tree)
-  override protected def addErrorStrategy(parser: TSqlParser): Unit = parser.setErrorHandler(new TSqlErrorStrategy)
+  override def parseLogicalPlan(parsing: Parsing): Transformation[LogicalPlan] = {
```
ericvergnaud, line 16: we need both the plan and the token stream to attach comments so had to move away from the original design

core/src/main/scala/com/databricks/labs/remorph/transpilers/Transpiler.scala
```diff
@@ -49,14 +48,12 @@ trait Formatter {
 
 abstract class BaseTranspiler extends Transpiler with Formatter with TransformationConstructors {
 
-  protected val planParser: PlanParser[_]
+  protected val planParser: PlanParser
```
ericvergnaud, line 51: drop dependency on ANTLR$ - no longer required

core/src/main/scala/com/databricks/labs/remorph/intermediate/trees.scala
```diff
@@ -252,31 +265,12 @@ abstract class TreeNode[BaseType <: TreeNode[BaseType]] extends Product {
...
-          // to older heuristic
-          false
-        } else {
-          val argsArray: Array[Class[_]] = allArgs.map(_.getClass)
-          isAssignable(argsArray, ctor.getParameterTypes)
-        }
-      }
-      .getOrElse(ctors.maxBy(_.getParameterTypes.length)) // fall back to older heuristic
+
+    val (defaultCtor, allArgs) = locateConstructor(ctors, newArgs, appendOrigin = false)
```
ericvergnaud, line 269: the constructor may or not require the Origin paranter from the 2nd parameter list, so try locating constructor with and without it

core/src/test/scala/com/databricks/labs/remorph/discovery/AnonymizerTest.scala
```diff
@@ -62,7 +62,7 @@ class AnonymizerTest extends AnyWordSpec with Matchers {
         Fingerprint(
           "id",
           new Timestamp(1725032011000L),
-          "290f4d72ca8faeb28873d8fff779ce93ed5cdb69",
+          "93f60d8795c8bffa2aafe174ae8a867b42235755",
```
ericvergnaud, line 65: To be discussed: should we generate a plan from a failed parse

core/src/main/scala/com/databricks/labs/remorph/intermediate/plans.scala
vil1: As per [the guidelines](https://github.com/databricks/scala-style-guide?tab=readme-ov-file#multi-param-list) we should avoid multiple parameter lists. Could we come up with a solution that doesn't involve adding these parameter lists?

ericvergnaud: I agree that mpl _should_ be `avoided`, but that's the only official way I know to exclude origin from comparisons in both TreeNode and all derived classes. I'm happy to consider any suggestion.

vil1: From the back of my head, I guess something along the lines of:

```scala
trait TreeNode {
  private [this] var _origin: Origin = Origin.empty

  def origin: Origin = _origin

  def withOrigin(updatedOrigin: Origin): this.type = {
    _origin = updatedOrigin
    this 
  }
}
```

should do the trick

ericvergnaud:
> This is quite a big change and somewhat challenging to review. Could you try and split this PR into smaller ones?
> 
> I believe we could at least have the changes related to `Origin` in a separate PR.
> 
> Stacking PRs is fine (PR `n` targeting the branch of PR `n - 1`).

ok, see https://github.com/databrickslabs/remorph/pull/1189

core/src/test/scala/com/databricks/labs/remorph/discovery/AnonymizerTest.scala
jimidle: Yes. Parse failures are generated as comments.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakePlanParser.scala
jimidle: > we need both the plan and the token stream to attach comments so had to move away from the original single-enity-transform design

Could we not just accumulate he comments in a structure that allows search by their location? Either after the lexer is called or just within the lexer directly? Then we would not need the token stream and this map or whatever it is can be retrieved from the common lexer, then passed to the generator. 

Perhaps we need to settle on a design for how we are resolving comment location before coding?

ericvergnaud:
superseded by multiple smaller PRs

PR #1189, opened 2024-11-12
Refactor origin in tree node

ericvergnaud:
There is a need to transpile SQL comments. This requires populating TreeNode.origin when building the ir AST.

This PR:
 - moves the `origin` field of `TreeNode` to a 2nd parameter list such that it is ignored during comparisons, and refactors accordingly

Progresses #869 

Future PRs will expand this one (see #1182)

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/intermediate/trees.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/intermediate/trees.scala
@@ -54,13 +54,15 @@ class TreeNodeException[TreeType <: TreeNode[_]](@transient val tree: TreeType, 
 }
 
 // scalastyle:off
-abstract class TreeNode[BaseType <: TreeNode[BaseType]](@JsonIgnore origin: Origin) extends Product {
+abstract class TreeNode[BaseType <: TreeNode[BaseType]](_origin: Option[Origin] = Option.empty) extends Product {
   // scalastyle:on
   self: BaseType =>
 
   @JsonIgnore lazy val containsChild: Set[TreeNode[_]] = children.toSet
   private lazy val _hashCode: Int = productHash(this, scala.util.hashing.MurmurHash3.productSeed)
   private lazy val allChildren: Set[TreeNode[_]] = (children ++ innerChildren).toSet[TreeNode[_]]
+
+  def origin: Option[Origin] = _origin
 
   /**
    * Returns a Seq of the children of this node. Children should not change. Immutability required for containsChild
--- a/core/src/main/scala/com/databricks/labs/remorph/intermediate/trees.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/intermediate/trees.scala
@@ -12,33 +12,30 @@ import scala.util.control.NonFatal
 private class MutableInt(var i: Int)
 
 case class Origin(
-    startLine: Option[Int] = None,
-    startColumn: Option[Int] = None,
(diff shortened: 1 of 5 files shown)


vil1 (changes requested):
This looks better than the previous encoding with multiple parameter lists.

However, with this new encoding, we won't be able to change a node's origin after creation. That means that every `visit*` method in our `Builder` classes will have to properly set the origin.

Again, I suggest we go for something like 

```scala
trait TreeNode[BaseType <: TreeNode[BaseType]] extends Product {
  private [this] var _origin: Origin = Origin.empty

  def origin: Origin = _origin

  def withOrigin(updatedOrigin: Origin): BaseType = {
    _origin = updatedOrigin
    self 
  }
}
```

so that we could at least have creation of IR nodes and call to their `withOrigin` method happen at different times

core/src/main/scala/com/databricks/labs/remorph/intermediate/trees.scala
```diff
@@ -10,36 +12,34 @@ import scala.util.control.NonFatal
...
-object CurrentOrigin {
-  private val value = new ThreadLocal[Origin]() {
-    override def initialValue: Origin = Origin()
-  }
+    startLine: Option[Int] = None,
+    startColumn: Option[Int] = None,
+    endLine: Option[Int] = None,
+    endColumn: Option[Int] = None,
+    startTokenIndex: Option[Int] = None,
+    endTokenIndex: Option[Int] = None)
```
vil1, line 20: Is it possible/meaningful for some of the fields to be `Some` while others are `None`, or is it either all fields `None` or all fields `Some` ?

In the latter case, we would be better to either `origin` be an `Option[Origin]` or, even better, define `Origin` as
```scala
sealed trait Origin
case object UnknownOrigin extends Origin
case class InputOrigin(
    startLine: Int,
    startColumn: Int,
    endLine: Int,
    endColumn: Int,
    startTokenIndex: Int,
    endTokenIndex: Int) extends Origin

// optionally, adding the following case for the node that are synthesized during optimization phase
// could be handy.
case class SyntheticOrigin(synthesizedBy: String) extends Origin
```

jimidle:
Also, don't forget this:

https://databricks.slack.com/archives/C070VK7G895/p1730991678494919

ericvergnaud:
> This looks better than the previous encoding with multiple parameter lists.
> 
> However, with this new encoding, we won't be able to change a node's origin after creation. That means that every `visit*` method in our `Builder` classes will have to properly set the origin.
> 
> Again, I suggest we go for something like
> 
> ```scala
> trait TreeNode[BaseType <: TreeNode[BaseType]] extends Product {
>   private [this] var _origin: Origin = Origin.empty
> 
>   def origin: Origin = _origin
> 
>   def withOrigin(updatedOrigin: Origin): BaseType = {
>     _origin = updatedOrigin
>     self 
>   }
> }
> ```
> 
> so that we could at least have creation of IR nodes and call to their `withOrigin` method happen at different times

Can you provide a scenario where this would be both useful and feasible ? I can't think of one. The only moment we can provide the origin is during conversion of ParserRuleContext nodes to ir nodes. Using 2 separate calls seems error prone (it's very easy to forget to call withOrigin). In scenarios where we don't have an origin, we should simply supply Origin.empty at ctor time.

ericvergnaud:
superseded by #1199\

PR #1196, opened 2024-11-12
Fixing presto strpos and array_average functions

ganeshdogiparthi-db:
In this PR we are handling both STRPOS and ARRAY_AVERAGE functions.

`STRPOS:`
--presto SQL:
SELECT strpos('Hello world', 'l', 2);

--databricks sql:
SELECT LOCATE('l', 'Hello world', 2);

`ARRAY_AVERAGE:`
Databricks doesn't have any built-in function to calculate the average of an array, so we are handling it using nested functions available in Databricks. It handles nulls, Integer and Double.

--presto SQL:
rray_average(arr)

--databricks SQL:
AGGREGATE(
      FILTER(arr, x -> x IS NOT NULL),
      NAMED_STRUCT('sum', CAST(0 AS DOUBLE), 'cnt', 0),
      (acc, x) -> NAMED_STRUCT('sum', acc.sum + x, 'cnt', acc.cnt + 1),
      acc -> TRY_DIVIDE(acc.sum, acc.cnt)
    )

### DIFF
--- a/src/databricks/labs/remorph/snow/local_expression.py
+++ b/src/databricks/labs/remorph/snow/local_expression.py
@@ -180,3 +180,11 @@ class MapKeys(Func):
 
 class ArrayExists(Func):
     arg_types = {"this": True, "expression": True}
+
+
+class Locate(Func):
+    arg_types = {"substring": True, "this": True, "position": False}
+
+
+class NamedStruct(Func):
+    arg_types = {"expressions": True}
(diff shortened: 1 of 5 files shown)


sundarshankar89 (approved):
LGTM

bishwajit-db (changes requested):

bishwajit-db (approved):
LGTM

PR #1199, opened 2024-11-13
Refactor origin in tree node

ericvergnaud:
There is a need to transpile SQL comments. This requires populating TreeNode.origin when building the ir AST.

This PR:
 - transforms the `origin` field of `TreeNode` to a method that returns a parameter that defaults 2nd parameter list such that it is ignored during comparisons, and refactors accordingly

Progresses #869 

Future PRs will expand this one (see #1182)

Supersedes #1189


ericvergnaud:
Cancelled

PR #1192, opened 2024-11-12
Drop PlanParser dependency on antlr4 

ericvergnaud:
Removes dependency on antlr4 in PlanParser

Requires https://github.com/databrickslabs/remorph/pull/1191

Progresses https://github.com/databrickslabs/remorph/issues/869

PR #1201, opened 2024-11-13
Drop plan parser dependency on antlr4 

ericvergnaud:
Our current PlanParser enforces use of ANTLR4 to parse input.
This dependency is undesirable, and does not provide any benefit.

Thanks to #1200, this is no longer required.

This PR:
Removes dependency on antlr4 in PlanParser

Requires https://github.com/databrickslabs/remorph/pull/1200

Supersedes #1192


jimidle (changes requested):
Making this less generic will force reimplementation of code in the specific plan parsers. I don't think that this is a good idea. Dependency on ANTLR is not really a problem for the parser I think.

Perhaps you can explain why you want to dothis?

PR #1203, opened 2024-11-13
Transpile snow flake select statement

ericvergnaud:
borrows technical idea from #1208, but applies it to a new field, thus avoiding any backwards compatibility issue.

Adds a test for checking transpilation of SnowFlake select statement, and fixes issues

Requires https://github.com/databrickslabs/remorph/pull/1201

Supersedes #1193


core/src/main/scala/com/databricks/labs/remorph/intermediate/plans.scala
```diff
@@ -138,14 +138,15 @@ abstract class Plan[PlanType <: Plan[PlanType]](_origin: Option[Origin] = Option
...
-abstract class LogicalPlan extends Plan[LogicalPlan] {
+abstract class LogicalPlan(_origin: Option[Origin] = Option.empty) extends Plan[LogicalPlan](_origin) {
 
   /**
    * Returns true if this expression and all its children have been resolved to a specific schema and false if it still
    * contains any unresolved placeholders. Implementations of LogicalPlan can override this (e.g.[[UnresolvedRelation]]
    * should return `false`).
    */
   lazy val resolved: Boolean = expressions.forall(_.resolved) && childrenResolved
+  var comments: Array[CommentNode] = Array()
```
ericvergnaud, line 149: to be discussed, mutating a node at parse time is no big deal, alternately we could insert comment nodes before first child (or after depending on context), but that also requires mutating the tree

jimidle (commented):
I think we have started running before we know the correct direction. As per previous comments, we need to design this as it is potentially a big change.

ericvergnaud:
> I think we have started running before we know the correct direction. As per previous comments, we need to design this as it is potentially a big change.

Indeed but feasability can only be proven by feasing...

Now we know waht can be done, and we have evidence of which problems are indtroduced

PR #1208, opened 2024-11-14
Alternate proposal for origin tracking

vil1:
Here's a proposal for tracking origin in IR nodes that:
- doesn't rely on maintaining a thread-local mutable variable (which is error prone as it is easy to forget to update said variable when needed)
- doesn't require any change in the structure of IR nodes

The proposed solution ensures that nodes return by the `visit*` methods of AST builders have their origin populated, by introducing a `WithKnownOrigin` marker trait.

Drawbacks are:
- addition of a mutable field in `TreeNode`, although said field is marked as `private [this]`.
- the proposed change is fairly huge

Remaining work:
- finish implementing actual origin computation (by taking inspiration from #1199)
- remove `CurrentOrigin`
- propagate origin when relevant in the `transform*` methods


sundarshankar89:
@ericvergnaud  already moved we can close this. @vil1  Thoughts?

PR #1213, opened 2024-11-18
Constrain the Databricks SDK: version 0.37.0 introduced breaking changes which affect our tests

asnare:
This PR updates the project dependencies to not use version 0.37.0 (or later) of the Databricks SDK: it introduced some breaking changes that affect our unit tests.

This is intended to be a temporary fix to unblock CI and new installations; these install the latest version.


asnare:
Closing pending a better understanding of the problem.

PR #1217, opened 2024-11-18
Limit LSQL to releases before 0.14.0

asnare:
The 0.14.0 release depends on Databricks SDK 0.37.0, but doesn't declare this. As such it can't be installed due to our dependency on Databricks SDK 0.29.0.

Reference: databrickslabs/lsql#330


vil1 (approved):
LGTM

sundarshankar89 (approved):
LGTM

PR #1218, opened 2024-11-18
Fix implementation of Snowflake's TO_TIME/TO_TIMESTAMP functions

vil1:
Improve the way `TO_TIME` and `TO_TIMESTAMP` are translated.

- When called with an explicit format, said format is properly translated (for instance, SF's format `YYYY-MM-DD HH24:MITZH:TZM` gets translated into DB's `yyyy-MM-dd'T'HH:mmZZZ`), be it a constant or a variable.
- When SF's TO_TIMESTAMP is called with the `TO_TIMESTAMP ( <numeric_expr>  , <scale> )` we simply truncate `numeric_expr` to a number-of-seconds-since-epoch, as it seems to be the best we can do.
- When TO_TIMESTAMP is called with no explicit format, we do our best as to handle the automatic formats that SF handles but DB doesn't by trying each explicitly until one matches
- the `DY` format part (corresponding to `EEE` in DB), ie. day-of-week, isn't allowed anymore since Spark 3.0. Since it isn't relevant in figuring out a specific point in time, we simply ignore it in both translated format and input string (resulting in huge amount of SQL being produced, see `tests/resources/functional/snowflake/core_engine/functions/conversion/to_timestamp_variable_format.sql` for an example)


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -62,6 +62,7 @@ class SnowflakeCallMapper extends ir.CallMapper with ir.IRHelpers {
       case ir.CallFunction("POSITION", args) => ir.CallFunction("LOCATE", args)
       case ir.CallFunction("REGEXP_LIKE", args) => ir.RLike(args.head, args(1))
       case ir.CallFunction("REGEXP_SUBSTR", args) => regexpExtract(args)
+      case ir.CallFunction("SHA2", args) => ir.Sha2(args.head, args.lift(1).getOrElse(ir.Literal(256)))
```
vil1, line 65: This is an unrelated fix that sorta slipped into this PR, I can remove it if requested

tests/resources/functional/snowflake/core_engine/functions/conversion/test_to_time_2.sql
```diff
@@ -3,4 +3,4 @@
 SELECT TO_TIME('2018-05-15 00:01:02');
 
 -- databricks sql:
-SELECT TO_TIMESTAMP('2018-05-15 00:01:02', 'yyyy-MM-dd HH:mm:ss');
+SELECT TO_TIMESTAMP('2018-05-15 00:01:02');
```
sundarshankar89, line 6: Can you check this https://docs.snowflake.com/en/sql-reference/functions/to_time, time only expects, parsable time format.

sundarshankar89: This should be the output.

```
SELECT DATE_FORMAT(TO_TIMESTAMP('00:01:02'), 'HH:mm:ss');

**Review by sundarshankar89 — CHANGES_REQUESTED** [review]

File: tests/resources/functional/snowflake/core_engine/functions/conversion/to_timestamp_variable_format.sql
Comment (vil1) [inline tests/resources/functional/snowflake/core_engine/functions/conversion/to_timestamp_variable_format.sql:1]: Quite a mouthful isn't it?

File: tests/resources/functional/snowflake/core_engine/functions/conversion/to_timestamp_variable_format.sql [hunk 3]
```diff
@@ -0,0 +1,179 @@
+-- snowflake sql:
+	SELECT TO_TIMESTAMP(str, fmt) FROM (VALUES ('2024-11-20T18:05:59.123456789', 'YYYY-MM-DD"T"HH24:MI:SS.FF'), ('Thu, 21 Dec 2000 04:01:07 PM +0200', 'DY, DD MON YYYY HH12:MI:SS AM TZHTZM') ) AS vals(str, fmt);
+
+-- databricks sql:
+SELECT IF(
```
sundarshankar89, line 5: We need to align better on these transformations long term approach.

sundarshankar89 (approved):
LGTM

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
jimidle: It's one line, so I think we can live ;)

jimidle (approved):
LGTM

Issue #1222, opened 2024-11-19
[LCA]: Ability replace Lateral Alias with actual column name and Expression


sundarshankar89:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Transpile

### Problem statement

Given the following Snowflake SQL:
```
SELECT column_a as customer_id
FROM mytable
WHERE customer_id = '123'
```

The following Spark SQL should be returned because of the different alias scoping used in Spark SQL

``` 
SELECT column_a as customer_id
FROM mytable
WHERE column_a = '123'
```

https://www.databricks.com/blog/introducing-support-lateral-column-alias

### Proposed Solution

NA

sundarshankar89:
@ericvergnaud  this might be an issue.

sundarshankar89:
more such examples are present under `lca_utils.py` on how we tackle we haven't covered all combinations but you can refer how we do preprocessing using python prototype.

ericvergnaud:
@sundarshankar89 re the output, did you mean:

```
SELECT column_a as customer_id
FROM mytable
WHERE column_a = '123'
```

?

sundarshankar89:
Just edited the description.

Also checkout `test_lca_utils.py` it has few more examples with LCA in sourcesystem.

asnare:
> Given the following Snowflake SQL:
> 
> ```
> SELECT column_a as customer_id
> FROM mytable
> WHERE customer_id = '123'
> ```
> 
> The following Spark SQL should be returned because of the different alias scoping used in Spark SQL
> 
> ```
> SELECT column_a as customer_id
> FROM mytable
> WHERE column_a = '123'

We need to be careful here because there's a problem with using this[^1] as a universal strategy. It's fine for simple column references, but if it's an expression then the semantics can be altered because:
 - Expressions aren't always deterministic.
 - Aliases imply 'evaluate once' semantics.
 - This can lead to the expression evaluating differently during projection and predicate evaluation.

A more accurate strategy would be something like:
```sql
SELECT expr_a AS customer_id
FROM mytable
WHERE customer_id = '123'
```
becomes
```sql
SELECT customer_id
FROM (
  SELECT expr_a AS customer_id
  FROM mytable
)
WHERE customer_id = '123'
```

[...]

PR #1225, opened 2024-11-20
Fix issue 1222

ericvergnaud:
Progresses #1222 by implementing `DealiasLCAs`, a transformation rule that replaces LCAs by the underlying alias expression in project filters (`where` clause of a `select` statement)


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAFilter.scala
```diff
@@ -0,0 +1,83 @@
...
+      case filter: Filter => dealiasProject(project, filter)
+      case _ => project
+    }
+  }
+
+  def dealiasProject(project: Project, filter: Filter): Project = {
+    val aliases = project.columns
+      .filter(col => col.isInstanceOf[Alias])
+      .map(col => col.asInstanceOf[Alias])
+      .filter(alias => alias.child.isInstanceOf[Id]) // TODO do we need to support more than that ?
```
jimidle, line 22: We should avoid isInstanceOf in favor of match

jimidle (changes requested):
Find a way to not use asInstanceOf

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAFilter.scala
ericvergnaud: I'd agree if we cared about more than 1 type, but in this case it would make the code much more complex without a clear benefit ?

jimidle: The benefit is not using asInstanceOf ;)

ericvergnaud: how is that a clear benefit ?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAFilter.scala
```diff
@@ -17,10 +17,9 @@ class DealiasLCAFilter extends Rule[LogicalPlan] with IRHelpers {
 
   def dealiasProject(project: Project, filter: Filter): Project = {
     val aliases = project.columns
-      .filter(col => col.isInstanceOf[Alias])
-      .map(col => col.asInstanceOf[Alias])
-      .filter(alias => alias.child.isInstanceOf[Id]) // TODO do we need to support more than that ?
-      .map(alias => alias.name.id -> alias.child.asInstanceOf[Id].id)
+      .collect { case a: Alias => a }
+      .filter(alias => !alias.child.isInstanceOf[Literal])
+      .map(alias => alias.name.id -> alias.child)
```
vil1:
```suggestion
      .collect { case Alias(e, name) if !e.isInstanceOf[Literal] => name.id -> e }
```

further distillation

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAFilter.scala
```diff
@@ -72,11 +81,7 @@ class DealiasLCAFilter extends Rule[LogicalPlan] with IRHelpers {
       if (alias.isEmpty) {
         item
       } else {
-        val replacement = alias.get._2
-        item transform {
-          case name: Name => name.makeCopy(Array(replacement))
-          case id: Id => id.makeCopy(Array(replacement.asInstanceOf[AnyRef], id.caseSensitive.asInstanceOf[AnyRef]))
-        }
+        alias.get._2
```
vil1: The whole function could become:
```scala
Option(item).collect{
  case Name(name) => name
  case Id(id, _) => id
}
.flatMap(aliases.get)
.getOrElse(item)
```
or


```scala 
val key = item match {
  case Name(name) => name
  case Id(id, _) => name
  case _ => null
}
aliases.getOrElse(key, item)
```

The former being slightly preferable, even though it is unlikely that `aliases` would have a entry with a `null` key.

vil1 (changes requested):
I think the pattern of calling `transform` and then checking the result with `eq` against the original value before calling `makeCopy` if there was a change is a premature optimization that unnecessarily hurts readability. Even more so if we consider that `transform` already performs a similar optimization on the descendants of the transformed tree.

As an example, `dealiasProject` would be more readable as:

```scala
private def dealiasProject(project: Project, filter: Filter, aliases: Map[String, Expression]): Project = project.copy(input = dealiasFilter(filter, aliases))
```

at the expense of a single object allocation.

vil1:
It wouldn't hurt to add a test where aliases appear deeply nested in an arbitrarily contrived expression involving boolean operators (including `NOT`), function calls, etc.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
vil1, line 1: The more I look into this, the more I think there might be an easier way to achieve the same using the right combination of `transform` and `transformAllExpressions`. By "simpler", I mean a solution that wouldn't require specifying every possible case (as in `dealiasWindow`, `dealiasIn`, etc).

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAFilter.scala
ericvergnaud: outdated

ericvergnaud:
> I think the pattern of calling `transform` and then checking the result with `eq` against the original value before calling `makeCopy` if there was a change is a premature optimization that unnecessarily hurts readability. Even more so if we consider that `transform` already performs a similar optimization on the descendants of the transformed tree.
> 
> As an example, `dealiasProject` would be more readable as:
> 
> ```scala
> private def dealiasProject(project: Project, filter: Filter, aliases: Map[String, Expression]): Project = project.copy(input = dealiasFilter(filter, aliases))
> ```
> 
> at the expense of a single object allocation.

Open for discussion, but the expense is much higher than _a single object allocation_ since you'd scan and transform a lot of stuff before discovering that you haven't changed anything (and might have lost information along the way).

vil1:
> Open for discussion, but the expense is much higher than _a single object allocation_ since you'd scan and transform a lot of stuff before discovering that you haven't changed anything (and might have lost information along the way).

If you look into `transform` implementation, you'll see that it already implements such "copy avoiding" mechanism, so wrapping the result of `transform` in a new node (ie, calling copy in our `Rule`s) is a marginal cost (that doesn't justify hurting readability/future refactoring IMO).

ericvergnaud:
> Moreover, as it relies on reflexion under the hood, `makeCopy` is prone to silently break if the order/types of constructor parameters gets changed in the future (silently = at runtime). If we insist on using `makeCopy` here, we should make ensure 100% test coverage to prevent such silent breaking from happening.

I prefer `makeCopy` for now because as we've seen, the constructor doesn't convey origin, comments etc... so I fear information would be lost

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
ericvergnaud: and I wouldn't _want_ to blindly transform expressions. Depending on where they are located, what we want to do may vary...

ericvergnaud:
> > Open for discussion, but the expense is much higher than _a single object allocation_ since you'd scan and transform a lot of stuff before discovering that you haven't changed anything (and might have lost information along the way).
> 
> If you look into `transform` implementation, you'll see that it already implements such "copy avoiding" mechanism, so wrapping the result of `transform` in a new node (ie, calling copy in our `Rule`s) is a marginal cost (that doesn't justify hurting readability/future refactoring IMO).

It would be interesting to measure that cost

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAFilter.scala
ericvergnaud: done

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
ericvergnaud: Just bumped into a showstopper for using transform 'as is'.
Say you have `Id("some_id")` transformed to `FunctionCall(Id("some_id"))` - there is a use case for this, then `transformDown` will cause a stack overflow because it will recursively call the same transform.

core/src/main/scala/com/databricks/labs/remorph/intermediate/expressions.scala
```diff
@@ -161,7 +161,25 @@ case class Cast(
     type_str: String = "",
     returnNullOnError: Boolean = false,
     timeZoneId: Option[String] = None)
-    extends Unary(expr)
+    extends Unary(expr) {
+
+  override def makeCopy(newArgs: Array[AnyRef]): Expression = {
```
vil1, line 166: I don't understand why we need overriding `makeCopy` here, is there a test somewhere showcasing the problem/solution ?

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
```diff
@@ -0,0 +1,24 @@
...
+  SUBSTR(ca_zip,1,5) AS ca_zip,
+  TRIM(name) AS name,
+  COUNT(*) OVER (
+    PARTITION BY
+      SUBSTR(ca_zip,1,5)
+  )
+FROM
+  customer_address
+WHERE
+  SUBSTR(ca_zip,1,5) IN ('89436', '30868');
```
vil1, line 24: This seems to be syntactically incorrect on Databricks (probably missing parenthesis around the subquery)

core/src/main/scala/com/databricks/labs/remorph/intermediate/expressions.scala
ericvergnaud: It crashes if you don't. `TreeNode.makeCopy` docs say:
```
Must be overridden by child classes that have
   * constructor arguments that are not present in the productIterator.

```
I don't know exactly why only 1 arg of the `Cast` ctor is seen by the `productIterator` (maybe because they are not `val`s), but it's what happens.
This is tested via the test cases in `resources/functional/snowflake/core_engine/lca`

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
ericvergnaud: The parenthesis issue is fixed in PR #1232.
This test is about homonyms, not subqueries, so I'm afraid I disagree with your comment re names.

ericvergnaud:
Superseded by #1241

PR #1226, opened 2024-11-20
add support for SnowFlake SUBSTR

ericvergnaud:
SnowFlake supports both SUBSTR and SUBSTRING, see https://docs.snowflake.com/fr/sql-reference/functions/substr.
This PR fills the missing SUBSTR


vil1 (changes requested):
It looks like `SUBSTRING` wasn't implemented either. 
Looking into `SnowflakeFunctionBuilder` we can see 

```scala
    case "SUBSTR , SUBSTRING" => FunctionDefinition.standard(3)
```
Which appears to be a remnant of the time I scrapped all the function names from the documentation. This malformed definition hasn't been fixed since then, now could be the right time.

If we are certain that `SUBSTR` and `SUBSTRING` will have the same signature in all dialects, we can get rid of that ill-formed definition in SnowflakeFunctionBuilder.

Moreover, to avoid duplicating code in the CallMapper, we can define things like so in the FunctionBuilder:
```scala
case "SUBSTR" => FunctionDefinition.standard(2, 3)
case "SUBSTRING" => FunctionDefinition.standard(2, 3).withConversionStrategy(SynonymOf("SUBSTR"))
```

That way, the CallMapper need only to care about `SUBSTR`.

ericvergnaud:
> If we are certain that `SUBSTR` and `SUBSTRING` will have the same signature in all dialects, we can get rid of that ill-formed definition in SnowflakeFunctionBuilder.

I would certainly not make that bet

jimidle:
Tests are not passing - I presume you saw that?

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
```diff
@@ -191,243 +191,248 @@ class FunctionBuilderSpec extends AnyFlatSpec with Matchers with TableDrivenProp
...
-      ("UPPER", Some(FunctionDefinition.standard(1))),
-      ("USER", Some(FunctionDefinition.standard(0))),
-      ("USER_ID", Some(FunctionDefinition.standard(0, 1))),
-      ("USER_NAME", Some(FunctionDefinition.standard(0, 1))),
-      ("VALUE", Some(FunctionDefinition.xml(2))),
-      ("VAR", Some(FunctionDefinition.standard(1))),
-      ("VARP", Some(FunctionDefinition.standard(1))),
-      ("XACT_STATE", Some(FunctionDefinition.standard(0))),
-      ("YEAR", Some(FunctionDefinition.standard(1))))
+      ("ABS", FunctionDefinition.standard(1)),
```
ericvergnaud, line 194: Not sure what we are testing here ? Not behavior. I suggest we get rid of this test altogether...

ericvergnaud:
> Tests are not passing - I presume you saw that?

Yes, it's fixed now

ericvergnaud:
Closed in favor of #1238

PR #1227, opened 2024-11-21
Implement remaining TSQL set operations.

### LINKED ISSUE #1102: [TODO] This will change when UNION is implemented correctly
See: https://github.com/databrickslabs/remorph/blob/661b68471a043d68b673fbca50b31684fa447310/core/src/test/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilderSpec.scala#L758-L763

### LINKED ISSUE #1126: [TODO] Implement set ops
See [more context](https://github.com/databrickslabs/remorph/blame/661b68471a043d68b673fbca50b31684fa447310/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala#L66-L71):
https://github.com/databrickslabs/remorph/blob/661b68471a043d68b673fbca50b31684fa447310/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala#L66-L71

asnare:
## Changes

This PR implements support for parsing set operations with TSql: `UNION [ALL]`, `EXCEPT` and `INTERSECT`.

The grammar previously supported these but they were not being converted to the IR.

### Linked issues

Resolves #1126.
Resolves #1102.

### Tests

- [ ] added unit tests
- [ ] added transpiler tests
- [ ] added functional tests

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
@@ -64,7 +64,9 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
     case Some(errorResult) => errorResult
     case None =>
       ctx match {
-        case qs if qs.querySpecification() != null => qs.querySpecification().accept(this) // TODO: Implement set ops
+        case qs if qs.querySpecification() != null =>
+          val lhs = qs.querySpecification().accept(this)
+          qs.sqlUnion().asScala.foldLeft(lhs)(buildSetOperation)
         case qe if qe.queryExpression() != null =>
           qe.queryExpression().asScala.map(_.accept(this)) match {
             case Seq(lhs) => lhs
@@ -73,6 +75,20 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
               ir.SetOperation(lhs, rhs, ir.UnionSetOp, is_all = isAll, by_name = false, allow_missing_columns = false)
           }
       }
+  }
+
+  private[tsql] def buildSetOperation(lhs: ir.LogicalPlan, ctx: TSqlParser.SqlUnionContext): ir.LogicalPlan = {
+    val setOp = ctx match {
+      case u if u.UNION() != null => ir.UnionSetOp
+      case e if e.EXCEPT() != null => ir.ExceptSetOp
+      case i if i.INTERSECT() != null => ir.IntersectSetOp
+    }
+    val isAll = ctx.ALL() != null
+    val rhs = ctx match {
+      case qs if qs.querySpecification() != null => qs.querySpecification().accept(this)
+      case qe if qe.queryExpression() != null => qe.queryExpression().accept(this)
+    }
+    ir.SetOperation(lhs, rhs, setOp, is_all = isAll, by_name = false, allow_missing_columns = false)
   }
 
   override def visitQuerySpecification(ctx: TSqlParser.QuerySpecificationContext): ir.LogicalPlan = errorCheck(
--- a/core/src/test/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilderSpec.scala
+++ b/core/src/test/scala/com/databricks/labs/remorph/parsers/tsql/TSqlAstBuilderSpec.scala
@@ -759,8 +759,13 @@ class TSqlAstBuilderSpec extends AnyWordSpec with TSqlParserTestCommon with Matc
           Some(Seq(Id("ID"), Id("Name"))),
           Project(
             TableAlias(
-              // TODO: This will change when UNION is implemented correctly
-              Project(namedTable("TableA"), Seq(simplyNamedColumn("ID"), simplyNamedColumn("Name"))),
+              SetOperation(
+                Project(namedTable("TableA"), Seq(simplyNamedColumn("ID"), simplyNamedColumn("Name"))),
+                Project(namedTable("TableB"), Seq(simplyNamedColumn("ID"), simplyNamedColumn("Name"))),
+                UnionSetOp,
+                is_all = false,
+                by_name = false,
+                allow_missing_columns = false),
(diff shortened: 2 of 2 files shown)


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -75,6 +77,20 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
       }
   }
 
+  private[tsql] def buildSetOperation(lhs: ir.LogicalPlan, ctx: TSqlParser.SqlUnionContext): ir.LogicalPlan = {
```
vil1, line 80: can be simply `private`, unless you need to call it explicitly from the tests

core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
```diff
@@ -2794,6 +2794,7 @@ queryExpression
     ;
 
 sqlUnion
+    // TODO: Handle INTERSECT precedence in the grammar; it has higher precedence than EXCEPT and UNION ALL.
```
jimidle, line 2797: I will look in to this - the TODO shoudl have raised an issue?

core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
```diff
@@ -2802,6 +2803,7 @@ querySpecification
     ;
 
 selectOptionalClauses
+    // TODO: Fix ORDER BY; it needs to be outside the set operations instead of between.
```
jimidle, line 2806: And also this - sometimes the original author implemented it just as reading the syntax diagram from MS, which does not always deal with such things

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -64,17 +65,74 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
...
+            ir.SetOperation(lhs, rhs, ir.UnionSetOp, is_all = isAll, by_name = false, allow_missing_columns = false)
           }
       }
   }
 
+  @tailrec
+  private[this] def buildIntersectOperations(
+      lhs: ir.LogicalPlan,
+      remainingContexts: Seq[TSqlParser.SqlUnionContext]): ir.LogicalPlan = {
+    /*
```
jimidle, line 83: Maybe this should be the the TODO as it has a better explanation?

jimidle (approved):
LGTM - other than maybe moving where the TODO is so we just get one created by the bot and it points to the explanatory text? But that is not a blocker for approval.

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -64,17 +65,74 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
...
+     * 1. Brackets.
+     * 2. INTERSECT
+     * 3. UNION and EXCEPT, from left to right.
+     *
+     * Specifically the grammar treats INTERSECT as having the same precedence as UNION and EXCEPT, so we need to
+     * handle that specially. (Brackets and left-right precedence is handled by the grammar.)
+     */
+    // Start by finding the contexts up to the first INTERSECT op (if any), and gather them into a plan covering the LHS
+    // of the INTERSECT op.
+    val (beforeFirstIntersect, fromFirstIntersect) = remainingContexts.span(_.INTERSECT() == null)
```
vil1, line 95: TIL about `span`, thanks!

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -64,17 +65,74 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
...
-              ir.SetOperation(lhs, rhs, ir.UnionSetOp, is_all = isAll, by_name = false, allow_missing_columns = false)
+          qe.queryExpression().asScala.map(_.accept(this)).reduceLeft { (lhs, rhs) =>
+            val isAll = qe.ALL() != null
+            ir.SetOperation(lhs, rhs, ir.UnionSetOp, is_all = isAll, by_name = false, allow_missing_columns = false)
           }
       }
   }
 
+  @tailrec
+  private[this] def buildIntersectOperations(
```
vil1, line 80: I wonder if we could simplify that a bit by first going through the the seq of context and split it at each instance of `INTERSECT`, producing a `Seq[Seq[SqlUnionContext]]` that we'd next traverse again to produce the actual `SetOperation(..., IntersectSetOp)`.

Something along the lines of:

```scala
def appendNext(agg: Seq[Seq[SqlUnionContext]], item: SqlUnionContext) = Seq(item) +: agg
def appendFirst (agg: Seq[Seq[SqlUnionContext]], item: SqlUnionContext) = 
  if (agg.isEmpty) Seq(Seq(item))
  else (item +: agg.head) +:  agg.tail

remainingContexts.foldLeft(Seq.empty[Seq[SqlUnionContext]]) {
  case (agg, item) if item.INTERSECT() == null => appendFirst(item)
  case (agg, item) => appendNext(item)
}
.map( // here you get seqs of SqlUnionContext that aren't INTERSECT, in reverse order, so you may foldRight/reduceRight )
.reduceRight(//here you build your SetOperation(..., IntersectSetOp, ...))
```

vil1: Having slept over it, maybe the existing @tailrec approach would still be more ... approachable. Your call.

vil1 (approved):

PR #1232, opened 2024-11-23
Enclose subqueries in parenthesis

ericvergnaud:
Generator does not currently enclose subqueries in parenthesis, thus generating incorrect code such as the following:

`SELECT * FROM SELECT * FROM t WHERE a > 'a'  WHERE a > 'b' `

This PR fixes the issue such that the generated code is now:

`SELECT * FROM (SELECT * FROM t WHERE a > 'a')  WHERE a > 'b' `

It takes special care of the `.. IN(SELECT...)` pattern, by avoiding doubling enlosing parenthesis

Test cases are included #1233 but not added here because they require changes unrelated to this PR

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
@@ -1,7 +1,7 @@
 package com.databricks.labs.remorph.generators.sql
 
 import com.databricks.labs.remorph.generators._
-import com.databricks.labs.remorph.{Generating, OkResult, Transformation, TransformationConstructors, intermediate => ir}
+import com.databricks.labs.remorph.{Generating, OkResult, TransformationConstructors, intermediate => ir}
 
 import java.time._
 import java.time.format.DateTimeFormatter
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -371,11 +371,29 @@ class ExpressionGenerator extends BaseSQLGenerator[ir.Expression] with Transform
...
+            if (s.charAt(0) == '(' && s.charAt(s.length - 1) == ')') {
+              OkResult((g, s))
+            } else {
+              OkResult((g, "(" + s + ")"))
+            }
+          case r => r
+        }
+      }
+      case p => values.run(p)
+    })
```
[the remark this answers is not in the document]
ericvergnaud: Done

jimidle (approved):
lgtm

core/src/main/scala/com/databricks/labs/remorph/generators/sql/ExpressionGenerator.scala
```diff
@@ -371,11 +371,21 @@ class ExpressionGenerator extends BaseSQLGenerator[ir.Expression] with Transform
 
   private def in(inExpr: ir.In): SQL = {
     val values = commas(inExpr.other)
-    code"${expression(inExpr.left)} IN ($values)"
+    val enclosed = values.map { sql =>
+      if (sql.charAt(0) == '(' && sql.charAt(sql.length - 1) == ')') {
```
vil1, line 375: nitpicking, as it is unlikely to happen, but `str.charAt(0) == '('` would fail on the empty string, as opposed to `str.startsWith("(")`

vil1: can be done in another PR fixing #1252

vil1 (approved):

PR #1235, opened 2024-11-25
Baseline IRs for translating ADF pipelines to workflows

ghanse:
### Summary
This PR adds intermediate representations of ADF pipelines, activities, and linked services from the APIs. This is a required baseline for translating ADF pipelines to Databricks workflows.

### Details

- Added the `orchestrators.adf` package
- Created `PipelineNode` for any nodes coming from ADF
- Created classes to represent the ADF API objects (e.g. Pipelines, Activities, and Linked Services)

### Notes

- This is a larger PR than I would like, but minimally implements the IRs needed to translate 1 pipeline
- Can you validate my pattern for implementing the `children` method? (e.g. in `DatabricksNotebookActivity`)


core/src/main/scala/com/databricks/labs/remorph/intermediate/adf/activities/Activity.scala
```diff
@@ -0,0 +1,29 @@
...
+    dependsOn: Seq[ActivityDependency],
+    description: Option[String],
+    linkedServiceName: Option[LinkedServiceReference],
+    name: Option[String],
+    onInactiveMarkAs: Option[OnInactiveMarkAs],
+    policy: Option[ActivityPolicy],
+    state: Option[ActivityState],
+    activityType: Option[String],
+    activityProperties: Option[ActivityProperties],
+    userProperties: Seq[UserProperty])
```
vil1, line 16: Case classes like this one where all fields are `Seq` or `Option` almost always indicate a design problem. As a rule of thumb, one should ask themselves "what would be the meaning of `Activity(Seq(), None, None, None, None, None, None, None, None, Seq())`?".

`Option` in a case class field should only be used when there is a meaningful situation where the field could be absent.

For example, is it possible for an `Activity` to have no `state`? Note that isn't exactly the same question as "can an `Activity` be neither `Active` nor `Inactive`?". If the answer to the latter question is "yes", then we should add a `case object Unknown extends Activity` for example and define the field as `state: ActivityState = Unknown`.

core/src/main/scala/com/databricks/labs/remorph/intermediate/adf/activities/LibraryDefinition.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import com.databricks.labs.remorph.intermediate.adf.PipelineNode
+
+case class LibraryDefinition(
+    jarFilePath: Option[String],
+    eggFilePath: Option[String],
+    whlFilePath: Option[String],
+    mavenSpecification: Option[MavenSpecification],
+    pyPiSpecification: Option[PyPiSpecification],
+    cranSpecification: Option[CranSpecification])
```
vil1, line 11: Following the comment above, here it is quite clear that we have three mutually exclusive cases mashed together. 
We should rather define it as
```scala
sealed trait LibraryDefinition
case class MavenLibrary(jarFilePath: String, mavenSpecification: MavenSpecification) extends LibraryDefinition
case class PyPiLibrary(eggFilePath: String, piPySpecification: PyPiSpecification) extends LibraryDefinition
case class CranLibrary(whlFilePath: String, cranSpecification: CranSpecification) extends LibraryDefinition
```

core/src/main/scala/com/databricks/labs/remorph/intermediate/adf/activities/ActivityProperties.scala
```diff
@@ -0,0 +1,7 @@
+package com.databricks.labs.remorph.intermediate.adf.activities
+
+import com.databricks.labs.remorph.intermediate.adf.PipelineNode
+
+abstract class ActivityProperties(name: Option[String]) extends PipelineNode {
+  override def children: Seq[PipelineNode] = Seq()
+}
```
vil1, line 7:
```suggestion
trait ActivityProperties extends PipelineNode {
  def name: String
}
```

core/src/main/scala/com/databricks/labs/remorph/intermediate/adf/activities/DatabricksNotebookActivity.scala
```diff
@@ -0,0 +1,12 @@
...
+import com.databricks.labs.remorph.intermediate.adf.PipelineNode
+
+case class DatabricksNotebookActivity(
+    name: Option[String],
+    baseParameters: Map[String, String],
+    libraries: Seq[LibraryDefinition],
+    notebookPath: Option[String])
+    extends ActivityProperties(name) {
+  override def children: Seq[PipelineNode] = super.children ++ libraries
+}
```
vil1, line 12:
```suggestion
case class DatabricksNotebookActivity(
    override val name: String,
    baseParameters: Map[String, String],
    libraries: Seq[LibraryDefinition],
    notebookPath: Option[String])
    extends ActivityProperties {
  override def children: Seq[PipelineNode] = libraries
}
```

core/src/main/scala/com/databricks/labs/remorph/intermediate/adf/activities/Activity.scala
```diff
@@ -0,0 +1,29 @@
...
+    extends PipelineNode {
+  override def children: Seq[PipelineNode] = Seq() ++ dependsOn ++ linkedServiceName ++
+    policy ++ userProperties
+}
```
vil1, line 20: If we are sure that `linkedServiceName`, `policy` and `userProperties` are optional

```suggestion
  override def children: Seq[PipelineNode] = dependsOn ++ linkedServiceName.toSeq ++
    policy.toSeq ++ userProperties.toSeq
}
```

otherwise, we should go for 
```scala
override def children: Seq[PipelineNode] = dependsOn ++ Seq(linkedServiceName, policy, userProperties)
```

vil1 (changes requested):
This needs to be refactored as to avoid `Option` fields as much as possible.

core/src/main/scala/com/databricks/labs/remorph/intermediate/adf/activities/Activity.scala
ghanse: I've made the `name`, `activityType`, `activityProperties`, and `state` fields required. Every activity should have at minimum a name, type, set of type-specific properties, and state.

Many of the other fields are configurations which will be optionally returned by the API client.

core/src/main/scala/com/databricks/labs/remorph/intermediate/adf/activities/Activity.scala
ghanse: This is done. I made fields required and used the last pattern.

PR #1238, opened 2024-11-26
Add support for SnowFlake `SUBSTR`

ericvergnaud:
SnowFlake supports both SUBSTR and SUBSTRING, see https://docs.snowflake.com/fr/sql-reference/functions/substr.
This PR fills the gap for missing SUBSTR

Supersedes https://github.com/databrickslabs/remorph/pull/1226 which was lacking gpg signature


core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
```diff
@@ -191,243 +191,248 @@ class FunctionBuilderSpec extends AnyFlatSpec with Matchers with TableDrivenProp
...
 
     val functionBuilder = new TSqlFunctionBuilder
 
-    forAll(functions) { (functionName: String, expectedArity: Option[FunctionDefinition]) =>
-      functionBuilder.functionDefinition(functionName) shouldEqual expectedArity
+    forAll(functions) { (functionName: String, expectedArity: FunctionDefinition) =>
+      {
+        val actual = functionBuilder.functionDefinition(functionName)
+        actual.nonEmpty shouldEqual true
+        actual.get.arity shouldEqual expectedArity.arity
```
vil1, line 434:
```suggestion
        actual shouldEqual Some(expectedArity)
```

vil1 (approved):

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
ericvergnaud: Your proposal is not comparing `arities` which is what the test is about. Comparing functions fails if the implemented function has other fields populated (such as a conversion strategy).

asnare, line 434: Minor nit, I think this will be more understandable if it fails.
```suggestion
        actual.map(_.arity) shouldEqual Some(expectedArity.arity)
```

Should we also be checking the function-type?
```suggestion
        actual.map(_.functionType) shouldEqual Some(StandardFunction)
```

asnare (commented):
Looks fine, with one question.

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
```diff
@@ -191,243 +191,248 @@ class FunctionBuilderSpec extends AnyFlatSpec with Matchers with TableDrivenProp
...
-      ("UPPER", Some(FunctionDefinition.standard(1))),
-      ("USER", Some(FunctionDefinition.standard(0))),
-      ("USER_ID", Some(FunctionDefinition.standard(0, 1))),
-      ("USER_NAME", Some(FunctionDefinition.standard(0, 1))),
-      ("VALUE", Some(FunctionDefinition.xml(2))),
-      ("VAR", Some(FunctionDefinition.standard(1))),
-      ("VARP", Some(FunctionDefinition.standard(1))),
-      ("XACT_STATE", Some(FunctionDefinition.standard(0))),
-      ("YEAR", Some(FunctionDefinition.standard(1))))
+      ("ABS", FunctionDefinition.standard(1)),
```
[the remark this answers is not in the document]
jimidle: Also, this area has also been changed in another PR, which also caused me to balk in the review of that one. PRs should have one single purpose and changes like this should not be split across multiple unrelated PRs.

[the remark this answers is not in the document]
ericvergnaud: @jimidle sorry, but isn't it required for a PR to pass existing tests ? That's why it's here

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
ericvergnaud: No we shouldn't! The test is called "FunctionBuilder" should "return correct arity for each function"
It was incorrectly comparing functions. Now it does what it says!

asnare: @ericvergnaud: thanks for the clarification; I understand the name of the test suggests this is the intent and that PR change is consistent with this.

However previously we _were_ checking the `functionType` and it's not clear to me that we're now checking it anywhere. I do see some adjacent tests for `functionType` on the TSQL side of things, but not on the SnowFlake side here. Have we lost something here unintentionally? Are we now missing a test?

ericvergnaud: IMHO this test is completely useless. It smells like testing java getters.

ericvergnaud: What we want to test is that a function behaves as expected, not that its contract complies with yet another spec.
But removing the test should go in another PR, if we all agree.

ericvergnaud: And we should add tests for each function. Many other PRs.

jimidle: > @ericvergnaud: thanks for the clarification; I understand the name of the test suggests this is the intent and that PR change is consistent with this.
> 
> However previously we _were_ checking the `functionType` and it's not clear to me that we're now checking it anywhere. I do see some adjacent tests for `functionType` on the TSQL side of things, but not on the SnowFlake side here. Have we lost something here unintentionally? Are we now missing a test?

I may have added the TSQL stuff and not added it for Snowflake because I did all the amalgamation of function calls into one or two grammar rules there first, then we did it in Snowflake after the fact. I think you are correct here.

ericvergnaud: So we're calling code without checking its behavior only to improve code coverage ratio ?

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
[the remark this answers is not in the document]
ericvergnaud: Sure, apologies for not clarifying the why again (I thought you had read [this above comment](https://github.com/databrickslabs/remorph/pull/1238#discussion_r1858873766))

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
asnare: Getting back to the changes, I understand the situation now. Thanks for the clarifications.

asnare (approved):

PR #1242, opened 2024-11-26
Add basic support for SnowFlake LCAs

ericvergnaud:
Implements DealiasLCAs, a transformation rule that replaces LCAs by the underlying alias expression whenever it's not supported by Databricks.

Replicates scala tests from tests/unit/snow/test_lca_utils.py 

Suffers from various limitations (already present in the Python implementation):
 - may call the same function multiple times which may be a problem if the function is not idempotent
 - may introduce precedence issues if an alias involves more than 2 expressions
 - does not raise warnings when the above is encountered

Fixes https://github.com/databrickslabs/remorph/issues/1222

Supersedes #1225 
Supersedes #1233


core/src/main/scala/com/databricks/labs/remorph/intermediate/expressions.scala
```diff
@@ -193,6 +211,14 @@ trait StarOrAlias
 
 case class Alias(expr: Expression, name: Id) extends Unary(expr) with StarOrAlias {
   override def dataType: DataType = expr.dataType
+
+  override def makeCopy(newArgs: Array[AnyRef]): Expression = {
```
ericvergnaud, line 215: `TreeNode.makeCopy` docs say:
```
Must be overridden by child classes that have
   * constructor arguments that are not present in the productIterator.

```
I don't know exactly why only 1 arg of the `Alias` ctor is seen by the `productIterator` (maybe because it's not a `val` ?), but it's what happens.

This is tested via existing test cases in `SnowflakeToDatabricksTranspilerTest`, which crash without this bug fix.

core/src/main/scala/com/databricks/labs/remorph/intermediate/extensions.scala
```diff
@@ -136,6 +136,14 @@ case class Timezone(expression: Expression, timeZone: Expression) extends Binary
 
 case class WithinGroup(expression: Expression, order: Seq[SortOrder]) extends Unary(expression) {
   override def dataType: DataType = expression.dataType
+
+  override def makeCopy(newArgs: Array[AnyRef]): Expression = {
```
ericvergnaud, line 140: `TreeNode.makeCopy` docs say:
```
Must be overridden by child classes that have
   * constructor arguments that are not present in the productIterator.

```
I don't know exactly why only 1 arg of the `WithinArg` ctor is seen by the `productIterator` (maybe because they are not `val`s ?), but it's what happens.

This is tested via existing test cases in `SnowflakeToDatabricksTranspilerTest`, which crash without this bug fix.

jimidle:
What's the status of this one now? We lost @vil1 's comments from the unsigned PR, but I think he was suggesting that there was a way to do this without catering for each different instance of alias?

ericvergnaud:
> What's the status of this one now? We lost @vil1 's comments from the unsigned PR, but I think he was suggesting that there was a way to do this without catering for each different instance of alias?

He was, but after we discussed he agreed that a generic approach would probably not work here. He needs a second look.

core/src/main/scala/com/databricks/labs/remorph/intermediate/expressions.scala
```diff
@@ -161,7 +161,25 @@ case class Cast(
     type_str: String = "",
     returnNullOnError: Boolean = false,
     timeZoneId: Option[String] = None)
-    extends Unary(expr)
+    extends Unary(expr) {
+
+  override def makeCopy(newArgs: Array[AnyRef]): Expression = {
```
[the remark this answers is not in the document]
vil1: Existing tests crashing? How's that not caught by CI?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
vil1, line 1: Most of the code in there seems like a reimplementation of `transform` _et al._ in disguise, there must be a solution not involving that amount of boilerplate.

ericvergnaud: I suspect that rather than 'most of the code', it's more like 30-40 lines. I'l happily consider a solution relying on `transform` if you can provide one.

vil1: see https://github.com/databrickslabs/remorph/pull/1245

jimidle:
> > What's the status of this one now? We lost @vil1 's comments from the unsigned PR, but I think he was suggesting that there was a way to do this without catering for each different instance of alias?
> 
> He was, but after we discussed he agreed that a generic approach would probably not work here. He needs a second look.

OK so with #1245 it seems our path forward is:
 - merge 1245 into this branch
 - remove the makeCopy stuff here as it seems it wasn't needed after all
 - then this PR can merge into main

core/src/main/scala/com/databricks/labs/remorph/intermediate/expressions.scala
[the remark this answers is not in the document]
ericvergnaud: A very similar issue is reproduced in `core/src/test/scala/com/databricks/labs/remorph/intermediate/TreeTest.scala` in branch `fix-makecopy`.
Strangely the fixes in this branch don't fix it.
I suspect the issue might come from using `Array` instead of `Seq`, and unlike the Scala compiler, `makeCopy` doesn't implement the required conversion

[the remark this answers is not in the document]
ericvergnaud: @vil1 
Further investigation shows that the issue originated from:

```
private def dealiasUnary(unary: Unary): Expression = {
    val child = dealiasExpression(unary.child)
    unary.makeCopy(Array(child))
  } 

```
All 3 problematic classes (`Alias`, `Cast` and `WithinGroup`) inherit from Unary but actually have more than 1 field

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
```diff
@@ -191,243 +191,248 @@ class FunctionBuilderSpec extends AnyFlatSpec with Matchers with TableDrivenProp
...
+      ("UNICODE", FunctionDefinition.standard(1)),
+      ("UPPER", FunctionDefinition.standard(1)),
+      ("USER", FunctionDefinition.standard(0)),
+      ("USER_ID", FunctionDefinition.standard(0, 1)),
+      ("USER_NAME", FunctionDefinition.standard(0, 1)),
+      ("VALUE", FunctionDefinition.xml(2)),
+      ("VAR", FunctionDefinition.standard(1)),
+      ("VARP", FunctionDefinition.standard(1)),
+      ("XACT_STATE", FunctionDefinition.standard(0)),
+      ("YEAR", FunctionDefinition.standard(1)))
```
jimidle, line 427: Why are you changing this in this PR? And why does it need to change?

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
```diff
@@ -191,243 +191,248 @@ class FunctionBuilderSpec extends AnyFlatSpec with Matchers with TableDrivenProp
...
 
     val functionBuilder = new TSqlFunctionBuilder
 
-    forAll(functions) { (functionName: String, expectedArity: Option[FunctionDefinition]) =>
-      functionBuilder.functionDefinition(functionName) shouldEqual expectedArity
+    forAll(functions) { (functionName: String, expectedArity: FunctionDefinition) =>
+      {
```
jimidle, line 431: Why in this PR?

jimidle (changes requested):
Why are we changing the definitions of functions in a PR about alias transformations. Such changes cause confusion when trying to review.

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
ericvergnaud: This is because I discovered bugs (fixed in other PRs) and pre-merged the corresponding PRs in order for the tests to pass. Now that these PRs have been merged to main, these changes no longer appear.

core/src/test/scala/com/databricks/labs/remorph/parsers/common/FunctionBuilderSpec.scala
ericvergnaud: This is because I discovered bugs (fixed in other PRs) and pre-merged the corresponding PRs in order for the tests to pass. Now that these PRs have been merged to main, these changes no longer appear.

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
```diff
@@ -263,6 +263,30 @@ class SnowflakeToDatabricksTranspilerTest extends AnyWordSpec with TranspilerTes
...
+    }
+
+    "transpile LCA with aliased table" in {
+      "SELECT t.col1, t.col2, t.col3 AS ca FROM table1 t WHERE ca in ('v1', 'v2');" transpilesTo
+        "SELECT t.col1, t.col2, t.col3 AS ca FROM table1 as t WHERE t.col3 in ('v1', 'v2');"
+    }
+
+    "transpile LCA with partition" in {
+      "SELECT t.col1 AS ca, ROW_NUMBER() OVER (PARTITION by ca ORDER BY ca) FROM table1 t;" transpilesTo
+        "SELECT t.col1 AS ca, ROW_NUMBER() OVER (PARTITION by t.col1 ORDER BY t.col1 ASC NULLS LAST) FROM table1 AS t;"
```
asnare, line 283: Please add a comment for this explaining that although LCAs are supported, LCAS within a window expression is not supported (and is therefore subject to dealiasing).

Along this line of thought I'd also like to see a test showing that normal LCAs are left alone (and not dealiased).

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
asnare, line 1: I would really like to see an example of a column expression being aliased. As discussed, inlining the aliases isn't safe in the general case and we should either be:
 - Raising an error; or
 - Leaving them as-is (which will produce an error downstream).

We need a test to demonstrate that we are _not_ changing the semantics of the query for the situations that we know are unsafe. A candidate Snowflake query would be:
```sql
SELECT *, random() as r
FROM the_table
WHERE r % 2 = 0;
```
This should fail or be left as-is (to fail downstream).

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
```diff
@@ -0,0 +1,27 @@
...
+SELECT
+  ca_zip
+FROM
+SELECT
+  SUBSTR(ca_zip,1,5) AS ca_zip,
+  TRIM(name) AS name,
+  COUNT(*) OVER (
+    PARTITION BY
+      SUBSTR(ca_zip,1,5)
+  )
```
asnare, line 23: I'm not sure whether this should have been dealiased or not: I believe that in Snowflake an unqualified reference such as this is assumed to be a column reference in preference to a column alias. Please test this to verify, and add a comment explaining the behaviour. Although I understand the intent of the query, I suspect that it's supposed to be interpreted as:
```suggestion
  customer_address.ca_zip
```

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
```diff
@@ -0,0 +1,27 @@
...
+FROM
+  customer_address
+WHERE
+  SUBSTR(ca_zip,1,5) IN ('89436', '30868');
```
asnare, line 27: Oooh. With which projection should the `WHERE` be associated? I _think_ it's okay?

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
```diff
@@ -0,0 +1,27 @@
+-- snowflake sql:
+select
+  ca_zip
+from (
+  SELECT
+    substr(ca_zip,1,5) ca_zip,
```
asnare, line 6: This is a nice edge-case where the column alias has the same name as an existing column.

tests/resources/functional/snowflake/core_engine/lca/lca_nested.sql
```diff
@@ -0,0 +1,13 @@
+-- snowflake sql:
+SELECT
+    b * c as new_b,
+    a - new_b as ab_diff
```
asnare, line 4: For fun, how about we swap those operators:
```suggestion
    b - c as new_b,
    a * new_b as ab_diff
```

tests/resources/functional/snowflake/core_engine/lca/lca_cte.sql
asnare, line 1: Does this (and every other new) functional test fail under Python? At present these are only being tested against the core engine. If they happen to pass with the Python implementation then we can move them there.

asnare (changes requested):
There are some questions sprinkled throughout as comments, and some tests I'd like to see.

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilderSpec.scala
```diff
@@ -38,6 +38,21 @@ class SnowflakeAstBuilderSpec extends AnyWordSpec with SnowflakeParserTestCommon
           Seq(Id("a"), Alias(Id("b"), Id("bb")), Id("c"))))
     }
 
+    "translate a SELECT query involving a table alias" in {
+      singleQueryExample(
+        query = "SELECT t.a FROM table_x t",
+        expectedAst =
+          Project(TableAlias(NamedTable("table_x", Map.empty, is_streaming = false), "t"), Seq(Dot(Id("t"), Id("a")))))
```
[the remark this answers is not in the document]
jimidle: IntelliJ does tell us about this, and it also tells us to name boolean arguments. I think it's small beer in the scheme of things, but probably worth doing. Some of this has arisen by virtue of one or more of us changing the parameters to have defaults when they did not before etc.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
```diff
@@ -0,0 +1,53 @@
...
+import com.databricks.labs.remorph.intermediate.{Expression, _}
+
+class DealiasLCAs extends Rule[LogicalPlan] with IRHelpers {
+
+  override def apply(plan: LogicalPlan): LogicalPlan = transformPlan(plan)
+
+  private def transformPlan(plan: LogicalPlan): LogicalPlan =
+    plan transform { case project: Project =>
+      dealiasProject(project)
+    }
```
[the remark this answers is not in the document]
ericvergnaud: @vil1 what do you think ?

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilderSpec.scala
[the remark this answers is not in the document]
ericvergnaud: isn't that covered by the functional tests ?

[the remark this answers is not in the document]
asnare: Not really?

The unit tests here are about the immediate IR and help verify where the dealiasing takes place.

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
ericvergnaud: > I'm not sure whether this should have been dealiased or not: I believe that in Snowflake an unqualified reference such as this is assumed to be a column reference in preference to a column alias. Please test this to verify, and add a comment explaining the behaviour. Although I understand the intent of the query, I suspect that it's supposed to be interpreted as:

The tests in [test_lca_utils.py](https://github.com/databrickslabs/remorph/blob/main/tests/unit/snow/test_lca_utils.py) do dealias window expressions. @sundarshankar89 can you opine ?

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
ericvergnaud: Indeed. In the absence of parenthesis, the WHERE clause is associated with the closest SELECT

tests/resources/functional/snowflake/core_engine/lca/lca_nested.sql
ericvergnaud: the behavior might indeed surprise... we'd probably want to raise a warning if dealiasing involves more than 2 expressions

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
asnare: To be clear the question is not about dealising in general within a window expression, it's to do with the ambiguity for this particular example: does the unqualified identifier (`ca_zip`) refer to the column alias or the original column?

Google suggests that in this case for Snowflake there's no way to refer to the column alias because there's a column with the same name.

I'm not sure of the full semantics for the tests in `test_lca_utils.py` but it looks like it doesn't verify output, merely whether an error occurs or not. In this case leaving the `ca_zip` reference as is would be fine and pass that test (because it's valid: it's a reference to the column).

ericvergnaud: See this test in [test_lca_utils.py](https://github.com/databrickslabs/remorph/blob/main/tests/unit/snow/test_lca_utils.py)

```
def test_fix_lca_with_lca_in_window(normalize_string):
    input_sql = """
        SELECT
            t.col1,
            t.col2,
            t.col3 AS ca,
            ROW_NUMBER() OVER (PARTITION by ca ORDER BY t.col2 DESC) AS rn
        FROM table1 t
    """
    expected_sql = """
        SELECT
            t.col1,
            t.col2,
            t.col3 AS ca,
            ROW_NUMBER() OVER (PARTITION by t.col3 ORDER BY t.col2 DESC) AS rn
        FROM table1 AS t
    """
```

ericvergnaud: Revisiting the spec is not in scope of this PR, but if you have doubts then we should definitely open a ticket.

tests/resources/functional/snowflake/core_engine/lca/lca_nested.sql
ericvergnaud: See last comment #1276

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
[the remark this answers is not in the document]
vil1: The style-guide advocates for it indeed, but we have deviated from this recommendation in other parts of the codebase already, so I don't think it's enough to block this PR any longer

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
asnare: To be clear, it's about `test_query_with_same_alias_and_column_name()` from that test suite, which I don't believe specifies the expected output for this example: merely that it won't produce an error.

The output we have here seems to be new (and might be the author's _intent_ when the query was written) but I don't believe it's how Snowflake will interpret the query.

I've tried to explain above why transpilation shouldn't produce an error, and also why dealiasing _this specific example_ is incorrect for Snowflake. Maybe we should jump on a call to go over this.

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilderSpec.scala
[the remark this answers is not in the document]
vil1: I agree with @asnare, you can borrow `DealiasLCAsSpec` from my other PR if you want.

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
vil1: LCAs are not supported inside window clauses, so the expected translation here seems correct to me.

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
vil1: There is a comment about that in the implementation of `DealiasLCAs`

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
asnare: I'll suggest we have a quick call in the morning, since I don't think we're understanding each other. My point is that for this specific example there's no LCA (to dealias).

core/src/test/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilderSpec.scala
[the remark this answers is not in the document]
ericvergnaud: Done, thanks @vil1

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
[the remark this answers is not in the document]
gueniai: For what it's worth, I would like to stick to using braces for methods (not a show stopper, but would like to see it changed if there are other changes going in)

vil1 (approved):
The amount of comments (56 before this one) and commits (60) as grown incredibly big for what ultimately boiled down to "just" a `+252-6` change. That makes this PR definitely epic! (and I see it as a sign of very healthy team work)

Our approach so far has been to incrementally improve our transpiler, while knowing that we are unlikely to get it right on our first attempt (in fact, we may never be able to get it 100% right).

In that respect, I think we should merge this PR and materialize our - legitimate - concerns about absolute correctness in separate issues.

tests/resources/functional/snowflake/core_engine/lca/lca_cte.sql
ericvergnaud: done

tests/resources/functional/snowflake/core_engine/lca/lca_homonym.sql
asnare: If we're going to leave the buggy behaviour in place, please change this functional test so that it represents an example that _is_ being transpiled correctly. Tests are supposed to indicate correct behaviour, and this example does not.

We already have a problem in some places where test fixtures are incorrect, and I am strongly against introducing another one.

asnare (commented):
I'm also for making progress, and understand why this PR is the way it is: it was intended to replicate some behaviour on the python-based transpiler, and we've since discovered that behaviour is incorrect under some circumstances.

When we discussed this as a team at the start of the week, a key outcome was that we agreed that when introducing a new behaviour we should not introduce "silent" errors that will go unnoticed. Put another way, when something in the query is unsupported we should either:

 * Raise an error; or
 * Leave it as-is (so that it can produce an error downstream).

I know we had a look at making it safe, and there were issues with that and scope creep etc. So we've ended up going back on our decision from the start of the week. I think there was another path: limit (for now) the LCAs that we expand to (for now) exclude functions.

I'm not going to block this PR, but do have one remaining change request, which I've also placed as a comment: please update/fix the example that we know we're mistranslating to be a correct example.

asnare (approved):

PR #1245, opened 2024-11-27
Simplify DealiasLCAs implementation using transform

vil1:


core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
```diff
@@ -268,11 +268,6 @@ class SnowflakeToDatabricksTranspilerTest extends AnyWordSpec with TranspilerTes
         "SELECT column_a as alias_a FROM table_a where column_a = '123';"
     }
 
-    "transpile LCA not replacing aliases" in {
```
vil1, line 271: The expected output query was wrong (failed on Databricks with `[UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter with name `alias_a` cannot be resolved.`)

jimidle: Does this mean that the transformation was incorrect and we were generating incorrect output, or that the expected output was incorrect and we had targeted producing this incorrect output. It would appear that the test output was not verified against Databricks in either case though. We need to make sure that everyone does this or we create tests that seem to pass but in fact generate bad output (which we have told the tests to expect).

jimidle: Explained as the test being incorrect and redundant.

jimidle (approved):
Looks good to me. 

So it looks like we:

  - merge this, which will update code in #1242 branch
  - pull the code from 1242
  - modify 1242 to not override makeCopy
  - merge 1242

sundarshankar89 (commented):
This simplifies tansformation process.

ericvergnaud:
> This simplifies tansformation process.

It's definitely more concise, wow ! But I have no clue how it works, so not sure who will be able to maintain this...

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
gueniai: For the record, the expected output is correct, see attached screenshot:
Screenshot 2024-11-27 at 12 17 42 PM
We should add the test back in, as it is definitely a valid twisted way someone can write things.

gueniai (changes requested):
I like the implementation, it does simplify things a lot. I agree with Eric that at first it is seemingly a lot of magic, but it probably is more a function on learning more about the codebase and what functions are already available. 
Left one small request to re-introduce the test since it's valid

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
ericvergnaud: Actually the test should read as follows:
  ```
  "transpile LCA not replacing aliases" in {
      "SELECT '123' as alias_a FROM table_a where alias_a = '123';" transpilesTo
        "SELECT '123' as alias_a FROM table_a where '123' = '123';"
    }
```
@gueniai your example is also valid but 'WITH TABLE...' creates a context which is not available in this test

**Review by ericvergnaud — CHANGES_REQUESTED** [review]

File: core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala (hunk 1 above)
Comment (vil1) [reply core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala:271]: @gueniai your example above works because of homonymy, the following won't work:

```sql
WITH 
  table_a(col1) AS (SELECT '123' as foo)
SELECT table_a.col1 as alias_a FROM table_a where alias_a = '123'
```

vil1: > Actually the test should read as follows:
> 
> ```
> "transpile LCA not replacing aliases" in {
>     "SELECT '123' as alias_a FROM table_a where alias_a = '123';" transpilesTo
>       "SELECT '123' as alias_a FROM table_a where '123' = '123';"
>   }
> ```
That is right, but it felt a bit redundant with the test above so I decided to delete it altogether

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
```diff
@@ -2,169 +2,43 @@ package com.databricks.labs.remorph.parsers.snowflake.rules
...
-      val sort_order = dealiasSortOrders(window.sort_order)
-      window.makeCopy(
-        Array(
-          window.window_function.asInstanceOf[AnyRef],
-          partition.asInstanceOf[AnyRef],
-          sort_order.asInstanceOf[AnyRef],
-          window.frame_spec.asInstanceOf[AnyRef],
-          window.ignore_nulls.asInstanceOf[AnyRef]))
+    val dealiasedInput = project.input transformDown { case Filter(in, cond) =>
+      Filter(in, dealiasExpression(cond, aliases))
```
ericvergnaud, line 23: This is losing information not present in the constructor

ericvergnaud, line 23: This is losing information not present in the constructor

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
```diff
@@ -2,169 +2,43 @@ package com.databricks.labs.remorph.parsers.snowflake.rules
...
-    if (aliases.isEmpty) {
-      project
-    } else {
-      val dealiaser = new Dealiaser(aliases)
-      dealiaser.dealiasProject(project)
+  def dealiasExpression(expr: Expression, aliases: Map[String, Expression]): Expression = {
+    expr transformUp {
+      case id: Id =>
+        aliases.getOrElse(id.id, id)
+      case e: Exists => Exists(apply(e.relation))
```
ericvergnaud, line 40: This is losing information not present in the constructor

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
```diff
@@ -2,169 +2,43 @@ package com.databricks.labs.remorph.parsers.snowflake.rules
...
+      case s: ScalarSubquery => ScalarSubquery(apply(s.plan))
```
ericvergnaud, line 41: This is losing information not present in the constructor

ericvergnaud (changes requested):

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
ericvergnaud: The one above doesn't use a literal as alias expression, so I'd say it's not redundant

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/DealiasLCAs.scala
```diff
@@ -2,169 +2,43 @@ package com.databricks.labs.remorph.parsers.snowflake.rules
...
-          val dealiaser = new Dealiaser(aliases)
-          dealiaser.dealiasProject(project)
-        }
-      }
-      case plan: LogicalPlan => plan // TODO log or raise error ?
-    }
-    subquery.makeCopy(Array(plan))
-  }
-
-  private def dealiasSortOrders(sort_order: Seq[SortOrder]): Seq[SortOrder] = {
```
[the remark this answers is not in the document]
ericvergnaud: Oh I missed Unary, my bad

ericvergnaud:
> > Member
> 
> Well, it takes a little time to learn a new codebase and the transforms are no exception mate.

> > Member
> 
> Well, it takes a little time to learn a new codebase and the transforms are no exception mate.

And some comments, thanks @vil1 for adding them

PR #1246, opened 2024-11-27
EditorConfig setup for project

asnare:
This PR sets up EditorConfig for the project. IntelliJ/PyCharm will pick this up automatically, as will many other editors.

In addition to the `.editorconfig` file, this PR updates existing files so that they:

 - have no trailing whitespace;
 - have an EOL at the end of the file.


sundarshankar89 (approved):
LGTM.

jimidle (commented):
While I agree with the intent, this is going to cause a change to 143 files just for whitespace. And I have already configured IntelliJ to do most of this.  It might be better to adopt the settings but let this happen a little more incrementally. 

For instance the ANTLR formatter currently does not add the trailing \n (I am writing a Java based version of that formatter that will do that, but that's in my spare weekend time, so who know when I will finish that). So when the formatting runs in CI, it will flag the files as being different - we have removed the ANTLR formatting from the CI though right now because it depends on Node, so may this is not a problem.

I am a bit wary of doing this.

gueniai (approved):
I like enforcing the standard in the repo for everybody and not depending on everyone's IDE settings. But to give us time to deal with any possible issues, let's merge this one on Monday.

asnare:
> While I agree with the intent, this is going to cause a change to 143 files just for whitespace. And I have already configured IntelliJ to do most of this. It might be better to adopt the settings but let this happen a little more incrementally.

I agree; it feels awkward and… churlish? to be changing so many files for something as pointless as whitespace, despite everyone (I think!) understanding and agreeing it's how it should be.

> For instance the ANTLR formatter currently does not add the trailing \n (I am writing a Java based version of that formatter that will do that, but that's in my spare weekend time, so who know when I will finish that). So when the formatting runs in CI, it will flag the files as being different - we have removed the ANTLR formatting from the CI though right now because it depends on Node, so may this is not a problem.

That's definitely an issue, and thanks for raising it: I definitely wasn't aware of the ANTLR formatter. However, bear in mind:
1. This only affects editors: it's not being enforced by CI or any other tool. Scalafmt and black ignore it, I don't think we have anything else?
2. It also only affects source files; generated code should be out of scope.

My main concern really is inadvertently altering test fixtures where we deliberately want to verify whitespace-related handling. Following that vein of thought, this PR does not 'fix' any tabs in our repository: a few of our functional tests have them. This was a mechanical PR, so I didn't want to spend time verifying whether the functional tests were intending to use them deliberately.

> I am a bit wary of doing this.

Understood, and rightly so. Thanks for being so diplomatic.

asnare:
> I like enforcing the standard in the repo for everybody and not depending on everyone's IDE settings. But to give us time to deal with any possible issues, let's merge this one on Monday.

Sounds good, thanks.

sundarshankar89 (approved):
LGTM

PR #1248, opened 2024-11-27
Tech debt: replace empty case classes with case objects

asnare:
This PR trivially refactors the situations where we are using cases classes without parameters to be case objects (as is idiomatic in Scala).


jimidle (approved):
lgtm

vil1 (approved):

PR #1249, opened 2024-11-27
Refactor `x.tail.foldleft(x.head)` chains as `x.reduceLeft`

asnare:
This PR refactors the situations where we use this:

```scala
x.tail.foldLeft(x.head)
```

to use the equivalent operation that does this:
```scala
x.reduceLeft
```

Following a review comment discussing the safety of reduce in these situations[^1], some further changes have been made to the surrounding code such that the behaviour remains the same but the safety can be determined via trivial inspection without needing to refer to the grammar.

[^1]: Not a new concern; the previous code had the same safety issue due to the use of `.head`.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/generators/py/StatementGenerator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/generators/py/StatementGenerator.scala
@@ -67,15 +67,9 @@ class StatementGenerator(private val exprs: ExpressionGenerator) extends BasePyt
   private def e(expr: Expression): Python = exprs.generate(expr)
 
   private def lines(statements: Seq[Statement], finish: String = "\n"): Python = {
-    if (statements.isEmpty) {
-      code""
-    } else {
-      val body = statements.map(generate(_))
-      val separatedItems = body.reduceLeft[Python] { case (agg, item) =>
-        code"$agg\n$item"
-      }
-      code"$separatedItems$finish"
-    }
+    val body = statements.map(generate)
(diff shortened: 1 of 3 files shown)


jimidle (approved):
LGTM - reasonable tidy up

vil1 (commented):
A step in the right direction !

The previous state of the code was already guilty of that, but we should guard these call to `reduceLeft` from being performed on empty sequences (as we should have done about calling `head` or `tail`).

In most the cases below, we know the list won't ever be empty due to the structure of the grammar but adding comment about that wouldn't hurt, especially for the education of our future less-experienced teammates.

asnare:
> A step in the right direction !
> 
> The previous state of the code was already guilty of that, but we should guard these call to `reduceLeft` from being performed on empty sequences (as we should have done about calling `head` or `tail`).
> 
> In most the cases below, we know the list won't ever be empty due to the structure of the grammar but adding comment about that wouldn't hurt, especially for the education of our future less-experienced teammates.

Agreed, it was a deliberate choice here not to change the _behaviour_ of the code but merely do a small tidy up that was clearly a (trivial) improvement, even if not the full extent to which improvements could be made. The existing code was safe, although this couldn't always be determined without referring to the grammar.

I've updated the code some more, adding a comment to explain why one of the situations is safe (due to the grammar). For the rest I've refactored things to be equivalent but where the safety apparent from inspecting the code without needing to refer to the grammar.

ericvergnaud (approved):
lgtm

vil1 (approved):
👍

PR #1258, opened 2024-11-29
Allow easier debugging of test failures

asnare:
This PR adjusts the way transpilation results are checked during tests so that the formatted values can be easily checked in a debugger. (IntelliJ/PyCharm have better ways of comparing output than the side-by-side view we display.)


ericvergnaud (approved):
lgtm

jimidle (approved):
LGTM - intermediate values are better for debugging.

Issue #1255, opened 2024-11-28
[TECH DEBT]: Handle the precedence of the `INTERSECT` set operation in the grammar(s) rather than in code


asnare:
### Current Behavior

As part of #1227 it became clear that our grammars for set operations treat `INTERSECT`, `UNION` and `EXCEPT` as equivalent. However it turns out that `INTERSECT` has higher precedence than the others. Although this can be dealt with in code during conversion to the IR, it would be better if this was reflected by the grammar (which also simplifies the conversion to IR).

### Expected Behavior

The grammars should treat `INTERSECT` as having higher precedence than `UNION` and `EXCEPT`.

The rules of precedence for set operations are:
1. Parentheses
2. `INTERSECT`
3. `EXCEPT` and `UNION` ordered from left-to-right.

Given that `INTERSECT` is commutative, ordering is unimportant.

Needs to be fixed in:

 - [x] TSQL (#1259)
 - [x] Snowflake (#1288)

References:
 - https://learn.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-except-and-intersect-transact-sql?view=sql-server-ver16#:~:text=following%20precedence
 - https://docs.snowflake.com/en/sql-reference/operators-query#:~:text=precedence

Relevant grammar rules:
https://github.com/databrickslabs/remorph/blob/932080a1fca53ed201fe7129a4679e1ab856dd23/core/src/main/antlr4/com/databricks/labs/remorph/parsers/snowflake/SnowflakeParser.g4#L3154-L3158
https://github.com/databrickslabs/remorph/blob/932080a1fca53ed201fe7129a4679e1ab856dd23/core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4#L2796-L2800

Changing the grammar will mean refactoring this code:
https://github.com/databrickslabs/remorph/blob/932080a1fca53ed201fe7129a4679e1ab856dd23/core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala#L79-L134

PR #1259, opened 2024-11-29
Update TSQL grammar so that INTERSECT precedence is handled there

asnare:
This PR updates the TSQL grammar and accompanying IR processing so that the precedence of `INTERSECT` is handled by the grammar instead of during transportation to our IR.

Relates #1255.


core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
```diff
@@ -2789,14 +2789,21 @@ predicate
     ;
 
 queryExpression
+    : unionExpression sqlIntersection*
```
ericvergnaud, line 2792: I suspect this and the below can be greatly simplified using something such as
```
queryExpression
 : querySpecification
 | queryExpression UNION ALL queryExpression
 | queryExpression INTERSECT queryExpression
etc...
```

ericvergnaud (changes requested):

core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
asnare: @ericvergnaud: This is interesting. I don't fully understand why the grammar is as it is. I already came up with this which also seems to be correct and is closer to what I would think of as the 'classical' way to express these rules:
```antlr
queryExpression
    : LPAREN queryExpression RPAREN
    | queryExpression (UNION ALL? | EXCEPT) queryExpression
    | queryExpression INTERSECT queryExpression
    | querySpecification
    ;
```

Are there performance/implementation reasons for preferring the existing folding (??) style over this?

jimidle: This is probably correct, precedence is generally best done as one rule with precedence then being explicit, then label the alts and the visitor is more simple as well

asnare: Thanks @jimidle, @vil1 indeed improved it as follows:
```antlr
queryExpression
    : LPAREN queryExpression RPAREN                                         #inParen
    | left = queryExpression (UNION ALL? | EXCEPT) right = queryExpression  #union
    | left = queryExpression INTERSECT right = queryExpression              #intersect
    | querySpecification                                                    #simple
    ;
```

jimidle: Generally, don' tuse the labels left/right as they generate more code. ctx.queryExpression(0), ctx.queryExpression(1) is more efficient. See the read me about working with antlr grammar, labels are useful for disambiguation, but even then it might mean splitting rules or labeling the alts.

jimidle: > @ericvergnaud: This is interesting. I don't fully understand why the grammar is as it is. I already came up with this which also seems to be correct and is closer to what I would think of as the 'classical' way to express these rules:
> 
> ```antlr
> queryExpression
>     : LPAREN queryExpression RPAREN
>     | queryExpression (UNION ALL? | EXCEPT) queryExpression
>     | queryExpression INTERSECT queryExpression
>     | querySpecification
>     ;
> ```
> 
> Are there performance/implementation reasons for preferring the existing folding (??) style over this?

BTW the grammar is like this because it was just typed in by reading the MS docs, which don't encapsulate such things. You should have seen it 6 months ago ;). These are the occasions where we get to fix it up.

asnare: Refactored to use this.

jimidle (approved):
LGTM

jimidle (approved):
LGTM

vil1 (approved):
LGTM

core/src/main/antlr4/com/databricks/labs/remorph/parsers/tsql/TSqlParser.g4
asnare: @ericvergnaud: could you check this? Merging is blocked currently because you requested this be changed, and it has been.

ericvergnaud (approved):
LGTM

PR #1263, opened 2024-12-02
Style: avoid unscoped `private` fields

asnare:
This PR replaces all private fields with either:

 - `private[this]`; 
 - or the intended scope.

This is recommended by two guidelines in the style guide:

 - https://github.com/databricks/scala-style-guide?tab=readme-ov-file#private-fields
 - https://github.com/databricks/scala-style-guide?tab=readme-ov-file#privatethis

(Aside from these justifications, my preferred reason is that by inspection I don't need to go looking for cross-instance access, which is normally a code smell.)

Some incidental changes:

 - A few field initialisers for mutable types were updated to assign the fully-initialised objects instead of partially-initialised instances that are then mutated to their intended state.


vil1 (approved):
LGTM

ericvergnaud (approved):
LGTM

PR #1265, opened 2024-12-03
Chore: remove some redundant (default) arguments on case classes in a test

asnare:
This PR tidies up a test by removing case class arguments that specify the default value. Although tests may wish to be explicit about this, none of these tests seem to be related to the argument so this cuts down on noise, and eliminates a bunch of IDE warnings.

An incidental change is the chopping down of a long line.


jimidle (approved):
lgtm

vil1 (approved):
👍

PR #1268, opened 2024-12-03
Added support for snowflake last_day, array_flatten functions

ganeshdogiparthi-db:


ganeshdogiparthi-db:
Attaching the screenshots for `ARRAY_FLATTEN` and `LAST_DAY` functions
Screenshot 2024-12-03 at 1 44 37 PM
Screenshot 2024-12-03 at 1 43 33 PM
Screenshot 2024-12-03 at 10 57 47 AM
Screenshot 2024-12-03 at 10 57 25 AM

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -62,6 +64,10 @@ class SnowflakeCallMapper extends ir.CallMapper with ir.IRHelpers {
       case ir.CallFunction("POSITION", args) => ir.CallFunction("LOCATE", args)
       case ir.CallFunction("REGEXP_LIKE", args) => ir.RLike(args.head, args(1))
       case ir.CallFunction("REGEXP_SUBSTR", args) => regexpExtract(args)
+      case ir.CallFunction("SEQ1", args) => ir.MonotonicallyIncreasingID()
```
sundarshankar89, line 67: I'm not convinced with this. given the output of each `seq` bytes are fixed.

ganeshdogiparthi-db: I agree, But Databricks doesn't have any logic other than MonotonicallyIncreasingID. I would be happy to modify the logic if we have any proper solutions in Databricks.

sundarshankar89: Let me think, 
fyi.. @vil1  another one to ponder on.

sundarshankar89 (approved):
LGTM
nit: Can you add a TODO with document link

ganeshdogiparthi-db:
Done @sundarshankar89 
Please have a look.

sundarshankar89 (approved):
LGTM

ericvergnaud:
Not blocking but is there a reason for not having 3 PRs, 1 for each added function ?

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -50,6 +51,7 @@ class SnowflakeCallMapper extends ir.CallMapper with ir.IRHelpers {
       case ir.CallFunction("IFNULL", args) => ir.Coalesce(args)
       case ir.CallFunction("IS_INTEGER", args) => isInteger(args)
       case ir.CallFunction("JSON_EXTRACT_PATH_TEXT", args) => getJsonObject(args)
+      case ir.CallFunction("LAST_DAY", args) => parse_last_day(args)
```
vil1, line 54: Please use camelCase in Scala
```suggestion
      case ir.CallFunction("LAST_DAY", args) => parseLastDay(args)
```

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -648,4 +658,24 @@ class SnowflakeCallMapper extends ir.CallMapper with ir.IRHelpers {
     irSortArray
   }
 
+  private def parse_last_day(args: Seq[ir.Expression]): ir.Expression = {
```
vil1, line 661:
```suggestion
  private def parseLastDay(args: Seq[ir.Expression]): ir.Expression = {
```

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -648,4 +658,24 @@ class SnowflakeCallMapper extends ir.CallMapper with ir.IRHelpers {
...
+    if (args.length == 1) {
+      return ir.LastDay(args.head)
+    }
+
+    val datePart = args(1) match {
+      case ir.StringLiteral(part) => part.toUpperCase()
+      case ir.Column(_, part) => part.toString.toUpperCase()
```
vil1, line 668: This may throw an (uncaught) `MatchError` if `arg(1)` is neither `StringLiteral` nor `Column`, please either handle the case or raise a `RemorphError`

tests/resources/functional/snowflake/arrays/array_flatten_1.sql
```diff
@@ -0,0 +1,11 @@
+-- snowflake sql:
+SELECT
+  ARRAY_FLATTEN([[1, 2, 3], [4], [5, 6]]) as col
+, ARRAY_FLATTEN([[[1, 2], [3]], [[4], [5]]]) as col1
+, ARRAY_FLATTEN([[1, 2, 3], NULL, [5, 6]]) as col3;
+
+-- databricks sql:
+select
+  flatten(array(array(1, 2, 3) , array(4) , array(5, 6))) as col,
```
vil1, line 9: please use uppercase

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
```diff
@@ -63,6 +65,14 @@ class SnowflakeCallMapper extends ir.CallMapper with ir.IRHelpers {
       case ir.CallFunction("POSITION", args) => ir.CallFunction("LOCATE", args)
       case ir.CallFunction("REGEXP_LIKE", args) => ir.RLike(args.head, args(1))
       case ir.CallFunction("REGEXP_SUBSTR", args) => regexpExtract(args)
+      case ir.CallFunction("SEQ1", args) =>
+        // TODO: Warn users about The seq[1/2/4/8] function in Snowflake and monotonically_increasing_id in Databricks
+        //  are equivalent in logic,
+        //  but the output size generated by seq[1/2/4/8] can be different. Please be cautious while using.
```
vil1, line 71: We could/should truncate and cast the `MonotonicallyIncreasingId` to the proper width (1, 2, 4 or 8 bytes).

vil1 (changes requested):
See comments above

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
vil1: `SEQ1(1)` -> `SELECT CAST (monotonically_increasing_id() % 128 AS TINYINT);` perhaps

bishwajit-db: We also need to handle this: If the optional sign argument is 0, the sequence continues at 0 after wrap-around. If the optional sign argument is 1, the sequence continues at the smallest representable number based on the given integer width.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
bishwajit-db: If the optional sign argument is 0, the sequence continues at 0 after wrap-around. If the optional sign argument is 1, the sequence continues at the smallest representable number based on the given integer width.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
vil1: Yeah, that will also need to be accounted for (didn't bother to do so in my above comment). It may also be slightly more complicated in terms of types as [this](https://docs.snowflake.com/en/sql-reference/data-types-numeric#int-integer-bigint-smallint-tinyint-byteint) suggests that numeric types like `TINYINT` are more of a "hint" in Snowflake.
But in any case, I'm confident we can come up with a faithful translation of `SEQ1` *et al.* using the proper mix of `MONOTONICALLY_INCREASING_ID` and basic arithmetics. 

What is almost sure though, is that this will likely fail any equality or hash based reconciliation process, as the only guarantee both `SEQ` and `MONOTONICALLY_INCREASING_ID` provide is that the number they produce increases monotonically (after having output `42` either one could correctly output any number greater or equal to `43`).

ganeshdogiparthi-db: I have removed the SEQ from this PR. I will create a new PR for the same

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
ganeshdogiparthi-db: Done

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/rules/SnowflakeCallMapper.scala
ganeshdogiparthi-db: I have removed the SEQ from this PR. I will create a new PR for the same

bishwajit-db (approved):
LGTM

vil1 (approved):
LGTM

Issue #1267, opened 2024-12-03
[BUG]: Precedence for CTEs with Snowflake is incorrect with respect to set operations


asnare:
### Current Behavior

At the moment CTEs with Snowflake are being scoped/bound to the first `SELECT` in a sequence of set operations. That is, this:
```sql
WITH cte AS (SELECT 1)
SELECT * FROM cte
UNION ALL
SELECT * FROM cte
```
is being transpiled as:
```sql
(WITH cte AS (SELECT 1)
 SELECT * FROM cte)
UNION ALL
SELECT * FROM cte
```

### Expected Behavior

This is incorrect, instead it should be equivalent to:
```sql
WITH cte AS (SELECT 1)
(SELECT * FROM cte
 UNION ALL
 SELECT * FROM cte)
```

PR #1273, opened 2024-12-03
Fix precedence of CTEs with respect to set operations in SnowFlake

asnare:
This PR fixes the handling of CTEs with respect to set operations; prior to this the CTE was associated with only the first `SELECT` in the sequence of set operations instead of the entire set.

Incidental changes:

 - Additional functional tests (that already passed) for a few of the CTEs.
 - Equivalent functional tests for TSQL to verify the behaviour was already correct.

Resolves: #1267


asnare:
Looks like some of the functional tests have cosmetic differences with the Python versions, specifically extra parentheses when transpiling set operations.

I'll move the affected tests into the core-specific test set.

vil1 (commented):
LGTM but some functional tests that are failing on python-sqlglot should be moved to `core_engine`

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
```diff
@@ -351,25 +351,41 @@ class SnowflakeToDatabricksTranspilerTest extends AnyWordSpec with TranspilerTes
...
-      |    b AS (2),
-      |    t (d, e) AS (SELECT 4, 5),
-      |    c AS (3)
-      |SELECT
-      |    a + b,
-      |    a * c,
-      |    a * t.d
-      |FROM t;""".stripMargin transpilesTo
+  "Common Table Expressions (CTEs)" should {
+    """support expressions""" in {
```
ericvergnaud, line 355: would "have precedence" be better ?

ericvergnaud (approved):

ericvergnaud:
> Looks like some of the functional tests have cosmetic differences with the Python versions, specifically extra parentheses when transpiling set operations.
> 
> I'll move the affected tests into the core-specific test set.

Not sure we want to drop the python tests altogether, notably n the light of the Multiplexer MVP scope ?

sundarshankar89:
> Looks like some of the functional tests have cosmetic differences with the Python versions, specifically extra parentheses when transpiling set operations.
> 
> I'll move the affected tests into the core-specific test set.

Move those tests under `core_engine` folder

tests/resources/functional/snowflake/cte/cte_set_operation_precedence.sql
```diff
@@ -0,0 +1,17 @@
...
+WITH a AS (SELECT 1, 2, 3)
+
+SELECT 4, 5, 6
+UNION
+SELECT * FROM a;
+
+-- databricks sql:
+WITH a AS (SELECT 1, 2, 3)
+
+(SELECT 4, 5, 6)
```
ericvergnaud, line 15: I'm not sure the parenthesis are required here

tests/resources/functional/snowflake/cte/multiple_cte.sql
```diff
@@ -0,0 +1,19 @@
...
+     c AS (SELECT * FROM a)
+SELECT * from b
+UNION
+SELECT * FROM c;
+
+-- databricks sql:
+WITH a AS (SELECT 1, 2, 3),
+     b AS (SELECT 4, 5, 6),
+     c AS (SELECT * FROM a)
+(SELECT * from b)
```
ericvergnaud, line 17: I'm not sure the parenthesis are required here

tests/resources/functional/tsql/cte/cte_set_operation_precedence.sql
```diff
@@ -0,0 +1,17 @@
...
+WITH a AS (SELECT 1, 2, 3)
+
+SELECT 4, 5, 6
+UNION
+SELECT * FROM a;
+
+-- databricks sql:
+WITH a AS (SELECT 1, 2, 3)
+
+(SELECT 4, 5, 6)
```
ericvergnaud, line 15: I'm not sure the parenthesis are required here

tests/resources/functional/tsql/cte/multiple_cte.sql
```diff
@@ -0,0 +1,19 @@
...
+     c AS (SELECT * FROM a)
+SELECT * from b
+UNION
+SELECT * FROM c;
+
+-- databricks sql:
+WITH a AS (SELECT 1, 2, 3),
+     b AS (SELECT 4, 5, 6),
+     c AS (SELECT * FROM a)
+(SELECT * from b)
```
ericvergnaud, line 17: I'm not sure the parenthesis are required here

tests/resources/functional/snowflake/cte/cte_set_operation_precedence.sql
asnare: With respect to the semantics of the query you are correct in that they're not _required_. However they are harmless (and unambiguous).

With respect to the functional tests, they are currently required because the output stage produces them. (I haven't looked into the output stage, but assume parentheses are there because this is the easiest "safe" and correct thing to do: eliminating redundant parentheses falls into the category of optimisations.)

tests/resources/functional/tsql/cte/multiple_cte.sql
asnare: Same situation as [here](https://github.com/databrickslabs/remorph/pull/1273/files/38fc3b3c0c96a2c654d2af20a42381ea2ff6e70c#r1869060713).

core/src/test/scala/com/databricks/labs/remorph/transpilers/SnowflakeToDatabricksTranspilerTest.scala
asnare: I think this comment was intended for the other test, and I appreciate the suggestion: the language was clumsy and I've fixed it in 63eb9bfacc656502d6209da2a58527a765086214.

asnare:
> > Looks like some of the functional tests have cosmetic differences with the Python versions, specifically extra parentheses when transpiling set operations.
> > I'll move the affected tests into the core-specific test set.
> 
> Not sure we want to drop the python tests altogether, notably n the light of the Multiplexer MVP scope ?

That's an interesting point, although my understanding is that our dev priorities at the moment are still focused on the core transpiler and not addressing issues in the python-based implementation. (Arguably all of the 'core engine' tests are issues that should be addressed in the python-based implementation.)

In this case the issue to address looks cosmetic, but might not be. The python implementation doesn't insert parentheses around the query expressions when set operations are present. I _know_ the core version will always be correct. What I don't know about the Python version:
 - Is this an optimisation that it makes?
 - Will parenthesis be inserted when it _does_ matter? (Specifically, there are mixes of parentheses combined with `UNION [ALL]` and `EXCEPT` where the parentheses probably can't be removed.)

At any rate aligning the python and core implementations here is beyond the scope of this PR.

tests/resources/functional/snowflake/cte/cte_set_operation_precedence.sql
ericvergnaud: Then shouldn't we create a ticket for it ?

vil1 (approved):
lgtm

tests/resources/functional/snowflake/cte/cte_set_operation_precedence.sql
asnare: That's a good question. I've created #1281 to cover this topic.

PR #1274, opened 2024-12-03
Implement jinja template processing in Snowflake

jimidle:
Here we implement changes to the ANTLR grammars and the visitors such that Snowflake parsing can deal with templates in the same way as TSQL


jimidle:
Moved to Morpheus

PR #1276, opened 2024-12-04
Move DealiasLCAs after CallMapper and rely on Fn's isIdempotent

vil1:
Fixes failing tests in the base branch by doing two things (the second being a temporary workaround).

1. Move `DealiasLCAs` after `SnowflakeCallMapper` so that it sees (mostly) "actual" functions (that is, subclasses of `Fn` that aren't `CallFunction`). That way, we can rely on more specific function definitions to decide whether they're idempotent or not (as an example `RAND` is idempotent if called with a specific seed, but isn't when called without).
2. Make `CallFunction.isIdempotent` return `true`. That is blatantly **wrong**, as we should be defensive and assume these functions that aren't caught by the CallMapper are not idempotent. In fact, I added a TODO item about that. The rationale behind this unsafe choice it that such functions that currently appear in our tests but are not caught by the CallMapper are mere missing implementation (found 26 such instances). As soon as we get all these functions implemented, we can make `CallMapper.isIdempotent` return `false` as it should.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/intermediate/functions.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/intermediate/functions.scala
@@ -6,41 +6,19 @@ import java.util.Locale
 
 trait Fn extends Expression {
   def prettyName: String
-  def isIdempotent: Boolean = false
+  def isIdempotent: Boolean = true
+}
+
+trait NonIdempotent extends Fn {
+  override final def isIdempotent: Boolean = false
 }
 
 case class CallFunction(function_name: String, arguments: Seq[Expression]) extends Expression with Fn {
   override def children: Seq[Expression] = arguments
   override def dataType: DataType = UnresolvedType
   override def prettyName: String = function_name.toUpperCase(Locale.getDefault)
-  override def isIdempotent: Boolean = {
-    if (arguments.exists {
-      case fn: Fn => !fn.isIdempotent
-      case _ => false
-    }) {
-      false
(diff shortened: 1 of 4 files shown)


ericvergnaud:
TBH I'm not super comfortable with the proposed approach. 
We want to raise warnings whenever replacing an alias is risky.
We need Rules to support Transformation to achieve that.
Until we have that, raising an exception when a function is not idempotent breaks tests.
As mentioned in the PR returning true from CallFunction.isIdempotent does not improve the situation since we're still hiding the risk from the user.
IMHO we should prioritize https://github.com/databrickslabs/remorph/issues/1278 and live with the existing "silent" risk

ericvergnaud:
I think we should make this part of a broader PR that improves dealiasing:
 - checks for single calls
 - checks for multiple not idempotent calls
 - possibly avoids multiple calls using a CTE
 - raises warnings if dealiasing involves more than 2 expressions (thus introducing precedence issues) 
 (requires #1278)

vil1:
> I think we should make this part of a broader PR that improves dealiasing:
> 
> * checks for single calls
> * checks for multiple not idempotent calls
> * possibly avoids multiple calls using a CTE
> * raises warnings if dealiasing involves more than 2 expressions (thus introducing precedence issues)
>   (requires [[FEATURE]: Support Transformation API for rules #1278](https://github.com/databrickslabs/remorph/issues/1278))

I'm fine with that, I'll approve #1242 then

PR #1279, opened 2024-12-04
Added support sha2 function in SQLGLOT engine

### LINKED ISSUE #951: [BUG]: SHA2 function is not having the digest size if it is not specified

### Category of Bug / Issue

TranspileValidationError

### Current Behavior

select sha2(test_col) from test_tbl

### Expected Behavior

select sha2(test_col, 256) from test_tbl

### Steps To Reproduce

If the digest size is not specified in a Snowflake query using SHA2, Snowflake will use 256 as the default.
In Databricks a digest size must be specified.
When transpile is being run on a Snowflake script if a sha2 function does not have its digest size specified it should be added, defaulting to 256 to match Snowflake.

_No response_

### Sample Query

```shell
select sha2(test_col) from test_tbl
```

### Operating System

macOS

### Version

latest via Databricks CLI

ganeshdogiparthi-db:
closes #951


sundarshankar89 (approved):
LGTM

vil1 (approved):
lgtm

PR #1288, opened 2024-12-05
Fix Snowflake precedence of `INTERSECT` with respect to other set operations

### LINKED ISSUE #1255: [TECH DEBT]: Handle the precedence of the `INTERSECT` set operation in the grammar(s) rather than in code

asnare:
This PR updates the Snowflake processing so that the precedence of `INTERSECT` is handled correctly. At present the set operations (`UNION`, `EXCEPT` and `INTERSECT`) are all handled left-to-right with equal precedence. However this is incorrect, Snowflake semantics are[^1], from highest to lowest:
1. Parenthesis
2. `INTERSECT`
3. `UNION [ALL]` and `MINUS`/`EXCEPT`

At the same level operators are processed from left to right.

Resolves #1255.

Tests coverage:
 - Unit tests.
 - Transpiler tests.
 - Functional tests.

Incidental changes include factoring out some of the tests that we run against TSQL so that the same tests can run against Snowflake.

[^1]: Reference: https://docs.snowflake.com/en/sql-reference/operators-query

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -12,7 +12,7 @@
   <properties>
     <httpclient.version>4.5.14</httpclient.version>
     <jackson.version>2.15.2</jackson.version>
-    <junit-bom.version>5.10.0</junit-bom.version>
+    <junit-bom.version>5.11.3</junit-bom.version>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <!-- 5.x starting from 5.0.0 requires JRE 11 -->


core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
```diff
@@ -67,10 +67,16 @@ class SnowflakeAstBuilder(override val vc: SnowflakeVisitorCoordinator)
...
+      val query = ctx.queryExpression().accept(this)
+      Option(ctx.withExpression()).foldRight(query)(buildCTE)
+    }
+  }
+
+  override def visitQueryExpression(ctx: QueryExpressionContext): ir.LogicalPlan = {
     errorCheck(ctx).getOrElse {
       val select = ctx.selectStatement().accept(vc.relationBuilder)
-      val queryExpression = ctx.setOperators().asScala.foldLeft(select)(buildSetOperator)
-      Option(ctx.withExpression()).foldRight(queryExpression)(buildCTE)
```
jimidle, line 73: Does this violate the parameter list stuff, or no?

https://github.com/databricks/scala-style-guide?tab=readme-ov-file#multiple-parameter-lists

jimidle (approved):
LGTM - but I wonder about the multiple parameter lists - not a blocker though.

core/src/main/scala/com/databricks/labs/remorph/parsers/snowflake/SnowflakeAstBuilder.scala
asnare: That's a good question. I interpret that guideline to be about declaring functions that use multiple parameter lists, which the `case class` is effectively doing.

In this example we're using them on a method declared by the standard library, which is not what I think the guideline is aimed at. (Such usage is unavoidable due to the way Scala libraries are often designed, especially if they take a function/closure as a parameter.)

ericvergnaud (approved):
LGTM

PR #1300, opened 2024-12-06
Fix TSQL precedence of `INTERSECT` with respect to `UNION`/`EXCEPT`

asnare:
This is an incredibly embarrassing PR, because it turns out that in #1126 and #1259 I totally misunderstood higher vs. lower precedence and implemented `INTERSECT` with the wrong precedence with respect to `UNION` and `EXCEPT`. This PR fixes this, and updates the associated tests.


vil1 (approved):
lgtm

jimidle (approved):
LGTM

Issue #1298, opened 2024-12-06
[FEATURE]: Multiplexer - create launcher from CLI


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Transpile

### Problem statement

Pluggable transpilers need to be launched from the CLI

### Proposed Solution

Add a flag to the CLI `transpile` command that will be used to locate the transpiler metadata

PR #1314, opened 2024-12-06
Refactor python transpiler code

ericvergnaud:
This PR implements a clear structure for existing python transpiler code, and also renames a few classes for clarity.

Progresses #1298


sundarshankar89 (commented):
Proposing an alternate structure.

```
sqlglot
       - dialects
                 - snowflake.py
                 - presto.py
                 - oracle.py
                 - teradata.py
```

Or

    
```
sqlglot
       - parser
                 - snowflake.py
                 - presto.py
                 - oracle.py
                 - teradata.py
           - generator
                  -  databricks.py
````

ericvergnaud:
> Proposing an alternate structure.
> 
> ```
> sqlglot
>        - dialects
>                  - snowflake.py
>                  - presto.py
>                  - oracle.py
>                  - teradata.py
> ```
> 
> Or
> 
> ```
> sqlglot
>        - parser
>                  - snowflake.py
>                  - presto.py
>                  - oracle.py
>                  - teradata.py
>            - generator
>                   -  databricks.py
> ```

Good suggestion, thanks, done.

src/databricks/labs/remorph/transpiler/sqlglot/generator/databricks.py
bishwajit-db, line 1: IMO, we should also keep databricks.py inside dialects. databricks(package) -> databricks.py instead of a separate generator package like other dialects. Because the generator related code is going to be contained inside a dialect only. Having a separate generator package doesn't add any additional value.

ericvergnaud: mmm... fair point, but it will never be used as a 'read' dialect, so wouldn't that be misleading ?
Also it might benefit in the future from generation related stuff that would need to sit in a dedicated package.

ericvergnaud: The package name should indicate what it does (as opposed to what it 'is'). To that end, I've renamed 'dialects' to 'parsers'.
(there are exceptions to the above rule, for example 'sqlglot' is a specialization of 'transpiler')

sundarshankar89: nit: having folders for each of them is overkill, isn't it? can we just have them have files inside?

ericvergnaud: moved the parsers up, doesn't hurt

sundarshankar89 (approved):
LGTM

sundarshankar89:
@ganeshdogiparthi-db I would like you to stamp this as you are more closer to user impact this has.

bishwajit-db (approved):
LGTM

PR #1315, opened 2024-12-06
Create launcher from cli

### LINKED ISSUE #1298: [FEATURE]: Multiplexer - create launcher from CLI

ericvergnaud:
This PR enhances the CLI as follows:
 - adds a `transpiler-config-path` arg to the CLI that points to a pluggable transpiler configuration file
 - defaults the transpiler-config-path` arg to 'sqlglot' such that built-in transpilers can still be invoked during the transition phase
 - makes the 'source-dialect' optional, it is only required if the transpiler supports more than 1 dialect

It introduces an LSPEngine, which acts as an alternative to the existing SqlglotEngine. It is not operational yet, this PR is about routing CLI calls, not executing them.

The LSPEngine is configured from the configuration stored at the above 'transpiler' path.
The configuration file itself is a yaml file with 2 sections:
```
remorph:
  version: 1
  dialects:
    - snowflake
  environment:
    - SOME_ENV: abc
  command_line:
    - python
    - lsp_server.py
custom:
  whatever: xyz

```
The `remorph` section is versioned and must strictly follow our spec.
The `custom` section is free-style, for use by the LSP server itself.

Both engines are loaded by the CLI, and used throughout the code, for both `transpile` and `lineage` methods.

Requires #1345 

Fixes #1298


src/databricks/labs/remorph/intermediate/root_tables.py
```diff
@@ -7,32 +7,36 @@
...
-            sql_content = read_file(filename)
-            self.engine_adapter.parse_sql_content(dag, sql_content, filename, engine)
+        for path in get_sql_file(self.input_path):
+            logger.debug(f"Generating Lineage file: {path}")
+            sql_content = read_file(path)
+            self._populate_dag(sql_content, path, dag)
 
         return dag
+
+    def _populate_dag(self, sql_content: str, path: Path, dag: DAG):
```
ericvergnaud, line 39: code copied from EngineAdapter

src/databricks/labs/remorph/lineage.py
```diff
@@ -20,13 +21,13 @@ def _generate_dot_file_contents(dag: DAG) -> str:
...
-def lineage_generator(source: str, input_sql: str, output_folder: str):
-    input_sql_path = Path(input_sql)
+def lineage_generator(engine: TranspileEngine, source_dialect: str, input_source: str, output_folder: str):
+    input_sql_path = Path(input_source)
     output_folder = output_folder if output_folder.endswith('/') else output_folder + '/'
 
     msg = f"Processing for SQLs at this location: {input_sql_path}"
     logger.info(msg)
-    root_table_identifier = RootTableIdentifier(source, input_sql_path)
+    root_table_identifier = RootTableLocator(engine, source_dialect, input_sql_path)
```
ericvergnaud, line 30: renamed for clarity

src/databricks/labs/remorph/transpiler/transpile_engine.py
```diff
@@ -7,10 +9,44 @@
...
+            return SqlglotEngine()
+        if not transpiler.exists():
+            raise ValueError(f"Error: Invalid value for '--transpiler': '{str(transpiler)}', file does not exist.")
+        # pylint: disable=import-outside-toplevel, cyclic-import
+        from databricks.labs.remorph.transpiler.lsp.lsp_engine import LSPEngine
+
+        return LSPEngine.with_config_path(transpiler)
+
+    @abc.abstractmethod
+    def analyse_table_lineage(
```
ericvergnaud, line 27: renamed for clarity

tests/unit/test_cli_lineage.py
```diff
@@ -0,0 +1,92 @@
+import datetime
```
ericvergnaud, line 1: tests moved from test_cli.py

tests/unit/test_cli_other.py
```diff
@@ -0,0 +1,36 @@
+from unittest.mock import patch
```
ericvergnaud, line 1: tests moved from test_cli.py

tests/unit/test_cli_transpile.py
```diff
@@ -1,148 +1,14 @@
-import datetime
-import io
-from unittest.mock import create_autospec, patch
+from unittest.mock import create_autospec, patch, PropertyMock, ANY
```
ericvergnaud, line 1: renamed from test_cli.py

src/databricks/labs/remorph/intermediate/engine_adapter.py
```diff
@@ -1,26 +0,0 @@
-import logging
```
ericvergnaud, line 1: code moved to RootTableExtractor

src/databricks/labs/remorph/config.py
```diff
@@ -14,7 +14,8 @@ class TranspileConfig:
     __file__ = "config.yml"
     __version__ = 1
```
sundarshankar89, line 15: @ericvergnaud  whenever we add new fields we bump up the version number. and build an upgrade script that will upgrade to a newer version. which can be separate PR.

src/databricks/labs/remorph/config.py
```diff
@@ -14,7 +14,8 @@ class TranspileConfig:
...
 
-    source_dialect: str
+    transpiler: str
```
sundarshankar89, line 17: Should we default to our sqlglot path?

src/databricks/labs/remorph/transpiler/transpile_engine.py
```diff
@@ -7,10 +9,44 @@
 
 class TranspileEngine(abc.ABC):
 
+    @classmethod
+    def load_engine(cls, transpiler: Path) -> TranspileEngine:
```
sundarshankar89, line 13: Why can't the load_engine be a abstract and implementation is pushed to the actual engine itself ?

tests/resources/lsp_transpiler/config.yml
```diff
@@ -0,0 +1,11 @@
+remorph:
```
sundarshankar89, line 1: TranspileConfig also has the same name, we need a different name for both or standardize both.

src/databricks/labs/remorph/config.py
ericvergnaud: Agreed in general, but this doesn't apply here because "config.yml" is just a test sample.
What we are versioning is the data structure sent to the remote server, and they don't have access to our code base, so we need a version number in the structure itself.

src/databricks/labs/remorph/config.py
ericvergnaud: Done here: https://github.com/databrickslabs/remorph/pull/1315/files?short_path=b335630#:~:text=transpiler%20%3D%20self,directory/file)%22)

src/databricks/labs/remorph/transpiler/transpile_engine.py
ericvergnaud: Not sure I understand your comment ? This class method is about instantiating the engine such that other methods can be abstract...

tests/resources/lsp_transpiler/config.yml
ericvergnaud: Not sure this is correct, I can't find a `remorph` field in `TranspileConfig`...
See https://github.com/databrickslabs/remorph/blob/main/src/databricks/labs/remorph/config.py#L12C1-L25C1
Also no longer in `RemorphConfigs`...

src/databricks/labs/remorph/config.py
ericvergnaud: ok, but that mechanism seems to apply to server-side stuff ? How do I upgrade a local `TranspileConfig` _before_ `Installation` tries (and probably fails) to load it ?

ericvergnaud: unbumped the version because it breaks the test
are we ok to do this in a separate PR ?
Created issue https://github.com/databrickslabs/remorph/issues/1374 to remember doing it...

tests/resources/lsp_transpiler/config.yml
sundarshankar89: I meant the file name config.yml 

https://github.com/databrickslabs/remorph/blob/main/src/databricks/labs/remorph/config.py#L14

ericvergnaud: ah. Fixed

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -51,7 +51,7 @@ def _process_file(
         validate_error_list.append(lca_error)
 
     transpiler_result: TranspilationResult = _transpile(
-        transpiler, config.source_dialect, config.target_dialect, source_sql, input_file, []
+        transpiler, config.source_dialect or "", config.target_dialect, source_sql, input_file, []
```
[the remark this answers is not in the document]
ericvergnaud: No, it's assuming that a given source dialect has been selected (as is already the case)

src/databricks/labs/remorph/config.py
ericvergnaud: Bumped the version and updated the expected version in the failing tests

src/databricks/labs/remorph/transpiler/transpile_engine.py
ericvergnaud: Why not but what problem would that solve ?
In the end state, the SQLGlot engine will go away (becoming a pluggable transpiler), so this code will become simpler.

sundarshankar89 (approved):
LGTM

ericvergnaud:
> LGTM @ericvergnaud is there a link to the pytest ticket which you have raised?

Yes but it's not related to this PR, see https://github.com/databrickslabs/remorph/issues/1363.

bishwajit-db (commented):
IMO, we should not make the 'source-dialect' optional. Let's make user's intention explicit and if the engine doesn't support it, let the engine throw an error.

ericvergnaud:
> IMO, we should not make the 'source-dialect' optional. Let's make user's intention explicit and if the engine doesn't support it, let the engine throw an error.

@gueniai what do you think ? 

Here is the proposed business logic in the PR:
 - user provides a path to the transpiler config (via the --transpiler CLI option)
 - if that config only supports 1 dialect, then we're good to go
 - if that config supports more than 1 dialect, then the user must also provide the source dialect (via the --source-dialect CLI option)

Bishwajit thinks that 'source-dialect' CLI option should aways be required.

README.md
```diff
@@ -117,8 +117,9 @@ Verify the successful installation by executing the provided command; confirmati
 2. The SQL file should encompass scripts intended for migration to Databricks SQL.
 
 Below is the detailed explanation on the arguments required for Transpile.
+- `transpiler [Required]` - Path to the transpiler configuration file.
```
gueniai, line 120: nit: this should be "transpiler-config-path", as we expect not the name of the transpiler nor the executable

src/databricks/labs/remorph/transpiler/transpile_engine.py
```diff
@@ -7,12 +9,39 @@
...
+        if str(transpiler) == "sqlglot":
```
gueniai: Add TODO here to modify this path once we have the SqlGlot plugin ready

README.md
ericvergnaud: Done

src/databricks/labs/remorph/transpiler/execute.py
gueniai, line 54: Now that this is required, there's no need to check for empty

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -255,7 +255,7 @@ def transpile_sql(
     transpiler: TranspileEngine = SqlglotEngine()
 
     transpiler_result = _transpile(
-        transpiler, config.source_dialect, config.target_dialect, source_sql, Path("inline_sql"), []
+        transpiler, config.source_dialect or "", config.target_dialect, source_sql, Path("inline_sql"), []
```
gueniai, line 258: Same here, no need to check for empty

gueniai (changes requested):

labs.yml
```diff
@@ -14,8 +14,12 @@ commands:
   - name: transpile
     description: Transpile SQL script to Databricks SQL
     flags:
+      - name: transpiler
+        description: Path to the transpiler configuration file
+        default: sqlglot
```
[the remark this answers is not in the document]
ericvergnaud: That's temporary, until we move sqlglot to a pluggable transpiler

src/databricks/labs/remorph/transpiler/transpile_engine.py
ericvergnaud: done

src/databricks/labs/remorph/transpiler/execute.py
ericvergnaud: done

src/databricks/labs/remorph/transpiler/execute.py
ericvergnaud: done

gueniai (approved):
LGTM

PR #1320, opened 2024-12-09
Isolate sqlglot dependencies

ericvergnaud:
Our current code uses sqlglot data structures in places where we should not depend on sqlglot.
This PR fixes the issue.

Progresses #1298


src/databricks/labs/remorph/config.py
```diff
@@ -1,40 +1,12 @@
...
+from pathlib import Path
 
 from databricks.labs.remorph.transpiler.transpile_status import ParserError
 from databricks.labs.remorph.reconcile.recon_config import Table
-from databricks.labs.remorph.transpiler.sqlglot.generator import databricks
-from databricks.labs.remorph.transpiler.sqlglot.parsers import oracle, presto, snowflake
-
-logger = logging.getLogger(__name__)
-
-SQLGLOT_DIALECTS: dict[str, DialectType] = {
```
ericvergnaud, line 13: moved to `sqlglot/dialect_utils.py`

src/databricks/labs/remorph/transpiler/sqlglot/lca_utils.py
```diff
@@ -15,20 +15,26 @@
 
 
 def check_for_unsupported_lca(
-    dialect: DialectType,
-    sql: str,
-    filename: str,
+    from_dialect: str,
```
sundarshankar89, line 18: Any reason we can't propagate dialect type directly here?
```suggestion
    from_dialect: DialectType,
```

This will avoid cyclic imports

sundarshankar89 (changes requested):

src/databricks/labs/remorph/transpiler/sqlglot/lca_utils.py
ericvergnaud: Agreed, done

src/databricks/labs/remorph/config.py
```diff
@@ -51,13 +23,19 @@ class TranspileConfig:
     schema_name: str = "transpiler"
     mode: str = "current"
 
-    def get_read_dialect(self):
-        return get_dialect(self.source_dialect)
+    @property
+    def input_path(self):
+        if self.input_source is None:
```
bishwajit-db, line 28:
```suggestion
        if not self.input_source:
```
This way we can also check for both empty strings and None

src/databricks/labs/remorph/config.py
```diff
@@ -51,13 +23,19 @@ class TranspileConfig:
...
+            raise ValueError("Missing input source!")
+        return Path(self.input_source)
+
+    @property
+    def output_path(self):
+        return None if self.output_folder is None else Path(self.output_folder)
```
bishwajit-db, line 34:
```suggestion
        return None if not self.output_folder else Path(self.output_folder)
```
This way we can also check for both empty strings and None

bishwajit-db (approved):
LGTM

PR #1322, opened 2024-12-09
Prevent Dependabot from upgrading Databricks Connect.

asnare:
This PR configures Dependabot so that it won't try to upgrade Databricks Connect.

Relates #1317.


sundarshankar89 (approved):
LGTM

PR #1338, opened 2024-12-10
Adjust Dependabot's maven update policies

asnare:
This PR adjusts the update policy for Dependabot:

 - We try to configure Dependabot to ignore `-preview` releases.
 - For Scala patch-release updates are allowed: it's the major/minor changes that are breaking that we wish to manage ourselves.

PR #1339, opened 2024-12-10
Fix dependabot configuration

asnare:
PR #1338 attempted to configure Dependabot to ignore (maven) versions with a specific suffix. However Dependabot doesn't allow that style for specifying (maven) versions to exclude. This PR reverts that change.


sundarshankar89 (approved):
LGTM

ericvergnaud (approved):

PR #1349, opened 2024-12-11
TSql: fix precedence of `ORDER BY` clause

### LINKED ISSUE #1256: [TECH DEBT]: Fix `ORDER BY` precedence in TSQL grammar
### Current Behavior

The TSQL grammar currently allows for `ORDER BY` expressions between set operations (`INTERSECT`, `EXCEPT`, `UNION`) and associates them with the intermediate expression.

### Expected Behavior

For TSQL the `ORDER BY` clause is applied _after_ all the set operations: it should not be attached to the immediately preceding expression. That is:

```sql
SELECT 10
UNION
SELECT 20
ORDER BY 1
```
is equivalent to:
```sql
(SELECT 10
 UNION
 SELECT 20)
ORDER BY 1
```
and _not_ this as currently is the case:
```sql
(SELECT 10)
UNION
(SELECT 20
 ORDER BY 1)
```

Reference: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-union-transact-sql?view=sql-server-ver16#c-using-union-of-two-select-statements-with-order-by

Further to this, the following is erroneous:
```sql
SELECT 10
ORDER BY 1
UNION
SELECT 10
ORDER BY 2
```

_The documentation refers to this as _incorrect_ which is a little ambiguous: it's open to read that rather than being an error it could be accepted but pointless. Testing shows that it is treated as an error:_
```sql
select 10
order by 1
union
select 20
order by 1 desc
```
_yields_
```
Failed to execute query. Error: Incorrect syntax near the keyword 'union'.
```

### Functional Test

This functional test should pass if `ORDER BY` has been implemented correctly.

```sql
--
-- ORDER BY appears at the end of a sequence of set operations, and applies to the result of the set operation.
--
-- tsql sql:

SELECT 1
UNION
SELECT 2

ORDER BY 1 DESC; -- Applies to the complete UNION, not just the final expression.

-- databricks sql:
((SELECT 1)
 UNION
 (SELECT 2))
ORDER BY 1 DESC;
```

asnare:
This PR updates where we handle `ORDER BY` clauses in TSql: these should be applied after set operations to the complete sequence, not to each individual query within a sequence of set operations.

References:

 - https://learn.microsoft.com/en-us/sql/t-sql/queries/select-transact-sql?view=sql-server-ver16#syntax
 - https://learn.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-union-transact-sql?view=sql-server-ver16#c-using-union-of-two-select-statements-with-order-by

Tested:

 - [x] Updated unit tests
 - [x] Transpiler tests
 - [X] Functional tests

Blocked by #1350.
Resolves #1256.


core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
```diff
@@ -50,14 +50,11 @@ class TSqlRelationBuilder(override val vc: TSqlVisitorCoordinator)
     case None =>
       // TODO: The FOR clause of TSQL is not supported in Databricks SQL as XML and JSON are not supported
```
ericvergnaud, line 51: Is there a ticket for this ?

ericvergnaud (approved):
LGTM

core/src/main/scala/com/databricks/labs/remorph/parsers/tsql/TSqlRelationBuilder.scala
asnare: Yes: #1123

jimidle (approved):
lgtm

PR #1350, opened 2024-12-11
Fix logical plan for `LIMIT` when combined with `OFFSET`

asnare:
This PR updates code emitted by the logical planner for IR of the form: $\mathop{\mathit{limit}}(\mathop{\mathit{offset}}(input, x), y)$

This is the form that should be used for the following SQL:
```sql
LIMIT x OFFSET y
```
or
```sql
OFFSET x ROWS FETCH FIRST y ROWS ONLY
```

Prior to this PR we emit the following code which is invalid:
```sql
OFFSET y LIMIT x
```

Further to fixing the logical planner, this PR marks the Snowflake builder where this isn't being handled correctly.

### DIFF
--- a/core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
+++ b/core/src/main/scala/com/databricks/labs/remorph/generators/sql/LogicalPlanGenerator.scala
@@ -25,6 +25,8 @@ class LogicalPlanGenerator(
         }
         code"${source} WHERE ${expr.generate(condition)}"
       }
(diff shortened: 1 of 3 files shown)


vil1 (changes requested):
Could you please go ahead and fix the implementation in SnowflakeRelationBuilder?

vil1 (approved):
Created https://github.com/databrickslabs/remorph/issues/1351 for the Snowflake fix

ericvergnaud (approved):
LGTM

PR #1354, opened 2024-12-12
Transpile using lsp

ericvergnaud:
This PR implements `LSPEngine`, a `TranspileEngine` that leverages the Language Server Protocol to launch and communicate with a pluggable transpiler, implemented as a LSP server.

In our prototype, we used the existing LSP `CodeAction` mechanism with `CodeActionKind.Refactor.` This design works well when using VSCode as a client, but is not ideal for batch transpile:
 - it does not provide a deterministic way to identify the command for transpiling to databricks
 - it requires 2 interactions with the server for each file (1 to fetch the command, 1 to execute it)
 - the transpile result is sent asynchronously, which complicates the client engine's job
 - the transpile errors are sent asynchronously, which complicates the client engine's job

For those reasons, this PR chooses a different design, relying on LSP custom capabilities.
The `LSPEngine` supports a "document/transpileToDatabricks" capability, that the server must register during initialization.
Once the registration is done, the LSPEngine can then safely invoke that capability, which returns all the results in a single response (changes and errors).

This mechanism is tested successfully against a test LSPServer (which for transpiling, simply sets the content to uppercase).

Worth noting, this PR works around a limitation (bug ?) in the lsprotocol Python library, where custom capability messages are incorrectly serialized. This is considered tech debt, for which an issue has been created: https://github.com/databrickslabs/remorph/issues/1378

Fixes #1299 
Requires #1364 
Requires #1390

PR #1362, opened 2024-12-12
Perform feasability checks as part of transpile

ericvergnaud:
Our current implementation performs pre-checks (such as non-transpilable LCAs) separately from the transpile itself.
This puts to much responsibility on remorph, and should be done by the TranspileEngine

This PR fixes the issue.

Progresses #1354 

Requires #1361


ericvergnaud:
Incorrect merge, superseded by #1364

Issue #1299, opened 2024-12-06
[FEATURE]: Multiplexer - have a pluggable transpiler transpile a file 


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Transpile

### Problem statement

Remorph must be able to have a file transpiled by a pluggable transpiler  

### Proposed Solution

Invoke the transpiler using LSP protocol

PR #1364, opened 2024-12-12
Perform checks during transpile

ericvergnaud:
Our current implementation performs pre-checks (such as `check_for_unsupported_lca`) separately from the `transpile` itself.
This puts to much responsibility on `remorph`, and should be done by the `TranspileEngine`.
This also creates complexity for Pluggable Transpilers, who need to implement several method instead of just a `transpile` one.

This PR fixes the issue.

Progresses #1299

Requires #1361


sundarshankar89 (approved):
LGTM

PR #1371, opened 2024-12-13
Allow python local imports if they solve a cyclic import problem

ericvergnaud:
Our current CI blocks merge of PRs if added python code contains comments such as:

`# pylint: disable=cyclic-import, import-outside-toplevel`

Local imports are discouraged, but are the most effective way of solving cyclic imports.

This PR solves the issue.

### DIFF
--- a/tests/unit/no_cheat.py
+++ b/tests/unit/no_cheat.py
@@ -20,7 +20,7 @@ def no_cheat(diff_text: str) -> str:
             continue
         codes = {_strip_code(code) for code in line[idx + len(DISABLE_TAG) :].split(',')}
         allowed_local_cyclic_imports = {'cyclic-import', 'import-outside-toplevel'}
-        if len(codes.intersection(allowed_local_cyclic_imports)) == 2:
+        if len(codes.intersection(allowed_local_cyclic_imports)) == len(allowed_local_cyclic_imports):
             codes = codes.difference(allowed_local_cyclic_imports)
         for code in codes:
             if line.startswith("-"):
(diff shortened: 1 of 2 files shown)


ericvergnaud:
@gueniai you'll have to force merge of the PR because the test file contains examples of patterns we're not letting pass

tests/unit/no_cheat.py
```diff
@@ -14,9 +18,11 @@ def no_cheat(diff_text: str) -> str:
         idx = line.find(DISABLE_TAG)
         if idx < 0:
             continue
-        codes = line[idx + len(DISABLE_TAG) :].split(',')
+        codes = {_strip_code(code) for code in line[idx + len(DISABLE_TAG) :].split(',')}
+        allowed_local_cyclic_imports = {'cyclic-import', 'import-outside-toplevel'}
+        if len(codes.intersection(allowed_local_cyclic_imports)) == 2:
```
vil1, line 23:
```suggestion
        if len(codes.intersection(allowed_local_cyclic_imports)) == len(allowed_local_cyclic_imports):
```
to make future changes easier

sundarshankar89:
@ericvergnaud I would like to hold on to this change till we absolutely need it. I think there is an alternate way where you introduce your local import. 

Thoughts?

ericvergnaud:
> @ericvergnaud I would like to hold on to this change till we absolutely need it. I think there is an alternate way where you introduce your local import.
> 
> Thoughts?

As discussed, this PR does not encourage local imports, rather it tolerates them in very specific use cases. PR reviews can still block PRs.

sundarshankar89 (approved):
LGTM

gueniai (approved):
LGTM

PR #1377, opened 2024-12-18
make changes to Makefile

sundarshankar89:
Adds alias in Make command to build Unit and Integration tests.


Makefile
```diff
@@ -19,18 +19,20 @@ fmt-python:
...
 
-test-python: setup_spark_remote
+test-python:
 	hatch run test
 
 test-scala:
 	mvn test -f pom.xml
 
-integration:
+integration-test: setup_spark_remote
```
ericvergnaud, line 35: maybw rename to `test-integration` ?

ericvergnaud (changes requested):

bishwajit-db (commented):
Just like unit tests, can we add integration-test-python, integration-test-scala? Later it would be convenient to add Scala integration tests and have uniform target naming convention. In fact, we can then add integration-test-all later to trigger both if needed.

sundarshankar89:
> Just like unit tests, can we add integration-test-python, integration-test-scala? Later it would be convenient to add Scala integration tests and have uniform target naming convention. In fact, we can then add integration-test-all later to trigger both if needed.

I have added test-integration and made changes to coverage as well. for now this should be enough considering things we plan to do down the line.

sundarshankar89:
@gueniai  need to add mandatory check for `integration-python`

bishwajit-db (approved):
LGTM

Makefile
sundarshankar89: Done

pyproject.toml
```diff
@@ -63,7 +63,7 @@ reconcile = "databricks.labs.remorph.reconcile.execute:main"
 
 [tool.hatch.envs.default.scripts]
 test        = "pytest --cov src --cov-report=xml tests/unit"
-coverage    = "pytest --cov src tests/unit --cov-report=html"
+coverage    = "pytest --cov src tests --cov-report=html"
```
ericvergnaud, line 66: not sure about this one, do we want to run coverage on both unit and integration tests

ericvergnaud (changes requested):

pyproject.toml
sundarshankar89: Yes, recon uses spark which uses integration tests for coverage. hence this change.

PR #1381, opened 2024-12-19
fix issue where unit tests would create files not ignored by git

ericvergnaud:
Running unit tests locally creates files in the source hierarchy which need to be removed each time before pushing to git
This PR fixes the issue


bishwajit-db (approved):
LGTM

vil1 (approved):
lgtm

PR #1390, opened 2024-12-27
Fix pytest-cov crash

### LINKED ISSUE #1363: [BUG]: pytest-cov crashes when collecting coverage report of async tests

### Category of Bug / Issue

Other

### Current Behavior

As part of implementing LSP, we're adding async functions and corresponding unit tests
When collecting coverage report for these tests, pytest-cov crashes

### Expected Behavior

Test coverage is collected without crashing

### Steps To Reproduce

See https://github.com/databrickslabs/remorph/actions/runs/12300437632/job/34329339753

_No response_

### Sample Query

_No response_

### Operating System

macOS

### Version

latest via Databricks CLI

ericvergnaud:
The `pytest-cov` plugin is used to gather test coverage info, using the `coverage` tool.
It suffers from a limitation where child processes created by tests do not inherit the `coverage` configuration from `pyproject.toml` because that file is not on the command line.
This limitation is described here https://github.com/nedbat/coveragepy/issues/512
With the introduction of LSP, we started encountering crashes when collecting coverage data, because LSP launches a child process.
The solution is to move `coverage` configuration into a dedicated `.coveragerc` file.

This PR does just that.

Fixes #1363

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,15 @@
+[run]
+branch = true
+parallel = true
+omit =
+    src/databricks/labs/remorph/coverage/*
+    src/databricks/labs/remorph/helpers/execution_time.py
+    tests/resources/lsp_transpiler/lsp_server.py
+    __about__.py
+
+[report]
+exclude_lines =
+    no cov
+    if __name__ == .__main__.:
+    if TYPE_CHECKING:
+    def main()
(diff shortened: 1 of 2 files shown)


sundarshankar89 (approved):
LGTM

Issue #1303, opened 2024-12-06
[FEATURE]: Multiplexer - convert Snowflake OSS transpiler


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Transpile

### Problem statement

As part of our MVP, we need an OSS example of a pluggable transpiler. It was decided to use the existing python Snowflake transpiler for that.

### Proposed Solution

Copy the existing code to a new repo and transform it to make it a pluggable transpiler

PR #1391, opened 2024-12-27
Make file path for transpile errors configurable

ericvergnaud:
Our current implementation hard codes the location of the file used to write transpile errors.
This is not in line with other transpile flags that the user can override.
It also exhibits undesirable side effects where error files are created in the source code hierarchy when running unit tests.

This PR fixes the issue by adding an optional `error-file `flag that users can provide.
It also renames the default error file name to use a timestamp rather than a pid.
It also updates tests that generate error files to put them in a tmp folder


src/databricks/labs/remorph/cli.py
```diff
@@ -33,12 +33,14 @@ def raise_validation_exception(msg: str) -> Exception:
 
 
 @remorph.command
+# pylint: disable=too-many-arguments
```
ericvergnaud, line 36: This will go away when implementing issue #1303

sundarshankar89 (approved):
LGTM

PR #1402, opened 2025-01-03
Labs Yaml Refactor

sundarshankar89:


src/databricks/labs/remorph/jvmproxy.py
```diff
@@ -18,6 +18,7 @@ def fn(**_):
 
 
 class JvmProxy:
+    # TODO Refactor this class to use morpheus instead
```
ericvergnaud, line 21: This is no longer required. Morpheus is a LSP server running in its own process, communication is done via LSP protocol

ericvergnaud (changes requested):

src/databricks/labs/remorph/jvmproxy.py
sundarshankar89: Agreed, I suspect a transition time where we will need this to support or trigger CLI should be removed when LSP communication is available for Morpheus.

vijaypavann-db (approved):
LGTM

PR #1404, opened 2025-01-06
Transpile using LSP

### LINKED ISSUE #1299: [FEATURE]: Multiplexer - have a pluggable transpiler transpile a file

ericvergnaud:
This PR implements `LSPEngine`, a `TranspileEngine` that leverages the Language Server Protocol to launch and communicate with a pluggable transpiler, implemented as a LSP server.

In our prototype, we used the existing LSP `CodeAction` mechanism with `CodeActionKind.Refactor.` This design works well when using VSCode as a client, but is not ideal for batch transpile:
 - it does not provide a deterministic way to identify the command for transpiling to databricks
 - it requires 2 interactions with the server for each file (1 to fetch the command, 1 to execute it)
 - the transpile result is sent asynchronously, which complicates the client engine's job
 - the transpile errors are sent asynchronously, which complicates the client engine's job

For those reasons, this PR chooses a different design, relying on LSP custom capabilities.
The `LSPEngine` supports a "document/transpileToDatabricks" capability, that the server must register during initialization.
Once the registration is done, the LSPEngine can then safely invoke that capability, which returns all the results in a single response (changes and errors).

This mechanism is tested successfully against a test LSPServer (which for transpiling, simply sets the content to uppercase).

Worth noting, this PR works around a limitation (bug ?) in the lsprotocol Python library, where custom capability messages are incorrectly serialized. This is considered tech debt, for which an issue has been created: https://github.com/databrickslabs/remorph/issues/1378

Fixes #1299 
Requires #1364 
Requires #1390
Supersedes #1354


ericvergnaud:
@gueniai the python-no-pylint-disable CI job fails as expected. All these breaches are required, see inline comments.

tests/resources/lsp_transpiler/lsp_server.py
```diff
@@ -0,0 +1,151 @@
+import os
+import sys
+from collections.abc import Sequence
+from typing import Any, Literal
+from uuid import uuid4
+
+import attrs
+
```
sundarshankar89, line 8: @ericvergnaud  Can you leave docs or notes on why this is needed for unit tests and why is this implementation under resources.

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -35,12 +68,175 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+            return func(self, params)
+
+        return wrapper
+
+
+class ChangeManager(abc.ABC):
+
+    @classmethod
+    def apply(cls, source_code: str, changes: Sequence[TextEdit]) -> str:
+        lines = source_code.split("\n")
```
sundarshankar89, line 190: @ericvergnaud  How will a SQL procedure, or any complex structure, be managed using this change manager?

Wouldn't the transpiler want the entire stream of syntax be parsed or analyzed to be able to make changes.

tests/resources/lsp_transpiler/lsp_server.py
ericvergnaud: done

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: The transpiler sits behind the protocol not within (the LSP protocol deals with text, not with complex structures).
The way it works is as follows:
  - client sends full text of a document to the server
  - client invokes a server capability or command
  - server returns a list of text edits that should be applied (this is a mandatory LSP optimization where only changes are sent, not the entire text)
  - client applies the changes
 The purpose of this class is to apply the changes.

sundarshankar89 (approved):
LGTM

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -35,12 +68,175 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+        for change in changes:
+            lines = cls._apply(lines, change)
+        return "\n".join(lines)
+
+    @classmethod
+    def _apply(cls, lines: list[str], change: TextEdit) -> list[str]:
+        new_lines = change.new_text.split("\n")
+        # special case where change covers the entire source code
+        if (
```
gueniai, line 199: This could be its own little function. i.e:

> def is_full_document_change(change, lines):
    return (
        change.range.start.line <= 0
        and change.range.start.character <= 0
        and change.range.end.line >= len(lines) - 1
        and change.range.end.character >= len(lines[-1])
    )

Then this could become
```suggestion
        if (is_full_document_change(change, lines):
            return new_lines
```

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -35,12 +68,175 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+            change.range.start.line <= 0
+            and change.range.start.character <= 0
+            and change.range.end.line >= len(lines) - 1
+            and change.range.end.character >= len(lines[-1])
+        ):
+            return new_lines
+        result: list[str] = []
+        # keep lines before
+        if change.range.start.line > 0:
+            result.extend(lines[0 : change.range.start.line])
```
gueniai, line 209:
```suggestion
            result = lines[:change.range.start.line]
```

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -35,12 +68,175 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+        # special case where change covers full lines
+        if change.range.start.character <= 0 and change.range.end.character >= len(lines[change.range.end.line]):
```
gueniai, line 211: `lines[change.range.end.line]` can go into a variable since it is used a few times

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -35,12 +68,175 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+            pass
+        # special case where change is within 1 line
+        elif change.range.start.line == change.range.end.line:
+            old_line = lines[change.range.start.line]
+            if change.range.start.character > 0:
+                new_lines[0] = old_line[0 : change.range.start.character] + new_lines[0]
```
gueniai, line 217: This would have better performance than vanilla string concat
```suggestion
                new_lines[0] = ''.join([old_line[0 : change.range.start.character], new_lines[0]])
```

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -35,12 +68,175 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+            if change.range.end.character < len(old_line):
+                new_lines[-1] += old_line[change.range.end.character :]
```
gueniai, line 219: Same here, although slightly more verbose
```suggestion
                new_lines[-1] = ''.join([new_lines[-1], old_line[change.range.end.character :]])
```

gueniai: essentially, everywhere that currently uses string concat

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -35,12 +68,175 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+
+    @classmethod
+    def apply(cls, source_code: str, changes: Sequence[TextEdit]) -> str:
+        lines = source_code.split("\n")
+        for change in changes:
+            lines = cls._apply(lines, change)
+        return "\n".join(lines)
+
+    @classmethod
+    def _apply(cls, lines: list[str], change: TextEdit) -> list[str]:
```
gueniai, line 196: we need some error handling in this function, we're assuming a lot of depth for `change` and not handling if anything went sideways.

gueniai (changes requested):
I'm not done yet, but wanted to send some initial feedback. For now I've only really gone through lsp_engine. 
Overall, we need more error handling everywhere, from file operations to potential complex objects not being completely loaded.

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -50,22 +246,97 @@ def _load_config(cls, config_path: Path) -> tuple[_LSPRemorphConfigV1, dict[str,
...
+    async def transpile(
+        self, source_dialect: str, target_dialect: str, source_code: str, file_path: Path
+    ) -> TranspileResult:
+        self.open_document(file_path, source_code=source_code)
+        response = await self.transpile_document(file_path)
+        self.close_document(file_path)
+        transpiled_code = ChangeManager.apply(source_code, response.changes)
+        return TranspileResult(transpiled_code, 1, [])
+
+    def open_document(self, file_path: Path, encoding="utf-8", source_code: str | None = None) -> None:
```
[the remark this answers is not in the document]
ericvergnaud: `document` is the LSP terminology (and API) and I believe we should stick to it in LSP-centric code (such as LSPEngine).

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: good suggestion

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: isn't it better practice to let such optimizations be done by the compiler/runtime ?

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: done

ericvergnaud: That's only true if concatenating more than 2 strings, see https://discuss.python.org/t/joining-strings-is-faster-than-adding-but-only-if-concatenating-3-strings/5293

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: See above

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: done

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: Agreed. I've added handling of IndexError which I think is the only built-in exception that can be caused by this code if data is incorrect

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
gueniai: yeah maybe? this is not a dealbreaker for me. I do think it helps with readability, but fine either way

gueniai (approved):
lgtm

Issue #1301, opened 2024-12-06
[FEATURE]: Multiplexer - display transpile errors in the CLI


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Transpile

### Problem statement

Errors returned by a pluggable transpiler should be displayed to the user

### Proposed Solution

Display the errors in the CLI

PR #1405, opened 2025-01-06
Refactor transpile error for LSP

ericvergnaud:
Our current implementation of transpile errors relies on error types i.e. `ParseError` and `ValidationError`.
This will not work with LSP because LSP cannot convey such class hierarchy, rather it conveys flat structures named `Diagnostics`.
This PR aligns our implementation with LSP `Diagnostics` by:
 - dropping `ParseError` and `ValidationError` in favor of their base class `TranspileError`
 - adding a mandatory error code
 - adding a mandatory error kind (with 4 kinds instead of 2)
 - adding a mandatory error severity
 - adding an optional code range to which the error applies

This is in preparation of #1301 
Requires #1404 
Supersedes #1382


ericvergnaud:
@gueniai @sundar this one relatively small ans is next in queue

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -191,8 +187,10 @@ def transpile(workspace_client: WorkspaceClient, engine: TranspileEngine, config
         {
             "total_files_processed": len(result.file_list),
             "total_queries_processed": result.no_of_transpiled_queries,
-            "no_of_sql_failed_while_parsing": result.parse_error_count,
-            "no_of_sql_failed_while_validating": result.validate_error_count,
+            "no_of_sql_failed_while_analysing": result.analysis_error_count,
```
sundarshankar89, line 190: only Morpheus might be able to surface this information rest of them might not. How are we thinking of surfacing this consistently.

sundarshankar89 (approved):
LGTM only small query.

src/databricks/labs/remorph/transpiler/execute.py
ericvergnaud: I believe that's up to each LSP transpiler provider to decide. If they can't specialize that much, then it should all go under `parsing` and `generating`, but that's just my 2p. We're not in charge of _their_ product.

gueniai: yeah we'll just make it available as part of the skeletons, and documented so that future plugins can make use of these

PR #1413, opened 2025-01-08
Teradata DDL support

sriram251-code:
This PR contains the features to support Teradata DDL.


sundarshankar89 (approved):
LGTM

ericvergnaud (changes requested):
Shouldn't this be contributed back to Sqlglot rather than overridden on our side ?
They make frequent releases so it wouldn't slow us down

sundarshankar89:
> Shouldn't this be contributed back to Sqlglot rather than overridden on our side ? They make frequent releases so it wouldn't slow us down

There is some context as to why this approach was taken. we will need to accommodate some legacy patterns till we contribute everything back to OSS Sqlglot itself. 
cc: @gueniai

sundarshankar89:
move to RCT.

PR #1423, opened 2025-01-21
Rename mock

ericvergnaud:
The current class name `Test` for what is actually a mock raises a warning when pytest collects tests.
This PR simply renames that class to avoid the warning


sundarshankar89 (approved):
LGTM

jimidle (approved):
LGTM

PR #1425, opened 2025-01-23
Patch User Agent

sundarshankar89:
Refactored using reflection so we can remove pylint disable

### DIFF
--- a/src/databricks/labs/remorph/transpiler/execute.py
+++ b/src/databricks/labs/remorph/transpiler/execute.py
@@ -215,11 +215,10 @@ def verify_workspace_client(workspace_client: WorkspaceClient) -> WorkspaceClien
 
     TODO: In future refactor this function so it can be used for reconcile module without cross access.
     """
-    product_info = workspace_client.config._product_info
-    if product_info[0] != "remorph":
-        product_info[0] = "remorph"
-    if product_info[1] != __version__:
-        product_info[1] = __version__
+    if workspace_client.config._product != "remorph":
+        workspace_client.config._product = "remorph"
+    if workspace_client.config._product_version != __version__:
+        workspace_client.config._product_version = __version__
     return workspace_client
 
 
--- a/src/databricks/labs/remorph/transpiler/execute.py
+++ b/src/databricks/labs/remorph/transpiler/execute.py
@@ -2,7 +2,7 @@ import asyncio
 import datetime
 import logging
 from pathlib import Path
-from typing import cast
+from typing import cast, Any
 
 from databricks.labs.remorph.__about__ import __version__
 from databricks.labs.remorph.config import (
@@ -33,7 +33,7 @@ from databricks.sdk import WorkspaceClient
 logger = logging.getLogger(__name__)
 
 
-def _process_file(
+async def _process_file(
     config: TranspileConfig,
     validator: Validator | None,
     transpiler: TranspileEngine,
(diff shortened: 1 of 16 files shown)


src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -215,11 +215,10 @@ def verify_workspace_client(workspace_client: WorkspaceClient) -> WorkspaceClien
...
-    product_info = workspace_client.config._product_info
-    if product_info[0] != "remorph":
-        product_info[0] = "remorph"
-    if product_info[1] != __version__:
-        product_info[1] = __version__
+    if workspace_client.config._product != "remorph":
+        workspace_client.config._product = "remorph"
+    if workspace_client.config._product_version != __version__:
+        workspace_client.config._product_version = __version__
     return workspace_client
```
sundarshankar89, line 222: Assignment of the tuple back to workspce_client is missing

```suggestion
    workspace_client.config._product_info = product_info
    return workspace_client
```

ericvergnaud: we're tapping directly into the `workspace_client.config._product_info` tuple so this is not necessary

ericvergnaud: If we have different structures between 'externally created' then we need a better fix

ericvergnaud: maybe you did not upgrade the sdk ?

ericvergnaud (changes requested):
I suspect the issue being fixed could be related to a stale environment ?

src/databricks/labs/remorph/transpiler/execute.py
sundarshankar89: @ericvergnaud  refactored to use reflection instead

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -209,17 +209,17 @@ async def _do_transpile(
...
     TODO: In future refactor this function so it can be used for reconcile module without cross access.
     """
-    product_info = workspace_client.config._product_info
+
+    # Using reflection to set right value for _product_info for telemetry
+    product_info = getattr(workspace_client.config, '_product_info')
     if product_info[0] != "remorph":
-        product_info[0] = "remorph"
-    if product_info[1] != __version__:
-        product_info[1] = __version__
```
asnare, line 222: It's not the point of this PR, given that we're replacing it, but I'm mildly surprised that this worked… I think it's a tuple, in which case assigning to the elements individually won't work.

asnare (commented):
I'm missing a bit of context on this PR, including:

 - Where the workspace comes from;
 - Why it's not being setup correctly during initialisation instead of patched here.

A slightly cleaner approach, by the way, would be to either:

1. Reconfigure the workspace client (make a copy of the `config`, replacing the product information, and assign that to the client; or
2. Create (and return) a new workspace client with the modified configuration from above: it's a bit to modify the client that the caller passed in.

src/databricks/labs/remorph/transpiler/execute.py
ericvergnaud: it probably didn't, but that code is never called anyway... this was to fix formatting alerts

ganeshdogiparthi-db (approved):
STAMP LGTM

PR #1433, opened 2025-01-28
generate recon config from table definition

sriram251-code:
This pull request introduces functionality to generate a JSON file from a `TableRecon` object and includes a unit test for this new feature.

### DIFF
--- a/src/databricks/labs/remorph/helpers/generate_recon_config.py
+++ b/src/databricks/labs/remorph/helpers/generate_recon_config.py
@@ -1,0 +1,40 @@
+import json
+from dataclasses import asdict
+
+from databricks.labs.remorph.config import TableRecon
+from databricks.labs.remorph.reconcile.recon_config import Table, Aggregate, JdbcReaderOptions, ColumnMapping, \
+    Transformation, ColumnThresholds, Filters, TableThresholds
+
+
+def generate_json_file(table_recon: TableRecon, file_name: str):
+    def serialize(obj):
+        """Helper function to handle optional complex objects"""
+        if isinstance(obj, (Aggregate, JdbcReaderOptions, ColumnMapping, Transformation, ColumnThresholds, Filters, TableThresholds)):
+            return asdict(obj)  # Serialize dataclass objects
+        raise TypeError(f"Type {type(obj)} not serializable")
+
+    table_recon_dict = asdict(table_recon)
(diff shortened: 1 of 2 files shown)


src/databricks/labs/remorph/helpers/generate_recon_config.py
```diff
@@ -0,0 +1,22 @@
...
+        raise TypeError(f"Type {type(obj)} not serializable")
+
+    table_recon_dict = asdict(table_recon)
+    json_data = json.dumps(table_recon_dict, default=serialize, indent=4)
+
+    try:
+        with open(file_name, 'w') as json_file:
+            json_file.write(json_data)
+    except (TypeError, IOError) as e:
+        print(f"Failed to generate JSON file: {e}")
```
sundarshankar89, line 21: Logger.

src/databricks/labs/remorph/helpers/generate_recon_config.py
```diff
@@ -0,0 +1,22 @@
+import json
+from dataclasses import asdict, is_dataclass
+
+from databricks.labs.remorph.config import TableRecon
+
+
+def generate_json_file(table_recon: TableRecon, file_name: str):
```
sundarshankar89, line 7: TableRecon is the target You will need to introduce TableDefinition.

``` python

from dataclasses import dataclass, field
from typing import List, Optional

@dataclass
class TableDefinition:
    catalog: str
    schema: str
    table: str
    location: Optional[str] = None
    tableFormat: Optional[str] = None
    viewText: Optional[str] = None
    columns: List[StructField] = field(default_factory=list)
    sizeGb: int = 0
    comment: Optional[str] = None

bishwajit-db: I would suggest to add `primary_keys: list[str] | None = None` to it too.

bishwajit-db: We may not use StructField for field information in TableDefinition. Because the dataType field in it is of type DataType ; which is tied to Spark. When we bring metadata from 3rd party systems, they may not have a direct equivalent for Spark data types. So we may want to use our own dataclass to store column information which is loosely structured around StructField but uses string for dataType. If needed, we can have our dataclass -> StructField transformation function.

PR #1434, opened 2025-01-30
Add language_id in TranspileDocumentResult

ericvergnaud:
In preparation of translating SQL stored procs to Python UDFs, Remorph needs to know the language to which the source was transpiled.
This PR adds that field.


sundarshankar89 (approved):
LGTM

PR #1452, opened 2025-02-18
Prune sqlglot transpiler and customization

ericvergnaud:
With the release of `Morpheus` and `remorph-community-transpiler`, `remorph` no longer requires its `sqlglot`-based transpiler and corresponding customizations.
This PR drops the corresponding code and related tests.


ericvergnaud:
@sundarshankar89 I might need your help to debug the failing integration tests

sundarshankar89:
@ericvergnaud Can you split this PR with code changes and then separate PR to remove the test examples SQL. 
That way it is easier to Review.

sundarshankar89:
The integration tests are failing as expected because we use custom dialects for our Reconcile target code generation and transformation.

ericvergnaud:
> The integration tests are failing as expected because we use custom dialects for our Reconcile target code generation and transformation.

I would expect this to also fail corresponding unit tests ?

sundarshankar89:
> > The integration tests are failing as expected because we use custom dialects for our Reconcile target code generation and transformation.
> 
> I would expect this to also fail corresponding unit tests ?

We had this conversation when this was all originally under unit and moved over to integration. few tests are unit and few really depend on spark infra.

ericvergnaud:
@sundarshankar89 I've brought back the custom dialects, si you should be able to validate this PR.

PR #1454, opened 2025-02-21
Prefer root uri in initialization

ericvergnaud:
As per LSP, rootUri is preferred over rootPath during initialization.
This PR fixes that and also tests loading a document at startup by the server.


sundarshankar89 (approved):
LGTM

PR #1455, opened 2025-02-23
Enhancement: Introduced mandatory password check for snowflake JDBC connection.

shsk:
PR: Introduced mandatory Password for snowflake connector when JDBC options are provided.

1. Added check for sfPassword while creating the JDBC URL for Snowflake connections.
2. If sfPassword is not found, it will fail with the following error 
`sfPassword is mandatory for jdbc connectivity with Snowflake.`

To test: Run recon with out sfPassword in Secret Scope along with jdbc_options.

Closes #issue-1449-mandatory_pwd_for_jdbc_options

### DIFF
--- a/src/databricks/labs/remorph/reconcile/connectors/snowflake.py
+++ b/src/databricks/labs/remorph/reconcile/connectors/snowflake.py
@@ -61,9 +61,16 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
 
     @property
     def get_jdbc_url(self) -> str:
+        try:
+            sf_password = self._get_secret('sfPassword')
+        except (NotFound, KeyError) as e:
+            message = "sfPassword is mandatory for jdbc connectivity with Snowflake."
+            logger.error(message)
+            raise NotFound(message) from e
+
         return (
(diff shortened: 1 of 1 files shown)


bishwajit-db (approved):
LGTM

src/databricks/labs/remorph/reconcile/connectors/snowflake.py
```diff
@@ -61,9 +61,16 @@ def __init__(
...
+            sf_password = self._get_secret('sfPassword')
+        except (NotFound, KeyError) as e:
+            message = "sfPassword is mandatory for jdbc connectivity with Snowflake."
+            logger.error(message)
+            raise NotFound(message) from e
+
         return (
             f"jdbc:{SnowflakeDataSource._DRIVER}://{self._get_secret('sfAccount')}.snowflakecomputing.com"
-            f"/?user={self._get_secret('sfUser')}&password={self._get_secret('sfPassword')}"
+            f"/?user={self._get_secret('sfUser')}&password={sf_password}"
```
sundarshankar89, line 73: it should be from secret only.

sundarshankar89 (changes requested):

sundarshankar89 (approved):
LGTM

PR #1456, opened 2025-02-24
Add support for DBT

ericvergnaud:
As part of supporting DBT, remorph needs to transpile not only SQL files but also dbt_project.yml files, that are used to populate DBT macros during transpilation.
This PR implements this change.


sundarshankar89 (approved):
LGTM
We need to work on some integration test for these. cc: @gueniai

sundarshankar89:
@ericvergnaud  can you resolve conflicts

jimidle (approved):
This PR is fine for looking at the base DBT project definition, but it will need to be extended as `.yml` files can be placed anywhere in the project hierarchy and can be called anything. Additionally, there are other files that may be seen in the project that are used by DBT but should be copied as is by the transpiler. There is an issue opened in Morpheus to track this, assigned to @ericvergnaud. So this can go in as is, but then needs to be expanded upon in a subsequent PR.

PR #1457, opened 2025-02-24
Transpiler logging improvements

asnare:
This PR updates some of the logging when transpiling, mainly to improve the readability and help with debugging. Changes include:

 - Remove the timer. (It doesn't work because we're using coroutines and therefore it always reports 0.00 seconds before anything has started.)
 - Debug logging for the command used to launch the LSP program. (This helps with debugging when the plugin doesn't work.)
 - Minor formatting fixes.
 - Improved debug/info logging during transpilation progress.
 - ~Final output uses the logging facility instead of `print()`.~ Left as-is, it seems this is expected.

### DIFF
--- a/src/databricks/labs/remorph/cli.py
+++ b/src/databricks/labs/remorph/cli.py
@@ -118,7 +118,7 @@ def transpile(
     status, errors = asyncio.run(do_transpile(ctx.workspace_client, engine, config))
 
     for error in errors:
-        logger.error(f"Transpile error: {error}")
+        print(str(error))
 
     logger.info(f"Finished transpilation: {status}")
 
--- a/src/databricks/labs/remorph/cli.py
+++ b/src/databricks/labs/remorph/cli.py
@@ -1,5 +1,4 @@
 import asyncio
-import json
 import os
 from pathlib import Path
 
--- a/src/databricks/labs/remorph/cli.py
+++ b/src/databricks/labs/remorph/cli.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 import os
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


src/databricks/labs/remorph/cli.py
```diff
@@ -119,9 +118,9 @@ def transpile(
     status, errors = asyncio.run(do_transpile(ctx.workspace_client, engine, config))
 
     for error in errors:
-        print(str(error))
+        logger.error(f"Transpile error: {error}")
```
ericvergnaud, line 121: IIRC the logging is done by do_transpile already
Now we want these errors in the console (this is the cli) without all the logged noise

ericvergnaud (changes requested):

src/databricks/labs/remorph/cli.py
```diff
@@ -121,7 +120,7 @@ def transpile(
     for error in errors:
         print(str(error))
 
-    print(json.dumps(status))
+    logger.info(f"Finished transpilation: {status}")
```
sundarshankar89, line 123: this is the part of table template in `labs.yml` this should have broken that template.

ericvergnaud (changes requested):

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -68,6 +68,7 @@ async def _process_one_file(
             w.write(transpile_result.transpiled_code)
             w.write("\n;\n")
 
+    logger.info(f"Processed file: {input_path} (success: {transpile_result.success_count}, errors: {len(error_list)})")
```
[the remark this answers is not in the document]
asnare: Hmm, looking at it a little closer it's unclear to me what `success_count` is really counting or means, so I've just removed it from the normal user-visible logging for now.

The (raw) transpile and validations results are now logged (at `DEBUG`) elsewhere.

ericvergnaud (approved):

sundarshankar89 (approved):
LGTM

PR #1471, opened 2025-03-04
Log lsp server stderr

ericvergnaud:
Currently we ignore stderr output from the LSP server.
This PR adds logging of LSP Server's stderr output, and a ~manual~ test for checking it.
Fixes https://github.com/databrickslabs/morpheus/issues/216


src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -229,6 +229,18 @@ def wrapper(params):
...
         return wrapper
 
+    async def start_io(self, cmd: str, *args, **kwargs):
+        await super().start_io(cmd, *args, **kwargs)
+        # forward stderr
+        task = asyncio.create_task(self.pipe_stderr())
+        self._async_tasks.append(task)
+
+    async def pipe_stderr(self):
+        while not self._stop_event.is_set():
```
asnare, line 239: This loop should also exit if you read a line that doesn't end in `\n`: that indicates EOF has been received. In general it would be okay for an LSP server to close stderr if it didn't plan to use it, and as things stand that would lead to a busy loop of infinite logging.

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -229,6 +229,18 @@ def wrapper(params):
...
+            data = await self._server.stderr.readline()
+            message = data.decode("utf-8").strip()
```
asnare, line 241: On EOF, an empty `bytes` object can be returned (to signal EOF): this shouldn't be logged.

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -229,6 +229,18 @@ def wrapper(params):
...
 
+    async def start_io(self, cmd: str, *args, **kwargs):
+        await super().start_io(cmd, *args, **kwargs)
+        # forward stderr
+        task = asyncio.create_task(self.pipe_stderr())
+        self._async_tasks.append(task)
+
+    async def pipe_stderr(self):
+        while not self._stop_event.is_set():
+            data = await self._server.stderr.readline()
```
asnare, line 240: On EOF, `.readline()` can return an empty `bytes` object (to signal EOF): this shouldn't be logged.

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: I believe that the server closing stderr would raise an exception on the client side, but it doesn't hurt. Done

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: Done

jimidle (approved):
It seems Andrew's change requests were addressed, so LGTM

sundarshankar89:
@asnare I will let you re-review this, since you got here first :)

tests/unit/transpiler/test_lsp_err.py
```diff
@@ -0,0 +1,33 @@
...
+from databricks.labs.blueprint.logger import install_logger
+from databricks.labs.remorph.config import TranspileConfig
+from databricks.labs.remorph.transpiler.lsp.lsp_engine import LSPEngine
+
+# this file's sole purpose is to test the LSP Engine's stderr pipe
+# it can't be ran as a unit test because pytest captures std i/o
+# so this script is a 'manual' test that can be run as required
+# it should display the following in red in the console:
+# Hello there! I'm the client!
+# 17:50:19 ERROR [d.l.r.t.lsp.lsp_engine] Running LSP Test Server...
```
asnare, line 14: Thanks for the explanation on why this is a manual test. That said, Pytest provides a facility for checking what has been written to stdout/stderr: https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function

Is there a reason we can't use this to make this part of the normal automated testing?

Although a manual is better than no test, my concern here is that a manual test:

 - won't be run as part of the normal CI/CD process, meaning regressions won't be spotted;
 - isn't visible to developers/maintainers, who probably won't realise there's a test to run if we notice problems in this area.

With this in mind I think it's worth really making sure that we can't automate it.

ericvergnaud: The above link explains how to capture stdout/stderr, which is the exact opposite of what we want for this test.
I think my explanation says it all, so not sure what other _reason we can't use this to make this part of the normal automated testing_ could be required ?
That said, if you can find a way to automate it, you're welcome.

asnare: @ericvergnaud: I'm unsure where the confusion seems to have arisen, but the link I supplied was exactly what you were looking for. If you'd used it, you would have seen that it captured the `Hello there! I'm the client!` message that you were looking for. There's also an associated mechanism for capturing logs, which can be used to capture the log output.

I've replaced the manual test with its automated equivalent. My changes boil down to:

 - Use `caplog` to capture the logs and verify the message we expect can be found.
 - Removing the `Hello there! I'm the client!` check: this was present only to help the manual testing.
 - Changing the log-level for stderr from `ERROR` to `INFO`: as I've noted, although the messages from the LSP server are written to stderr, they don't necessarily represent errors and are really informational in nature.

Given I've now authored these changes, please ask someone else to review this.

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -241,7 +241,9 @@ async def pipe_stderr(self):
             if not data:
                 return
             message = data.decode("utf-8").strip()
-            logger.error(message)
+            # Although information may arrive via stderr, it's generally informational in nature and doesn't necessarily
```
ericvergnaud, line 249: That's an incorrect statement

asnare: I'm not sure why you say that, but I will clarify what I mean to avoid misunderstanding: the fact that something is arriving via stderr says nothing about the severity level of the information, from the perspective of the remorph CLI. The purpose of stderr and the purpose of the `ERROR` logging level are unrelated, except by accident in the same sense that a broken clock is correct twice a day.

The output on stderr from the LSP server can be anything, including but not limited to:
 - line-based logging (itself with its own intended level of importance);
 - structured logging, where the same could apply.
 - Completely free-form, such as the startup message that we look for in the test.

The LSP server is simply using stderr because stdout is already taken and unavailable.

As such the closest logging level, if we must assign one at all, is `INFO` because we're simply passing on information from the LSP server to the user.

ericvergnaud: mmm... I think the original intent was to only send errors. Things seem to have drifted and, at least from Morpheus, we're sending all the logs. 
We'd need to write the raw message to the log directly. Otherwise, whichever channel we use, we'll end up with duplicate and misleading logs such as:
`INFO 2025-04-12-15:13:17 ERROR 2025-04-12-15:13:14 Something went wrong`
I'd suggest using a dedicated Remorph logger configured to not store metadata ?

ericvergnaud: FWIW the freeform startup message was for manual testing only, and proving that we were indeed receiving whatever the LSP server was sending to stderr.

PR #1480, opened 2025-03-07
Only install LSP add-ons if requested

ericvergnaud:
Our current code attempts to install LSP add-ons as part of installing Remorph. This could fail the entire install if not successful.
This PR ensures that installing LSP add-ons is only attempted when explicitly requested.


src/databricks/labs/remorph/install.py
```diff
@@ -261,8 +261,9 @@ def run(
         module: str,
         config: RemorphConfigs | None = None,
     ) -> RemorphConfigs:
-        self.install_rct()
-        self.install_morpheus()
+        if module in {"transpiler", "all"}:
```
sundarshankar89, line 264: I covered during the hotfix, I have handled it differently.

https://github.com/databrickslabs/remorph/pull/1476

ericvergnaud: Not sure re-running install tasks during config is good ? Do you not prefer the fix in this PR ?

sundarshankar89:
@ericvergnaud  Can we close I think I rebased this branch on to my hotfix branch

ericvergnaud:
Already merged, closing

PR #1485, opened 2025-03-10
Fix missing dependencies

ericvergnaud:
This PR:
 - adds `standard-distutils` conditional dependency. Starting with Python 3.11, `distutils` is no longer part of python, so needs to be installed.


pyproject.toml
```diff
@@ -14,6 +14,8 @@ classifiers = [
 ]
 dependencies = [
   "databricks-sdk>=0.29,<0.42",
+  "databricks-connect==15.1",
```
sundarshankar89, line 17: Need to understand why this change is needed, databricks-connect is dependant on cluster_id or warehouse_id, so we can't fix it to specific version, it is driven by `labs.yml` configuration hence it is a test only depadancy.

Also why is the standard-distuilts is tagged to python version above the base line 3.10 we support?

sundarshankar89 (changes requested):

pyproject.toml
ericvergnaud: `distutils` is part of python 3.10, but removed starting with 3.11. If running python > 3.10 it needs to be added.

ericvergnaud: `databricks-connect` is required when running the cli locally. Happy to remove it if it's in the way, but please provide a means to run the cli locally

sundarshankar89: labs.yml should have this, but I want to understand why is databricks-connect needed for transpile.

```
install:
  min_runtime_version: 13.3
  require_running_cluster: false
  require_databricks_connect: true
```

ericvergnaud: IIRC it's needed during `install-transpile` , because it loads all configs and `recon` related code

sundarshankar89: Thats why moving it as part of the configure instead of run will avoid this need.

`def configure(self, module: str) -> RemorphConfigs:`

ericvergnaud: Noted. Updated this PR accordingly.

ericvergnaud: @sundarshankar89 ready to merge

sundarshankar89 (approved):
LGTM

jimidle (approved):
LGTM

PR #1487, opened 2025-03-10
Configure lsp custom options

### LINKED ISSUE #1478: [FEATURE]: Let remorph manage options specific to a transpiler

Transpile

### Problem statement

Our current implementation supports passing `custom` configuration to the transpiler, but it has to be hardcoded  in the transpiler config file.
With RCT, there is only 1 option: 'experimental', which is True by default, so it's not penalizing.
With Morpheus, there is no option.
With BladeRunner, there are many options, and the set of valid ones depends on the selected dialect.
We need a way for remorph to ask the user for additional options once the dialect and transpiler are selected.

### Proposed Solution

This could be done by enhancing the LSP config file of each transpiler.
The selected options would be stored in the transpiler config, together with the mandatory ones.
They would then be passed to the transpiler as part of the `custom` section

### Additional Context

_No response_

ericvergnaud:
Each transpiler may require options that `remorph` doesn't know about, but need to be configured via the CLI before 
calling the transpiler.

This PR implements a mechanism for doing so, as follows:
 - the `TranspileConfig` structure now comprises a `transpiler_options` entry that can hold key/value pairs
 - the transpiler `config.yml` now comprises an optional `options` section which describes CLI actions for populating the `transpiler_options` above. 4 actions are supported: `FORCE`, `CONFIRM`, `QUESTION` and `CHOICE`. 
 - the CLI uses the above section to interact with the user and gather configuration options
 - the LSP client passes these options to the server as part of initialization

Fixes #1478

### DIFF
--- a/src/databricks/labs/remorph/config.py
+++ b/src/databricks/labs/remorph/config.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 
 class LSPPromptMethod(Enum):
-    FORCE = auto() # for mandatory values that are specific to a dialect
+    FORCE = auto()  # for mandatory values that are specific to a dialect
     QUESTION = auto()
     CHOICE = auto()
     CONFIRM = auto()
@@ -30,7 +30,7 @@ class LSPConfigOptionV1:
 
     @classmethod
     def parse_all(cls, data: dict[str, Any]) -> dict[str, list[LSPConfigOptionV1]]:
-        return { key: list(LSPConfigOptionV1.parse(item) for item in value) for (key, value) in data.items() }
+        return {key: list(LSPConfigOptionV1.parse(item) for item in value) for (key, value) in data.items()}
 
     @classmethod
     def parse(cls, data: Any) -> LSPConfigOptionV1:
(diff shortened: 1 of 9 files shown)


sundarshankar89 (commented):
This looks good, can we add an integration tests as a followup to run this E2E.

ericvergnaud:
> This looks good, can we add an integration tests as a followup to run this E2E.

There already is one, see https://github.com/databrickslabs/remorph/pull/1487/files#diff-8222ddc240f5203ac243870cb96b655cfee85b351485787fe25b556f8dafb68f:~:text=tests/unit/transpiler/test_lsp_engine.py

sundarshankar89:
> > This looks good, can we add an integration tests as a followup to run this E2E.
> 
> There already is one, see https://github.com/databrickslabs/remorph/pull/1487/files#diff-8222ddc240f5203ac243870cb96b655cfee85b351485787fe25b556f8dafb68f:~:text=tests/unit/transpiler/test_lsp_engine.py

I was talking about integration tests with actual side effects.

sundarshankar89 (approved):
LGTM, Approving this since that doesn't block this PR, but we need E2E integration test now we have env up and running.

ericvergnaud:
> > > This looks good, can we add an integration tests as a followup to run this E2E.
> > 
> > 
> > There already is one, see https://github.com/databrickslabs/remorph/pull/1487/files#diff-8222ddc240f5203ac243870cb96b655cfee85b351485787fe25b556f8dafb68f:~:text=tests/unit/transpiler/test_lsp_engine.py
> 
> I was talking about integration tests with actual side effects.

The side effect is in the logs generated by the LSP server. What other side effect do you have in mind ?

sundarshankar89:
> > > > This looks good, can we add an integration tests as a followup to run this E2E.
> > > 
> > > 
> > > There already is one, see https://github.com/databrickslabs/remorph/pull/1487/files#diff-8222ddc240f5203ac243870cb96b655cfee85b351485787fe25b556f8dafb68f:~:text=tests/unit/transpiler/test_lsp_engine.py
> > 
> > 
> > I was talking about integration tests with actual side effects.
> 
> The side effect is in the logs generated by the LSP server. What other side effect do you have in mind ?

Entire Installation journey, with state and configs.

PR #1484, opened 2025-03-10
Improve installer resiliency

ericvergnaud:
This PR includes #1480 
It fixes various issues where `install transpile` would crash if external requirements were not met, such as access to pypi and maven, or availability of required packages.

PR #1472, opened 2025-03-04
Pass log level of "databricks" logger to LSP server

ericvergnaud:
Or current code does not manage the log level of the LSP Server
This PR fixes #1464 by passing the current "databricks" log level to the LSP Server as part of command line

PR #1488, opened 2025-03-10
Make transpile config flexible

ericvergnaud:
Our current code requires various transpile config fields at `install` time, but requires them as parameters when running `transpile`.
This PR fixes the discrepancy by:
 - making all fields optional at `install` time
 - making all parameters optional at `transpile` time
 - checking the config completeness at `transpile` time
 
Builds on top of the following PRs:
#1491 
#1490 
#1487 
#1485 
#1484 
#1480 
#1472 
#1456 

Also requires https://github.com/databrickslabs/blueprint/pull/189

### DIFF
--- a/.github/scripts/setup_spark_remote.sh
+++ b/.github/scripts/setup_spark_remote.sh
@@ -5,7 +5,7 @@ set -xve
 mkdir -p "$HOME"/spark
 cd "$HOME"/spark || exit 1
 
-version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark- | tr -d / | sort -r --version-sort | head -1)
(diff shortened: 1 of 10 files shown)


src/databricks/labs/remorph/config.py
```diff
@@ -9,20 +13,59 @@
...
-    source_dialect: str
+    source_dialect: str | None = None
     input_source: str | None = None
     output_folder: str | None = None
     error_file_path: str | None = None
     sdk_config: dict[str, str] | None = None
     skip_validation: bool = False
     catalog_name: str = "remorph"
     schema_name: str = "transpiler"
+    transpiler_options: dict[str, Any] | None = None
```
asnare, line 68: What are the types of the values in this dictionary supposed to be?

asnare:
```
JSONValue: typing.TypeAlias = None | bool | int | float | str | list["JSONValue"] | dict[str, "JSONValue"]
```

PR #1490, opened 2025-03-11
Fix labs.yml

ericvergnaud:
Our current `labs.yml` uses the literal `None` for default values.
However, the labs installer is written in `go` and expects go literals only. As such it incorrectly interprets `None` as a string literal, leading to all sorts of issues...
This PR fixes the issue by replacing `None` with `null`.


sundarshankar89 (approved):
LGTM

jimidle (approved):
LGTM

PR #1491, opened 2025-03-11
Improve lsp resiliency

ericvergnaud:
Increase timeout and other minor improvements

### DIFF
--- a/src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
+++ b/src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
@@ -375,6 +375,11 @@ class LSPEngine(TranspileEngine):
         env = deepcopy(os.environ)
         for name, value in self._config.remorph.env_vars.items():
             env[name] = value
+        # ensure modules are searched locally before being searched in remorph
+        if "PYTHONPATH" in env.keys():
+            env["PYTHONPATH"] = str(self._workdir) + os.pathsep + env["PYTHONPATH"]
+        else:
+            env["PYTHONPATH"] = str(self._workdir)
         args = self._config.remorph.command_line[1:]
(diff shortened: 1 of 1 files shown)


src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
```diff
@@ -375,6 +375,11 @@ async def _do_initialize(self, config: TranspileConfig) -> None:
         env = deepcopy(os.environ)
         for name, value in self._config.remorph.env_vars.items():
             env[name] = value
+        # ensure modules are searched locally before being searched in remorph
```
sundarshankar89, line 378: Unfortunately, I don't follow this change, can you explain a bit more, we create a venv when remorph is installed, that is the Path we should be pointing to.

sundarshankar89 (commented):
LGTM

src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
ericvergnaud: we don't -yet- create a `venv` for the LSP server. We need to ensure that the LSP server's imports are resolved against its codebase rather than against `remorph`'s code base. The `PYTHONPATH` here is for the LSP Server only, it doesn't affect `remorph`.

ericvergnaud: Then we need to prioritize having independent installers per pluggable transpiler. I have created a ticket for this: https://github.com/databrickslabs/remorph/issues/1507.
In the meantime, this PR limits risks and side-effects.

sundarshankar89: Done

sundarshankar89 (approved):
LGTM

biswadeepupadhyay-db (approved):
LGTM

PR #1521, opened 2025-04-23
CI: verify checksum of Debian package before installing

asnare:
This PR updates the setup for the acceptance tests during CI by verifying the checksum of the downloaded package before we try to install it.

This ensures that we notice if the package is changed instead of just blindly installing something we've downloaded.


sundarshankar89 (approved):
LGTM

gueniai (approved):

PR #1523, opened 2025-04-24
Constrain dependency versions for project

asnare:
At present our project specifies some dependencies without any version restrictions. This means that future (unreleased) versions of the dependencies could be installed, and these might not be compatible with a released version of this project.

To help limit the risk here, this PR adds a version constraint to all dependencies that don't already have one specified. The version constraint in all cases is for those "compatible" with the latest release at the time of this PR. This assumes the dependencies themselves use semantic versioning, and is a little conservative, but the PyPi mechanism doesn't let us withdraw a release or update dependencies to exclude a problematic dependency so this is probably the safest thing to go with.


sundarshankar89 (approved):
LGTM

gueniai (approved):
LGTM

PR #1524, opened 2025-04-24
Improve `Makefile` project setup

asnare:
This PR updates the `dev` target in our `Makefile`:

 - It no longer tries to install hatch if it's already installed.
 - When invoking pip, it uses `pip3` instead of `pip`: the latter is more reliably available than `pip`, which is often skipped.

Tested with:

 - dash
 - bash
 - zsh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,10 @@ setup_python:
 
 
 dev:
-	pip install hatch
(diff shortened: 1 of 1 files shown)


sundarshankar89 (approved):
LGTM.

gueniai (approved):
LGTM

PR #1527, opened 2025-04-28
Patched Reconcile Install

sundarshankar89:
Patches wheels context generation usage while upgrades.
`AttributeError: 'WheelsV2' object has no attribute '_local_wheel'`


bishwajit-db (approved):
LGTM

biswadeepupadhyay-db (approved):
LGTM

sundarshankar89:
> The fix seems unrelated to the PR main comment... Maybe change it to: "Upload wheels 1 by 1 to avoid ..." ?

Object is expected to use as context manager inside blueprint, so we are fixing this.

sundarshankar89:
> > The fix seems unrelated to the PR main comment... Maybe change it to: "Upload wheels 1 by 1 to avoid ..." ?
> 
> Object is expected to use as context manager inside blueprint, so we are fixing this or else _local_wheel doesn't get initialized from our singleton.

ericvergnaud (approved):
I still don't understand this change, but not worth bocking

PR #1529, opened 2025-04-28
Update transpiler configurations to use a dictionary to specify the environment

asnare:
This PR updates the way environment variables are specified in our transpiler configurations. Specifically the configuration now takes a mapping of environment variables, rather than a list of mappings that are merged during load.

Although an incompatible change:

 - Our transpiler configurations don't use this part of the configuration file;
 - We haven't yet released.

As such I think it's worth the benefit of tidying this up before we make a release.

### DIFF
--- a/docs/remorph/docs/transpile_guide/pluggable_transpilers.mdx
+++ b/docs/remorph/docs/transpile_guide/pluggable_transpilers.mdx
@@ -50,10 +50,10 @@ remorph:
     - ...
     - <sql dialect _n_>
   environment: # this section is optional, variables are set prior to launching the transpiler
-    - <name 1>: <value 1>
-    - <name 2>: <value 2>
-    - ...
-    - <name _n_>: <value _n_>
+    <name 1>: <value 1>
+    <name 2>: <value 2>
+    ...
+    <name _n_>: <value _n_>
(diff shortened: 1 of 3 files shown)


ericvergnaud:
Please also check and update if necessary:
 - docs
 - plugin starter
 - RCT
 - Morpheus
 - Bladerunner

asnare:
> Please also check and update if necessary:
> 
> * docs
> * plugin starter
> * RCT
> * Morpheus
> * Bladerunner

This PR already includes the necessary documentation changes, I think? Is there documentation somewhere else that I've missed? (The instructions in the Morpheus project do not need updating.)

As of right now, the Remorph CLI embeds the configuration files for RCT and Morpheus: these do not make use of environment variables, and therefore do not need updating.

Currently the plugin starter does not contain any configuration files (or relevant instructions) so there are no updates required there. The same applies for Bladerunner, as far as I can tell. (Please point me in the right direction if I've missed something.)

ericvergnaud:
Thanks for checking. There are some upcoming config files in some of these repos but as of writing they're still in draft branches

ericvergnaud (approved):

sundarshankar89 (approved):
LGTM

PR #1550, opened 2025-05-01
Mark `test_execute_query_without_connection` as allowed to fail locally if ODBC isn't available

### LINKED ISSUE #1549: [BUG]: Test failure if system ODBC libraries are not installed: `test_execute_query_without_connection`

### Category of Bug / Issue

Other

### Current Behavior

The `test_execute_query_without_connection` unit test fails if the system libraries for ODBC are not unavailable. This is an undocumented project requirement, with an `ImportError` being triggered if the native libraries cannot be found.

### Expected Behavior

This test should be skipped if:
1. Not running under the CI environment;
2. The ODBC libraries are unavailable.

Alternatively, the project setup/requirements for local development should be updated. (Installing the ODBC libraries is not within the scope of hatch/pip: they are limited to installing Python libraries in the local virtual environment.)

### Steps To Reproduce

1. On a macOS, verify that `unixodbc` is _not_ installed.
2. Run the project tests:
   ```bash
   HATCH_PYTHON=$(which python3.10) make clean dev test
   ```
3. The `test_execute_query_without_connection` test will fail with an `ImportError` because `pyodbc` cannot load the native library.

```shell
FAILED
tests/unit/connections/test_database_manager.py:47 (test_execute_query_without_connection)
def test_execute_query_without_connection():
>       db_manager = DatabaseManager("mssql", sample_config)

tests/unit/connections/test_database_manager.py:49: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/databricks/labs/remorph/connections/database_manager.py:83: in __init__
src/databricks/labs/remorph/connections/database_manager.py:54: in _create_connector
    return connector_class(config)
src/databricks/labs/remorph/connections/database_manager.py:28: in __init__
    self.engine: Engine = self._connect()
src/databricks/labs/remorph/connections/database_manager.py:78: in _connect
<string>:2: in create_engine
    ???
.venv/lib/python3.10/site-packages/sqlalchemy/util/deprecations.py:281: in warned
    return fn(*args, **kwargs)  # type: ignore[no-any-return]
.venv/lib/python3.10/site-packages/sqlalchemy/engine/create.py:602: in create_engine
    dbapi = dbapi_meth(**dbapi_args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'sqlalchemy.dialects.mssql.pyodbc.MSDialect_pyodbc'>

    @classmethod
    def import_dbapi(cls) -> ModuleType:
>       return __import__("pyodbc")
E       ImportError: dlopen(/Users/**REDACTED**/dev/remorph-2/.venv/lib/python3.10/site-packages/pyodbc.cpython-310-darwin.so, 0x0002): Library not loaded: /usr/local/opt/unixodbc/lib/libodbc.2.dylib
E         Referenced from: <13D74168-289F-36F6-8E48-C5C94BC764C8> /Users/**REDACTED**/dev/remorph-2/.venv/lib/python3.10/site-packages/pyodbc.cpython-310-darwin.so
E         Reason: tried: '/usr/local/opt/unixodbc/lib/libodbc.2.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/opt/unixodbc/lib/libodbc.2.dylib' (no such file), '/usr/local/opt/unixodbc/lib/libodbc.2.dylib' (no such file), '/usr/local/lib/libodbc.2.dylib' (no such file), '/usr/lib/libodbc.2.dylib' (no such file, not in dyld cache)

.venv/lib/python3.10/site-packages/sqlalchemy/connectors/pyodbc.py:60: ImportError
```

### Operating System

macOS

### Version

latest via Databricks CLI

asnare:
This PR resolves #1549 by allowing the `test_execute_query_without_connection` test as expected to fail, but only locally and when ODBC isn't available.

There's currently an undocumented platform dependency that requires the system have native ODBC libraries installed for this test to pass. This cannot be satisfied by Hatch, and rather than complicate setup instructions, this PR marks the test as expected to fail, but only:

 - If running locally; and
 - The `pyodbc` module cannot be imported.

Note that on CI this test must pass, and currently does because the CI environment sets up ODBC before running the tests.

### DIFF
--- a/tests/unit/connections/test_database_manager.py
+++ b/tests/unit/connections/test_database_manager.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 from unittest.mock import MagicMock, patch
 from databricks.labs.remorph.connections.database_manager import DatabaseManager
@@ -45,6 +46,30 @@ def test_execute_query(mock_mssql_connector):
     mock_connector_instance.execute_query.assert_called_once_with(query)
 
 
+def running_on_ci() -> bool:
+    """Return True if the tests are running within a CI environment."""
+    env_vars = {"CI", "BUILD_NUMBER"}
+    return any(var in os.environ for var in env_vars)
+
+
+def odbc_available() -> bool:
+    """Return True of the ODBC driver is available."""
+    try:
+        import pyodbc  # type: ignore[import]
+
+        assert pyodbc.version is not None
+
+        return True
+    except ImportError:
+        return False
+
+
+# Only allow this to be skipped during local development: the CI environment must be set up to run this test.
+@pytest.mark.xfail(
+    not running_on_ci() and not odbc_available(),
+    reason="This test needs native ODBC libraries to be installed.",
+    raises=ImportError,
+)
 def test_execute_query_without_connection():
     db_manager = DatabaseManager("mssql", sample_config)
 


sundarshankar89 (approved):
LGTM

asnare:
> Interesting thing is in CI we don't install any thing additional apart from `pyodbc` itself, there seem to some pre-provisioned in the CI runners which dosesn't allow this tests to fail in CI.

I assumed it was this: https://github.com/databrickslabs/remorph/blob/4e6fbf5dcf8795fc2b98938828888bf73a045309/.github/workflows/acceptance.yml#L47

That said, the failing test is a unit test and it runs from other workflows: it's a bit of a mystery why it works there. Maybe that needs investigating, but I don't think it affects the scope or purpose of this PR which is limited to local development.

sundarshankar89:
> > Interesting thing is in CI we don't install any thing additional apart from `pyodbc` itself, there seem to some pre-provisioned in the CI runners which dosesn't allow this tests to fail in CI.
> 
> I assumed it was this:
> 
> https://github.com/databrickslabs/remorph/blob/4e6fbf5dcf8795fc2b98938828888bf73a045309/.github/workflows/acceptance.yml#L47
> 
> That said, the failing test is a unit test and it runs from other workflows: it's a bit of a mystery why it works there. Maybe that needs investigating, but I don't think it affects the scope or purpose of this PR which is limited to local development.

Agreed.

jimidle (approved):
LGTM

PR #1551, opened 2025-05-01
Constrain the Databricks SDK to a single version

asnare:
As part of #1523 most dependencies were constrained to the specific version currently in use. Unfortunately I overlooked the version for `databricks-sdk` and left it as a wide range rather than pinning it to the latest release (or a patch-release thereof).

This PR fixes this by pinning the `databricks-sdk` to a single version, plus any patch releases that are made for it.


sundarshankar89 (approved):
LGTM

jimidle (approved):
LGTM

PR #1566, opened 2025-05-08
Fix spark version detection in CI integration 

ericvergnaud:
Our CI suddenly fails because the spark download page has new entries that our version extractor does not properly parse.
See https://github.com/databrickslabs/remorph/actions/runs/14901901650/job/41855507048
This PR fixes that.

### DIFF
--- a/.github/scripts/setup_spark_remote.sh
+++ b/.github/scripts/setup_spark_remote.sh
@@ -5,7 +5,7 @@ set -xve
 mkdir -p "$HOME"/spark
 cd "$HOME"/spark || exit 1
 
-version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark- | tr -d / | sort -r --version-sort | head -1)
+version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | rev | cut -d - -f 1 - | tr -d / | rev | sort -r --version-sort | head -1)
 if [ -z "$version" ]; then
   echo "Failed to extract Spark version"
    exit 1
--- a/.github/scripts/setup_spark_remote.sh
+++ b/.github/scripts/setup_spark_remote.sh
@@ -5,7 +5,7 @@ set -xve
 mkdir -p "$HOME"/spark
 cd "$HOME"/spark || exit 1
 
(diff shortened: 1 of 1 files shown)


.github/scripts/setup_spark_remote.sh
```diff
@@ -5,7 +5,7 @@ set -xve
 mkdir -p "$HOME"/spark
 cd "$HOME"/spark || exit 1
 
-version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark- | tr -d / | sort -r --version-sort | head -1)
+version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | rev | cut -d - -f 1 - | tr -d / | rev | sort -r --version-sort | head -1)
```
asnare, line 8: The problem is that non-spark things have been published, and they're not being filtered out. The intention is that when there are several spark releases available, the latest will be chosen. This change works with the current listing, but will break as soon as anything else in the listing has a higher version. (For example: `spark-connect-swift-10.3`).

Maybe let's address the underlying issue by only looking at the spark releases, with something like:
```suggestion
version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark-[0-9.]*/"'  | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark-/ | sort -r --version-sort | head -1)
```

asnare (changes requested):

.github/scripts/setup_spark_remote.sh
```diff
@@ -5,7 +5,7 @@ set -xve
 mkdir -p "$HOME"/spark
 cd "$HOME"/spark || exit 1
 
-version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark' | grep -v 'preview' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | rev | cut -d - -f 1 - | tr -d / | rev | sort -r --version-sort | head -1)
+version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark-[0-9.]*/"'  | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark-/ | sort -r --version-sort | head -1)
```
asnare, line 8: Sigh, I guess it needs to be:
```suggestion
version=$(wget -O - https://dlcdn.apache.org/spark/ | grep 'href="spark-[0-9.]*/"' | sed 's:</a>:\n:g' | sed -n 's/.*>//p' | tr -d spark/- | sort -r --version-sort | head -1)
```

asnare (approved):

gueniai (approved):

PR #1568, opened 2025-05-08
Ensure the Remorph CLI provides product information to the pluggable transpilers during initialisation.

### LINKED ISSUE #1553: [FEATURE]: Remorph should fill in client name and version when initialising transpiler plugins

Transpile

### Problem statement

The LSP protocol allows the client to indicate its name and version during the initialisation handshake, but at present the Remorph CLI does not include this information. Aside from the diagnostics benefit, this sort of information may be needed as a "break glass" mechanism to deal with issues further down the track if a pluggable transpiler needs to alter its behaviour based on the version of Remorph that it's running under.

### Proposed Solution

Update the CLI to use the blueprint mechanism for determining its name and version, and include this in the in LSP initialisation sequence.

asnare:
This PR updates the LSP engine so that the product information is supplied to the pluggable transpilers during initialisation.

Name and version information are provided using the normal blueprint manner, and are currently `remorph` and `0.9.0` respectively.

Resolves #1553.

### DIFF
--- a/src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
+++ b/src/databricks/labs/remorph/transpiler/lsp/lsp_engine.py
@@ -388,6 +388,7 @@ class LSPEngine(TranspileEngine):
         root_path = input_path if input_path.is_dir() else input_path.parent
         params = InitializeParams(
             capabilities=self._client_capabilities(),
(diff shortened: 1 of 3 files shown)


ericvergnaud (approved):

sundarshankar89 (approved):
LGTM

PR #1569, opened 2025-05-08
Ensure test LSP server is shut down when tests fail

asnare:
## Changes

This PR updates the fixture for some of our LSP engine tests to ensure that the underlying LSP server (and its process) is shut down properly even if the test fails.

### Tests

- updated unit tests


gueniai (approved):

sundarshankar89 (approved):
LGTM

PR #1571, opened 2025-05-08
Wait for transpile capability during LSP initialisation

asnare:
## Changes

This PR updates the initialisation sequence when starting up the LSP server: initialisation now waits until it knows the transpiler capability is available before returning.

### Relevant implementation details

Currently we are using LSPs dynamic registration mechanism for the server (=transpiler plugin), after the initialisation sequence, to notify the client (=remorph CLI) that transpilation is supported. This normally happens quickly, but previously the first call to transpile would wait up to 10 seconds for the capability to be declared. As noted on #1542, this meant that a failure that prevented it from arriving would be noted on the call to transpile, which is a little misleading because something has really gone wrong during startup.

### Caveats/things to watch out for when reviewing:

We don't currently have any way to verify the sequence of LSP interactions at a fine-grained level: this would require mocking the LSP part of the stack, and be a significant amount of work. Instead the approach until now is to use a small test LSP server. Testing for this PR is therefore a little racy: we can't strictly verify that initialisation waits for the capability, because it might occur after initialisation but before the test performs its check. A passing test therefore doesn't _prove_ that it works in the strict sense, although a failure (including flakiness) would demonstrate that it's not working as intended.

### Linked issues

Resolves #1542.

### Tests

- update tests

PR #1574, opened 2025-05-12
LSP: include the client's process ID during the initialisation sequence

asnare:
## Changes

This PR updates the LSP initialisation sequence for pluggable transpilers so that the client (=Remorph CLI) notifies the server (=pluggable transpiler) of its process ID. The [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize) mentions this:
```typescript
/**
 * The process Id of the parent process that started the server. Is null if
 * the process has not been started by another process. If the parent
 * process is not alive then the server should exit (see exit notification)
 * its process.
 */
```

The intent here is that sometimes a server cannot rely on orderly shutdown of the LSP server (including the LSP messages or stdin/stdout being closed) to detect a client that has crashed, and a mitigation here is that the server should periodically check that this process is still present.

There's a line from another test removed: this is from a test added in #1571, and #1569 would have removed the line but both of those PRs were merged in parallel.

### Functionality

 - updated initialisation sequence during transpile requests.

### Tests

 - added tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
 ]
 dependencies = [
-  "databricks-sdk~=0.44.0",
+  "databricks-sdk~=0.51",
   "standard-distutils~=3.11.9; python_version>='3.11'",
   "sqlglot==26.1.3",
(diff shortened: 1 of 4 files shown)


sundarshankar89 (approved):
LGTM

ericvergnaud (approved):

PR #1582, opened 2025-05-14
Improve CLI 'transpile' UX

ericvergnaud:
In order to transpile code, the user must call:
`databricks labs remorph transpile ...`
Our current implementation requires the user to pass all `transpile` config options as parameters to the above command, which is painful, and leads to wonder what's the point in configuring the transpiler...
PR #1488 makes it less painful by making all parameters optional, and using their configured counterpart if not provided.
However, if the resulting config is inconsistent, it will fail, which can be frustrating.
This PR goes a step further by prompting the user if a config value is missing or inconsistent.

(not tested yet)

Requires #1488

### DIFF
--- a/src/databricks/labs/remorph/cli.py
+++ b/src/databricks/labs/remorph/cli.py
@@ -5,19 +5,20 @@ import os
 import time
 import sys
 from pathlib import Path
+from typing import cast
 
 from databricks.labs.blueprint.cli import App
 from databricks.labs.blueprint.entrypoint import get_logger
 
 from databricks.labs.remorph.assessments.configure_assessment import ConfigureAssessment
-from databricks.labs.remorph.config import TranspileConfig
+from databricks.labs.remorph.config import TranspileConfig, LSPConfigOptionV1, JsonType
 from databricks.labs.remorph.connections.credential_manager import create_credential_manager
 from databricks.labs.remorph.connections.env_getter import EnvGetter
 from databricks.labs.remorph.contexts.application import ApplicationContext
 from databricks.labs.blueprint.tui import Prompts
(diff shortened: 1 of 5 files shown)


tests/unit/conftest.py
```diff
@@ -220,7 +220,7 @@ def path_to_resource(*args: str) -> str:
 
 
 @pytest.fixture
-def mock_workspace_client_cli():
+def mock_workspace_client():
```
ericvergnaud, line 223: renamed for clarity, it's not a cli

tests/unit/test_cli_transpile.py
```diff
@@ -36,26 +41,30 @@ def test_transpile_with_missing_installation():
         )
 
 
-def patch_do_transpile():
+@pytest.fixture
+def mock_cli_for_transpile(mock_workspace_client):
```
ericvergnaud, line 45: factorize test code

src/databricks/labs/remorph/cli.py
```diff
@@ -107,70 +113,175 @@ def transpile(
...
+    checker.check_output_folder(output_folder)
+    checker.check_error_file_path(error_file_path)
+    checker.check_skip_validation(skip_validation)
+    checker.check_catalog_name(catalog_name)
+    checker.check_schema_name(schema_name)
     config, engine = checker.check()
     asyncio.run(_transpile(ctx, config, engine))
 
 
 class _TranspileConfigChecker:
```
sundarshankar89, line 131: nit: I like this to moved inside a singleton Application Context!!! to keep the cli.py small.
non blocking

sundarshankar89 (approved):
LGTM, small non blocking nits:

PR #1586, opened 2025-05-16
Improve the output in case of errors

vil1:
## Changes

Add a header to the produced files that states from which input they were transpiled, and in case of error, summarize the errors. Errors are displayed with a position (line:column) in the output, which gets decorated with line numbers.

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] modified existing command: `databricks labs remorph ...`
- [ ] ... +add your own

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/remorph/transpiler/execute.py
+++ b/src/databricks/labs/remorph/transpiler/execute.py
@@ -1,6 +1,7 @@
 import asyncio
 import datetime
 import logging
+import math
 from pathlib import Path
 from typing import cast, Any
 
@@ -64,10 +65,9 @@ async def _process_one_file(
     if error_list:
         with_line_numbers = ""
         lines = output_code.split("\n")
-        line_number = 1
-        for line in lines:
-            with_line_numbers += f"/* {line_number:4d} */  {line}\n"
-            line_number += 1
+        line_number_width = math.floor(math.log(len(lines), 10)) + 1
+        for line_number, line in enumerate(lines, start=1):
+            with_line_numbers += f"/* {line_number:{line_number_width}d} */  {line}\n"
         output_code = with_line_numbers
 
     if validator and not error_list:
--- a/src/databricks/labs/remorph/transpiler/execute.py
+++ b/src/databricks/labs/remorph/transpiler/execute.py
@@ -100,16 +100,20 @@ def _make_header(file_path: Path, errors: list[TranspileError]) -> str:
     header = f"/*\n    Transpiled from {file_path}\n"
     failed_producing_output = False
     if errors:
-        header += "\n    Following errors were found while transpiling:\n"
+        header += "\n    The following errors were found while transpiling:\n"
         for diag in errors:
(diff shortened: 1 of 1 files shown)


src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -56,39 +56,72 @@ async def _process_one_file(
...
 
     logger.info(f"Processed file: {input_path} (errors: {len(error_list)})")
     return transpile_result.success_count, error_list
 
 
+def _make_header(file_path: Path, errors: list[TranspileError]) -> str:
+    header = f"/*\n    Transpiled from {file_path}\n"
+    failed_producing_output = False
+    if errors:
+        header += "\n    Following errors were found while transpiling:\n"
```
jimidle, line 103: Better English:

`The following errors....`

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -56,39 +56,72 @@ async def _process_one_file(
...
+        for diag in errors:
+            if diag.range:
+                header += f"      - [{diag.range.start.line + 1}:{diag.range.start.character + 1}] {diag.message}\n"
+            else:
+                header += f"      - {diag.message}\n"
+            failed_producing_output = failed_producing_output or diag.kind == ErrorKind.PARSING
+
+    if failed_producing_output:
+        header += "\n\n    Parsing errors prevented the converter to translate the input query.\n"
```
jimidle, line 112: Better English:

`...the converter from translating the input query.`

jimidle (approved):
LGTM but with some corrections to the English grammar

src/databricks/labs/remorph/transpiler/execute.py
```diff
@@ -56,39 +56,76 @@ async def _process_one_file(
 
     output_path.parent.mkdir(parents=True, exist_ok=True)
 
+    output_code = transpile_result.transpiled_code
+
+    if any(err.kind == ErrorKind.PARSING for err in error_list):
```
sundarshankar89, line 61: How do we handle ErrorKind.Internal?  the example of Recursive CTE?

vil1: That is a good question. The `if any(...)` above detects cases where the output code is known to be incomplete (due to parse errors), which would therefore be useless to the user. In which case we display the input unchanged instead. 

Other kinds of errors are dealt with the "regular" way, with the output displayed under the header. Are there other cases (apart from parsing errors) where we should display the input?

ericvergnaud: I would (somewhat hypocritically) convert them to something like "Your source code relies on features not yet supported by Lakebridge. Contact support for help."

sundarshankar89 (approved):
My comments are non blocking

PR #1592, opened 2025-05-23
Remove sqlglot dependency for count queries

ericvergnaud:
As part of our roadmap, we want to remove our dependency on sqlglot.
This is a step towards that.


ericvergnaud:
superseded by #1696

PR #1609, opened 2025-05-28
Modify Analyzer CLI

sundarshankar89:
### What does this PR do? 

Modified the current CLI for `analyze` to use arguments aka flags instead of prompts. 

### Functionality

- modified existing command: `databricks labs remorph analyze`

### Tests

- updated unit tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
   "standard-distutils~=3.11.9; python_version>='3.11'",
   "databricks-bb-analyzer~=0.1.4",
   "sqlglot==26.1.3",
-  "databricks-labs-blueprint[yaml]>=0.10.1",
+  "databricks-labs-blueprint[yaml]>=0.11.0,<0.12.0",
   "databricks-labs-lsql==0.16.0",
   "cryptography>=44.0.2,<45.1.0",
   "pyodbc~=5.2.0",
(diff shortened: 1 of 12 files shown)


labs.yml
```diff
@@ -10,6 +10,11 @@ min_python: 3.10
 commands:
   - name: analyze
     description: Analyze source code from existing technology
+    flags:
+      - name: source-directory
+        description: Source Directory to Analyze
+      - name: output-file-name
+        description: Output File Name for Analysis Report, including full path, includes file extension(.xlsx)
```
asnare, line 17: What do you think about this?
```suggestion
      - name: report-file
        description: Name of (.xlsx) report file to write
```
 - `report`: maybe more intuitive?
 - No `-name`: redundant for a command argument.

asnare (approved):
Just manually tested, and it worked for me.

I'd maybe think a bit about the name of the flag (see the comment) but that's not blocking.

Other observations:
 - We should do something in general to make the UX a bit nicer if a mandatory argument is not provided. Currently the error message is a bit cryptic:
   ```
   TypeError: analyze() missing 2 required positional arguments: 'source_directory' and 'output_file_name'
   ```
   That's a general issue though, and I only mention it here to aid with awareness.
 - Maybe the source-tech can also be an option?
 - Should we maybe prompt if the option is omitted instead of failing?

labs.yml
sundarshankar89: Done

PR #1611, opened 2025-05-28
Update maven installer to not use `search.maven.org`

### LINKED ISSUE #1602: [BUG]: Don't use `search.maven.org` when installing morpheus

### Category of Bug / Issue

Other

### Current Behavior

Currently when installing maven-based transpilers (eg. morpheus) we use `search.maven.org` to locate the latest version. If unavailable, installation fails.

The [availability of this service](https://status.maven.org) over the last 90 days is ~95% meaning that on average ~1/20 installs will fail. (The service is described on the status page as "legacy".) Given this and the fact that this service is intended for interactive use by humans (rather than as an API) we should not be using it.

### Expected Behavior

We don't need to use or rely on Maven's search service. If we need to know the latest version we should use the artefact metadata (`maven-metadata.xml`) to figure this out.

Documentation can be found [here](https://maven.apache.org/repositories/metadata.html), and an example [here](https://repo.maven.apache.org/maven2/org/apache/spark/spark-core_2.13/maven-metadata.xml).

Some versions in the metadata are labelled. If present, we should use (in order of priority):

 - `release`
 - `latest`

Otherwise it's the highest (version-sorted) listed version.

### Steps To Reproduce

1. To simulate `search.maven.org` unavailability:
   ```sh
   printf '0.0.0.0 search.maven.org\n' >> /etc/hosts
   ```
2. Work around a potential (and unrelated) installer bug:
   ```sh
   mkdir -p ~/.databricks/labs/remorph-transpilers/remorph-community-transpiler
   ```
2. Try to install the transpilers:
   ```sh
   databricks labs remorph install-transpile
   ```
3. Observe failure to install:
   ```
   11:40:17    ERROR [d.l.remorph.install-transpile] URLError: <urlopen error [Errno 61] Connection refused>
   ```
   (If the API is reachable but down, this can be different.)
5. Don't forget to remove the `search.maven.org` line from your `/etc/hosts` file.

```shell
11:38:49    ERROR [d.l.remorph.install] Error while fetching maven metadata: com.databricks.labs:databricks-morph-plugin
Traceback (most recent call last):
  File "**REDACTED**/remorph/src/databricks/labs/remorph/install.py", line 79, in get_maven_version
    with request.urlopen(url) as server:
  File "**REDACTED**/python/lib/python3.10/urllib/request.py", line 216, in urlopen
    return opener.open(url, data, timeout)
  File "**REDACTED**/python/lib/python3.10/urllib/request.py", line 525, in open
    response = meth(req, response)
  File "**REDACTED**/python/lib/python3.10/urllib/request.py", line 634, in http_response
    response = self.parent.error(
  File "**REDACTED**/python/lib/python3.10/urllib/request.py", line 563, in error
    return self._call_chain(*args)
  File "**REDACTED**/python/lib/python3.10/urllib/request.py", line 496, in _call_chain
    result = func(*args)
  File "**REDACTED**/python/lib/python3.10/urllib/request.py", line 643, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 504: Gateway Time-out
```

### Operating System

All

### Version

`main` via `databricks labs install .`

asnare:
### What does this PR do?

This PR updates `MavenInstaller` so that it no longer uses `search.maven.org` to either:

 - Locate the current version of an artefact; or
 - Download the actual artefact.

The installer now fetches the metadata needed for detecting the version, as well as the artefact itself, from the maven central repository directly. This is the CDN-backed service where artefacts are supposed to be fetched from, and it is much more reliable than the search service.

In addition to this, the installer had some logic that could abort the installation if a file already exists at the target location on the local filesystem. This check now happens before the download so that it aborts immediately rather than first performing the (ultimately futile) download.

In addition to using the repository directly, this PR starts plumbing through support for providing the artefact classifier. ~Another PR will be needed to:~

~1. Expose the classifier as a parameter.~
~2. Update `install_morpheus()` to provide the classifier.~

~(Without these the normal installer will fail even after we publish morpheus because we're not currently specifying the classifier.)~

[It looks like we're not publishing with a classifier, so this isn't needed.]

### Linked issues

Resolves #1602 

### Functionality

### Tests

- updated integration tests

### DIFF
--- a/src/databricks/labs/remorph/install.py
+++ b/src/databricks/labs/remorph/install.py
@@ -14,6 +14,7 @@ from urllib.error import URLError, HTTPError
 import webbrowser
 from datetime import datetime, timezone
 from pathlib import Path
+import xml.etree.ElementTree as ET
 from zipfile import ZipFile
 
 from databricks.labs.blueprint.installation import Installation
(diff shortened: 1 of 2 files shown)


sundarshankar89 (approved):
LGTM

sundarshankar89 (changes requested):
@asnare while testing this manually this worked correctly by identifying the new release as 0.3.0 unlike the old search but 
os.rename(self._product_path, backup_path) failed with a non empty directory error.

gueniai (approved):
LGTM

PR #1615, opened 2025-05-30
Fix timestamp in morpheus transpiler state file

asnare:
### What does this PR do? 

This PR fixes a bug in the morpheus installer whereby the state-file that it writes has a timestamp that isn't in the expected format: it should include timezone information.

The same issue was caught on the PyPi side during review of the initial implementation but was missed on the Maven side. This PR factors out the code so that it's shared between the implementations, and also adds a unit-test to verify it.

To deal with some type-checking issues, this PR also removes the `_current_version` and `_latest_version` properties from the installers. It was difficult to reason about whether they had been set properly (or were maybe `None`) so plumbing them through the call-chain lets the type-checker eliminate state-related bugs in this area.

### Functionality

### Tests

- manually tested
- added unit tests
- existing integration tests

### DIFF
--- a/src/databricks/labs/remorph/install.py
+++ b/src/databricks/labs/remorph/install.py
@@ -2,7 +2,7 @@ import abc
 import dataclasses
 import shutil
 from collections.abc import Iterable
-from json import loads, dumps
+from json import loads, dump
 import logging
 import os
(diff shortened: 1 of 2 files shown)


gueniai (approved):
LGTM

PR #1616, opened 2025-05-30
Drop post-install support for Java-based transpilers

asnare:
### What does this PR do? 

This PR updates the maven-based installer so that:

1. It no longer copies all `lsp/` resources from the JAR into the transpiler directory;
2. It no longer supports running `lsp/install.sh` after the install.
3. It fails immediately instead of proceeding if the `config.yml` file is missing from the JAR.

This PR does not affect the PyPi-based installer for BB.

Aside from the above behaviour changes, some type-related changes are present:

 - Additional type-hints in a few places.
 - Paths are passed around as-is for longer, with conversion to string only when necessary.
 - Some methods now indicate success with a boolean rather than using an integer.

### Linked issues

This is stacked on top of #1615, which should be dealt with first.

### Functionality

- modified existing command: `databricks labs remorph install-transpiler`

### Tests

- manually tested
- existing unit tests
- existing integration tests

### DIFF
--- a/src/databricks/labs/remorph/install.py
+++ b/src/databricks/labs/remorph/install.py
@@ -613,10 +613,10 @@ class WorkspaceInstaller:
         self._save_config(config)
         return config
 
-    def _all_installed_dialects(self):
+    def _all_installed_dialects(self) -> list[str]:
         return sorted(TranspilerInstaller.all_dialects())
 
-    def _transpilers_with_dialect(self, dialect: str):
+    def _transpilers_with_dialect(self, dialect: str) -> list[str]:
         return sorted(TranspilerInstaller.transpilers_with_dialect(dialect))
 
     def _transpiler_config_path(self, transpiler: str) -> Path:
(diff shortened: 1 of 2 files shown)


src/databricks/labs/remorph/install.py
```diff
@@ -360,21 +360,21 @@ def download_artifact_from_maven(
         target: Path,
         classifier: str | None = None,
         extension: str = "jar",
-    ) -> int:
+    ) -> bool:
         if target.exists():
             logger.warning(f"Skipping download of {group_id}:{artifact_id}:{version}; target already exists: {target}")
```
sundarshankar89, line 365: nit: if we are doing this we need an option for force install or update.

sundarshankar89 (approved):
LGTM

gueniai (approved):

src/databricks/labs/remorph/install.py
asnare: Yeah, I agree. There needs to be a downgrade path, and right now that's basically impossible.

PR #1622, opened 2025-06-03
Add a test for checking transpilation of dbt sub dbt_project.yml files

ericvergnaud:
Our current code supports transpilation of `dbt_project.yml` files, but this is not tested.
The PR fixes that by adding a relevant test.
Closes [#240](https://github.com/databrickslabs/morpheus/issues/240)
(for tactical reasons, this test sits on Remorph side rather than Morpheus side, this should be changed)

### DIFF
--- a/src/databricks/labs/remorph/cli.py
+++ b/src/databricks/labs/remorph/cli.py
@@ -113,7 +113,7 @@ def transpile(
 ):
     """Transpiles source dialect to databricks dialect"""
     ctx = ApplicationContext(w)
-    print(f"{ctx.transpile_config!s}")
+    logger.debug(f"Application transpiler config: {ctx.transpile_config}")
     checker = _TranspileConfigChecker(ctx.transpile_config, ctx.prompts)
     checker.check_input_source(input_source)
     checker.check_source_dialect(source_dialect)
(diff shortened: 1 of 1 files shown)


tests/integration/transpile/test_dbt.py
```diff
@@ -0,0 +1,30 @@
...
+        transpile_config = TranspileConfig(
+            transpiler_config_path=str(config_path),
+            source_dialect="snowflake",
+            input_source=str(input_source),
+            output_folder=output_folder,
+            skip_validation=False,
+            catalog_name="catalog",
+            schema_name="schema",
+        )
+        await transpile(ws, lsp_engine, transpile_config)
```
gueniai: @ericvergnaud Can you add TODOs to this to move this test to Morpheus and also to test for the contents of the files?

gueniai (approved):

tests/integration/transpile/test_dbt.py
ericvergnaud: @gueniai done

src/databricks/labs/remorph/cli.py
```diff
@@ -449,4 +450,6 @@ def analyze(w: WorkspaceClient, source_directory: str, report_file: str):
 
 
 if __name__ == "__main__":
+    if "--debug" in sys.argv:
```
gueniai, line 453: This was fixed as part of https://github.com/databrickslabs/remorph/pull/1644, can be removed now.

gueniai (changes requested):
This PR needs to be rebased from main, or recreated, plus there some code that should be removed

ericvergnaud:
Superseded by https://github.com/databrickslabs/remorph/pull/1650

PR #1632, opened 2025-06-05
Ci/fix scope of integration tests

ericvergnaud:
Our integration tests are ran from 2 workflows: push and acceptance (with DB platform access)
However, which specific integration tests are ran depends on a pytest hook.
This PR adjusts this hook such that only tests requiring DB platform access are ran as part of acceptance.
It also moves them to larger runners.


sundarshankar89 (approved):
LGTM

gueniai (approved):
LGTM

PR #1639, opened 2025-06-05
Properly pin the Databricks SDK.

asnare:
### What does this PR do? 

This PR properly pins the project to a release of the Databricks SDK (or patch releases), which happens to resolve a linting error on `main` due to a breaking change in a later SDK release.

The `~=` syntax means the _last_ digit can be incremented safely. Because the `.0` was missing, were allowing `0.51+` instead of `0.51.x`.


sundarshankar89 (approved):
LGTM

jimidle (approved):
LGTM

PR #1637, opened 2025-06-05
Update CLI argument handling for `transpile`

asnare:
### What does this PR do? 

This PR updates the way command-line arguments for the `transpile` subcommand are processed and validated, including error-handling.

Changes include:

 - Differentiating between an invalid command-line argument, an invalid configuration file setting, and a missing argument because it's neither part of the configuration file nor provided as a command-line argument.
 - Updating existing tests to look for the new message(s).
 - The error file is now optional. If configured or provided, its parent must exist.
 - The output folder argument no longer has a default value: it must be configured or provided. At runtime its parent must exist.
 - The _parent_ of the output folder must already exist.
 - Relative paths for input/output/error are now relative to the current directory when you run the `transpile` command.
 - The workspace configuration file is no longer required: a warning is issued if it's missing, but arguments can be provided via the command-line when running `transpile`.
 - _… to be continued…_

### Relevant implementation details

_To be provided…_

### Caveats/things to watch out for when reviewing:

_To be provided…_

### Linked issues

See also: #1634

Depends on: #1641 (which deals with #1640).

### Functionality

- modified existing commands:

    - `databricks labs remorph transpile`
    - `Databricks Labs remorph generate-lineage`

### Tests

- manually tested
- updated unit tests
- added integration tests

PR #1641, opened 2025-06-05
Make transpiler configuration path optional in the workspace transpile configuration

### LINKED ISSUE #1640: [BUG]: `Set it later` for dialect during transpiler install puts `"None"` in the `config.yml`
### Current Behavior

During `install-transpile`, the user is prompted to configure the default dialect to use. If they choose `Set it later` then the string `"None"` is placed in the `config.yml` under that setting.

### Expected Behavior

If we don't have a dialect configured, this property should be set to `None` (the value, not a string) and the property end up omitted from the configuration file.

### Steps To Reproduce

1. `databricks labs install remorph install-transpile` (use `--artifact` if necessary).
2. If prompted to replace the configuration, choose `yes`.
3. When prompted to _Select the source dialect_, choose option 0: `Set it later`
4. Accept the defaults for the rest of the questions, exception for the last question to open the configuration file in our browser.
5. When the configuration file opens, notice that it contains `transpiler_config_path: None` which is a _string_ value, and not null.

asnare:
### What does this PR do? 

This PR makes the `transpiler_path_config` setting optional in `TranspilerConfig`, normally set during the post-install configuration questionnaire as part of the `install-transpile` subcommand.

This is in turn means we don't need to store `"None"` (the string) as a value in the file, in turn meaning we don't need downstream workarounds when reading the configuration.

In addition, when upgrading a v2 configuration we no longer need to set a value that will fail: we can just leave it out. Even though this is uncommon, leaving it unfilled rules out a source of issues downstream.

Making an existing field optional is forward-compatible, so we don't need to introduce a new version of the configuration.

### Linked issues

Resolves #1640.

Relates to #1637.

### Functionality

### Tests

- manually tested
- existing tests

### DIFF
--- a/src/databricks/labs/remorph/cli.py
+++ b/src/databricks/labs/remorph/cli.py
@@ -176,11 +176,12 @@ class _TranspileConfigChecker:
     def check(self) -> tuple[TranspileConfig, TranspileEngine]:
         logger.debug(f"Checking config: {self!s}")
         # not using os.path.exists because it sometimes fails mysteriously...
-        if not self._config.transpiler_config_path or not Path(self._config.transpiler_config_path).exists():
+        transpiler_path = self._config.transpiler_path
+        if not transpiler_path or not transpiler_path.exists():
             raise_validation_exception(
                 f"Invalid value for '--transpiler-config-path': Path '{self._config.transpiler_config_path}' does not exist."
             )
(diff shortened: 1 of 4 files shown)


gueniai (approved):
LGTM

sundarshankar89 (approved):
LGTM

PR #1642, opened 2025-06-05
Support combined output when converting ETL source code

ericvergnaud:
### What does this PR do? 
Add support for multi-file transpiler output in the form of a mime/multipart file
This is required for ETL conversion by `bladerunner` 

None

There is no detailed error handling when processing combined output, like there is for single output.
Since combined output can only come from bladerunner it wouldn't add value on the short term because:
- bladerunner does not support code fragment location
- bladerunner only fails on technical issues

An end-to-end integration test would be welcome, but given the urgency, it might make sense to merge this PR without it.

### Linked issues
Requires https://github.com/databrickslabs/bladerunner/pull/288 on bladerunner side

### Functionality
None

### Tests
- [x] added unit tests


ericvergnaud:
Superseded by #1651

PR #1652, opened 2025-06-06
skip invisible files

ericvergnaud:
### What does this PR do? 
Skips invisible files (starting with '.') when transpiling folders

### Tests
- [x] not tested, just doesn't break anything


gueniai (approved):
LGTM

PR #1655, opened 2025-06-06
[BUGFIX] Fix bladerunner install from pypi

ericvergnaud:
Adds a test for installing from PyPI and fixes corresponding bugs.


src/databricks/labs/remorph/install.py
```diff
@@ -253,12 +253,12 @@ def _unsafe_install_latest_version(self, version: str) -> Path | None:
     def _create_venv(self) -> None:
         cwd = os.getcwd()
         try:
+            os.chdir(self._install_path)
```
asnare, line 256: This looks like an improvement.

src/databricks/labs/remorph/install.py
```diff
@@ -312,7 +312,8 @@ def _install_with_pip(self) -> None:
     def _install_local_artifact(self) -> None:
         pip = self._locate_pip()
         pip = pip.relative_to(self._install_path)
-        target = self._site_packages.relative_to(self._install_path)
+        target = self._site_packages
+        target = target.relative_to(self._install_path)
```
asnare, line 316: This looks like it doesn't change anything functionally, although splitting it up like this helps with debugging or narrowing down the cause of an exception.

src/databricks/labs/remorph/install.py
```diff
@@ -325,13 +326,14 @@ def _install_local_artifact(self) -> None:
     def _install_remote_artifact(self) -> None:
         pip = self._locate_pip()
         pip = pip.relative_to(self._install_path)
-        target = self._site_packages.relative_to(self._install_path)
+        target = self._site_packages
+        target = target.relative_to(self._install_path)
```
asnare, line 330: This looks like it doesn't change anything functionally, although splitting it up like this helps with debugging or narrowing down the cause of an exception.

src/databricks/labs/remorph/install.py
```diff
@@ -325,13 +326,14 @@ def _install_local_artifact(self) -> None:
...
         if sys.platform == "win32":
             args = [str(pip), "install", self._pypi_name, "-t", str(target)]
             completed = run(args, stdin=sys.stdin, stdout=sys.stdout, stderr=sys.stderr, shell=False, check=False)
         else:
-            args = [f"'{pip!s}'", "install", self._pypi_name, "-t", f"'{target!s}'"]
-            completed = run(args, stdin=sys.stdin, stdout=sys.stdout, stderr=sys.stderr, shell=True, check=False)
+            command = f"'{pip!s}' install {self._pypi_name} -t '{target!s}'"
+            completed = run(command, stdin=sys.stdin, stdout=sys.stdout, stderr=sys.stderr, shell=True, check=False)
```
asnare, line 336: This looks like it might be addressing a problem. Was the problem that `self._pypi_name` or `target` was not being escaped properly for the shell?

If indeed the problem is related to shell-escaping, I think the root cause is that we're running with `shell=True`. If this is indeed the case, I suspect this might both address the issue, and be cleaner because it avoids the class of shell-escaping problems:
```suggestion
            args = [str(pip), "install", self._pypi_name, "-t", str(target)]
            completed = run(args, stdin=sys.stdin, stdout=sys.stdout, stderr=sys.stderr, shell=False, check=False)
```
(This would make it exactly the same as the `win32` branch above.)

tests/integration/install/test_install.py
```diff
@@ -220,6 +220,43 @@ async def test_installs_and_runs_local_bladerunner(bladerunner_artifact):
             assert transpiled == sql_code
 
 
+async def test_installs_and_runs_pypi_bladerunner():
+    # Note: This test currently uses the user's home-directory, and doesn't really test the install process if the
+    # transpiler is already installed there: many install paths are a no-op if the transpiler is already installed.
+    # TODO: Fix to use a temporary location instead of the user's home directory.
+    bladerunner = TranspilerInstaller.transpilers_path() / "bladerunner"
+    if bladerunner.exists():
+        shutil.rmtree(bladerunner)
```
asnare: Warning: this is not a test-specific location, but is rather `~/.databricks/labs/remorph-transpilers/bladerunner` for the developer (or CI?) that is running this test.

tests/integration/install/test_install.py
```diff
@@ -220,6 +220,43 @@ async def test_installs_and_runs_local_bladerunner(bladerunner_artifact):
             assert transpiled == sql_code
 
 
+async def test_installs_and_runs_pypi_bladerunner():
+    # Note: This test currently uses the user's home-directory, and doesn't really test the install process if the
+    # transpiler is already installed there: many install paths are a no-op if the transpiler is already installed.
```
asnare: This test is different to those where I attached this comment, so the comment is misleading: this test is never a no-op because it always overwrites whatever the user previously had in their home-directory.

asnare (commented):
I've added some comments intended to help with further review; they are not a request for changes or adjustment.

One thing that would help with review is something describing what the (technical) problem was that is being fixed.

PR #1685, opened 2025-06-10
Use diagnostics' severity to log their message using the corresponding log level

vil1:
### What does this PR do?

This avoids reporting warning as errors to the console.

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/cli.py
+++ b/src/databricks/labs/lakebridge/cli.py
@@ -40,7 +40,6 @@ from databricks.labs.lakebridge.transpiler.lsp.lsp_engine import LSPConfig
 from databricks.labs.lakebridge.transpiler.sqlglot.sqlglot_engine import SqlglotEngine
 from databricks.labs.lakebridge.transpiler.transpile_engine import TranspileEngine
 
-from src.databricks.labs.lakebridge.transpiler.transpile_status import ErrorSeverity
 
 lakebridge = App(__file__)
 logger = get_logger(__file__)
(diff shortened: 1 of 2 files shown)


src/databricks/labs/lakebridge/cli.py
```diff
@@ -307,7 +307,13 @@ async def _transpile(ctx: ApplicationContext, config: TranspileConfig, engine: T
...
     status, errors = await do_transpile(ctx.workspace_client, engine, config)
     for error in errors:
-        logger.error(f"Error Transpiling: {str(error)}")
+        msg = f"{error.path}: {error.message}"
+        if error.severity.name == "ERROR":
+            logger.error(msg)
+        elif error.severity.name == "WARNING":
+            logger.warning(msg)
+        elif error.severity.name == "INFO":
+            logger.info(msg)
```
asnare: Maybe we should log it at `DEBUG` if we don't recognise the level, rather than dropping it? (Can this happen?)

Something like this, maybe?
```python
    levels = {"ERROR": logging.ERROR, "WARNING": logging.WARNING, "DEBUG": logging.DEBUG}
    for error in errors:
        level = levels.get(error.severity.name, logging.DEBUG)
        logger.log(level, f"{error.path}: {error.message}")
```

asnare (approved):
Looks fine, although I think I'd prefer that unknown levels are mapped to `DEBUG` instead of dropped.

PR #1697, opened 2025-06-12
Refactor hash queries to not use sqlglot

ericvergnaud:
### What does this PR do?
Removes dependency on sqlglot for hash queries

### Worth noting
There is an inherited probable bug here https://github.com/databrickslabs/lakebridge/blob/52dc7ede799099cd59a2865a32660f421a46e6cc/src/databricks/labs/lakebridge/reconcile/execute.py#L510
where the same source is used for both sides of the reconciliation
This PR does not fix this bug, but there is a need to look into it.

### Tests
- [x] refactored and ran unit tests


sundarshankar89:
Created #2026  so closing this

PR #1705, opened 2025-06-16
Type hints

asnare:
### What does this PR do?

This PR adds some type-hints, extracted from #1637 in an effort to make that PR smaller in scope. The type-hints here cover:

 - `TranspilerConfig`
 - The unit tests for the CLI.
 - Some of the test fixtures.

When a function signature doesn't have any type hints, the linter does _not_ check types in the body at all. This means it's always important to use `-> None` if nothing else is present to ensure the body is linted.

### Linked issues

Relates: #1637

### Tests

- updated unit tests


gueniai (approved):
lgtm

PR #1751, opened 2025-06-23
Disable header comments

vil1:
Disable the generation of header comment summarizing the encountered errors/warnings as they may make `.py` or `.json` files incorrect.


asnare (approved):
In the interests of expediency this is fine, although I'd like a marker for why it's disabled.

sundarshankar89:
@vil1  andrew wants a comment to the comment on why we commented that out 🙂

vil1:
> @vil1 andrew wants a comment to the comment on why we commented out 🙂

Which makes the above a comment about a comment asking for commenting why code have been commented out ><.

sundarshankar89 (approved):
LGTM

PR #1774, opened 2025-06-27
Patches End To End Install Test in main

sundarshankar89:
### What does this PR do?
This PR modified the behaviour of the existing tests to remove async and install with a temporary directory.

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/tests/integration/install/test_install_and_run.py
+++ b/tests/integration/install/test_install_and_run.py
@@ -76,7 +76,6 @@ def _install_and_run_local_bladebridge(bladebridge_artifact: Path) -> None:
             transpiled = format_transpiled(result.transpiled_code)
             assert transpiled == sql_code
 
-
 def test_installs_and_runs_pypi_bladebridge():
     os.chdir(base_cwd)
     if sys.platform == "win32":
--- a/tests/integration/install/test_install_and_run.py
+++ b/tests/integration/install/test_install_and_run.py
@@ -33,8 +33,6 @@ async def run_lsp_operations_sync(
     return await run_lsp_operations()  # Await the coroutine
 
 
-
-
 def test_installs_and_runs_local_bladebridge(bladebridge_artifact: Path) -> None:
     if sys.platform == "win32":
         _install_and_run_pypi_bladebridge()
--- a/tests/integration/install/test_install_and_run.py
+++ b/tests/integration/install/test_install_and_run.py
@@ -1,4 +1,3 @@
-import os
 import asyncio
 import shutil
 import sys
@@ -34,11 +33,9 @@ async def run_lsp_operations_sync(
     return await run_lsp_operations()  # Await the coroutine
 
 
-base_cwd = os.getcwd()
 
 
 def test_installs_and_runs_local_bladebridge(bladebridge_artifact: Path) -> None:
-    os.chdir(base_cwd)
     if sys.platform == "win32":
         _install_and_run_pypi_bladebridge()
     else:
@@ -78,7 +75,6 @@ def _install_and_run_local_bladebridge(bladebridge_artifact: Path) -> None:
 
 
 def test_installs_and_runs_pypi_bladebridge():
-    os.chdir(base_cwd)
     if sys.platform == "win32":
         _install_and_run_pypi_bladebridge()
     else:
@@ -118,7 +114,6 @@ def _install_and_run_pypi_bladebridge() -> None:
 
 
 def test_installs_and_runs_local_morpheus(morpheus_artifact):
-    os.chdir(base_cwd)
     if sys.platform == "win32":
         _install_and_run_pypi_bladebridge()
     else:
--- a/tests/integration/install/test_install_and_run.py
+++ b/tests/integration/install/test_install_and_run.py
@@ -21,6 +21,19 @@ def format_transpiled(sql: str) -> str:
     return sql
 
 
+def run_lsp_operations_sync(lsp_engine, transpile_config, input_source, sql_code):
+    """Helper function to run LSP operations synchronously"""
+    async def run_lsp_operations():
+        await lsp_engine.initialize(transpile_config)
+        dialect = transpile_config.source_dialect
+        input_file = Path(input_source) / "some_query.sql"
+        result = await lsp_engine.transpile(dialect, "databricks", sql_code, input_file)
+        await lsp_engine.shutdown()
+        return result
+    
+    return asyncio.run(run_lsp_operations())
+
+
 base_cwd = os.getcwd()
 
 
@@ -66,16 +79,8 @@ def _install_and_run_local_bladebridge(bladebridge_artifact: Path):
                 schema_name="schema",
             )
             
-            async def run_lsp_operations():
-                await lsp_engine.initialize(transpile_config)
-                dialect = cast(str, transpile_config.source_dialect)
-                input_file = Path(input_source) / "some_query.sql"
-                sql_code = "select * from employees"
-                result = await lsp_engine.transpile(dialect, "databricks", sql_code, input_file)
-                await lsp_engine.shutdown()
-                return result
-            
-            result = asyncio.run(run_lsp_operations())
+            sql_code = "select * from employees"
+            result = run_lsp_operations_sync(lsp_engine, transpile_config, input_source, sql_code)
             transpiled = format_transpiled(result.transpiled_code)
             assert transpiled == sql_code
 
@@ -122,16 +127,8 @@ def _install_and_run_pypi_bladebridge():
                 schema_name="schema",
             )
             
-            async def run_lsp_operations():
-                await lsp_engine.initialize(transpile_config)
-                dialect = transpile_config.source_dialect
-                input_file = Path(input_source) / "some_query.sql"
-                sql_code = "select * from employees"
-                result = await lsp_engine.transpile(dialect, "databricks", sql_code, input_file)
-                await lsp_engine.shutdown()
-                return result
-            
-            result = asyncio.run(run_lsp_operations())
+            sql_code = "select * from employees"
+            result = run_lsp_operations_sync(lsp_engine, transpile_config, input_source, sql_code)
             transpiled = format_transpiled(result.transpiled_code)
             assert transpiled == sql_code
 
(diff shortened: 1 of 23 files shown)


sundarshankar89:
closes #1781

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
     parts = sql.lower().split("\n")
     stripped = [s.strip() for s in parts]
     sql = " ".join(stripped)
-    sql = sql.replace(";;", ";")
+    # Remove any trailing semicolons Since the bug was fixed in the LSP engine, we can safely remove them
```
asnare: There's something wrong with the wording here.

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
...
+    # sql = sql.replace(";;", ";")
     return sql
 
 
-# don't run installers in parallel
-installer_lock = asyncio.Lock()
+def run_lsp_operations_sync(lsp_engine, transpile_config, input_source, sql_code):
```
asnare: Type hints, please. ;)

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
...
+    """Helper function to run LSP operations synchronously"""
+
+    async def run_lsp_operations():
```
asnare: Type hints here too.

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
...
+
+    return asyncio.run(run_lsp_operations())
+
+
 base_cwd = os.getcwd()
 
 
-async def test_installs_and_runs_local_bladebridge(bladebridge_artifact):
-    await installer_lock.acquire()
+def test_installs_and_runs_local_bladebridge(bladebridge_artifact):
```
asnare: Let's type-hint this while we're here?

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
...
+        return result
+
+    return asyncio.run(run_lsp_operations())
+
+
 base_cwd = os.getcwd()
 
 
-async def test_installs_and_runs_local_bladebridge(bladebridge_artifact):
-    await installer_lock.acquire()
```
asnare, line 28: 💯

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
...
-
-
-async def _install_and_run_local_bladebridge(bladebridge_artifact: Path):
+    # TODO temporary workaround for RecursionError with temp dirs on Windows
+    if sys.platform == "win32":
+        _install_and_run_pypi_bladebridge()
+    else:
+        with TemporaryDirectory() as tmpdir:
+            with patch.object(TranspilerInstaller, "labs_path", return_value=Path(tmpdir)):
+                _install_and_run_local_bladebridge(bladebridge_artifact)
```
asnare: This is a huge code-smell, there's something fundamentally wrong with what's going on here and how we're testing. Fixing it is beyond the scope of this PR I think, but suffice to say I don't think we should be having platform-guards like this in the middle of tests: it means they're testing different things.

If we absolutely need to keep platform-specific things, let's try to isolate them to a test and use the skip functionality to run them appropriately.

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
...
+
+
+def _install_and_run_local_bladebridge(bladebridge_artifact: Path):
```
asnare:
```suggestion
def _install_and_run_local_bladebridge(bladebridge_artifact: Path) -> None:
```

tests/integration/install/test_install_and_run.py
```diff
@@ -17,31 +16,40 @@ def format_transpiled(sql: str) -> str:
...
+
+    async def run_lsp_operations():
+        await lsp_engine.initialize(transpile_config)
+        dialect = transpile_config.source_dialect
+        input_file = Path(input_source) / "some_query.sql"
+        result = await lsp_engine.transpile(dialect, "databricks", sql_code, input_file)
+        await lsp_engine.shutdown()
+        return result
+
+    return asyncio.run(run_lsp_operations())
```
asnare: I don't think there's anything wrong with having async tests, and in some ways it's better for pytest to manage that than for us to do it ourselves. It's a good idea to factor out the common sequence though…
```suggestion
async def run_lsp_operations_sync(lsp_engine: ..., transpile_config: ..., input_source: Path, sql_code: ...) -> ...:
    """Common LSP sequence used by tests."""

    await lsp_engine.initialize(transpile_config)
    dialect = transpile_config.source_dialect
    input_file = input_source / "some_query.sql"
    result = await lsp_engine.transpile(dialect, "databricks", sql_code, input_file)
    await lsp_engine.shutdown()
    return result
```

asnare (commented):
Firstly, removing those `asyncio` locks is a good idea: they shouldn't be here.

Generally speaking, for `async` code I think it's best to leave the test as `async`: we're using `pytest-asyncio` which takes care of things for us and reduces clutter.

One big code-smell with these tests that we need to address is the platform-specific guards: these mean we're testing different things on different platforms, and it's not immediately clear what and why. It's beyond the scope of this PR, I think, but I think we need to:

 - Revisit why the platform-specific code is in place. If it's truly necessary, we should try to just have different platform-specific tests (with `skip=` guards) rather than a single test that does different things on different platforms.
 - Currently these tests all share and overwrite `.databricks` in the user's home directory, and I strongly suspect this is the source of a lot of the problems that led to what we see in here. The way we should solve this, without patching, is to set up a temporary `DATABRICKS_HOME` folder for each test. This will let tests run concurrently and isolated without interfering with each other.

tests/integration/install/test_install_and_run.py
sundarshankar89: That was a bug in Morpheus, issue is long fixed so should remove this from our code

tests/integration/install/test_install_and_run.py
sundarshankar89: Done

tests/integration/install/test_install_and_run.py
sundarshankar89: Done

tests/integration/install/test_install_and_run.py
sundarshankar89: Done

tests/integration/install/test_install_and_run.py
sundarshankar89: Done

tests/integration/install/test_install_and_run.py
sundarshankar89: Done

tests/integration/install/test_install_and_run.py
```diff
@@ -16,45 +14,44 @@ def format_transpiled(sql: str) -> str:
...
 
 
-# don't run installers in parallel
-installer_lock = asyncio.Lock()
+async def run_lsp_operations_sync(
+    lsp_engine: LSPEngine, transpile_config: TranspileConfig, input_source: str, sql_code: str
+) -> TranspileResult:
+    """Helper function to run LSP operations synchronously"""
+
+    async def run_lsp_operations() -> TranspileResult:
```
asnare: I know why this was originally here, but I think we can inline it now.

tests/integration/install/test_install_and_run.py
```diff
@@ -16,45 +14,44 @@ def format_transpiled(sql: str) -> str:
...
-            await _install_and_run_pypi_bladebridge()
-        else:
-            with TemporaryDirectory() as tmpdir:
-                with patch.object(TranspilerInstaller, "labs_path", return_value=Path(tmpdir)):
-                    await _install_and_run_local_bladebridge(bladebridge_artifact)
-    finally:
-        installer_lock.release()
+def test_installs_and_runs_local_bladebridge(bladebridge_artifact: Path) -> None:
+    if sys.platform == "win32":
+        _install_and_run_pypi_bladebridge()
```
asnare: This branch means that we duplicate `test_installs_and_runs_pypi_bladebridge`, making it redundant.

We should remove the branch and just mark the test to be skipped if the platform is `win32`.

tests/integration/install/test_install_and_run.py
```diff
@@ -16,45 +14,44 @@ def format_transpiled(sql: str) -> str:
...
 
-async def _install_and_run_local_bladebridge(bladebridge_artifact: Path):
-    # TODO: Test that running with existing install does nothing
-    # TODO: Test that running with legacy install upgrades it
-    # check new install
+def _install_and_run_local_bladebridge(bladebridge_artifact: Path) -> None:
     bladebridge = TranspilerInstaller.transpilers_path() / "bladebridge"
-    # TODO temporary workaround for RecursionError with temp dirs on Windows
     if sys.platform == "win32" and bladebridge.exists():
         shutil.rmtree(bladebridge)
```
asnare: This looks like dead code, I think: we don't run this on win32.

tests/integration/install/test_install_and_run.py
```diff
@@ -68,47 +65,32 @@ async def _install_and_run_local_bladebridge(bladebridge_artifact: Path):
...
-    # TODO: Test that running with existing install does nothing
-    # TODO: Test that running with legacy install upgrades it
-    # check new install
+def test_installs_and_runs_pypi_bladebridge():
+    if sys.platform == "win32":
+        _install_and_run_pypi_bladebridge()
+    else:
+        with TemporaryDirectory() as tmpdir:
+            with patch.object(TranspilerInstaller, "labs_path", return_value=Path(tmpdir)):
+                _install_and_run_pypi_bladebridge()
```
asnare, line 72: This is a massive smell, and I believe is the reason we still have flaky tests on Windows: they're sharing the user's `~/.databricks` directory instead of using new/clean locations.

For the non-Windows situation there's the patch in place to avoid this, but it's unclear why this isn't also being done for Windows.

tests/integration/install/test_install_and_run.py
```diff
@@ -122,49 +104,35 @@ async def _install_and_run_pypi_bladebridge():
...
-        installer_lock.release()
-
-
-async def _install_and_run_local_morpheus(morpheus_artifact):
-    # TODO: Test that running with existing install does nothing
-    # TODO: Test that running with legacy install upgrades it
-    # check new install
+def test_installs_and_runs_local_morpheus(morpheus_artifact):
+    if sys.platform == "win32":
+        _install_and_run_pypi_bladebridge()
```
asnare: This looks like copypasta, and isn't supposed to be here. (Does this means we're not integration-testing morpheus at all on Windows?)

asnare (approved):
This PR does improve things and I think on that basis we should merge it, bearing in mind that there are still significant issues:

 - It looks like we're not testing morpheus at all on Windows.
 - The `win32` paths are a bit of a mess, and it looks like they're still both sharing and wiping the `~/.databricks/remorph-transpilers/` directory.

tests/integration/install/test_install_and_run.py
sundarshankar89: testing a fix

gueniai (approved):
lgtm

PR #1784, opened 2025-06-30
Improve diagnostics if the Java check fails prior to installing morpheus

asnare:
### What does this PR do?

This PR updates the check for Java when installing Morpheus so that we differentiate between the different causes of failure, and provide more information to the user if it fails. All code-paths now have a `WARNING` that describes the problem, followed by (the same) `ERROR` indicating that we need Java 11 or later.

### Linked issues

Resolves #1768 

### Functionality

- modified existing command: `databricks labs lakebridge install-transpile`

### Tests

- manually tested
- updated unit tests
- updated integration tests

#### Sample output

When Java cannot be found at all:
```
17:07:53  WARNING [d.l.lakebridge.install] No Java executable found in the system PATH.
17:07:53    ERROR [d.l.lakebridge.install] The morpheus transpiler requires Java 11 or above. Please install Java and re-run 'install-transpile'.
```

When Java can be found, but running it fails:
```
17:14:04    DEBUG [d.l.lakebridge.install] Using java executable: PosixPath('/tmp/x/java')
17:14:04    DEBUG [d.l.lakebridge.install] Failed to run (1, ['/tmp/x/java', '-version']) (exit-code=1, stdout=b'', stderr=b'This is not the Java you are looking for.\n')
Traceback (most recent call last):
  File "/Users/**REDACTED**/dev/remorph-2/src/databricks/labs/lakebridge/install.py", line 624, in find_java
    completed = run([str(java_executable_path), "-version"], shell=False, capture_output=True, check=True)
  File "/Users/**REDACTED**/Library/Application Support/hatch/env/virtual/.pythons/3.10/python/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/tmp/x/java', '-version']' returned non-zero exit status 1.
17:14:04  WARNING [d.l.lakebridge.install] Java found, but could not determine the version: /tmp/x/java.
17:14:04    ERROR [d.l.lakebridge.install] The morpheus transpiler requires Java 11 or above. Please install Java and re-run 'install-transpile'.
```

When Java can be found but the version can't be parsed, for example with Java 1.8:
```
15:55:07    DEBUG [d.l.lakebridge.install] Using java executable: PosixPath('/usr/bin/java')
[...]

PR #1785, opened 2025-06-30
Improve LSP client directory handling

asnare:
### What does this PR do?

This PR factors out some common code that we use for temporarily changing the current-working-directory into a context manager. It also updates the initialisation sequence of the LSP client to use it; previously the current working directory was being left changed. (This in turn often left the environment for unit tests corrupted because this was happening in a temporary directory that was subsequently removed.)

The LSP engine had some code to restore the working directory, but it only took effect if an error occurred: the happy flow did not restore the working directory.

Some integration tests had defensive code in place to try to re-initialize the working directory; this was probably in place because those tests were otherwise flaky and failing because something _else_ had corrupted the test environment. These effectively masked the issue, so I've removed it: I'd prefer the underlying issue be fixed.

### Linked issues

Blocks #1784 (integration tests are flaky because the working directory sometimes doesn't exist)

### Functionality

### Tests

- added unit tests
- updated integration tests

### DIFF
--- a/tests/unit/test_cli_transpile.py
+++ b/tests/unit/test_cli_transpile.py
@@ -2,7 +2,6 @@ from collections.abc import Generator, Callable
 from unittest.mock import create_autospec, patch, ANY, MagicMock
 from pathlib import Path
 import dataclasses
-import os.path
 
 import pytest
 import yaml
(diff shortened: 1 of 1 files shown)


sundarshankar89 (approved):
LGTM

gueniai (approved):
LGTM

PR #1802, opened 2025-07-03
Workaround issue loading transpiler configuration with python ≥ 3.12.4

asnare:
### What does this PR do?

Due to an upstream issue in Blueprint, loading dataclasses doesn't work on versions of Python from 3.12.4 if the type annotations are a forward reference. This PR introduces a workaround so that we avoid this problem.

A forward reference is when the type annotation has the type in quotes, for example `"TranspilerConfig"`. Using the `from __future__ import annotations` directive makes Python rewrite _all_ type annotations in this way. As such removing the directive fixed the problem.

### Linked issues

Relates: databrickslabs/blueprint#252

### Functionality

- modified existing commands:

   - `databricks labs lakebridge install-transpile`
   - `databricks labs lakebridge transpile`

(Both of these load the configuration from the workspace.)

### Tests

- manually tested
- added unit tests


gueniai (approved):
LGTM!

Issue #63, opened 2023-11-17
Create test to check for file with BOM


nsenno-dbr:
reading a file with a Byte Order Mark created parsing issues for the original code_transpiler. We should add a test to remorph to ensure it can parse these types of files correctly 

See the original PR https://github.com/databricks/code_transpiler/pull/144

PR #1843, opened 2025-07-18
Improve end-to-end tests with non-UTF-8 sources

asnare:
### What does this PR do?

This PR improves the end-to-end integration tests that we perform with sources that are not UTF-8 encoded. This includes:

 - Morpheus: a SQL file encoded with UTF-16.

(A Bladebridge example is not included: it turns out this fails, and will be addressed in a subsequent PR.)

Although it looks like there are changes to `wf_m_employees_load.XML`, there aren't really: it's a side-effect of configuring git to ensure the file has the encoding/line/file-ending that we intend. (Git prefers to store everything internally as UTF-8, and assumes files when checked out also are UTF-8… this PR configures git to encode the checked-out file properly.)

### Linked issues

Relates #63

### Tests

- updated integration tests


tests/resources/functional/informatica/wf_m_employees_load.XML
asnare, line 1: The content of this file has not changed, but rather the way it's stored internally has (which triggers the diff).

sundarshankar89 (approved):
LGTM

Issue #1603, opened 2025-05-27
Installer Improvements


sundarshankar89:
As things stand with this PR, there are three main behaviours that it introduces that need to be addressed. In order of priority:
 * [X] Update to not use the Debian CDN. This is an external service unaffiliated with PyPi and intended for use by the Debian project. We should not be using it.  
       Issue created: #1600  
       Resolved by: #1809
 * [ ] Update the installer to: a) not copy the entire `lsp/` directory out of the installed artefact, but instead just the `config.yml` file; b) not try to run `install.sh`.
 * [X] Tech-debt: Use the `venv` module to set up the virtual environment rather than shell out, as well use it for all the path manipulation instead of doing it ourselves. Use `python3 -m pip` to run pip within the virtualenv, rather than trying to figure out the entry-point ourselves. (The path to `python3` comes from the `venv` module.) This will eliminate a lot of platform-specific code. Dealt with in: #1894.

I don't really mind if the above items are addressed in this PR or if an issue is created to cover it, so long as the issue is assigned so that it's clear who will be doing to work.

Some to this some additional tickets are needed:

 * [ ] Tech-debt: When launching the LSP server, just use the `venv` module to grab the `bin/` directory if it exists and put it on the `PATH` so that the command from `config.yml` just works. This will eliminate most (all?) of python- and java-specific code, along with the platform-specific code. The `PYTHONPATH` manipulation might also no longer be necessary.  
       Dealt with in: #1907
 * [ ] Wheels are architecture-specific, with `any` being a catch-all. We should support architecture-specific wheels, not just blindly use the `any` one.
 * [ ] Do we need to detect the version of the plugin we're about to install? If we do, we need to document why. (Maybe I missed it somewhere?) If we don't, just use pip within the virtualenv to find the latest version and install it. Probably solves the prior point, and also resolves an issue that many customers will have where they are behind a corporate firewall/proxy or have to use an internal PyPi mirror: pip can be configured to work.
 * [ ] Improvement: For maven, prefer `mvn` if installed and use `mvn org.apache.maven.plugins:maven-dependency-plugin:2.8:get` to get the latest version. This will also deal with the common problems of firewalls/proxies, because `mvn` can be configured to work with internal mirrors whereas we can't. (Related issue #1602: remove our dependency on the `search.maven.org` service.)
[...]

PR #1861, opened 2025-07-24
Internal improvements to code responsible for installing transpilers

asnare:
## Changes

This PR refactors the class hierarchy (and methods) used for installing transpilers and locating their configuration, allowing the location of the transpilers to be injectable. This eliminates a major source of problems with testing:

 - Monkeypatching is no longer needed, meaning tests can safely run concurrently.
 - Integration tests no longer share (and wipe) the user's transpiler directory, removing a major source of flakiness.
 - Some platform-specific codepaths have been eliminated.

Incidental improvements included in this PR:

 - Type hinting in various places, along with accompanying fixes.
 - Some tests that monkey-patched classes without restoring the original implementation have been fixed, partly because the monkey-patching is no longer required.
 - Some tests that ran their own async event-loop now use the pytest support for this.
 - Some tests that managed their own temporary directories now use the pytest fixture that supports this.
 - A new integration test against the published version of Morpheus has also been added.
 - Some TODO markers for future work/improvements.

The existing `TranspilerInstaller` has been split into two parts:

 - `TranspilerRepository`: manages the location where transpilers are installed, and locating installed transpilers and their configured.
 - A new `TranspilerInstaller` which is just the common code (as a base class) for `WheelInstaller` and `MavenInstaller`.

Composition is used to combine these rather than inheritance which makes injection and testing much easier and more reliable. 

Some Win32-specific code

A few other minor improvements along with way are present.

### Linked issues

Relates #1603

### Functionality

### Tests

- [ ] manually tested
- [X] updated unit tests
- [X] added and updated integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/install.py
+++ b/src/databricks/labs/lakebridge/install.py
@@ -123,13 +123,15 @@ class TranspilerInstaller(abc.ABC):
 
     @classmethod
     def transpiler_config_path(cls, transpiler_name) -> Path:
-        config = cls.all_transpiler_configs().get(transpiler_name, None)
-        if not config:
-            raise ValueError(f"No such transpiler: {transpiler_name}")
+        # Note: Can't just go straight to the directory: the transpiler names don't exactly match the directory names.
+        try:
+            config = next(c for c in cls._all_transpiler_configs() if c.name == transpiler_name)
+        except StopIteration as e:
+            raise ValueError(f"No such transpiler: {transpiler_name}") from e
         return config.path
 
     @classmethod
-    def transpiler_config_options(cls, transpiler_name, source_dialect) -> list[LSPConfigOptionV1]:
+    def transpiler_config_options(cls, transpiler_name: str, source_dialect: str) -> list[LSPConfigOptionV1]:
         config = cls.all_transpiler_configs().get(transpiler_name, None)
         if not config:
             return []  # gracefully returns an empty list, since this can only happen during testing
@@ -208,14 +210,12 @@ class WheelInstaller(TranspilerInstaller):
 
     def _install_latest_version(self, version: str) -> Path | None:
         logger.info(f"Installing Databricks {self._product_name} transpiler v{version}")
-        # use type(self) to workaround a mock bug on class methods
-        self._product_path = type(self).transpilers_path() / self._product_name
+        self._product_path = self.transpilers_path() / self._product_name
         backup_path = Path(f"{self._product_path!s}-saved")
         if self._product_path.exists():
             os.rename(self._product_path, backup_path)
-        self._product_path.mkdir(parents=True, exist_ok=True)
         self._install_path = self._product_path / "lib"
-        self._install_path.mkdir(exist_ok=True)
+        self._install_path.mkdir(parents=True, exist_ok=True)
         try:
             result = self._unsafe_install_latest_version(version)
             logger.info(f"Successfully installed {self._pypi_name} v{version}")
--- a/src/databricks/labs/lakebridge/install.py
+++ b/src/databricks/labs/lakebridge/install.py
@@ -46,31 +46,25 @@ logger = logging.getLogger(__name__)
 TRANSPILER_WAREHOUSE_PREFIX = "Lakebridge Transpiler Validation"
 
 
-class TranspilerInstaller(abc.ABC):
-
-    @classmethod
-    def labs_path(cls) -> Path:
+class TranspilerRepository:
+    @staticmethod
+    def default_labs_path() -> Path:
+        """Return the default path where labs applications are installed."""
         return Path.home() / ".databricks" / "labs"
 
-    @classmethod
-    def transpilers_path(cls) -> Path:
-        return cls.labs_path() / "remorph-transpilers"
+    @staticmethod
+    def user_home() -> "TranspilerRepository":
+        """The default repository for the in the current user's home directory."""
+        return _DEFAULT_REPOSITORY
 
-    @classmethod
-    def install_from_pypi(cls, product_name: str, pypi_name: str, artifact: Path | None = None) -> Path | None:
-        installer = WheelInstaller(product_name, pypi_name, artifact)
-        return installer.install()
+    def __init__(self, labs_path: Path) -> None:
+        self._labs_path = labs_path
 
-    @classmethod
-    def install_from_maven(
-        cls, product_name: str, group_id: str, artifact_id: str, artifact: Path | None = None
-    ) -> Path | None:
-        installer = MavenInstaller(product_name, group_id, artifact_id, artifact)
-        return installer.install()
+    def transpilers_path(self) -> Path:
+        return self._labs_path / "remorph-transpilers"
 
-    @classmethod
-    def get_installed_version(cls, product_name: str, is_transpiler=True) -> str | None:
-        product_path = (cls.transpilers_path() if is_transpiler else cls.labs_path()) / product_name
+    def get_installed_version(self, product_name: str) -> str | None:
+        product_path = self.transpilers_path() / product_name
         current_version_path = product_path / "state" / "version.json"
         if not current_version_path.exists():
             return None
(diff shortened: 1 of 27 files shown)


src/databricks/labs/lakebridge/install.py
```diff
@@ -158,6 +133,32 @@ def _transpiler_config(cls, path: Path) -> LSPConfig | None:
             logger.error(f"Could not load config: {path!s}", exc_info=e)
             return None
 
+
+_DEFAULT_REPOSITORY: TranspilerRepository = TranspilerRepository(TranspilerRepository.default_labs_path())
+
+
+class TranspilerInstaller(abc.ABC):
```
m-abulazm, line 161: I would split into two files for better readability. 
I would encourage the practice of one service per file

src/databricks/labs/lakebridge/install.py
```diff
@@ -46,31 +46,29 @@
...
+    @staticmethod
+    def default_labs_path() -> Path:
+        """Return the default path where labs applications are installed."""
         return Path.home() / ".databricks" / "labs"
 
-    @classmethod
-    def transpilers_path(cls) -> Path:
-        return cls.labs_path() / "remorph-transpilers"
+    @staticmethod
+    def user_home() -> "TranspilerRepository":
```
m-abulazm: why not using an empty constructor? since the default labs path is a well known property

for better readability I would introduce `TranspilerInstallerFactory#create_default()`

src/databricks/labs/lakebridge/install.py
```diff
@@ -158,6 +133,32 @@ def _transpiler_config(cls, path: Path) -> LSPConfig | None:
...
 
+
+_DEFAULT_REPOSITORY: TranspilerRepository = TranspilerRepository(TranspilerRepository.default_labs_path())
+
+
+class TranspilerInstaller(abc.ABC):
+    def __init__(self, repository: TranspilerRepository) -> None:
+        self._repository = repository
+
+    _version_pattern = re.compile(r"[_-](\d+(?:[.\-_]\w*\d+)+)")
```
m-abulazm, line 165: use as constant outside class

tests/integration/install/test_install_and_run.py
```diff
@@ -35,125 +38,185 @@ def format_transpiled(sql: str) -> str:
...
+        else:
+            log_dir_name = log_dir.relative_to(transpilers_path)
+            logger.info(f"No logs found for transpiler: {log_dir_name}")
+
+
+@pytest.fixture(name="transpiler_repository")
+def log_capturing_transpiler_repository(tmp_path: Path) -> Generator[TranspilerRepository, None, None]:
+    labs_path = tmp_path / "labs"
+    transpiler_repository = TranspilerRepository(labs_path=labs_path)
+    try:
```
m-abulazm: you dont need try finally according to https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended or am I missing something

tests/integration/install/test_install_and_run.py
```diff
@@ -35,125 +38,185 @@ def format_transpiled(sql: str) -> str:
...
+    assert transpiled == sql_code
+
+
+async def test_installs_and_runs_pypi_bladebridge(transpiler_repository: TranspilerRepository, tmp_path: Path) -> None:
+    input_source = tmp_path / "input_source"
+    output_folder = tmp_path / "output_folder"
+    await _install_and_run_pypi_bladebridge(transpiler_repository, input_source, output_folder)
+
+
+async def _install_and_run_pypi_bladebridge(
```
m-abulazm: I am not getting this pattern as in here of a public and private methods that I see in a lot of places in production code 

but that is a discussion for another time I guess

tests/integration/install/test_install_and_run.py
```diff
@@ -35,125 +38,185 @@ def format_transpiled(sql: str) -> str:
...
+        sdk_config={"cluster_id": "test_cluster"},
+        skip_validation=False,
+        catalog_name="catalog",
+        schema_name="schema",
+    )
+
+    sql_code = "select * from employees"
+    result = await run_lsp_operations(lsp_engine, transpile_config, input_source, sql_code)
+    transpiled = process_email_content(result.transpiled_code)
+    assert transpiled == sql_code
```
m-abulazm, line 111: this is similar for a few of the tests, can be dry-ed with something like e.g.
```
_test_install_and_run_transpiler(transpiler_repository: TranspilerRepository,
    bladebridge_artifact: Path,
    input_source: Path,
    output_folder: Path,
    installer: Installer,
    transpiler_name: str)
```

tests/integration/transpile/test_bladebridge.py
```diff
@@ -1,36 +1,37 @@
...
-    ws: WorkspaceClient, bladebridge_artifact: Path, tmp_path: Path
+    ws: WorkspaceClient,
+    bladebridge_artifact: Path,
+    tmp_path: Path,
 ) -> None:
     labs_path = tmp_path / "labs"
     output_folder = tmp_path / "output"
-    with patch.object(TranspilerInstaller, "labs_path", return_value=labs_path):
-        await _transpile_informatica_with_sparksql(ws, bladebridge_artifact, output_folder)
+    transpiler_repository = TranspilerRepository(labs_path)
```
m-abulazm, line 22: really great. much better

src/databricks/labs/lakebridge/install.py
asnare: I think I'd prefer to leave it here:

 - As a class attribute, it's still initialised during loading (and effectively constant). This is deliberate, to ensure that errors in the regex are detected early (during import) rather than lazily upon search).
 - It's tightly coupled to the method below, and local to the class.

With this in mind I'd prefer to leave it here (scoped to the class) rather than move it to module-level, also bearing in mind that this is moved code rather than new code.

m-abulazm: won't the build fail on broken regex if the constant is outside?
anyway I am fine with what you prefer. feel free to resolve

src/databricks/labs/lakebridge/install.py
asnare: This is here because I specifically want to discourage the callers from creating an instance unless they specify the path. The idea here is that:

 - This is a single instance for application use.
 - Tests can initialise new instances with a different path.

To help clarify this I've updated things a bit:

 - The docstrings now make the expected usage clear.
 - I've moved the singleton into the class so that it's clearer through inspection what's going on.
 - I've updated the `__init__` implementation to raise an exception if the caller is trying to create a new (default) instance.

Does that seem better?

src/databricks/labs/lakebridge/install.py
asnare: This PR was about splitting up an existing class hierarchy a bit, moving in the right direction. Keeping things here has helped a little bit with review, although the diffs are still a bit misleading because some existing code has been moved too far for the diff algorithm to realise it's not new.

I think the right modularity here is to move `TranspilerRepository` into its own module, but leave the installers here for now. I've added a TODO marker for this, and will do it in a separate PR once this is finalised. Does that sound okay?

tests/integration/install/test_install_and_run.py
asnare: I agree, it's a bit weird. I suspect that originally the test wasn't async but the `_method` was. Now it's just noise though. Would you prefer to see that addressed here, or in a separate PR?

tests/integration/transpile/test_bladebridge.py
asnare: Indeed, this was the target… pretty hairy Yak though.

tests/integration/install/test_install_and_run.py
asnare: Good catch, the documentation isn't explicit about what happens with a _failed_ test but I just tested it and failures/exceptions are caught by pytest so that the (coroutine) fixture continues. Fixed.

tests/integration/install/test_install_and_run.py
asnare: Nevermind, I just refactored it… just lines were already touched and the indirection was entirely pointless.

src/databricks/labs/lakebridge/install.py
m-abulazm: sounds good

tests/integration/install/test_install_and_run.py
m-abulazm: thanks it was a lot of lines of codes

src/databricks/labs/lakebridge/install.py
m-abulazm: better thank you

m-abulazm (approved):
in a pretty good state already and one previous review comment left but since it is not production code I am approving :)

tests/integration/install/test_install_and_run.py
```diff
@@ -35,125 +38,150 @@ def format_transpiled(sql: str) -> str:
     return sql
 
 
-async def run_lsp_operations_sync(
-    lsp_engine: LSPEngine, transpile_config: TranspileConfig, input_source: str, sql_code: str
+def _capture_transpiler_logs(transpiler_repository: TranspilerRepository) -> None:
```
sundarshankar89, line 41: we need this in BB repo too.
@sundarshankar89 note to self.

sundarshankar89 (approved):
LGTM

src/databricks/labs/lakebridge/install.py
asnare: See #1879.

PR #1867, opened 2025-07-25
Remove circular dependency between `TranspileEngine` and `LSPEngine`

m-abulazm:
### What does this PR do?
Remove circular dependency between `LspEngine` and `TranspileEngine`

this dependency was easy to get rid of by introducing a factory

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/cli.py
+++ b/src/databricks/labs/lakebridge/cli.py
@@ -426,7 +426,7 @@ class _TranspileConfigChecker:
         #
 
         # Step 1: Check the transpiler config path.
-        engine: TranspileEngine | None = None
+        engine: TranspileEngine | None
         transpiler_config_path = self._config.transpiler_config_path
         if transpiler_config_path is not None:
             self._validate_transpiler_config_path(
@@ -435,6 +435,8 @@ class _TranspileConfigChecker:
             )
             path = Path(transpiler_config_path)
             engine = LSPEngine.from_config_path(path)
+        else:
+            engine = None
         del transpiler_config_path
 
         # Step 2: Check the source dialect, assuming it has been specified, and infer the transpiler config path if necessary.


src/databricks/labs/lakebridge/transpiler/transpile_engine_factory.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+
+from databricks.labs.lakebridge.transpiler.lsp.lsp_engine import LSPEngine
+from databricks.labs.lakebridge.transpiler.sqlglot.sqlglot_engine import SqlglotEngine
+from databricks.labs.lakebridge.transpiler.transpile_engine import TranspileEngine
+
+
+class TranspileEngineFactory:
+    @staticmethod
+    def build_transpile_engine(transpiler_config_path: Path) -> TranspileEngine:
```
sundarshankar89: Personally, I'm not against this; it helps with understanding; however, we lose the flexibility of DI to test this better. 

And also, SqlglotEngine will eventually be removed from our code base, in favour of pluggable transpilers and a newer reconciliation pattern.

sundarshankar89: @asnare thoughts?

m-abulazm: @sundarshankar89 since this is a static method, it can be passed as argument and would be the functional way of doing things instead of the current mocking approach

I can implement the change 
```
from collections.abc import Callable

class TranspileEngine:
    def _configure_transpiler_config_path(self, 
                                          source_dialect: str,
                                          factory_func: Callable[Path, TranspileEngine]
                                          ) -> TranspileEngine | None:
...
```

asnare (changes requested):
I think this is worth fixing, but it's simpler than what we have here. I just had a poke around, and the only callers to that factory method expect `LSPEngine`. With this in mind, we can:
1. Move that path validation check into the calling class: given the error it raises, that's where it belongs.
2. The caller can just invoke `LSPEngine.from_config_path(path)`.
3. The method we're refactoring here can just go away.

m-abulazm:
@asnare done as reviewed. 
but just to get this through: so `SqlglotEngine` is not used at the moment?

src/databricks/labs/lakebridge/transpiler/transpile_engine_factory.py
m-abulazm: resolved with https://github.com/databrickslabs/lakebridge/pull/1867#pullrequestreview-3062961251

asnare:
> @asnare done as reviewed. but just to get this through: so `SqlglotEngine` is not used at the moment?

Correct, for transpiling we don't use it at all at the moment. Historically we used only it, and the transition was a little chaotic so there's unfortunately a lot of tech debt to deal with.

src/databricks/labs/lakebridge/transpiler/lsp/lsp_engine.py
```diff
@@ -373,6 +373,11 @@ class LSPEngine(TranspileEngine):
 
     @classmethod
     def from_config_path(cls, config_path: Path) -> LSPEngine:
+        if not config_path.exists():
+            raise ValueError(
+                f"Error: Invalid value for '--transpiler-config-path': '{str(config_path)}', file does not exist."
+            )
+
```
asnare: Can we move this into cli?

The niggle I have here is that the error message is specific to the CLI, so raising it here doesn't seem right: this class doesn't really know it's being called by the CLI.

m-abulazm: that was very lazy on my part.
the callers already do that check. so I changed the error message in the caller.

asnare (approved):

gueniai (approved):
LGTM

PR #1877, opened 2025-07-29
[1745] Support snowflake arrays

### LINKED ISSUE #1745: [BUG]: Snowflake to Databricks ARRAY data type schema mismatch

### Category of Bug / Issue

Reconcile bug

### Current Behavior

Snowflake to Databricks schema comparison with `ARRAY` data type is not working. 

Of course there are some different concepts of this data type:
- A Snowflake `ARRAY` does not specify the data type of its elements.
- A Databricks `ARRAY` always specifies the data type of its elements: `ARRAY<STRING>`, `ARRAY<INT>` etc.

But there should be a possibility to match them at least on higher data type level, not diving in the array nested structure data types. Other way it is just impossible to get success on schema comparison.

(I saw there is #403 issue with close topic, but it is 1 year without any response. And here I tried to give more detailed and directed description)

Image

### Expected Behavior

For example Snowflake `ARRAY` and Databricks `ARRAY<STRING>` - are matched.

### Steps To Reproduce

To run reconcile for Snowflake and Databricks tables that has a column of `array` data type.

m-abulazm:
### What does this PR do?
Improve schema reconcilation to support snowflake arrays. reconciling snowflake arrays schema did not succeed as databricks arrays are typed and snowflake arrays are untyped.

this is solved by adding a second check that parses and converts the databricks SQL create table statement to source which will catch this corner case. e.g databricks `array<string>` converts to snowflake `array` and schema reconcilation succeeds. 

### Linked issues

Resolves #1745 

### Functionality

- [ ] added new CLI command
- [X] modified existing command: `databricks labs lakebridge reconcile`

### Tests

- [ ] manually tested
- [ ] added unit tests
- [X] added integration tests


gueniai (approved):
LGTM

PR #1890, opened 2025-08-01
Fix semi-broken unit tests on `main` for `installer.py`

asnare:
## Changes

This PR updates the fixture used by the installer unit tests, fixing an error introduced in #1861 whereby it accesses (and modifies) the user's `~/.databricks/labs/remorph-transpilers`. As a unit test, it should do this. Coincidentally the tests pass if there is nothing installed in that location, but can fail if there is. 

### Linked issues

Follows #1861

### Tests

- fixed unit tests

### DIFF
--- a/tests/unit/test_install.py
+++ b/tests/unit/test_install.py
@@ -55,6 +55,20 @@ PATH_TO_TRANSPILER_CONFIG = "/some/path/to/config.yml"
 def ws_installer() -> Generator[Callable[..., WorkspaceInstaller], None, None]:
 
     class TestWorkspaceInstaller(WorkspaceInstaller):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            # Ensure that the transpiler repository is mocked for unit tests instead of being the real thing.
+            if self._transpiler_repository == TranspilerRepository.user_home():
+                self._transpiler_repository = create_autospec(TranspilerRepository)
+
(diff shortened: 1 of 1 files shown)


tests/unit/test_install.py
```diff
@@ -55,6 +55,20 @@ def ws() -> WorkspaceClient:
 def ws_installer() -> Generator[Callable[..., WorkspaceInstaller], None, None]:
 
     class TestWorkspaceInstaller(WorkspaceInstaller):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            # Ensure that the transpiler repository is mocked for unit tests instead of being the real thing.
+            if self._transpiler_repository == TranspilerRepository.user_home():
```
m-abulazm, line 61: i find something like this better and better readable
```
import pytest
from unittest.mock import create_autospec, Mock

@pytest.fixture()
def transpiler_repository():
    return create_autospec(TranspilerRepository)

@pytest.fixture()
def test_installer(transpiler_repository, monkeypatch):
    installer = WorkspaceInstaller(repo=transpiler_repository)
    monkeypatch.setattr(installer, "install_bladebridge", Mock())
    monkeypatch.setattr(installer, "install_morpheus", Mock())
    return installer
```

this is also a sign that we are missing an interface somewhere and use DI e.g. `MorpheusInstaller`

I will not block this PR since mocking vs extending is a matter of taste

sundarshankar89 (approved):
LGTM, this moves things forward.

Issue #1866, opened 2025-07-25
[BUG]: TSQL Does not accept special character in Reconciler Column Mapping


williamweixibi:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Category of Bug / Issue

Reconcile bug

### Current Behavior

Actively using the lakebridge reconciler, they currently using TSQL as source (Synapse)
In their column mappings, from both their Synapse and Databricks side, column names contain special characters, and this is causing the reconciler to fail and is preventing them from using the module, is there a workaround to enable special character handling as input

they tried replacing the special character with a number on the Databricks target side and is still generating the error

### Expected Behavior

Show allow mapping columns with special character handling as input without crashing reconciler

### Steps To Reproduce

pick an arbitrary column that contains special character and run reconciler against it

### Relevant log output or Exception details

```shell
[ReconcileTableOutput(target_table_name=, source_table_name= status=StatusOutput(row=None, column=None, schema=None, aggregate=None),
exception_message='Runtime exception occurred while fetching data using
SELECT custnum AS ``cust#`` FROM : \n
[PARSE_SYNTAX_ERROR] Syntax error at or near cust. SQLSTATE: 42601 (line 1, pos 20)\n\n== SQL ==\nSELECT custnum AS ``cust#``
ReconcileOutput(recon_id=', results=[ReconcileTableOutput(target_table_name=, source_table_name=, status=StatusOutput(row=None, column=None, schema=None, aggregate=None),
exception_message='Runtime exception occurred while fetching data using
SELECT custnum AS cust# FROM  : \n
[PARSE_SYNTAX_ERROR] Syntax error at or near #
```

### Logs Confirmation

- [x] I ran the command line with `--debug`
- [x] I have attached the `lsp-server.log` under USER_HOME/.databricks/labs/remorph-transpilers/<converter_name>/lib/lsp-server.log

### Sample Query

```shell

```

### Operating System

Windows

### Version

latest via Databricks CLI

PR #1892, opened 2025-08-01
Refactor/recon/execute main

m-abulazm:
### What does this PR do?
This is a cherry pick of https://github.com/databrickslabs/lakebridge/pull/1886 since the refactoring can go to main and result in a smaller PR

some functions in recon are too big and pylint rejects and modifications to them due to variables count inside function and there are a few code smells like brain files that need to be refactored and split up in order to be more readable and extendable. This is just a start at better modularity and there is still more work to do in order to be better readable

this is a prerequisite to the work being done for https://github.com/databrickslabs/lakebridge/issues/1866 as this recon part is being extended with normalization 

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/reconcile/execute.py
```diff
@@ -154,7 +99,8 @@ def _trigger_reconcile_aggregates(
       after logging the error details.
     """
     try:
-        recon_output = reconcile_aggregates(
+        reconcile_config.report_type = "aggregate"
```
asnare, line 102: Is this something new?

src/databricks/labs/lakebridge/reconcile/execute.py
```diff
@@ -1,81 +1,26 @@
 import logging
```
sundarshankar89, line 1: This is good refactoring, but we have broken our documented execution instructions we just need to update them. not an issue.

src/databricks/labs/lakebridge/reconcile/utils.py
```diff
@@ -0,0 +1,38 @@
...
+from databricks.sdk import WorkspaceClient
+
+from databricks.labs.lakebridge.config import ReconcileMetadataConfig
+from databricks.labs.lakebridge.reconcile.connectors.source_adapter import create_adapter
+from databricks.labs.lakebridge.reconcile.exception import InvalidInputException
+from databricks.labs.lakebridge.reconcile.recon_config import Table
+from databricks.labs.lakebridge.transpiler.sqlglot.dialect_utils import get_dialect
+
+logger = logging.getLogger(__name__)
+
```
sundarshankar89, line 14: Can we reimagine this. utils like this will become a place holder for undefined scope with static functions. which is difficult to manage in future.

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
```diff
@@ -0,0 +1,249 @@
...
+            recon_process_duration=recon_process_duration,
+            record_count=reconciler.get_record_count(table_conf, reconciler.report_type),
+        )
+        if reconciler.report_type != "schema":
+            ReconIntermediatePersist(
+                spark=spark, path=utils.generate_volume_path(table_conf, reconcile_config.metadata_config)
+            ).clean_unmatched_df_from_volume()
+
+    @staticmethod
+    def verify_successful_reconciliation(
```
sundarshankar89, line 232: why can't we have this as private?

src/databricks/labs/lakebridge/reconcile/utils.py
m-abulazm: you are right and I will have a followup PR for that
would that be ok?

src/databricks/labs/lakebridge/reconcile/execute.py
m-abulazm: fixed

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
m-abulazm: it is shared and reused by recon aggregate

asnare (approved):
I'm okay with this is as a step forward. Some observations:

 - Breaking the bag of top-level functions up into groups encapsulated within a set of classes is fine.
 - I know we prefer one class per file in JVM-based languages, but in Python a file is a module and a module can contain multiple classes. This is often useful, especially if they function together as a group. (One class per module is often over-doing things.)
 - A class that contains only static methods is a little weird, especially if those methods all have a common subset of parameters. That's ripe for improvement by, for example, passing the `WorkspaceClient` into `__init__` and just using normal methods.
 - Similarly, in Python almost always it makes more sense to use class methods instead of static methods. (These don't have a counterpart in other languages, and tend to make it easier to test the methods without monkey-patching.)

src/databricks/labs/lakebridge/reconcile/execute.py
sundarshankar89: there will another file in reconcile automation as well, and also a dbc file. you will have to update them

docs/lakebridge/docs/reconcile/index.mdx
```diff
@@ -99,14 +99,14 @@ You can execute the reconciliation process by executing the below command in a n
...
-from databricks.labs.lakebridge.reconcile.execute import recon
+from databricks.labs.lakebridge.reconcile.trigger_recon_service import TriggerReconService
 from databricks.labs.lakebridge.reconcile.exception import ReconciliationException
 
 ws = WorkspaceClient(product="lakebridge", product_version=__version__)
 
 
-try:
-  result = recon(
+try:s
```
sundarshankar89:
```suggestion
try:
```

sundarshankar89 (changes requested):

src/databricks/labs/lakebridge/reconcile/execute.py
m-abulazm: no not really I did a full text search in the repo

src/databricks/labs/lakebridge/reconcile/execute.py
m-abulazm: yes but it does not affect the user. the old impl never checked it if the use triggered recon aggregate

the new impl that now shares code between recon aggregate and other recon needs this to be set correctly

m-abulazm:
> * I know we prefer one class per file in JVM-based languages, but in Python a file is a module and a module can contain multiple classes. This is often useful, especially if they function together as a group. (One class per module is often over-doing things.)

this is really about not having files with 1000+ lines. And multiple files make dependencies visible and lead to better code structure, modularity and maintainability. I see the python way of one file is for scripts not production code. 
I would follow JVM standards over python as it leads to better maintainable code

> * A class that contains only static methods is a little weird, especially if those methods all have a common subset of parameters. That's ripe for improvement by, for example, passing the `WorkspaceClient` into `__init__` and just using normal methods.

I understand your point - This is more functional programming than python. On the other hand, functions  usage has the advantage of avoiding local states.

> * Similarly, in Python almost always it makes more sense to use class methods instead of static methods. (These don't have a counterpart in other languages, and tend to make it easier to test the methods without monkey-patching.)

do you have an example for me?

sundarshankar89 (approved):
LGTM, we will fix forward document updates if necesary

Issue #1891, opened 2025-08-01
[BUG]: Installing BB transpiler fails if the backup directory already exists


asnare:
### Current Behavior

When installing the transpilers, a (temporary) backup of any existing transpiler is made to rollback to if something goes wrong. Making the backup fails if there is already a backup present. This can happen, for example, if a prior installation was interrupted.

### Expected Behavior

The maven installer (for Morpheus) already clears the backup ahead of time; the wheel installer (for Bladebridge) should do the same.

PR #1893, opened 2025-08-01
Fix `install-transpile` if a transpiler backup already exists

asnare:
## Changes

This PR:

 - Fixes an issue in the wheel installer whereby it would fail if there was already a backup directory (eg. `~/.databricks/labs/remorph-saved/`) present when it started.
 - Improves logging when installing transpilers, particularly when directories are being moved/deleted/restored.
 - Factors out the common code that was present in both the maven and wheel installers.

There's still a lot of cruft in place, in particular with respect to what the instance properties or return values mean. That notwithstanding I think this is an improvement.

### Linked issues

Resolves: #1891

### Functionality

### Tests

- existing integration tests


src/databricks/labs/lakebridge/install.py
```diff
@@ -75,6 +83,45 @@ def _store_product_state(cls, product_path: Path, version: str) -> None:
...
 
+    def _install_version_with_backup(self, version: str) -> Path | None:
+        """Install a specific version of the transpiler, with backup handling."""
+        logger.info(f"Installing Databricks {self._product_name} transpiler (v{version})")
+        product_path = self._repository.transpilers_path() / self._product_name
+        backup_path: Path | None = product_path.with_name(f"{product_path.name}-saved")
+        assert backup_path is not None
+        if backup_path.exists():
+            logger.debug(f"Existing backup of {self._product_name} installation found, removing: {backup_path}")
+            rmtree(backup_path)
```
m-abulazm: would refactor to e.g.` _calculate_fresh_backup_path()`

src/databricks/labs/lakebridge/install.py
```diff
@@ -75,6 +83,45 @@ def _store_product_state(cls, product_path: Path, version: str) -> None:
...
+        if product_path.exists():
+            logger.debug(f"Backing up existing {self._product_name} installation: {product_path} -> {backup_path}")
+            os.rename(product_path, backup_path)
+        else:
+            backup_path = None
```
m-abulazm: same e.g. `backup_exisiting_product_else_not()`

src/databricks/labs/lakebridge/install.py
```diff
@@ -75,6 +83,45 @@ def _store_product_state(cls, product_path: Path, version: str) -> None:
...
+        self._install_path = product_path / "lib"
+        self._install_path.mkdir(parents=True, exist_ok=True)
+        try:
+            result = self._install_version(version)
+        except (CalledProcessError, KeyError, ValueError) as e:
+            logger.error(f"Failed to install {self._product_name} transpiler (v{version})", exc_info=e)
+            result = False
```
m-abulazm: smaller method `_do_install()`

src/databricks/labs/lakebridge/install.py
```diff
@@ -75,6 +83,45 @@ def _store_product_state(cls, product_path: Path, version: str) -> None:
...
+        if result:
+            logger.info(f"Successfully installed {self._product_name} transpiler (v{version})")
+            self._store_product_state(product_path=product_path, version=version)
+            if backup_path is not None:
+                rmtree(backup_path)
+            self._product_path = product_path
+            return product_path
```
m-abulazm: e.g. `on_install_success()`

src/databricks/labs/lakebridge/install.py
```diff
@@ -75,6 +83,45 @@ def _store_product_state(cls, product_path: Path, version: str) -> None:
...
+        logger.debug(f"Removing path for failed {self._product_name} installation: {product_path}")
+        rmtree(product_path)
+        if backup_path is not None:
+            logger.debug(f"Restoring previous {self._product_name} installation: {backup_path} -> {product_path}")
+            os.rename(backup_path, product_path)
+        return None
```
m-abulazm: e.g. `on_install_failure`

m-abulazm (changes requested):
very useful PR and I added some comments to split the new method into smaller ones

asnare:
> very useful PR and I added some comments to split the new method into smaller ones

@m-abulazm:
The challenge with splitting in this case is that there's shared state that needs to be distributed to the methods, and the methods are tightly coupled: the implementations of the methods of depend on each other. Because they're only being called once this isn't a repetition issue: it's more about being able to understand the implementation and verify that it's working as intended.

With this in mind, I've pushed a change in f901656 that extracts the backup/restore behaviour out into a separate class (which holds the state). Do you feel it's an improvement? (I'm not sure, to be frank.)

Part of the challenge is that success/failure is reported by both return code _or_ whether an exception is thrown. This is existing code, which I'm trying to incrementally clean up, and drawing the boundaries of an incremental change is difficult: pulling on each thread unravels more.

src/databricks/labs/lakebridge/install.py
```diff
@@ -74,6 +137,31 @@ def _store_product_state(cls, product_path: Path, version: str) -> None:
...
+        product_path = self._repository.transpilers_path() / self._product_name
+        with _PathBackup(product_path) as backup:
+            self._install_path = product_path / "lib"
+            self._install_path.mkdir(parents=True, exist_ok=True)
+            try:
+                result = self._install_version(version)
+            except (CalledProcessError, KeyError, ValueError) as e:
+                logger.error(f"Failed to install {self._product_name} transpiler (v{version})", exc_info=e)
+                result = False
+            if result:
```
m-abulazm, line 155: new line before

src/databricks/labs/lakebridge/install.py
```diff
@@ -74,6 +137,31 @@ def _store_product_state(cls, product_path: Path, version: str) -> None:
...
+                self._store_product_state(product_path=product_path, version=version)
+                backup.commit()
+                self._product_path = product_path
+                return product_path
+            backup.rollback()
+        return None
+
+    @abc.abstractmethod
+    def _install_version(self, version: str) -> bool:
+        """Install a specific version of the transpiler, returning True if successful."""
```
m-abulazm, line 165: python uses `pass` for this

src/databricks/labs/lakebridge/install.py
```diff
@@ -42,9 +42,72 @@
...
+
+
 class TranspilerInstaller(abc.ABC):
-    def __init__(self, repository: TranspilerRepository) -> None:
+
+    _install_path: Path
+    """The path where the transpiler is being installed, once this starts."""
+
+    _product_path: Path
+    """The path where the transpiler is installed, after this has finished."""
```
m-abulazm: are these really needed? I would avoid this state state and no side effects

src/databricks/labs/lakebridge/install.py
asnare: Our pylint configuration doesn't allow `pass` here: it's unnecessary (because we have a docstring).

src/databricks/labs/lakebridge/install.py
asnare: They were present (and necessary) before this PR, here I've just added type annotations and documented their lifecycle.

That said, due to the changes in this PR we no longer need `_product_path` so I've removed it. Unfortunately `_install_path` is still necessary due to use elsewhere in the subclasses. I've added a TODO marker for it, because I expect it can also be removed when I remove the support for post-install hooks from this class. (That's the next set of changes I'll be making.)

m-abulazm: thanks

m-abulazm (approved):
LGTM

PR #1894, opened 2025-08-01
Refactor virtualenv setup when installing BB

asnare:
### What does this PR do?

This PR refactors the code responsible for setting up the virtualenv and installing the BB wheels during `install-transpile`. Changes include:

 - We use the `venv` module directly instead of invoking a subprocess to create the virtual environment. This ensures that errors are captured directly without having to worry about capturing them from the subprocess.
 - We invoke `pip` within the virtualenv in the manner [suggested by the documentation](https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program). This lets us eliminate some of the alternate code paths. In addition we don't use the shell to run the command, which resolves some quoting issues that were present.
 - ~If something does go wrong in a subprocess, we now log its output to help with diagnosing the problem.~ [Moved to #1898.]
 - Integration tests now fail earlier if there was an issue with the install, rather than having the test conversion fail with a weird error.

### Linked issues

Resolves some of the items in #1603.

### Functionality

### Tests

- existing integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/install.py
+++ b/src/databricks/labs/lakebridge/install.py
@@ -1,24 +1,23 @@
 import re
 import abc
 import dataclasses
-import shutil
-from json import loads, dump
 import logging
 import os
+import shutil
+import sys
+import webbrowser
+import xml.etree.ElementTree as ET
+from datetime import datetime, timezone
+from json import loads, dump
+from pathlib import Path
 from shutil import rmtree, move
 from subprocess import run, CalledProcessError
-import sys
 from typing import Any, cast
 from urllib import request
 from urllib.error import URLError, HTTPError
-import webbrowser
-from datetime import datetime, timezone
-from pathlib import Path
-import xml.etree.ElementTree as ET
 from zipfile import ZipFile
 
-from databricks.labs.blueprint.installation import Installation, JsonValue
-from databricks.labs.blueprint.installation import SerdeError
+from databricks.labs.blueprint.installation import Installation, JsonValue, SerdeError
 from databricks.labs.blueprint.installer import InstallState
 from databricks.labs.blueprint.tui import Prompts
 from databricks.labs.blueprint.wheels import ProductInfo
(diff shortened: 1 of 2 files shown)


src/databricks/labs/lakebridge/install.py
```diff
@@ -134,12 +139,19 @@ def _install_latest_version(self, version: str) -> Path | None:
             if backup_path.exists():
                 rmtree(backup_path)
             return result
-        except (CalledProcessError, ValueError) as e:
+        except CalledProcessError as e:
+            # Warning: if you end up here under the IntelliJ/PyCharm debugger, it can be because the debugger is
+            # trying to inject itself into the subprocess. Try disabling:
+            #   Settings | Build, Execution, Deployment | Python Debugger | Attach to subprocess automatically while debugging
```
asnare: @sundarshankar89: This is what you're looking for. ;)

src/databricks/labs/lakebridge/install.py
```diff
@@ -134,12 +139,19 @@ def _install_latest_version(self, version: str) -> Path | None:
...
             logger.error(f"Failed to install {self._pypi_name} v{version}", exc_info=e)
-            rmtree(self._product_path)
-            if backup_path.exists():
-                os.rename(backup_path, self._product_path)
-            return None
+            logger.debug(f"Output (stdout) from failed install command: {e.stdout.decode('utf-8', errors='replace')}")
+            logger.debug(f"Output (stderr) from failed install command: {e.stderr.decode('utf-8', errors='replace')}")
+        except ValueError as e:
+            logger.error(f"Failed to install {self._pypi_name} v{version}", exc_info=e)
+        rmtree(self._product_path)
```
m-abulazm: I am missing blank lines in this method so I can identify logical code blocks e.g. before lines 151, 136

asnare: Given this was an incidental change, I've moved it to #1898 so we can deal with it there and hopefully not block the rest of the changes.

m-abulazm: sure :) the changes are quite important

src/databricks/labs/lakebridge/install.py
```diff
@@ -148,86 +153,36 @@ def _unsafe_install_latest_version(self, version: str) -> Path | None:
...
+        context = builder.ensure_directories(venv_path)
+        logger.debug(f"Created virtual environment with context: {context}")
+        self._venv_exec_cmd = context.env_exec_cmd
+        self._site_packages = lib_path
 
-    def _locate_pip(self) -> Path:
-        return self._venv / "Scripts" / "pip3.exe" if sys.platform == "win32" else self._venv / "bin" / "pip3"
+    def _install_with_pip(self) -> None:
+        # Based on: https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program
+        # (But with venv_exec_cmd instead of sys.executable, so that we use the venv's pip.)
```
sundarshankar89, line 174: This is interesting.

PR #1900, opened 2025-08-04
Introduce `schema_fixture_factory`

m-abulazm:
### What does this PR do?
`Schema` will be extended to introduce normalization. as the constructor is used in a few places, the changes will be big so in order to have a smaller PR, this isolated change can go already to main. a fixture factory is introduced that the tests go through so we have a central place to extend later


tests/conftest.py
```diff
@@ -210,6 +210,10 @@ def report_tables_schema():
     return recon_schema, metrics_schema, details_schema
 
 
+def schema_fixture_factory(column_name: str, data_type: str) -> Schema:
```
sundarshankar89, line 213:
```suggestion
# Schema function is expected to be extended with normalisation attributes covered in PR#<XYZ>
def schema_fixture_factory(column_name: str, data_type: str) -> Schema:
```

PR #1907, opened 2025-08-06
Refactor LSP server launch code

asnare:
## Changes

This PR refactors the code that deals with figuring out how to launch the LSP server when running the `transpile` command, and mirrors the changes in #1894.

We now use a slightly different approach to before:

 - If there is a virtual environment associated with the transpiler, we ensure that its _script directory_ (where `python3`, etc. are installed) is placed on the search path first.
 - We resolve the first argument of the command line (the executable to run) using the search path. This eliminates some problematic platform-specific behaviour that the Python [documentation warns about](https://docs.python.org/3/library/subprocess.html#popen-constructor).

As a bonus: wheel-based plugins can now declare proper entry points, and they will now work.

Windows virtual environments introduce a few complications:

 - When creating or working with them, `use_symlinks` needs to be `False`. (The default is `True`.) This mimics the behaviour of the `pip` utility.
 - The `Scripts/` directory contains `python.exe` but _not_ `python3.exe`, which differs from on macOS/Linux where `bin/` contains both `python` and `python3`. Our BladeRunner transpiler specifies `python3` as the launch command, which means we need to have special handling for it to work.

### Linked issues

Resolves an item in #1603.

### Functionality

### Tests

- existing unit tests
- existing integration tests


tests/resources/lsp_transpiler/lsp_config.yml
```diff
@@ -6,7 +6,7 @@ remorph:
   environment:
     SOME_ENV: abc
   command_line:
-    - python
+    - python3
```
sundarshankar89: this fails in windows, so we need better way to manage this version.

asnare: That's interesting. :) Thanks!

gueniai (approved):
LGTM

PR #1912, opened 2025-08-06
[TECH DEBT]: Remove unnecessary `chdir()` during `transpile`

asnare:
## Changes

This PR eliminates an unnecessary call to `chdir()` when launching a LSP server. Aside from being unnecessary, the current working directory is a form of global state, assuming it can lead to code smells and concurrency issues (especially during testing).

A unit test is also removed: it's an example of code that assumed a specific working directory, and only passed cooincidentally. Rather than update it I removed it because the functionality it was testing is covered by other tests.

### Functionality

### Tests

- existing unit tests
- existing integration tests


m-abulazm (approved):
LGTM

PR #1920, opened 2025-08-11
Add identifier normalization support in `DataSource`

m-abulazm:
### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/tests/unit/reconcile/connectors/test_sql_server.py
+++ b/tests/unit/reconcile/connectors/test_sql_server.py
@@ -177,10 +177,11 @@ def test_get_schema_exception_handling():
     ):
         data_source.get_schema("org", "schema", "supplier")
 
+
 def test_normalize_identifier():
     engine, spark, ws, scope = initial_setup()
     data_source = TSQLServerDataSource(engine, spark, ws, scope)
 
     assert data_source.normalize_identifier("col1") == "`col1`"
-    assert data_source.normalize_identifier("[col1]") == "`col1`" # T-SQL delimiter
-    assert data_source.normalize_identifier("`col1`") == "`col1`" # ANSI SQL delimiter
+    assert data_source.normalize_identifier("[col1]") == "`col1`"  # T-SQL delimiter
+    assert data_source.normalize_identifier("`col1`") == "`col1`"  # ANSI SQL delimiter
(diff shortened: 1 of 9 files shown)


src/databricks/labs/lakebridge/reconcile/connectors/dialect_utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class DialectUtils:
+    _ANSI_IDENTIFIER_DELIMITER = "`"
+
+    @staticmethod
+    def normalize_identifier(
+        identifier: str, source_start_delimiter: str, source_end_delimiter: str
+    ) -> NormalizedIdentifier:
+        identifier = identifier.strip().lower()
```
sundarshankar89, line 36: Are you sure about this what if the input is already escaped?

tests/unit/reconcile/connectors/test_sql_server.py
```diff
@@ -176,3 +177,12 @@ def test_get_schema_exception_handling():
         ),
     ):
         data_source.get_schema("org", "schema", "supplier")
+
+
+def test_normalize_identifier():
+    engine, spark, ws, scope = initial_setup()
+    data_source = TSQLServerDataSource(engine, spark, ws, scope)
+
+    assert data_source.normalize_identifier("col1") == NormalizedIdentifier("`col1`", "[col1]")
```
sundarshankar89: test are little basic
Can you add other special characters few examples for TSQL are documented here

https://github.com/databrickslabs/morpheus/issues/305

src/databricks/labs/lakebridge/reconcile/connectors/dialect_utils.py
m-abulazm: good point. till now we lowercase every identifier and it works probably because we did not encounter case sensitive collations. but this needs to be handled properly anyway.

can we move this to a different PR? as this is the current behavior and let's implement case sensitivity as its own feature

sundarshankar89: Can you create a ticket so that we don't miss out on this

tests/unit/reconcile/connectors/test_sql_server.py
m-abulazm: implemented. pushing

src/databricks/labs/lakebridge/reconcile/connectors/dialect_utils.py
m-abulazm: yup https://github.com/databrickslabs/lakebridge/issues/1941

PR #1921, opened 2025-08-11
Support columns names having special characters in schema compare

### LINKED ISSUE #1866: [BUG]: TSQL Does not accept special character in Reconciler Column Mapping

m-abulazm:
### What does this PR do?
**Big Picture:** 
Add support in lakebridge reconciler for columns with special chars e.g. `Cust#` as at the moment reconciler will fail if such columns are passed. 

**This PR:**
Second step is to support the normalized columns names in schema compare. We need to update all sqlglot calls with normalized column names in case the user is using special chars. 

* Schema compare uses sqlglot which expects source compatible dialect so we cannot use ansi everywhere since e.g. snowflake or oracle use other characters and not backticks to delimit identifiers

This is a big change that had the implementation split between 6 PRs to make reviews manageable. this cannot go to main without the rest of the PRs. this is PR 2/6

### Linked issues

Resolves #1866 

### Functionality

- [ ] added new CLI command

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/reconcile/query_builder/base.py
+++ b/src/databricks/labs/lakebridge/reconcile/query_builder/base.py
@@ -4,6 +4,7 @@ from abc import ABC
 import sqlglot.expressions as exp
 from sqlglot import Dialect, parse_one
 
+from databricks.labs.lakebridge.reconcile.connectors.data_source import DataSource
 from databricks.labs.lakebridge.reconcile.exception import InvalidInputException
 from databricks.labs.lakebridge.reconcile.query_builder.expression_generator import (
     DataType_transform_mapping,
@@ -16,17 +17,12 @@ logger = logging.getLogger(__name__)
 
 
 class QueryBuilder(ABC):
-    def __init__(
-        self,
-        table_conf: Table,
-        schema: list[Schema],
-        layer: str,
-        engine: Dialect,
-    ):
+    def __init__(self, table_conf: Table, schema: list[Schema], layer: str, engine: Dialect, data_source: DataSource):
         self._table_conf = table_conf
         self._schema = schema
         self._layer = layer
         self._engine = engine
+        self._data_source = data_source
 
     @property
     def engine(self) -> Dialect:
--- a/src/databricks/labs/lakebridge/reconcile/reconciliation.py
+++ b/src/databricks/labs/lakebridge/reconcile/reconciliation.py
@@ -122,12 +122,12 @@ class Reconciliation:
         src_schema,
         tgt_schema,
     ):
-        src_hash_query = HashQueryBuilder(table_conf, src_schema, "source", self._source_engine).build_query(
-            report_type=self._report_type
-        )
-        tgt_hash_query = HashQueryBuilder(table_conf, tgt_schema, "target", self._source_engine).build_query(
-            report_type=self._report_type
-        )
+        src_hash_query = HashQueryBuilder(
+            table_conf, src_schema, "source", self._source_engine, self._source
+        ).build_query(report_type=self._report_type)
+        tgt_hash_query = HashQueryBuilder(
+            table_conf, tgt_schema, "target", self._source_engine, self._target
+        ).build_query(report_type=self._report_type)
         src_data = self._source.read_data(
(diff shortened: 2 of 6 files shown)


src/databricks/labs/lakebridge/reconcile/recon_output_config.py
```diff
@@ -31,7 +31,8 @@ class DataReconcileOutput:
 
 @dataclass
 class SchemaMatchResult:
-    source_column: str
```
bishwajit-db, line 34: Do we need the escaping to be displayed in the result? From what I understand, we escape the identifiers because running queries against the source with unescaped identifiers will cause them to fail. This is an implementation detail concerning how we manage the escaping. In my opinion, the result should only reflect what is exactly in the source.

m-abulazm: good catch

m-abulazm: will implement after merging https://github.com/databrickslabs/lakebridge/pull/1906

tests/integration/reconcile/test_schema_compare.py
```diff
@@ -80,6 +86,7 @@ def databricks_databricks_schema():
         schema_fixture_factory("col_num10", "decimal(10,1)"),
         schema_fixture_factory("col_dec", "decimal(20,2)"),
         schema_fixture_factory("col_numeric_2", "decimal(38,0)"),
+        schema_fixture_factory("col_escaped", "string", "`col_escaped`", "`col_escaped`"),
```
sundarshankar89: add more complex test examples not just escaped but with special characters to test the behaviour.

sundarshankar89 (changes requested):

tests/integration/reconcile/test_schema_compare.py
m-abulazm: done

src/databricks/labs/lakebridge/reconcile/recon_output_config.py
sundarshankar89: Just add a TODO so we can remove it later for this small nit else this PR is ready to be merged

sundarshankar89 (approved):
LGTM with small nit

src/databricks/labs/lakebridge/reconcile/recon_output_config.py
m-abulazm: done

PR #1937, opened 2025-08-14
improve test support of test mssql instance

m-abulazm:
### What does this PR do?
1) Makes integrations tests with sqlserver more straightforward and easier to develop by removing a patch that was needed
2) extends `DatabaseManager` with commit behavior if needed otherwise all database updates are rolledback

### Functionality

- [ ] added new CLI command

### Tests

- [ ] manually tested
- [X] added unit tests
- [X] added integration tests


sundarshankar89 (approved):
LGTM small nit:

tests/integration/assessments/test_pipeline.py
```diff
@@ -55,8 +49,8 @@ def python_failure_config():
     return config
 
 
-def test_run_pipeline(extractor, pipeline_config, get_logger):
-    pipeline = PipelineClass(config=pipeline_config, executor=extractor)
+def test_run_pipeline(test_sqlserver, pipeline_config, get_logger):
+    pipeline = PipelineClass(config=pipeline_config, executor=test_sqlserver)
     results = pipeline.execute()
     print("*******************\n")
```
sundarshankar89: small nit can you remove these lines

m-abulazm: done

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -40,8 +40,10 @@ def get_secret(self, key: str) -> str:
 class CredentialManager:
     def __init__(self, credential_loader: dict, secret_providers: dict):
         self._credentials = credential_loader
-        self._secret_providers = secret_providers
         self._default_vault = self._credentials.get('secret_vault_type', 'local').lower()
+        self._provider = secret_providers.get(self._default_vault)
+        if not self._provider:
```
asnare, line 45: Nit, not blocking:
```suggestion
        if self._provider is None:
```

Side-note: this code snippet is an example of something that's hard to verify by inspection. Because the type of `secret_providers` is only partially annotated, I can't strictly tell as a reviewer whether `None` is an allowed value in that dictionary, in which case this code isn't correct and to be safe it should be a `try/except` that remaps `KeyError`. However given the context I've:

 - Assumed that `None` is not a valid value.
 - Suggested the `is` check to be a little stricter, but it's not blocking because I assume there aren't values in the dictionary such as `0`, `""` or `False` that would mistakenly trigger the `ValueError`.

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -18,7 +18,7 @@ def _connect(self) -> Engine:
         pass
 
     @abstractmethod
-    def execute_query(self, query: str) -> Result[Any]:
+    def execute_query(self, query: str, commit: bool = False) -> Result[Any]:
```
asnare: Nit, not blocking, to force boolean argument with keyword:
```suggestion
    def execute_query(self, query: str, *, commit: bool = False) -> Result[Any]:
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -30,12 +30,15 @@ def __init__(self, config: dict[str, Any]):
     def _connect(self) -> Engine:
         raise NotImplementedError("Subclasses should implement this method")
 
-    def execute_query(self, query: str) -> Result[Any]:
+    def execute_query(self, query: str, commit: bool = False) -> Result[Any]:
```
asnare: Nit, not blocking, to force boolean argument with keyword:
```suggestion
    def execute_query(self, query: str, *, commit: bool = False) -> Result[Any]:
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -30,12 +30,15 @@ def __init__(self, config: dict[str, Any]):
...
         if not self.engine:
             raise ConnectionError("Not connected to the database.")
         session = sessionmaker(bind=self.engine)
         connection = session()
-        return connection.execute(text(query))
+        result = connection.execute(text(query))
+        if commit:
+            connection.commit()
```
asnare: This interface is problematic: in general it's not safe to access results from a query beyond the boundary of the transaction. On some databases this works, on others it doesn't, it depends a bit on some of the internal implementation details. So for a general interface this is something we need to avoid.

I see 2 paths to dealing with this, although maybe there are more:

 - Proactively fetch the results, so we safely end the transaction.
 - Defer the commit/rollback so that the caller can safely consume the results.

A final point: it's conventional to normally ensure that a transaction either commits or rolls back, partly to establish the boundary of the transaction as an invariant. Without this it becomes hard to reason about because depending on which branch is taken the boundary can be in different places.

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -82,9 +85,9 @@ class DatabaseManager:
     def __init__(self, db_type: str, config: dict[str, Any]):
         self.connector = _create_connector(db_type, config)
 
-    def execute_query(self, query: str) -> Result[Any]:
+    def execute_query(self, query: str, commit: bool = False) -> Result[Any]:
         try:
-            return self.connector.execute_query(query)
+            return self.connector.execute_query(query, commit)
```
asnare: See earlier comment: I'm worried about this interface because I don't think it's safe to access results after the transaction has ended. Can you look into this?

tests/integration/conftest.py
```diff
@@ -53,48 +55,26 @@ def mock_spark() -> SparkSession:
...
-                },
-            },
-        },
-    ):
-        yield
+@pytest.fixture()
+def sandbox_sqlserver_config():
+    env = TestEnvGetter(True)
+    db_url = env.get("TEST_TSQL_JDBC").removeprefix("jdbc:")
+    base_url, params = db_url.replace("jdbc:", "", 1).split(";", 1)
```
asnare: I think the `.replace()` bit here is redundant, given the `.removeprefix()` above?

tests/integration/conftest.py
```diff
@@ -53,48 +55,26 @@ def mock_spark() -> SparkSession:
...
-                    'redact_sql_pools_sql_text': False,
-                },
-            },
-        },
-    ):
-        yield
+@pytest.fixture()
+def sandbox_sqlserver_config():
+    env = TestEnvGetter(True)
+    db_url = env.get("TEST_TSQL_JDBC").removeprefix("jdbc:")
```
asnare, line 61: Here I think it's probably worth:

1. Splitting (max 3 parts) on `:`;
2. Pattern matching on `["jdbc", "sqlserver", rest]`.

The JDBC specification says that the format of the URI has to be: `jdbc:subprotocol:subname`

This means we can really only work on `surname` once we've verified `subprocotol` is what we expect (`"sqlserver"`). From there on we can assume it's in the SQL Server format…

tests/integration/conftest.py
```diff
@@ -53,48 +55,26 @@ def mock_spark() -> SparkSession:
...
-                    'exclude_spark_pools': False,
-                    'exclude_monitoring_metrics': False,
-                    'redact_sql_pools_sql_text': False,
-                },
-            },
-        },
-    ):
-        yield
+@pytest.fixture()
+def sandbox_sqlserver_config():
```
asnare: Please annotate the return type.

tests/integration/conftest.py
```diff
@@ -53,48 +55,26 @@ def mock_spark() -> SparkSession:
...
+    env = TestEnvGetter(True)
+    db_url = env.get("TEST_TSQL_JDBC").removeprefix("jdbc:")
+    base_url, params = db_url.replace("jdbc:", "", 1).split(";", 1)
+    url_parts = urlparse(base_url)
+    server = url_parts.hostname
+    query_params = dict(param.split("=", 1) for param in params.split(";") if "=" in param)
+    database = query_params.get("database", "")
```
asnare, line 66: What happens to the non-database properties in the URL? Are they just ignored?

tests/integration/conftest.py
```diff
@@ -53,48 +55,26 @@ def mock_spark() -> SparkSession:
...
+        "password": env.get("TEST_TSQL_PASS"),
+        "server": server,
+        "database": database,
+        "driver": "ODBC Driver 18 for SQL Server",
+    }
+    return config
+
+
+@pytest.fixture()
+def sandbox_sqlserver(sandbox_sqlserver_config):
```
asnare: Please annotate the return type.

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: I added missing type parameters and the check stays the same. does it look better now?

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: removed commit flag

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: removed commit flag

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: the previous tests never ran into this by luck. Now it wont happen because I changed the implementation to prefetch the results

tests/integration/conftest.py
m-abulazm: since this is a jdbc url; all query params are database properties. and some params will be ignored for now. if needed later, we can add them

I looked for a more industrial way of converting from jdbc url to pyodbc but did not find something in sqlalchemy at least because I didnt want extra dependencies

tests/integration/conftest.py
m-abulazm: I can add the check but I dont see the necessity of being defensive here since this env is created by "us"

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: excellent points. totally agree

> Proactively fetch the results, so we safely end the transaction.

adopted and did rename the method to `fetch` to be clear of what is happening here.
the previous impl was very problematic; we also never closed the session. this is handled now as well

do you find it better now?

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -163,15 +162,10 @@ def _save_to_db(self, result, step_name: str, mode: str, batch_size: int = 1000)
...
+            placeholders = ', '.join(['?' for _ in result.columns])
             insert_query = f"INSERT INTO {step_name} VALUES ({placeholders})"
 
-            # Fetch and insert rows in batches
-            while True:
-                rows = result.fetchmany(batch_size)
-                if not rows:
-                    break
-                conn.executemany(insert_query, rows)
+            conn.executemany(insert_query, result.rows)
```
m-abulazm, line 253: this will probably break. we should expose a more flexible api e.g. return the session

asnare: Ah, previously we only had one batch in flight at any time, now it's just bulk-inserting the lot. Is that your concern?

It's out-of-scope, but it also looks like we're in auto-commit mode and it's unclear to me whether `.executemany()` is running an `INSERT` for each row or a single `INSERT` with all of them.

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -30,12 +39,13 @@ def __init__(self, config: dict[str, Any]):
...
 
-    def execute_query(self, query: str) -> Result[Any]:
+    def fetch(self, query: str) -> FetchResult:
         if not self.engine:
             raise ConnectionError("Not connected to the database.")
-        session = sessionmaker(bind=self.engine)
-        connection = session()
-        return connection.execute(text(query))
+        Session = sessionmaker(self.engine)  # pylint: disable=invalid-name
+        with Session.begin() as session:  # pylint: disable=no-member
```
asnare: Pylint suppressions need special approval, so it's best to avoid them. Here I think you can just go with this?
```suggestion
        with sessionmaker(self.engine).begin() as session:
```

tests/unit/connections/test_database_manager.py
```diff
@@ -30,20 +30,37 @@ def test_mssql_connector(mock_mssql_connector):
...
-    result = db_manager.execute_query(query)
+    result = db_manager.fetch(query)
 
     assert result == mock_result
-    mock_connector_instance.execute_query.assert_called_once_with(query)
+    mock_connector_instance.fetch.assert_called_once_with(query)
+
+
+@patch('databricks.labs.lakebridge.connections.database_manager.MSSQLConnector')
+def test_fetch_commit(mock_mssql_connector):
```
asnare: So that this function is type-checked:
```suggestion
def test_fetch_commit(mock_mssql_connector) -> None:
```

src/databricks/labs/lakebridge/assessments/pipeline.py
m-abulazm: if a query returns thousands and thousands of data, we loaded it all into memory which might cause OOM errors

src/databricks/labs/lakebridge/discovery/tsql_table_definition.py
asnare, line 1: Something has gone wrong with this: it shouldn't be an empty file?

tests/integration/discovery/test_tsql_table_definition.py
asnare, line 1: Here's another empty file.

tests/unit/discovery/test_tsql_table_definition_service.py
asnare, line 1: And another.

m-abulazm: I did not merge this correctly. this is one of the three files removed by https://github.com/databrickslabs/lakebridge/pull/2075 removing

tests/integration/discovery/test_tsql_table_definition.py
m-abulazm: I did not merge this correctly. this is one of the three files removed by https://github.com/databrickslabs/lakebridge/pull/2075 removing

src/databricks/labs/lakebridge/discovery/tsql_table_definition.py
m-abulazm: I did not merge this correctly. this is one of the three files removed by https://github.com/databrickslabs/lakebridge/pull/2075 removing

tests/unit/connections/test_database_manager.py
m-abulazm: done

sundarshankar89 (approved):
LGTM

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -30,12 +39,13 @@ def __init__(self, config: dict[str, Any]):
...
 
-    def execute_query(self, query: str) -> Result[Any]:
+    def fetch(self, query: str) -> FetchResult:
         if not self.engine:
             raise ConnectionError("Not connected to the database.")
-        session = sessionmaker(bind=self.engine)
-        connection = session()
-        return connection.execute(text(query))
+
+        with Session(self.engine) as session, session.begin():
```
sundarshankar89, line 46: I'm ok with this as long as this method since we will only fetch limited data.

m-abulazm: yes this does not support large expensive queries. we should move to the stream api for such queries https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncConnection.stream

src/databricks/labs/lakebridge/connections/database_manager.py
asnare: Yup, that's much better. Thanks for putting the effort into solving it.

PR #1963, opened 2025-08-25
CLI entry points, enforce keyword arguments

asnare:
## Changes

This PR updates the CLI entry points so that they must be called with keyword arguments: this is how the `@command` plumbing works, so this makes it a bit clearer what's going on. (It also raises the bar to accidental breakage.)

### Tests

- existing unit tests
- existing integration tests


m-abulazm (approved):
Looks good

sundarshankar89 (approved):
LGTM

PR #1978, opened 2025-08-29
Upgrade transpilers during CLI upgrade

asnare:
## Changes

This PR updates the CLI install process so that during an install/upgrade of Lakebridge we:

 - Detect any installed transpilers and upgrade them to the latest version;
 - Run the transpiler configuration questions if the transpilers were updated.

The intention here is that during an upgrade we also want to ensure the transpilers are updated. Some limitations we face:

 - There's no dedicate hook (via labs.yml) for `databricks labs upgrade` versus `databricks labs install`; upgrades are treated as installations.
 - The existing transpiler installers check versions and don't do anything if the latest version is already installed.

Incidental changes needed for this to work and be testable:

 - There is a lot of transpiler-specific code that was in the main `install.py` module; this PR moves it into the `transpiles` module.
 - The existing `TranspilerInstaller` class was renamed to `ArtifactInstaller`: it is used for installing PyPi or Maven artefacts.
 - The code for installing Bladebridge and Morpheus has been moved out of methods on the workspace installer into its own hierarchy, based on a new `TranspilerInstaller`.
 - Accompanying tests were moved as well.

### Linked issues

Resolves #1386

### Functionality

- [X] modified existing commands:

      - `databricks labs install lakebridge`
      - `databricks labs upgrade lakebridge`

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/install.py
+++ b/src/databricks/labs/lakebridge/install.py
@@ -1,22 +1,10 @@
-import re
-import abc
 import dataclasses
 import logging
 import os
-import shutil
-import sys
-import venv
 import webbrowser
-import xml.etree.ElementTree as ET
-from datetime import datetime, timezone
-from json import loads, dump
+from collections.abc import Set, Callable, Sequence
 from pathlib import Path
-from shutil import rmtree, move
-from subprocess import run, CalledProcessError
-from typing import Any, Literal, cast
-from urllib import request
-from urllib.error import URLError, HTTPError
-from zipfile import ZipFile
+from typing import Any, cast
 
 from databricks.labs.blueprint.installation import Installation, JsonValue, SerdeError
 from databricks.labs.blueprint.installer import InstallState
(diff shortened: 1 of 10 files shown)


src/databricks/labs/lakebridge/install.py
```diff
@@ -447,123 +93,36 @@ def run(
...
-                return True
-
-    def install_artifact(self, artifact: str):
+        upgraded = False
+        for transpiler_installer in self._transpiler_installers:
+            name = transpiler_installer.name
+            if name in installed_transpilers:
+                logger.info(f"Checking for {name} upgrades...")
+                upgraded |= transpiler_installer.install()
+        # If we upgraded anything, the configuration process needs to run again.
```
sundarshankar89, line 107: I like this nudge

tests/resources/lsp_transpiler/test-lsp-server.log
```diff
@@ -0,0 +1,43 @@
+INFO:pygls.feature_manager:Registered builtin feature exit
```
sundarshankar89: @asnare do we need this log checked in?

sundarshankar89 (changes requested):

tests/resources/lsp_transpiler/test-lsp-server.log
asnare: @sundarshankar89: Thanks for picking that up, indeed it shouldn't be there. (Logs being dropped in the source tree is something I noted in #1603, I should have noticed it here.)

src/databricks/labs/lakebridge/install.py
asnare: It's more than a nudge, it's also necessary because after an update the previous configuration might not be valid any more (eg. dialect removed), so this forces it into a valid state.

tests/unit/test_install.py
```diff
@@ -1358,118 +1346,152 @@ def test_runs_and_stores_choice_config_option(
...
+    # The setup here is:
+    #   - 'foo' and 'bar' are installed already.
+    #   - 'bar' and 'baz' are known transpilers.
+    # It should therefore try to install/upgrade bar but _not_ baz.
+
+    mock_repository = create_autospec(TranspilerRepository)
+    mock_repository.all_transpiler_names.return_value = {"foo", "bar"}
+    ctx = ApplicationContext(ws).replace(
+        product_info=ProductInfo.for_testing(LakebridgeConfiguration),
+        prompts=(MockPrompts({r"Do you want to override the existing installation?": "no"})),
```
m-abulazm, line 1391: is override different than upgrade?

tests/unit/transpiler/test_installers.py
```diff
@@ -0,0 +1,109 @@
...
+
+import pytest
+
+from databricks.labs.lakebridge.transpiler.installers import ArtifactInstaller, MorpheusInstaller
+
+
+def test_store_product_state(tmp_path) -> None:
+    """Verify the product state is stored after installing is correct."""
+
+    class MockArtifectInstaller(ArtifactInstaller):
```
m-abulazm:
```suggestion
    class MockArtifactInstaller(ArtifactInstaller):
```

m-abulazm: is this Dutch? :)

src/databricks/labs/lakebridge/install.py
```diff
@@ -447,123 +93,36 @@ def run(
...
+        if upgraded:
+            config = self.configure("transpile")
```
m-abulazm, line 109: I understand that we need to reconfigure. but this really complicates the upgrade user journey - a log to the user is a better UX. 
or we dont upgrade lakebridge and ask user to rerun `upgrade` with `--upgrade-transpilers`

tests/unit/transpiler/test_installers.py
asnare: Well spotted, fixed.

tests/unit/test_install.py
asnare: The settings are stored in the user-workspace, and this question is about whether the user wants to update them or not. This happens during a normal install of the transpilers (if they already have a configuration), and also here when we re-run the configuration questionnaire. (If they say no, the existing configuration is left as-is and no more questions are asked.)

src/databricks/labs/lakebridge/install.py
asnare: I suspect that if we log something, it will be ignored.

Unfortunately forcing them to re-run with a flag is not possible: the `upgrade` command we're hooked into is part of the Databricks CLI and outside the control of this tool and arguments are not passed in. (In fact we can't even distinguish between a fresh install and an upgrade, which is why we have this logic here.)

The goal here is to ensure the transpilers (if present) are upgraded along with the `lakebridge` CLI, with the minimum friction possible. We assume that when doing this users will want the latest version of the transpilers as well, without having to run additional commands.

sundarshankar89 (approved):
LGTM..

This improves the process but has potential to refactor further incrementally one step at a time.

asnare:
> If there already exists a config file, why do we go through the transpiler installation questions again?

If the transpilers are updated the set of available options can change, in a similar manner to during `install-transpile`.

Using `install-transpire` is the manual process for upgrading transpilers, and I've implemented this as a conditional action when the CLI is being upgraded: if our transpilers are already installed, trigger `install-transpire` to also force them to upgrade.

Stepping back a bit there's definitely room to improve the configuration process. Aside from needing a machine-compatible flow for the UI:

 - The first question isn't great: _Do you want to override the existing configuration?_  
   The term _override_ is weird, and ordinary users won't know what this means or how to confidently answer it. They will just hit enter to accept the default ("No").
   Maybe `s/override/update/g`.
 - We don't do anything to check if a configuration is valid or not, which would help some of these journeys: if it's invalid we shouldn't be asking the first question and probably need to force the questionnaire.
 - When updating an existing configuration, arguably the default should be the current value (if valid).
 - The whole 'set it later' thing is awkward. "_Provide_ it later" maybe?
 - There's no non-interactive mode, needed for scripting and other use-cases.

tests/unit/test_install.py
m-abulazm: ok got it. so the user installs/upgrades the transpilers. then choose to reconfigure if they want

src/databricks/labs/lakebridge/install.py
m-abulazm: https://github.com/databrickslabs/lakebridge/pull/1978#discussion_r2321359962 explains it to me.
after an upgrade, configure runs but the user can still "opt-out"

another approach without the flag or log is to prompt the user: e.g. `Found new version of transpiler. Upgrade [Y/n]`

PR #1982, opened 2025-09-02
Add transpiler version to telemetry

m-abulazm:
### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/cli.py
```diff
@@ -115,6 +111,31 @@ def transpile(
     print(json.dumps(result))
 
 
+def _add_user_agent_extras(
```
gueniai: I recommend to rename this to `_add_user_agent_extras_transpile` (or something like that), since we also need to set a few extras on the `analyze` command

m-abulazm: good point. done

PR #1989, opened 2025-09-04
Fix missing user agent extras

m-abulazm:
### What does this PR do?
Recreate `WorkspaceClient` on demand in our app to get around `user_agent_extras` limitations

we use `user_agent_extras` to pass telemetry information. This is provided by the Databricks SDK with the drawback that it gets evaluated only when `_BaseClient` gets created and not on every http call. 

In lakebridge this initialization happens only once because `blueprint` takes care of initializing the `WorkspaceClient` at the beginning of the command. when we change the extras later in the app. it wont take effect. `cmd` extra works because `blueprint` sets it before initializing `WorkspaceClient`.

This is probably a bug in sdk and should be fixed by implementing proper handling of the user agent extras in `_BaseClient`

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [X] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/contexts/application.py
+++ b/src/databricks/labs/lakebridge/contexts/application.py
@@ -140,9 +140,8 @@ class ApplicationContext:
         return LakebridgeAnalyzer(self.current_user, self.prompts, is_debug)
 
     def add_user_agent_extra(self, key: str, value: str) -> None:
-        new_config = self._ws.config.with_user_agent_extra(key, value)
+        new_config = self._ws.config.copy().with_user_agent_extra(key, value)
         logger.debug(f"Added User-Agent extra {key}={value}")
 
-        # Recreate the WorkspaceClient from the same class to preserve type information
-        new_client = type(self._ws)(config=new_config)
-        self._ws = new_client
+        # Recreate the WorkspaceClient from the same type to preserve type information
+        self._ws = type(self._ws)(config=new_config)


src/databricks/labs/lakebridge/contexts/application.py
```diff
@@ -34,9 +35,10 @@ def replace(self, **kwargs):
             self.__dict__[key] = value
         return self
 
-    @cached_property
+    @property
     def workspace_client(self) -> WorkspaceClient:
-        return self._ws
+        w = WorkspaceClient(product="lakebridge", product_version=__version__)
```
sundarshankar89: why should we recreate the workspace client again?

sundarshankar89: we should check existing and add the product and product version like we do in `verify_workspace_client`

sundarshankar89 (changes requested):

src/databricks/labs/lakebridge/contexts/application.py
m-abulazm: if it is not recreated, the new value of `user_agent_extras` will not be used because the SDK `_BaseClient` evaluates and caches user_agent_extras only during its creation.

This means any later updates or changes to `user_agent_extras` after the client has been instantiated will be ignored in subsequent HTTP requests, since the values are not re-evaluated for each call but rather fixed for the lifetime of that client instance

sundarshankar89: ```return self._ws```

where we are using the client,
we need to `w.config.with_user_agent_extra("cmd", "execute-transpile")`

sundarshankar89 (changes requested):

src/databricks/labs/lakebridge/contexts/application.py
```diff
@@ -34,9 +35,10 @@ def replace(self, **kwargs):
             self.__dict__[key] = value
         return self
 
-    @cached_property
+    @property
```
asnare, line 37: This is correct; it should never have been `@cached_property`.

src/databricks/labs/lakebridge/contexts/application.py
```diff
@@ -34,9 +35,10 @@ def replace(self, **kwargs):
...
     def workspace_client(self) -> WorkspaceClient:
-        return self._ws
+        w = WorkspaceClient(product="lakebridge", product_version=__version__)
+        return w
```
asnare: I think this should be left as-is:
```suggestion
        return self._ws
```

Instead, I would prefer it if we added a method to `ApplicationContext` that updates the user-agent, something like:
```python
    def add_user_agent_extra(self, key: str, value: str) -> None:
        ....
```
This should:
1. Create a new `client.Config()` instance using the `.with_user_agent_extra(key, value)` method.
2. Instantiate a new `WorkspaceClient` using the new `Config` and assign it to `self._ws` so that the property just returns the new instance.

asnare (changes requested):
I've suggested a way to do that should ensure we re-use the same `WorkspaceClient` instance as much as possible.

In addition it should obviate the need for most of the changes in the unit tests. (In tests, we really want to avoid monkey-patching as much as possible.)

src/databricks/labs/lakebridge/contexts/application.py
m-abulazm: implemented

src/databricks/labs/lakebridge/contexts/application.py
m-abulazm: done

tests/unit/test_cli_analyze.py
```diff
@@ -11,13 +11,15 @@
 # TODO: These should be moved to the integration tests.
 
 
+@patch("databricks.labs.lakebridge.contexts.application.ApplicationContext.add_user_agent_extra", new=MagicMock())
```
sundarshankar89: why do you need to patch given the config is already inside `workspace_client` and we already have a mock for it.

All we need is to add an additional assert may be to check the value for that key?

sundarshankar89 (changes requested):

tests/unit/test_cli_analyze.py
m-abulazm: because `add_user_agent_extra()` recreates WorkspaceClient so we get a real WorkspaceClient not the mock anymore

asnare: I'm testing a simpler fix, hold on.

src/databricks/labs/lakebridge/contexts/application.py
```diff
@@ -70,7 +70,7 @@ def recon_config(self) -> ReconcileConfig | None:
     def remorph_config(self) -> LakebridgeConfiguration:
         return LakebridgeConfiguration(transpile=self.transpile_config, reconcile=self.recon_config)
 
-    @cached_property
+    @property
```
asnare, line 73: Also nicely spotted.

src/databricks/labs/lakebridge/contexts/application.py
```diff
@@ -138,3 +138,11 @@ def upgrades(self):
     def analyzer(self):
         is_debug = logger.getEffectiveLevel() == logging.DEBUG
         return LakebridgeAnalyzer(self.current_user, self.prompts, is_debug)
+
+    def add_user_agent_extra(self, key: str, value: str) -> None:
+        with_user_agent_extra(key, value)
```
asnare: I don't think we can (or should) update the global extras here:

 - It's not necessary: our extras are already available via the API client.
 - Integration (and some unit) tests end up all confused, because it's a side-effect that accumulates across tests without properly being reset between tests. This in turn leads to tests injecting noise upstream.
 - Arguably the scope of the call should be limited to the `ApplicationContext` that we made the call on. (If we create a new `ApplicationContext` it might be surprising that it inherits changes enacted on another instance.)

I understand the goal was to ensure that clients created via some other mechanism end up with the extra pieces. However I think the above outweighs this goal.

src/databricks/labs/lakebridge/contexts/application.py
```diff
@@ -138,3 +138,11 @@ def upgrades(self):
...
+        new_config = self._ws.config.with_user_agent_extra(key, value)
+        logger.debug(f"Added User-Agent extra {key}={value}")
+
+        new_client = WorkspaceClient(config=new_config)
+        self._ws = new_client
```
asnare: I understand this is annoying for the situation with mocks, but I think we can use this?
```suggestion
        # Copy the existing client with the new config, allowing for it possibly being a mock.
        self._ws = type(self._ws)(config=new_config)
```

tests/unit/test_cli_analyze.py
asnare: I don't think these are needed if we clone as described above.

asnare (changes requested):
It's looking much better, thanks. I think we're very close now. :)

src/databricks/labs/lakebridge/contexts/application.py
```diff
@@ -138,3 +138,11 @@ def upgrades(self):
     def analyzer(self):
         is_debug = logger.getEffectiveLevel() == logging.DEBUG
         return LakebridgeAnalyzer(self.current_user, self.prompts, is_debug)
+
+    def add_user_agent_extra(self, key: str, value: str) -> None:
+        with_user_agent_extra(key, value)
+        new_config = self._ws.config.with_user_agent_extra(key, value)
```
asnare: One thing I just noticed is that `with_user_agent_extra()` modifies the configuration rather than returning a new instance. As such I think I'd prefer to see this as:
```suggestion
        new_config = self._ws.config.copy().with_user_agent_extra(key, value)
```

src/databricks/labs/lakebridge/contexts/application.py
m-abulazm: agree and changed

src/databricks/labs/lakebridge/contexts/application.py
m-abulazm: done

tests/unit/test_cli_analyze.py
m-abulazm: not needed anymore

asnare (approved):

sundarshankar89 (approved):
LGTM

PR #1994, opened 2025-09-05
Improve handling of encoding-related errors (#1858)

### LINKED ISSUE #1858: [FEATURE]: Improve handling of encoding-related errors

Transpiler

### Problem statement

Transpiling (source) files that are not encoded "correctly" leads to an exception and crashes the process.

An example of this would be when `LANG=en_US.UTF-8` but sources are encoded as Latin1 (or something else): these will fail to decode as UTF-8 if any non-ASCII characters are present.

### Proposed Solution

Instead of crashing we should:

 - Handle it as an error, warning the user that the file is not encoded as expected.
 - Continue to transpile other files.

### Additional Context

When transpiling we read files and deal with their encoding:

 - XML: according to the rules in the XML specification.
 - Files that have a Unicode BOM at the start: decoded as indicated by the BOM.
 - Otherwise we use the default "platform" encoding, normally specified via the `LANG`/`LC_*` environment variables.

On Windows, the locale is not controlled by the `LANG`/`LC_*` environment variables. To check the current locale, use:
```python
python3 -c 'import locale; print(locale.getpreferredencoding())'
```

As a workaround, including on Windows, when the locale is _not_ Unicode-aware, set the environment variable: `PYTHONUTF8=1`  
(Python will then assume that all files are encoded as UTF-8.)

BrianDeacon:
### What does this PR do?
Fixes transpiler crashes when encountering files with unsupported or mismatched encodings. Previously, the transpiler would crash with `UnicodeDecodeError` or `LookupError` when processing files that are not encoded as expected (e.g., Latin-1 files with non-ASCII characters when system `LANG=en_US.UTF-8`). Now these errors are handled gracefully, warning the user about problematic files and continuing to transpile other files.

- Added exception handling in `_process_one_file()` to catch `UnicodeDecodeError` and `LookupError`
- `UnicodeDecodeError`: Handles file content that doesn't match expected encoding
- `LookupError`: Handles XML files declaring unknown/invalid encoding names
- Errors are logged as warnings and reported via `TranspileError` with severity ERROR
- Transpilation continues with remaining files instead of crashing entirely

- Only catches specific encoding-related exceptions to avoid masking other system errors
- Files with encoding issues are counted as "processed" but with 0 successful queries
- Error messages include the original exception details for debugging

### Linked issues
Resolves #1858

### Functionality
- [X] modified existing command: `databricks labs lakebridge transpile`

### Tests
- [X] manually tested
- [X] added unit tests

Added comprehensive test coverage for encoding error scenarios.

### DIFF
--- a/src/databricks/labs/lakebridge/transpiler/execute.py
+++ b/src/databricks/labs/lakebridge/transpiler/execute.py
@@ -87,7 +87,21 @@ async def _process_one_file(context: TranspilingContext) -> tuple[int, list[Tran
     # local platform encoding.
     sniff_xml_encoding = input_path.suffix.lower() == ".xml"
 
-    source_code = read_text(input_path, detect_xml=sniff_xml_encoding)
+    try:
+        source_code = read_text(input_path, detect_xml=sniff_xml_encoding)
+    except (UnicodeDecodeError, LookupError) as e:
+        # UnicodeDecodeError can be thrown by bad encodings. LookupError can be thrown
+        # when detect_xml is True and the XML specifies an unknown encoding.
+        error = TranspileError(
+            code="encoding-error",
+            kind=ErrorKind.INTERNAL,
+            severity=ErrorSeverity.ERROR,
+            path=context.input_path,
+            message=f"File encoding error: {e}. File may use or specify an unsupported encoding.",
+        )
+        logger.warning(f"Skipping file {input_path} due to encoding error: {e}")
+        return 0, [error]
+
     context = dataclasses.replace(context, source_code=source_code)
 
     transpile_result = await _transpile(
--- a/tests/unit/transpiler/test_execute.py
+++ b/tests/unit/transpiler/test_execute.py
@@ -443,6 +443,96 @@ def test_server_decombines_workflow_output(mock_workspace_client, lsp_engine, tr
         assert any(Path(output_folder).glob("*.json")), "No .json file found in output_folder"
 
 
+def test_encoding_error_unicode_decode_error(input_source, output_folder, mock_workspace_client: WorkspaceClient):
+    """Test UnicodeDecodeError handling when file encoding doesn't match expected encoding."""
+    # Create a file with Latin-1 encoding containing non-ASCII characters
+    # When read_text() tries to decode this as UTF-8, it will fail with UnicodeDecodeError
+    problematic_file = input_source / "latin1_file.sql"
+    with open(problematic_file, "wb") as f:
+        # Write Latin-1 encoded content with non-ASCII characters that will fail as UTF-8
+        f.write("SELECT 'héllo wörld' AS greeting;".encode("latin-1"))
+
+    transpile_config = TranspileConfig(
+        transpiler_config_path=None,
+        source_dialect="identity",
+        input_source=str(problematic_file),
+        output_folder=str(output_folder),
+        skip_validation=True,
+    )
+
+    status, errors = transpile(mock_workspace_client, IdentityTranspileEngine(), transpile_config)
+
+    # Verify error handling
+    assert status.get("total_files_processed") == 1  # File was processed (but failed)
+    assert status.get("total_queries_processed") == 0  # No queries successfully processed
+    assert len(errors) == 1
+    assert errors[0].code == "encoding-error"
+    assert errors[0].severity == ErrorSeverity.ERROR
+    assert "codec can't decode" in errors[0].message
+
+
+def test_encoding_error_lookup_error(input_source, output_folder, mock_workspace_client: WorkspaceClient):
+    """Test LookupError handling when XML file declares an unknown encoding."""
+    # Create an XML file that declares an invalid encoding name
+    # When read_text() tries to use this encoding, it will fail with LookupError
+    xml_file = input_source / "invalid_encoding.xml"
+    with open(xml_file, "wb") as f:
+        # XML declaration with non-existent encoding - will trigger LookupError
+        xml_content = '<?xml version="1.0" encoding="definitely-invalid-codec"?><root>data</root>'
+        f.write(xml_content.encode("utf-8"))
+
+    transpile_config = TranspileConfig(
+        transpiler_config_path=None,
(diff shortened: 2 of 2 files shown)


BrianDeacon:
Not sure if you guys are using the same process as in https://github.com/databrickslabs/tempo, but I signed off on the databricks CLA already to get cleared to contribute there.  It was this commit for reference: https://github.com/databrickslabs/tempo/commit/74c2b07b381f8eef95f6cb894967dff5e52a0d94

I think I got the gpg signing right. First time for me. Let me know if there's any other process we need to do.

:)

src/databricks/labs/lakebridge/transpiler/execute.py
```diff
@@ -87,7 +87,21 @@ async def _process_one_file(context: TranspilingContext) -> tuple[int, list[Tran
...
 
-    source_code = read_text(input_path, detect_xml=sniff_xml_encoding)
+    try:
+        source_code = read_text(input_path, detect_xml=sniff_xml_encoding)
+    except (UnicodeDecodeError, LookupError) as e:
+        # UnicodeDecodeError can be thrown by bad encodings. LookupError can be thrown
+        # when detect_xml is True and the XML specifies an unknown encoding.
+        error = TranspileError(
+            code="encoding-error",
+            kind=ErrorKind.INTERNAL,
```
asnare, line 97: I'm on the fence about this, but I think this is fine for now.

We don't really document these, but `INTERNAL` in other cases typically refers to an error that occurs due to a bug or some other error outside the control of the user. In this case however it's a problem with the input we've received from the user. I don't think it's worth introducing a new category, and the only other category of error that might be relevant is `PARSING`.

tests/unit/transpiler/test_execute.py
```diff
@@ -443,6 +443,96 @@ def test_server_decombines_workflow_output(mock_workspace_client, lsp_engine, tr
         assert any(Path(output_folder).glob("*.json")), "No .json file found in output_folder"
 
 
+def test_encoding_error_unicode_decode_error(input_source, output_folder, mock_workspace_client: WorkspaceClient):
```
asnare, line 446: Nit: I think `tmp_path` would be fine here, for this and the other tests: `input_source` sets up a lot of different inputs, and we don't need (or care) about them for what we're testing here.

tests/unit/transpiler/test_execute.py
```diff
@@ -443,6 +443,96 @@ def test_server_decombines_workflow_output(mock_workspace_client, lsp_engine, tr
...
+    """Test UnicodeDecodeError handling when file encoding doesn't match expected encoding."""
+    # Create a file with Latin-1 encoding containing non-ASCII characters
+    # When read_text() tries to decode this as UTF-8, it will fail with UnicodeDecodeError
+    problematic_file = input_source / "latin1_file.sql"
+    with open(problematic_file, "wb") as f:
+        # Write Latin-1 encoded content with non-ASCII characters that will fail as UTF-8
+        f.write("SELECT 'héllo wörld' AS greeting;".encode("latin-1"))
```
asnare, line 453: There's nothing wrong with this, although I tend to prefer to specify the encoding at the file level (during `open()`) rather than individual write, mainly because normally the encoding remains the same for the entire file.

tests/unit/transpiler/test_execute.py
```diff
@@ -443,6 +443,96 @@ def test_server_decombines_workflow_output(mock_workspace_client, lsp_engine, tr
...
+
+    transpile_config = TranspileConfig(
+        transpiler_config_path=None,
+        source_dialect="identity",
+        input_source=str(problematic_file),
+        output_folder=str(output_folder),
+        skip_validation=True,
+    )
+
+    status, errors = transpile(mock_workspace_client, IdentityTranspileEngine(), transpile_config)
```
asnare, line 463: The call to `transpile()` will assume the user's locale when reading files, which this test assumes will always be `UTF-8`. Although that's the case for Linux/macOS, on Windows the encoding is often `Latin1` which means this test will fail.

Behaviour of this test[^1] can be checked under macOS by setting this in the environment:
```
LC_CTYPE=en_US.ISO8859-1
```
(Verify by inspecting `locale.getpreferredencoding()`.)

From memory it's difficult to inject/alter the value of `locale.getpreferredencoding()` during a test, which somewhat limits what we can do here, but I think it's a combination of:

 - Ensure we have a version of the test that runs for the common values of `locale.getpreferredencoding()`.
 - Mark as skipped when running on the 'wrong' value.

A heavier lift, not really possible within this PR, would be to plumb the encoding through so that the test can control the value.

[^1]: Somewhat embarrassingly, there are some other unit tests that fail if the local encoding isn't UTF-8. But let's not make it worse.

tests/unit/transpiler/test_execute.py
```diff
@@ -443,6 +443,96 @@ def test_server_decombines_workflow_output(mock_workspace_client, lsp_engine, tr
...
+    # Verify error handling
+    assert status.get("total_files_processed") == 1  # File was processed (but failed)
+    assert status.get("total_queries_processed") == 0  # No queries successfully processed
+    assert len(errors) == 1
+    assert errors[0].code == "encoding-error"
+    assert errors[0].severity == ErrorSeverity.ERROR
+    assert "encoding" in errors[0].message
+
+
+def test_encoding_error_continues_with_other_files(input_source, output_folder, mock_workspace_client):
```
asnare, line 503: Can we type-hint this signature? That ensures the body is type-checked.

asnare (changes requested):
@BrianDeacon: Thanks for the contribution, it looks good. I appreciate in particular the tests that cover the changed behaviour.

I've noted some nits, but the main thing I think needs to be looked at is `test_encoding_error_unicode_decode_error` when the user/system locale encoding isn't UTF-8. Are you able to look at this?

(Some other tests also fail in this situation, I'll address those myself.)

asnare:
@BrianDeacon I extended this PR as #2059, and the work in here has been merged although because it was squashed GitHub isn't aware of that. Thanks for your contribution!

PR #1996, opened 2025-09-08
Add more tests to `test_cli_transpile`

m-abulazm:
### What does this PR do?
Add more test cases for the cli `transpile` command. This makes sure bugfix https://github.com/databrickslabs/lakebridge/pull/1985 is well-tested and does not regress

* change tests to use `TranspilerRepository` correctly as the current setup did not reflect real usage
* Add test cases when the config is missing that was fixed in production in #1985 

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -283,7 +283,7 @@ argument-naming-style = "snake_case"
 
 # Regular expression matching correct argument names. Overrides argument-naming-
 # style. If left empty, argument names will be checked with the set naming style.
-argument-rgx = "[a-z_][a-z0-9_]{2,33}$"
+argument-rgx = "[a-z_][a-z0-9_]{2,30}$"
 
 # Naming style matching correct attribute names.
 attr-naming-style = "snake_case"
--- a/tests/unit/test_cli_transpile.py
+++ b/tests/unit/test_cli_transpile.py
@@ -22,17 +22,12 @@ TRANSPILERS_PATH = Path(__file__).parent.parent / "resources" / "transpiler_conf
 
 
 @pytest.fixture()
-def transpilers_repository_path(tmp_path):
-    return tmp_path
-
-
-@pytest.fixture()
-def transpiler_repository(transpilers_repository_path) -> TranspilerRepository:
-    return TranspilerRepository(transpilers_repository_path)
+def transpiler_repository(tmp_path: Path) -> TranspilerRepository:
+    return TranspilerRepository(tmp_path)
 
 
 @pytest.fixture(name="transpiler_config_path")
-def stubbed_transpiler_config_path(transpilers_repository_path) -> Path:
+def stubbed_transpiler_config_path(tmp_path: Path) -> Path:
     """Path to a stubbed LSP transpiler configuration file."""
     # Refer to LSPConfig.load() for the expected structure; there is no .save() method.
     lsp_configuration = {
@@ -53,7 +48,7 @@ def stubbed_transpiler_config_path(transpilers_repository_path) -> Path:
         },
     }
 
-    config_path = transpilers_repository_path / "remorph-transpilers" / "stub-transpiler" / "lib" / "config.yml"
+    config_path = tmp_path / "remorph-transpilers" / "stub-transpiler" / "lib" / "config.yml"
     config_path.parent.mkdir(parents=True)
     with config_path.open("w") as f:
         yaml.dump(lsp_configuration, f)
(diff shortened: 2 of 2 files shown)


tests/integration/transpile/test_morpheus.py
asnare, line 1: I think this is the fix for the broken integration test on `main`, and most important fix in here? Can we put this in its own PR so it can be merged as a priority?

sundarshankar89 (dismissed):
@m-abulazm This is doing two things, can you split them to get the main working first and then add the enhacement tests as a seperate PR.

tests/unit/test_cli_transpile.py
```diff
@@ -269,12 +325,51 @@ def test_transpile_prints_errors(caplog, tmp_path: Path, mock_workspace_client:
...
+            error_file_path=cfg.error_file_path,
+            sdk_config=cfg.sdk_config,
+            skip_validation=cfg.skip_validation,
+            catalog_name=cfg.catalog_name,
+            schema_name=cfg.schema_name,
+        ),
+    )
+
+
+def test_transpile_no_config(mock_cli_for_transpile_no_config, transpiler_repository, transpiler_config_path) -> None:
```
sundarshankar89: This should be different PR.

tests/integration/transpile/test_morpheus.py
m-abulazm: this one now adds more transpile cli test cases

pyproject.toml
```diff
@@ -283,7 +283,7 @@ argument-naming-style = "snake_case"
 
 # Regular expression matching correct argument names. Overrides argument-naming-
 # style. If left empty, argument names will be checked with the set naming style.
-argument-rgx = "[a-z_][a-z0-9_]{2,30}$"
+argument-rgx = "[a-z_][a-z0-9_]{2,33}$"
```
asnare: Hmm. Do we really need this? :)

We use the same linting setup across all our projects… shall we avoid diverging?

tests/unit/test_cli_transpile.py
```diff
@@ -15,13 +15,24 @@
...
 from databricks.labs.lakebridge.contexts.application import ApplicationContext
+from databricks.labs.lakebridge.transpiler.repository import TranspilerRepository
 from tests.unit.conftest import path_to_resource
 
 TRANSPILERS_PATH = Path(__file__).parent.parent / "resources" / "transpiler_configs"
 
 
+@pytest.fixture()
+def transpilers_repository_path(tmp_path):
+    return tmp_path
```
asnare: I'm not really sure what value this adds. (In some ways this can be misleading, with `transpilers_repository_path` and `tmp_path` both giving the same value.)

tests/unit/test_cli_transpile.py
```diff
@@ -15,13 +15,24 @@
...
+
+
+@pytest.fixture()
+def transpiler_repository(transpilers_repository_path) -> TranspilerRepository:
```
asnare: Don't forget the type annotation for the argument.

tests/unit/test_cli_transpile.py
```diff
@@ -15,13 +15,24 @@
...
+    return TranspilerRepository(transpilers_repository_path)
+
+
 @pytest.fixture(name="transpiler_config_path")
-def stubbed_transpiler_config_path(tmp_path: Path) -> Path:
+def stubbed_transpiler_config_path(transpilers_repository_path) -> Path:
```
asnare: (Here as well: argument type annotation.)

asnare (commented):
I like updating the tests to ensure they're using their own repository instead of the default (user) one, especially in unit tests.

Please don't forget to annotate the types of arguments for tests; this is missing in a lot of places and annotations make it easier to review by inspection.

I'm not too happy about the monkeypatching of shared resources, but it's a pretty heavy lift and requires a dedicated PR to deal with mocking `ApplicationContext` and stubbing `do_transpile`. Can you create a tech-debt issue to cover these? I think this would involve:
 - Moving most of the `cli.py` methods into a class.
 - Maybe making the `@command` methods class methods on that class. This would allow them to be instantiated as now, and also from tests with overrides for the things that need to be mocked.

pyproject.toml
m-abulazm: argument `mock_cli_for_transpile_no_config` used in tests is 33 chars.

I will shorten the name

tests/unit/test_cli_transpile.py
m-abulazm: I was not sure that the value of `tmp_path` will be the same across fixtures but it stays the same per test invocation

tests/unit/test_cli_transpile.py
m-abulazm: done

gueniai (approved):
LGTM

PR #1997, opened 2025-09-08
Support basic `analyzer` telemetry

m-abulazm:
### What does this PR do?
Add `analyzer_source_tech` telemetry

* the existing impl never sent the information because it used a cached API
* Call the API only once at the end of the run to send all telemetry data
* the `WorkspaceClient` needs to be recreated so the `user_agent_extras` work

### Functionality

- [ ] added new CLI command

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -47,7 +45,6 @@ def _get_source_tech(self, platform: str | None = None) -> str:
                 logger.warning(f"Invalid source technology {platform}")
             platform = self._prompts.choice("Select the source technology", self.supported_source_technologies())
             with_user_agent_extra("analyzer_source_tech", make_alphanum_or_semver(platform))
```
sundarshankar89: this needs to be adjusted.

m-abulazm: this still works since `ctx.add_user_agent_extra("cmd", "analyze")` gets called afterwards

the global "extras" that we access and change here will be used -part of the SDK- when creating a new `WorkspaceClient`

m-abulazm: and we can't pass `ApplicationContext` here since this will be a cyclic dependency

sundarshankar89: > this still works since `ctx.add_user_agent_extra("cmd", "analyze")` gets called afterwards
> 
> the global "extras" that we access and change here will be used -part of the SDK- when creating a new `WorkspaceClient`

Can you add a comment so in future we don't trip up

m-abulazm: this did have to change as it was conditional and added comment

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -46,8 +44,11 @@ def _get_source_tech(self, platform: str | None = None) -> str:
             if platform is not None:
                 logger.warning(f"Invalid source technology {platform}")
             platform = self._prompts.choice("Select the source technology", self.supported_source_technologies())
-            with_user_agent_extra("analyzer_source_tech", make_alphanum_or_semver(platform))
-            logger.debug(f"User: {self._current_user}")
+
+        # Need to call `ApplicationContext#add_user_agent_extra()` afterwards
+        # in order for this info to be captured correctly in telemetry
+        # which the cli does
+        with_user_agent_extra("analyzer_source_tech", make_alphanum_or_semver(platform))
```
asnare: Should this still be here?

(I don't think it will have an effect?)

src/databricks/labs/lakebridge/cli.py
```diff
@@ -654,10 +654,11 @@ def analyze(
 ):
     """Run the Analyzer"""
     ctx = ApplicationContext(w)
-    ctx.add_user_agent_extra("cmd", "analyze")
+    ctx.analyzer.run_analyzer(source_directory, report_file, source_tech)
 
+    # run_analyzer() adds its own user agent extras so we just need to add the command here
+    ctx.add_user_agent_extra("cmd", "analyze")
     logger.debug(f"User: {ctx.current_user}")
-    ctx.analyzer.run_analyzer(source_directory, report_file, source_tech)
```
asnare: Previously we were trying to ensure telemetry was collected before running the analyser, and now we do it afterwards? Can you elaborate on the change?

I would expect it to be before, if for no other reason than it still happening if the analyser raises an exception or fails for some reason.

m-abulazm: previously only `cmd` worked. 

I agree if an error happens we do not send telemetry. but after the user fixes the error and retries (this typical user behavior), we will get the information. I will add this to a finally block so we always collect the telemetry even on errors

this also relates to gathering more telemetry like `command_error` which can only be figured out after running the command and we should collect this info.

The current analyzer exposes only `run_analyzer()` which makes it harder to collect the telemetry before running and a more flexible interface is needed. I would not refactor the interface part of this PR

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
m-abulazm: this will have effect because the global extras are used when a new `WorkspaceClient` is created which in the impl happens after this per https://github.com/databrickslabs/lakebridge/pull/1997#discussion_r2334049964

asnare: I know you didn't put this here originally and are trying to find the minimum set of changes to make it work, but I think we can do better than this. The main issues I see are:

 - Global side-effects, which we've covered elsewhere. (Doesn't reset between tests, etc.)
 - It relies on an unrelated piece of code doing the right thing. In this case it relies on something else constructing a new client and making an API call. Although documented, that's a form of coupling that is brittle.
 - ~This isn't the only code-path for running the analyser with a specific dialect, so it won't capture everything.~ [Missed a call, this is on both code-paths.]

I'll propose something that I think addresses this and is a bit cleaner…

src/databricks/labs/lakebridge/cli.py
```diff
@@ -654,10 +655,15 @@ def analyze(
...
-    ctx.add_user_agent_extra("cmd", "analyze")
-
-    logger.debug(f"User: {ctx.current_user}")
-    ctx.analyzer.run_analyzer(source_directory, report_file, source_tech)
+    try:
+        ctx.analyzer.run_analyzer(source_directory, report_file, source_tech)
+    finally:
+        exception_cls, _, _ = sys.exc_info()
+        if exception_cls is not None:
+            ctx.add_user_agent_extra("analyzer_error", exception_cls.__name__)
```
asnare, line 683: I understand why we have this, but I think this is stretching the use of the user-agent beyond its intended purpose.

@sundarshankar89: I'm fine with this for now, but it's an example of something we need to move out to the proper CLI-integrated telemetry.

asnare (commented):
I think we should adjust the place where the agent is updated, the purpose being:

 - Make things cleaner, because we don't need to adjust globals.
 - Keep the telemetry code in a single location.
 - Ensure that all the code-paths that run the analyser are covered. (If the dialect is passed via command-line argument, it's currently missed.)

My proposal for this is:
1. Update `LakebridgeAnalyzer.run_analyzer()` to return the source dialect. (This will involve updating the 2 `_run*` methods to get it.)
2. Update the finally block to use the return value, updating the agent on the `ApplicationContext` as you already do.

What do you think?

src/databricks/labs/lakebridge/cli.py
sundarshankar89: I would rather have detail logs with our long term cli sidecar based telemtry than this.

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
m-abulazm: > This isn't the only code-path for running the analyser with a specific dialect, so it won't capture everything.

I don't see that path. what I see is both `_run_arg_analyzer` and `_run_prompt_analyzer` calling the method here so it should capture everything. is there a third path?

asnare: > I don't see that path. what I see is both `_run_arg_analyzer` and `_run_prompt_analyzer` calling the method here so it should capture everything. is there a third path?

You're right, I missed the call in `_run_arg_analyzer` .

I think the other points stand though even without this.

m-abulazm: changed so that telemetry use workspace client config

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -40,15 +43,15 @@ def _get_result_file_path(self, directory: Path) -> Path:
...
+    def _get_source_system(self, platform: str | None = None) -> str:
         """Validate source technology or prompt for a valid source"""
         if platform is None or platform not in self.supported_source_technologies():
             if platform is not None:
                 logger.warning(f"Invalid source technology {platform}")
             platform = self._prompts.choice("Select the source technology", self.supported_source_technologies())
-            with_user_agent_extra("analyzer_source_tech", make_alphanum_or_semver(platform))
-            logger.debug(f"User: {self._current_user}")
-        return _PLATFORM_TO_SOURCE_TECHNOLOGY[platform]
+        assert _PLATFORM_TO_SOURCE_TECHNOLOGY[platform], f"Unsupported platform: {platform}"
```
asnare: Nit:
```suggestion
        assert platform in _PLATFORM_TO_SOURCE_TECHNOLOGY
```

tests/unit/analyzer/test_analyzer.py
```diff
@@ -0,0 +1,51 @@
...
+    mock_prompts = MockPrompts({})
+    input_path = tmp_path / "in"
+    output_path = tmp_path / "out"
+    tech = "Synapse"
+    analyzer = LakebridgeAnalyzer(mock_prompts, True)
+
+    with (
+        patch.object(analyzer, "_run_binary", return_value=None),
+        patch("databricks.labs.lakebridge.analyzer.lakebridge_analyzer.move_tmp_file", return_value=None),
+    ):
```
asnare: I'd really like to see us do this without monkey-patching, especially the global. To start with…

```python
class _LakebridgeAnalyzer(LakebridgeAnalyzer):

    def _run_binary(self, directory: Path, result: Path, technology: str, is_debug: bool) -> None:
        return None
```

tests/unit/analyzer/test_analyzer.py
```diff
@@ -0,0 +1,51 @@
...
+            "Enter full path to the source directory": str(input_path),
+            "Enter report file name or custom export path including file name without extension": str(output_path),
+        }
+    )
+    analyzer = LakebridgeAnalyzer(mock_prompts, True)
+
+    with (
+        patch.object(analyzer, "_run_binary", return_value=None),
+        patch("databricks.labs.lakebridge.analyzer.lakebridge_analyzer.move_tmp_file", return_value=None),
+    ):
```
asnare: The same applies here. (I think we can do this without monkeypatching.)

tests/unit/analyzer/test_analyzer.py
m-abulazm: I did fix this (although a different direction than the review) through DI to avoid extending a non abstract class or overriding private methods

the type `AnalyzerRunner` is now responsible of running the analyzer and gets swapped in the tests to avoid monkey patching

PR #1999, opened 2025-09-08
Document internal transpiler repository interfaces

asnare:
## Changes

The primary goal of this PR is to clarify (and document) the different types of name a transpiler can have in the existing code-base. A secondary goal was to tighten some of the tight hints, in particular to limit collections to their read-only abstract interfaces.

Changes include:

 - Internal documentation for existing interfaces related to working with the transpiler repository.
 - Tighten type-hints to the read-only abstract collections.
 - Some minor code changes. (Called out inline, so they're visible.)

Out of scope: dealing with the inconsistencies and quirks of the naming that's in use.

### Tests

- existing unit tests
- existing integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/cli.py
+++ b/src/databricks/labs/lakebridge/cli.py
@@ -318,7 +318,7 @@ class _TranspileConfigChecker:
                 return None
             case 1:
                 # Only one transpiler available for the specified dialect, use it.
-                transpiler_name = compatible_transpilers.pop()
+                transpiler_name = next(iter(compatible_transpilers))
                 logger.debug(f"Using only transpiler available for dialect {source_dialect!r}: {transpiler_name!r}")
             case _:
                 # Multiple transpilers available for the specified dialect, prompt for which to use.
@@ -361,7 +361,7 @@ class _TranspileConfigChecker:
                 raise_validation_exception(msg)
             case 1:
                 # Only one dialect available, use it.
-                source_dialect = supported_dialects.pop()
+                source_dialect = next(iter(supported_dialects))
                 logger.debug(f"Using only source dialect available: {source_dialect!r}")
             case _:
                 # Multiple dialects available, prompt for which to use.
--- a/src/databricks/labs/lakebridge/transpiler/lsp/lsp_engine.py
+++ b/src/databricks/labs/lakebridge/transpiler/lsp/lsp_engine.py
@@ -53,12 +53,12 @@ logger = logging.getLogger(__name__)
 @dataclass
 class _LSPRemorphConfigV1:
     name: str
-    dialects: list[str]
-    env_vars: dict[str, str]
-    command_line: list[str]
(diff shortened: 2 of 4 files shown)


src/databricks/labs/lakebridge/cli.py
```diff
@@ -318,7 +318,7 @@ def _configure_transpiler_config_path(self, source_dialect: str) -> TranspileEng
                 return None
             case 1:
                 # Only one transpiler available for the specified dialect, use it.
-                transpiler_name = compatible_transpilers.pop()
+                transpiler_name = next(iter(compatible_transpilers))
```
asnare, line 321: Pythonic way to retrieve the only member of a collection.

src/databricks/labs/lakebridge/cli.py
```diff
@@ -361,7 +361,7 @@ def _prompt_source_dialect(self) -> TranspileEngine:
                 raise_validation_exception(msg)
             case 1:
                 # Only one dialect available, use it.
-                source_dialect = supported_dialects.pop()
+                source_dialect = next(iter(supported_dialects))
```
asnare, line 364: Pythonic way to retrieve the only member of a collection.

src/databricks/labs/lakebridge/transpiler/lsp/lsp_engine.py
```diff
@@ -79,15 +79,15 @@ def parse(cls, data: dict[str, Any]) -> _LSPRemorphConfigV1:
...
+    custom: Mapping[str, Any]
 
     @property
     def name(self):
         return self.remorph.name
 
-    def options_for_dialect(self, source_dialect: str) -> list[LSPConfigOptionV1]:
-        return self.options.get("all", []) + self.options.get(source_dialect, [])
+    def options_for_dialect(self, source_dialect: str) -> Sequence[LSPConfigOptionV1]:
+        return [*self.options.get("all", []), *self.options.get(source_dialect, [])]
```
asnare, line 90: Defensive way to construct a list: doesn't assume that the first argument is a sequence.

src/databricks/labs/lakebridge/transpiler/repository.py
```diff
@@ -56,45 +73,91 @@ def transpilers_path(self) -> Path:
...
+        """
         # Warning: product_name here (eg. 'morpheus') and transpiler_name elsewhere (eg. Morpheus) are not the same!
         product_path = self.transpilers_path() / product_name
         current_version_path = product_path / "state" / "version.json"
-        if not current_version_path.exists():
+        try:
+            text = current_version_path.read_text("utf-8")
+        except FileNotFoundError:
             return None
-        text = current_version_path.read_text("utf-8")
```
asnare, line 64: Generally when opening files it's an anti-pattern to try and predict whether it can actually be opened. Instead it's preferable to simply try opening it and handle failure.

PR #2008, opened 2025-09-10
Handle Special Character in filenames

sundarshankar89:
### What does this PR do?

- Adds support for processing filenames with special characters in email attachments.
- Ensures encoded and percent-encoded filenames are properly decoded.
- Introduces a new decoding utility function to transparently handle these cases.

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [X] manually tested
- [ ] added unit tests
- [X] added integration tests


tests/integration/transpile/common_utils.py
```diff
@@ -14,6 +14,9 @@ def assert_sql_outputs(output_folder: Path, expected_sql: str, expected_failure_
         actual_failure_sql = f.read()
     assert actual_failure_sql.strip() == expected_failure_sql.strip()
 
+    # Added to check only for the file name handling with special characters
+    assert (output_folder / "test_(filename)_comment@.sql").exists()
```
asnare, line 18: Can we include `%` somewhere in the filename? That will shake out some other gremlins that might be lurking.

src/databricks/labs/lakebridge/transpiler/lsp/lsp_engine.py
```diff
@@ -516,18 +516,21 @@ async def transpile(
 
     def open_document(self, file_path: Path, source_code: str) -> None:
         text_document = TextDocumentItem(
-            uri=file_path.absolute().as_uri(), language_id=LanguageKind.Sql, version=1, text=source_code
+            uri=str(file_path.absolute()), language_id=LanguageKind.Sql, version=1, text=source_code
```
asnare, line 519: I don't think this is the right way to solve this problem, and will explain further in the review summary.

asnare (changes requested):
These changes modify the way we're using LSP so that instead of the URI property being a URI, it's instead just a file path. This is misleading, and not the intention of how it's supposed to work. For reference:

 - [Here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem) is the definition of `TextDocumentIdentifier`.
 - [Here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri) is an explanation of how URIs are supposed to work.

They don't state it explicity, but URIs are intended to be %-encoded. (This is implied by the equivalent examples they mention.)

This means that in Python, in terms of going between URIs and `Path` the approach is:
 - `path.absolute().as_uri()`: Yields a %-encoded URI, what we need.
 - `Path.from_uri(uri)`: Goes from a %-encoded URI back to Path, also what we need.

To demonstrate:
```python
Python 3.13.7 (main, Aug 14 2025, 11:12:11) [Clang 17.0.0 (clang-1700.0.13.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pathlib import Path
>>> p = Path("/(horrible%path)")
>>> p
PosixPath('/(horrible%path)')
>>> u = p.as_uri()
>>> u
'file:///%28horrible%25path%29'
>>> q = Path.from_uri(u)
>>> q
PosixPath('/(horrible%path)')
>>> p == q
True
```

The errors we're seeing with special files imply that somewhere we're not using `.as_uri()` or `Path.from_uri()` when going back and forth. It might not be in the CLI (our LSP client), it could equally be in the plugins (our LSP server).

Maybe I'm misunderstanding the situation, if so please help me out because on protocol matters like this for interop we all need to be on the same page.

sundarshankar89:
@asnare Implemented this [here](https://github.com/databrickslabs/bladerunner/pull/535)

sundarshankar89:
This issue will be tackled as part of `databricks-bb-plugin`

PR #2009, opened 2025-09-10
Standardize schema queries

m-abulazm:
### What does this PR do?
Fix reading schemas in reconcile 
* Use the same column names in all the data sources to DRY and allow using one mapper
* Add a sql server intergation spec
* Adapt local spark setup and add missing jars so integration spec can work
* the mssql spark connector is legacy and the install is not well documented. the jar alone locally gives `ClassNotFoundException` and fixed by adding some missing dependencies
* In the future we should use our staging databricks for the integration tests to simplify local setup
### Tests

- [ ] manually tested
- [ ] added unit tests
- [X] added integration tests

### DIFF
--- a/tests/integration/reconcile/connectors/test_read_schema.py
+++ b/tests/integration/reconcile/connectors/test_read_schema.py
@@ -1,0 +1,27 @@
+from unittest.mock import create_autospec
+
+from databricks.labs.lakebridge.reconcile.connectors.tsql import TSQLServerDataSource
+from databricks.labs.lakebridge.transpiler.sqlglot.dialect_utils import get_dialect
+
+from databricks.sdk import WorkspaceClient
+
+from integration.connections.debug_envgetter import TestEnvGetter
+
+
+class TSQLServerDataSourceUnderTest(TSQLServerDataSource):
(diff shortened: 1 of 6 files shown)


tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -0,0 +1,33 @@
+from unittest.mock import create_autospec
+
+import pytest
+
+from databricks.labs.lakebridge.reconcile.connectors.tsql import TSQLServerDataSource
+from databricks.labs.lakebridge.transpiler.sqlglot.dialect_utils import get_dialect
+
+from databricks.sdk import WorkspaceClient
```
sundarshankar89, line 8: This is a good start but I would rather have e2e reconcile tests this tests only Schema Extract Query.

sundarshankar89 (approved):
LGTM, but this PR could have been atleast 3 PRs col_name change, TSQL quoted fixes and adding integration tests.

gueniai (approved):
LGTM

bishwajit-db (approved):
LGTM

PR #2011, opened 2025-09-10
Support for the new `--generate-json` switch for analyzer to generate json file

### LINKED ISSUE #2010: [FEATURE]:  Add CLI option for analyzer in Lakebridge to support the json file generation

Analyzer

### Problem statement

Adding json file generation flag to analyzer
--generate-json
This flag will force the analyzer to generate a json file in addition to the excel file.  The filename will be the same as xlsx, except with the .json extension.

### Proposed Solution

Modify a set of files to support the new option

### Additional Context

_No response_

simone-dbx-labs:
### What does this PR do?
Adds a new `--generate-json` flag to the `databricks labs lakebridge analyze` command to generate a structured JSON representation of analyzer results alongside the existing Excel report.

### Key Features

**JSON Generation Behavior:**
- **Success case**: When `--generate-json true` is specified, a JSON file is generated in the same directory as the Excel report, with the same base filename but `.json` extension
  - Example: `--report-file /path/to/output.xlsx` → produces both `output.xlsx` and `output.json`
- **Failure case**: If JSON schema validation fails, the command exits with a clear error message indicating which field caused the validation failure
- **No flag**: When `--generate-json` is not specified or set to `false`, only the Excel report is generated (existing behavior)

**JSON Schema Validation:**
- Uses the Python `jsonschema` library (Draft 2020-12 specification)
- Validation occurs **after** the Perl analyzer completes execution but **before** files are moved to the final output directory
- Two schemas available depending on source technology:
  - `analyzer-sql-schema.json` - For SQL-based technologies (SQL Server, Teradata, Oracle, etc.)
  - `analyzer-etl-schema.json` - For ETL technologies (Informatica, DataStage, ADF, SSIS, SAS, etc.)
- Validation errors provide precise JSON path information for debugging

**Example validation error message:**
```
ERROR    JSON Schema Validation Failed: Analyzer output does not conform to schema at $.inventory[15].jobParameters[39]: Additional properties are not allowed ('expression' was unexpected)
ValueError: JSON Schema Validation Failed: Analyzer output does not conform to schema at $.inventory[15].jobParameters[39]: Additional properties are not allowed ('expression' was unexpected)
```

Or for type mismatches:
```
ERROR    JSON Schema Validation Failed: Analyzer output does not conform to schema at $.subJobInfo[73].childType: None is not of type 'string'
ValueError: JSON Schema Validation Failed: Analyzer output does not conform to schema at $.subJobInfo[73].childType: None is not of type 'string'
```

### Relevant Implementation Details

**Lakebridge changes:**
- Updated `cli.py` to accept and pass the `--generate-json` flag
- Modified `labs.yml` to define the new CLI flag
- Enhanced `file_utils.py::move_tmp_file()` to handle both `.xlsx` and `.json` files from temporary directory
- Updated `lakebridge_analyzer.py` to propagate the flag through the analyzer execution chain

**Bladerunner/Bladespector changes** (dependency: `databricks-bb-analyzer>=0.1.24`):
- Modified `analyzer.py` to invoke Perl analyzer with `-j` flag when JSON generation is requested
- Added post-execution JSON schema validation with improved error messages
- Updated `analyzer-etl-schema.json` to accommodate real-world analyzer output:
  - Made `fullyQualifiedName` and `folder` optional/nullable (file-based ETL tools like SAS, SSIS don't use folders)
  - Added `COMPLEX` and `VERY_COMPLEX` to `complexityLevel` enum
  - Changed `jobParameter.additionalProperties` to `true` (allows technology-specific fields like SSIS `expression`)
  - Made `subJobInfoItem.parentType` and `childType` nullable (some relationships don't specify types)

### Testing

**Validation Testing - 7 Source Technologies:**
All tests performed on macOS with representative production metadata samples:

| Technology | Test Dataset | JSON Size | Schema | Result |
|------------|--------------|-----------|---------|--------|
| Informatica PowerCenter | ~50 workflows | 692KB | ETL | ✅ PASS |
| SQL/Teradata | ~200 SQL objects | 72KB | SQL | ✅ PASS |
| SAS | ~30 programs | 47KB | ETL | ✅ PASS |
| Azure Data Factory | ~40 pipelines | 113KB | ETL | ✅ PASS |
| Informatica Big Data Edition | ~60 mappings | 182KB | ETL | ✅ PASS |
| Informatica Cloud | ~20 taskflows | 49KB | ETL | ✅ PASS |
| SSIS | (previous test) | ~XKB | ETL | ✅ PASS |

**Test coverage:**
- ✅ With `--generate-json true`: Both Excel + JSON produced and validated
- ✅ Without `--generate-json`: Only Excel produced (backward compatible)
- ✅ Schema validation failures: Multiple real validation errors were caught during extended testing across ETL tools (missing required `fullyQualifiedName`, `null` values for string fields, invalid enum values `COMPLEX`/`VERY_COMPLEX`, additional properties like `expression` in SSIS, `null` parent/child types in Informatica Cloud). All issues were identified through the validation mechanism and fixed in the schema. No explicit negative tests needed.
- ✅ File placement: JSON correctly placed in same directory as Excel output
- ✅ Error handling: Descriptive error messages with JSON paths on validation failures

### Linked Issues
Resolves #2010

### Functionality
- [x] modified existing command: `databricks labs lakebridge analyze --generate-json true`
- [x] added comprehensive validation using JSON Schema
- [x] extensively tested across 7 source technologies


src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -56,24 +57,31 @@ def get_source_system(self, platform: str | None = None) -> str:
...
     ):
         self._runnable = runnable
         self._move_file = move_file
         self._is_debug = is_debug
 
     @classmethod
     def create(cls, is_debug: bool = False) -> "AnalyzerRunner":
-        return cls(Analyzer.analyze, move_tmp_file, is_debug)
+        return cls(
+            cast(Callable[[Path, Path, str, bool, bool], None], Analyzer.analyze),
```
asnare: In general you shouldn't need to use `cast()`. I think in this case it's masking a bug? The tests have caught something that might be related?

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -56,24 +57,31 @@ def get_source_system(self, platform: str | None = None) -> str:
...
+            move_tmp_file,
+            is_debug,
+        )
 
-    def run(self, source_dir: Path, results_dir: Path, platform: str) -> AnalyzerResult:
-        logger.debug(f"Starting analyzer execution in {source_dir} for {platform}")
+    def run(self, source_dir: Path, results_dir: Path, platform: str, generate_json: bool = False) -> AnalyzerResult:
+        logger.debug(f"Starting analyzer execution in ${source_dir} for ${platform}")
```
asnare: Something inserted spurious `$` here.
```suggestion
        logger.debug(f"Starting analyzer execution in {source_dir} for {platform}")
```

CHANGELOG.md
```diff
@@ -278,6 +278,10 @@ Introducing the new Switch LLM converter, expanding Lakebridge's conversion capa
 
 ---
 ## Other updates
+## Analyzer
+
+**JSON Output Generation**
+Added `--generate-json` flag to the analyzer command to generate a JSON representation of the analysis results alongside the Excel report. This enables programmatic processing and integration with other tools. Both files are now correctly written to the output directory when JSON generation is enabled.
```
asnare: This file is updated during the release process; no changes are needed here.

src/databricks/labs/lakebridge/helpers/file_utils.py
```diff
@@ -87,15 +87,38 @@ def check_path(path: Path | str) -> bool:
...
     finally:
-        tmp_path.parent.rmdir()
-        logger.info(f"Results store at {output_path}")
+        # Clean up temp directory (will only succeed if empty)
+        try:
+            tmp_dir.rmdir()
+        except OSError:
+            # Directory not empty or other issue - log but don't fail
+            logger.debug(f"Could not remove temp directory {tmp_dir}, may contain other files")
+        logger.info(f"Results stored at {output_path}")
```
asnare: This method was already quite poor, and unfortunately the need to also move a `.json` file has exposed that. Despite its name, it's not generic and is very specific to the Analyser output. I will fix this in a separate PR, after which this will need to be updated (but will be simpler).

asnare (changes requested):
A major question I've got is whether we should still produce the Excel file if `--generate-json` is provided. I don't think so?

I've added some comments regarding the implementation; I'll fix up `move_tmp_file` in a separate PR.

Separate to my comments:

 - Tests (and linting) are failing; these need to be fixed.
 - We're missing automated tests to check it works properly (and stays working).

CHANGELOG.md
simone-dbx-labs: Removing my change.

sundarshankar89:
@simone-dbx-labs  we did have the new version released can you update and check if this is working as expected?

docs/lakebridge/docs/assessment/analyzer/index.mdx
```diff
@@ -37,7 +37,8 @@ Usage:
     databricks labs lakebridge analyze [flags]
 
 Flags:
-    -h, --help help for analyze
+    --generate-json string    (Optional) Generate JSON file alongside the Excel report
```
asnare:
```suggestion
    --generate-json true|false    (Optional) Generate JSON file alongside the Excel report
```

labs.yml
```diff
@@ -16,7 +16,9 @@ commands:
       - name: report-file
         description: (Optional) Local filesystem `path` of the analysis report file to write
       - name: source-tech
-        description: (Optional) The technology/platform of the sources to analyze
+        description: (Optional) Name of the Source System Technology you want to analyze
+      - name: generate-json
+        description: (Optional) Generate JSON file alongside the Excel report
```
asnare:
```suggestion
        description: (Optional) Generate JSON file alongside the Excel report (`true|false`)
        default: "false"
```

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -54,15 +55,17 @@ def get_source_system(self, platform: str | None = None) -> str:
...
 class AnalyzerRunner:
-    def __init__(self, runnable: Callable[[Path, Path, str, bool], None], is_debug: bool) -> None:
+    def __init__(self, runnable: Callable[[Path, Path, str, bool, Path | None], None], is_debug: bool) -> None:
         self._runnable = runnable
         self._is_debug = is_debug
 
     @classmethod
     def create(cls, is_debug: bool = False) -> "AnalyzerRunner":
-        return cls(Analyzer.analyze, is_debug)
+        return cls(cast(Callable[[Path, Path, str, bool, Path | None], None], Analyzer.analyze), is_debug)
```
asnare: I noticed this in my original review, but the comment seems to have disappeared:
> In general you shouldn't need to use cast(). I think in this case it's masking a bug? The tests have caught something that might be related?

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -76,30 +79,45 @@ def run(self, source_dir: Path, results_file_path: Path, platform: str) -> Analy
...
+            staging_json = Path(tmp_dir) / "staging-report.json" if json_result else None
+            self._runnable(source_dir, staging_path, platform, is_debug, staging_json)
             # On Windows, can't overwrite via move() so first need to remove the target if it exists.
             results_file_path.unlink(missing_ok=True)
             shutil.move(staging_path, results_file_path)
             logger.debug(f"Report moved from staging to requested location: {staging_path} -> {results_file_path}")
+            if json_result and staging_json and staging_json.exists():
+                json_result.unlink(missing_ok=True)
+                shutil.move(staging_json, json_result)
+                logger.debug(f"JSON report moved: {staging_json} -> {json_result}")
```
asnare: Something here is not right:

 - The reason this method exists in this form is because we couldn't pass the path we want for the JSON. If we can do that, we should do that instead of setting up a staging location to write to. (That's the workaround mentioned in the `TODO` marker.)
 - The version of `databricks-bb-analyzer` (0.2.2) we're using doesn't support this API.

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
simone-dbx-labs: Addressed - removed cast() entirely. Analyzer.analyze is now passed directly to cls(). Types match naturally with the new bladespector release.

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
simone-dbx-labs: Addressed - removed staging_json and all the staging/move logic for JSON.

PR #2014, opened 2025-09-11
Ensure we output transpiled code

gueniai:
### What does this PR do?
Currently, if there's a parsing error (usually, these mean the output is unusable) we copy the input code into the output file. This leads to a confusing interaction, because there is no indication that the file is the same as the input, and the user is left wondering what happened. 

This PR changes the output, so that the transpiled code, including all issues we found, are available in the "transpiled" file. 

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/tests/unit/transpiler/test_execute.py
+++ b/tests/unit/transpiler/test_execute.py
@@ -181,36 +181,75 @@ def test_with_file(input_source, error_file, mock_workspace_client):
     check_error_lines(status["error_log_file"], expected_errors)
 
 
-class IdentityTranspileEngine(TranspileEngine):
-    """A simple "identity" transpiler that does not change the source it is given."""
+class ConfigurableTestEngine(TranspileEngine):
+    """Expand test transpiler engine."""
+
+    def __init__(
+        self,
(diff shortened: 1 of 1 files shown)


src/databricks/labs/lakebridge/transpiler/execute.py
```diff
@@ -165,7 +165,7 @@ def _process_non_mime_result(context: TranspilingContext, error_list: list[Trans
     output_path = cast(Path, context.output_path)
 
     if any(err.kind == ErrorKind.PARSING for err in error_list):
-        output_code = context.source_code or ""
+        output_code = context.transpiled_code or ""
```
vil1: we should remove the whole `if`

vil1 (changes requested):
Ready to approve once my little comment is addressed and https://github.com/databrickslabs/morpheus/pull/509 is merged

vil1 (approved):

sundarshankar89 (approved):
LGTM, I made changes to the test

Issue #1972, opened 2025-08-27
[BUG]: Default jobs created for reconcile fails as could not find databricks_labs_remorph package


patel-viral:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Category of Bug / Issue

Reconcile bug

### Current Behavior

Job created post reconcile config with name "LAKEBRIDGE_Reconciliation_Runner" points to incorrect python wheel. The name shall be I believe "databricks_labs_lakebridge" instead of "databricks_labs_remorph". When manually changed the process works.

`Run failed with error message
 Python wheel with name databricks_labs_remorph could not be found. Please check the driver logs for more details`

### Expected Behavior

Created job shall run fine without any error.

Probable fix I would see is changing below 2 files,
[src/databricks/labs/lakebridge/deployment/job.py](https://github.com/databrickslabs/lakebridge/blob/6b51afd865c1f854abec1c0450431690333d2ee4/src/databricks/labs/lakebridge/deployment/job.py#L129)
and [tests/integration/reconcile/test_aggregates_reconcile.py](https://github.com/databrickslabs/lakebridge/blob/6b51afd865c1f854abec1c0450431690333d2ee4/tests/integration/reconcile/test_aggregates_reconcile.py#L354)
to use package_name: databricks_labs_lakebridge instead of package_name: databricks_labs_remorph

### Steps To Reproduce

`databricks labs lakebridge configure-reconcile --profile <profile-name>`

```
Select the Data Source:
[0] databricks
[1] mssql
[2] oracle
[3] snowflake
[4] synapse
Enter a number between 0 and 4: 0
Select the report type:
[0] all
[1] data
[2] row
[3] schema
Enter a number between 0 and 3: 0
Enter Secret scope name to store `Databricks` connection details / secrets (default: remorph_databricks): 
Enter source schema name for `Databricks`: source_schema
Enter target catalog name for Databricks: hive_metastore
Enter target schema name for Databricks: target_schema
Enter catalog name (default: remorph): dbw_kxinooykoucey
Enter schema name (default: reconcile): reconcile
Enter volume name (default: reconcile_volume): reconcile_volume
Open config file https://adb-<workspaceurl>.azuredatabricks.net/#workspace/Users/username@domain.com/.lakebridge/reconcile.yml in the browser? (default: no): yes
```

Below is the job created post above.
```
resources:
  jobs:
    LAKEBRIDGE_Reconciliation_Runner:
[...]

Issue #1971, opened 2025-08-27
[BUG]: Links to docs in reconciler error handling code is incorrect


patel-viral:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Category of Bug / Issue

Reconcile bug

### Current Behavior

Links defined within code here src/databricks/labs/lakebridge/reconcile/runner.py is not working and shall point to valid url. Hence is not helpful in debugging and pointing to correct issue and resolution when actual errors are raised.
`_RECON_README_URL = "https://github.com/databrickslabs/lakebridge/blob/main/docs/recon_configurations/README.md"`

### Steps To Reproduce

Error is obvious if you browse the code at src/databricks/labs/lakebridge/reconcile/runner.py as of today 27-Aug-2025 and click on the link which is codded.

`_RECON_README_URL = "https://github.com/databrickslabs/lakebridge/blob/main/docs/recon_configurations/README.md"`

In case wondering how I noticed it, below are the steps. I will be fixing my configuration to re run it.
Configured the reconciler for databricks to databricks.
`databricks labs lakebridge configure-reconcile --profile <profile-name>`

- data source: databricks [0]
- report type: all [0]
- Left the secret scope name as default and tried removing as well.
- Other settings pointing to correct schema and catalogs where asked.

Config created fine.

Run
`databricks labs lakebridge reconcile --profile <profile-name>`

Error
`ERROR [d.l.l.reconcile.runner] Cannot find recon table configuration in existing `reconcile` installation. Please provide the configuration file recon_config_databricks_source_schema_all.json in the workspace.. For more details, please refer to https://github.com/databrickslabs/lakebridge/blob/main/docs/recon_configurations/README.md`

### Relevant log output or Exception details

```shell

```

### Logs Confirmation

- [ ] I ran the command line with `--debug`
- [ ] I have attached the `lsp-server.log` under USER_HOME/.databricks/labs/remorph-transpilers/<converter_name>/lib/lsp-server.log

### Sample Query

```shell

```

### Operating System

Windows

### Version

latest via Databricks CLI

PR #2039, opened 2025-09-23
Fix deploy reconcile jobs  

m-abulazm:
### What does this PR do?
* fix broken docs link when running reconcile from cli
* fix deployed reconcile jobs to use correct wheel when running from cli

* the name of the python task is parsed from the wheel path to be future proof. the parsing asserts that lakebridge is part of the name

### Linked issues

Resolves https://github.com/databrickslabs/lakebridge/issues/1972 and https://github.com/databrickslabs/lakebridge/issues/1971

### Functionality

- [X] modified existing command: 
* `databricks labs lakebridge reconcile`
* `databricks labs lakebridge configure-reconcile`

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/deployment/installation.py
```diff
@@ -87,16 +87,15 @@ def _apply_upgrades(self):
     def _upload_wheel(self):
```
asnare: Hmm, I know this was missing before but can we add this while we're here?
```suggestion
    def _upload_wheel(self) -> str:
```

tests/unit/deployment/test_job.py
```diff
@@ -96,6 +96,27 @@ def test_deploy_missing_job(snowflake_recon_config):
     install_state = InstallState.from_installation(installation)
     product_info = ProductInfo.for_testing(LakebridgeConfiguration)
     job_deployer = JobDeployment(workspace_client, installation, install_state, product_info)
-    job_deployer.deploy_recon_job(name, snowflake_recon_config, "remorph-x.y.z-py3-none-any.whl")
+    job_deployer.deploy_recon_job(name, snowflake_recon_config, "lakebridge-x.y.z-py3-none-any.whl")
     workspace_client.jobs.create.assert_called_once()
     assert install_state.jobs[name] == str(job.job_id)
+
+
+def test_parse_package_name():
```
asnare: Nit:
```suggestion
def test_parse_package_name() -> None:
```

src/databricks/labs/lakebridge/deployment/job.py
```diff
@@ -145,3 +146,18 @@ def _get_default_node_type_id(self) -> str:
     def _name_with_prefix(self, name: str) -> str:
         prefix = self._installation.product()
         return f"{prefix.upper()}_{name}".replace(" ", "_")
+
+    def parse_package_name(self, wheel_path: str) -> str:
+        default_name = "databricks_labs_lakebridge"
+
+        try:
+            name = wheel_path.split("/")[-1].split("-")[0]
+        except IndexError:
```
asnare: If `split()` always returns a sequence with at least the original value, I don't think this exception can ever be raised?

src/databricks/labs/lakebridge/deployment/job.py
```diff
@@ -145,3 +146,18 @@ def _get_default_node_type_id(self) -> str:
...
+            logger.warning(f"Cannot parse package name from wheel path {wheel_path}, using default.")
+            name = default_name
+
+        if self._product_info.product_name() not in name:
+            logger.warning(f"Parsed package name {name} does not match product name, using default.")
+            name = default_name
```
asnare, line 157: I'm curious about when/why this occurs?

asnare (commented):
Looking good. I have some small questions that I've noted inline.

src/databricks/labs/lakebridge/deployment/job.py
m-abulazm: true, it will not be raised for any strings. I can remove it although I feel some form of error handling is needed and using `Exception` is too generic

src/databricks/labs/lakebridge/deployment/job.py
m-abulazm: we reply on third party code to build and upload the wheel: "blueprint" and "wheel". if the implementation there changes, it might occur

src/databricks/labs/lakebridge/deployment/job.py
m-abulazm: I removed the exception handling and added validation before. does it look better now?

src/databricks/labs/lakebridge/deployment/job.py
```diff
@@ -145,3 +147,21 @@ def _get_default_node_type_id(self) -> str:
...
+
+    def parse_package_name(self, wheel_path: str) -> str:
+        default_name = "databricks_labs_lakebridge"
+
+        if not isinstance(wheel_path, Path):
+            wheel_path = str(wheel_path)
+
+        if not isinstance(wheel_path, str):
+            logger.warning("wheel path is not a string, using default.")
+            return default_name
```
asnare: I think these checks are dead code, the type hint on the argument means that:

 - `wheel_path` cannot be a `Path` instance.
 - `wheel_path` must be a `str` instance.

Although at runtime the type hints are not enforced, we don't need to be defensive here: we're allowed to rely on the type-checker.

src/databricks/labs/lakebridge/deployment/job.py
asnare: Yup on removing the exception handling, although I didn't see this until I already commented elsewhere on the validation you added.

src/databricks/labs/lakebridge/deployment/job.py
m-abulazm: ok removing

sundarshankar89 (approved):
LGTM

PR #2041, opened 2025-09-23
Fix type hints on LSP configuration classes and improve validation when loading

asnare:
## Changes

This PR updates type hints on the configuration classes, and implements proper checking of the configuration during loading. Changes include:

 - We now distinguish between missing attributes and attributes that are present but malformed.
 - With the options for a dialect, the prompt is now optional instead of stubbing with an empty string. (The prompt is not needed for the `FIXED` option type.)
 - For options, the combinations of choices/default are now dealt with properly depending on the kind of option it is.
 - Some tests that were using invalid fixtures have been fixed.

During deserialisation, in many places we now handle unhappy paths via exceptions rather than checking in advance: the latter is unambiguous whereas its easy to have subtle issues when the check doesn't quite correctly anticipate or correspond to the actual usage.

Sadly, most of this has been done by hand even though there are libraries that do this sort of thing for us.

### Tests

- existing unit tests

### DIFF
--- a/docs/lakebridge/docs/reconcile/reconcile_automation.mdx
+++ b/docs/lakebridge/docs/reconcile/reconcile_automation.mdx
@@ -92,7 +92,7 @@ import LakebridgeTabs from '@site/src/components/ReconcileTabs';
 
 <LakebridgeTabs />
 
-[Link to the notebook](/lakebridge_reconciliation.dbc) - Unzip the downloaded file and upload the notebook file to your Databricks workspace.
+<a href={useBaseUrl('/lakebridge_reconciliation.dbc')} target="_blank" rel="noopener noreferrer">Link to the notebook</a> - Unzip the downloaded file and upload the notebook file to your Databricks workspace.
 
 The utility consists of three key Databricks notebooks:
 


src/databricks/labs/lakebridge/config.py
```diff
@@ -20,39 +21,124 @@ class LSPPromptMethod(Enum):
...
     @classmethod
-    def parse(cls, data: Any) -> "LSPConfigOptionV1":
+    def _get_prompt_field(cls, data: Mapping[str, JsonValue], prompt_method: LSPPromptMethod) -> str | None:
+        # Whether a prompt is mandatory or not depends on the prompt method.
+        try:
+            prompt = data["prompt"]
+            if not isinstance(prompt, str):
+                msg = f"Invalid 'prompt' entry in {data}, expecting a string: {prompt}"
+                raise ValueError(msg)
+            return prompt
```
m-abulazm:
```suggestion
            return get_string_field(data, "prompt")
```

src/databricks/labs/lakebridge/config.py
```diff
@@ -20,39 +21,124 @@ class LSPPromptMethod(Enum):
...
+            return None
+
+    @classmethod
+    def _get_choices_field(cls, data: Mapping[str, JsonValue], prompt_method: LSPPromptMethod) -> list[str] | None:
+        try:
+            choices_unsafe = data["choices"]
+            if not isinstance(choices_unsafe, list) or not all(isinstance(item, str) for item in choices_unsafe):
+                msg = f"Invalid 'choices' entry in {data}, expecting a list of strings: {choices_unsafe}"
+                raise ValueError(msg)
+            return cast(list[str], choices_unsafe)
```
m-abulazm:
```suggestion
            cast(list[str], choices_unsafe)
```

casting implies this check and `ValueError` will be raised otherwise

src/databricks/labs/lakebridge/config.py
```diff
@@ -20,39 +21,124 @@ class LSPPromptMethod(Enum):
...
+            return None
+
+    @classmethod
+    def _get_default_field(cls, data: Mapping[str, JsonValue]) -> str | None:
+        try:
+            default = data["default"]
+            if default is not None and not isinstance(default, str):
+                msg = f"Invalid 'default' entry in {data}, expecting a string: {default}"
+                raise ValueError(msg)
+            return default
```
m-abulazm:
```suggestion
            return get_string_field(data, "default")
```

src/databricks/labs/lakebridge/config.py
```diff
@@ -20,39 +21,124 @@ class LSPPromptMethod(Enum):
     CONFIRM = auto()
 
 
-@dataclass
+E = TypeVar("E", bound=Enum)
+
+
+def get_string_field(data: Mapping[str, JsonValue], name: str) -> str:
```
m-abulazm:
```suggestion
def get_string_field_or_throw(data: Mapping[str, JsonValue], name: str) -> str:
```

src/databricks/labs/lakebridge/config.py
```diff
@@ -20,39 +21,124 @@ class LSPPromptMethod(Enum):
...
+        value = data[name]
+    except KeyError as e:
+        raise ValueError(f"Missing '{name}' attribute in {data}") from e
+    if not isinstance(value, str) or not value:
+        msg = f"Invalid '{name}' attribute, must be a non-empty string: {value}"
+        raise ValueError(msg)
+    return value
+
+
+def get_enum_field(data: Mapping[str, JsonValue], name: str, enum_type: type[E]) -> E:
```
m-abulazm:
```suggestion
def get_enum_field_or_throw(data: Mapping[str, JsonValue], name: str, enum_type: type[E]) -> E:
```

m-abulazm (changes requested):
this greatly improves the user experience. thanks for this

but I would have not implemented it this way. if you are working on this part, why didnt you try to use `cattrs` or something similar?

I also added a few suggestions to improve readability.

src/databricks/labs/lakebridge/config.py
asnare: In Python, [`cast()`](https://docs.python.org/3/library/typing.html#typing.cast) is merely a hint for the type-checker, in this case narrowing to the validated type. Because it returns the value as-is and does not perform any checks, these still need to be done manually.

src/databricks/labs/lakebridge/config.py
asnare: Sadly we can't use `get_string_field()` here because whether the field is required or not depends on a related attribute, and once it returns we can't tell if the exception was raised because the field is invalid or because it was missing.

I've adjusted the pattern for this method to make it a bit clearer what the difference is when compared to the body of `get_string_field()`.

src/databricks/labs/lakebridge/config.py
asnare: Similarly, `get_string_field()` isn't appropriate here because the attribute is optional.

(It also turns out I could just use `get("default", None)` instead of the try/except pattern, so I've updated this as well.)

asnare:
> why didnt you try to use `cattrs` or something similar?

That's a good question, and we've asked ourselves the same question especially with respect to similar code in `blueprint`. I think there's a preference for Pydantic over `cattrs`, but the point stands.

The general situation with respect to dependencies is:

 - Any new dependencies need to go through legal review. (In this case it would probably be fine, and from a practical viewpoint I think it's a transitive dependency anyway. But still legal need to look at it.)
 - We need to consider the situation with respect to other dependencies that use the library and how likely conflicts are. This tends to be a bit messy, and cause headaches.
 - It needs to be a library that we think that all developers can use and understand.
 - Needs to be a strong benefit to including it.

Back to the situation at hand: I agree we should be using a library to do this, and have created #2050 to cover this.

src/databricks/labs/lakebridge/config.py
asnare: I see what you're getting at.

Putting `_or_throw` in the name doesn't feel great (eg. Python terminology is raise rather than throw) but they're doing more than just getting. I've renamed all of these to be `extract_*` instead of `get_*` to hint at them being more than just simple getters. Does that seem better?

src/databricks/labs/lakebridge/config.py
m-abulazm: why not add `is_required` argument to `extract_string_field()` to have a more reusable impl. 
the code is quire similar and a more flexible interface will remove a lot of duplicated code

the usage would look like
```
is_required = prompt_method == LSPPromptMethod.FORCE:
return extract_string_field(data, "prompt", is_required)
```

src/databricks/labs/lakebridge/config.py
m-abulazm: got it

src/databricks/labs/lakebridge/config.py
m-abulazm: should also use `is_required` similar to the previous comment

src/databricks/labs/lakebridge/config.py
asnare: I thought about that, but then the return signature becomes `| None` and most of the callers need to add an `assert x is not None` on the return value because the type checker can't infer that (based on the parameter). That's also pretty messy, so I went with this as the least bad option.

src/databricks/labs/lakebridge/config.py
m-abulazm: that is fine as well and just to share my motivation for this initial comment:

raising/throwing errors should be clear to the callers of the method. this follows clean code practices of code doing what is supposed to do and no suprises. 
regarding terminology, I might be biased but `throw` is more universal than python `raise`.

asnare: Yup, thanks for the clarification: I understood what you were getting at.

src/databricks/labs/lakebridge/config.py
asnare: (Renamed `extract_enum_field`, discussed elsewhere.)

src/databricks/labs/lakebridge/config.py
asnare: Due to the differing return types, I've pulled the code for this (and `"default"`) out into a `_maybe_extract_string_field()`: does that look better?

src/databricks/labs/lakebridge/config.py
m-abulazm: agree

src/databricks/labs/lakebridge/config.py
m-abulazm: yes. any reasons why this method is not next to the other extract methods? 
also `extract_string_field()` can reuse `_maybe_extract_string_field()` and type cast/assert

asnare: I've updated it to work this way, but I think we're approaching the point of diminishing returns.

We're not quite there yet, but please bear in mind that at some point strict DRY (and some of Uncle Bob's other rules) can make code less maintainable due to the additional indirection they introduce, as well as dead code paths via combinations of branches. (I've tried hard to avoid this.)

m-abulazm: yes I understand, there is a no silver bullet.  but on the other hand, I see a lot of duplicate code so just trying to improve and I am open minded how we can get rid of duplicate code.

These extractors have the potential to be reused a lot so I was more careful in this review

PR #2042, opened 2025-09-23
Unit tests for `TranspilerRepository`, and additional integration tests

asnare:
## Changes

This PR implements unit tests for the `TranspilerRepository` class, as well as extending the coverage of the integration tests. This is aimed mainly at capturing existing behaviour and ensuring bugs are not introduced.

### Tests

- added unit tests
- added integration tests

### DIFF
--- a/tests/unit/transpiler/test_repository.py
+++ b/tests/unit/transpiler/test_repository.py
@@ -126,7 +126,8 @@ def test_all_transpiler_configurations() -> None:
     """Verify that the repository yields the set of transpiler configurations that are available."""
     a_transpiler = mock_transpiler("a_transpiler", transpiler_config("A Transpiler", dialects=("foo",)))
     b_transpiler = mock_transpiler("b_transpiler", transpiler_config("B Transpiler", dialects=("bar",)))
-    labs_dir = mock_labs_path_with_registry((a_transpiler, b_transpiler))
+    another_b = mock_transpiler("another_b", transpiler_config("B Transpiler", dialects=("baz",)))
+    labs_dir = mock_labs_path_with_registry((a_transpiler, b_transpiler, another_b))
     transpiler_repository = TranspilerRepository(labs_dir)
 
     transpiler_configurations = transpiler_repository.all_transpiler_configs()
@@ -135,6 +136,7 @@ def test_all_transpiler_configurations() -> None:
     assert transpiler_configurations == {
         "a_transpiler": LSPConfig.load(a_transpiler / "lib" / "config.yml"),
         "b_transpiler": LSPConfig.load(b_transpiler / "lib" / "config.yml"),
+        "another_b": LSPConfig.load(another_b / "lib" / "config.yml"),
     }
 
 


asnare:
> ❌ test_describe_installed_transpilers: AssertionError: […]

This is a flaky test on `main`, addressed in #2060.

src/databricks/labs/lakebridge/transpiler/repository.py
```diff
@@ -109,14 +107,29 @@ def all_transpiler_configs(self) -> Mapping[str, LSPConfig]:
         Returns:
           A mapping of configurations, keyed by their transpiler names.
         """
-        all_configs = self._all_transpiler_configs()
-        return {config.name: config for _, config in all_configs}
+        return {config.name: config for _, config in self._all_transpiler_configs()}
```
m-abulazm: I feel we really need top standardize this. as discussed before, transpiler path is the ultimate identifier of the transpiler but all the methods that deal with transpiler pathes are private.
a few methods return transpiler names instead of path then the public interface is mostly using transpiler names not pathes. 
in some places it is called product name and in other name or transpiler name. we should keep one going forward. please check my other comment

tests/unit/transpiler/test_repository.py
```diff
@@ -0,0 +1,300 @@
...
+    """Verify that the transpiler names (not product) for installed transpilers is available."""
+    a_transpiler = mock_transpiler("a_transpiler", transpiler_config("A Transpiler", dialects=("foo",)))
+    b_transpiler = mock_transpiler("b_transpiler", transpiler_config("B Transpiler", dialects=("bar",)))
+    another_b = mock_transpiler("another_b", transpiler_config("B Transpiler", dialects=("baz",)))
+    labs_dir = mock_labs_path_with_registry((a_transpiler, b_transpiler, another_b))
+    transpiler_repository = TranspilerRepository(labs_dir)
+
+    transpiler_names = transpiler_repository.all_transpiler_names()
+
+    assert transpiler_names == {"A Transpiler", "B Transpiler"}
```
m-abulazm, line 179: should we add a test for `all_transpiler_configs()` for this scenario

tests/unit/transpiler/test_repository.py
```diff
@@ -0,0 +1,300 @@
...
+    labs_dir = mock_labs_path_with_registry((a_transpiler, b_transpiler))
+    transpiler_repository = TranspilerRepository(labs_dir)
+
+    options = transpiler_repository.transpiler_config_options(transpiler_name="B Transpiler", source_dialect="bar")
+
+    assert options == b_bar_options
+
+
+def test_get_installed_transpiler_version() -> None:
+    """Verify that the installed version of a transpiler can be queried."""
```
m-abulazm: in some places we treat product names as transpiler names and in other places not

tests/integration/transpile/test_repository.py
```diff
@@ -76,3 +87,50 @@ def test_lists_dialect_transpilers(transpiler_repository: TranspilerRepository)
     assert transpilers == {'Morpheus', 'Bladebridge'}
     transpilers = transpiler_repository.transpilers_with_dialect("datastage")
     assert transpilers == {'Bladebridge'}
+
+
+@pytest.mark.parametrize(("product_name", "version"), (("morpheus", "0.4.0"), ("bladebridge", "0.1.9")))
```
m-abulazm, line 92: is this a moving target?

tests/integration/transpile/test_repository.py
```diff
@@ -76,3 +87,50 @@ def test_lists_dialect_transpilers(transpiler_repository: TranspilerRepository)
...
+
+
+@pytest.mark.parametrize("transpiler_name", ("Morpheus", "Bladebridge"))
+def test_transpilers_config_path(transpiler_repository: TranspilerRepository, transpiler_name: str) -> None:
+    config_path = transpiler_repository.transpiler_config_path(transpiler_name)
+    assert config_path.name == "config.yml" and config_path.parent.name == "lib"
+    assert config_path.is_file()
+
+
+@pytest.mark.parametrize(
```
m-abulazm, line 105: same

tests/integration/transpile/test_repository.py
asnare: Not really.

There are some (early) versions of the transpilers embedded in the repo. (Yes, they should not be there.) These tests set up a thin version of a repository containing just the metadata from these transpilers, which is why the version information is fixed.

tests/integration/transpile/test_repository.py
asnare: Not really.

There are some (early) versions of the transpilers embedded in the repo. (Yes, they should not be there.) These tests set up a thin version of a repository containing just the metadata from these transpilers, which is why the version information is fixed.

tests/unit/transpiler/test_repository.py
asnare: We do, see: [`test_all_transpiler_configurations`](https://github.com/databrickslabs/lakebridge/blob/6a19d2d53169c1218bec787e4cf501d6a8b0dd8c/tests/unit/transpiler/test_repository.py#L125)

Or did you mean something else?

tests/unit/transpiler/test_repository.py
asnare: Can you given an example?

At the moment we have two types of name:

 - _Transpiler_ Name: what the user sees (e.g. `Morpheus` or `Bladebridge`), comes from the configuration file embedded in the LSP server.
 - _Product_ Name: the name of the directory under which the LSP server is installed.

I've tried to be very careful to always and ensure that docstrings and parameter names we currently use always refer to either _transpiler_ or _product_ name, and never ambiguously just 'name'.

src/databricks/labs/lakebridge/transpiler/repository.py
asnare: I agree it's a mess, and should be solved. There are essentially two things:

 - What _we_ (the CLI) refer to the transpiler as, when installing. This is the product name, and it's used as the directory name into which the LSP server will be installed. This is not user-facing.
 - What the LSP server calls itself, from its embedded configuration file. This is the transpiler name, and it's used in the UI and documentation to refer to the transpiler. This _is_ user-facing.

These details are difficult to understand, and I agree the inconsistencies don't help things at all. I'd also like it solved. Some sharp edges here include:

 - Anything we do needs to be forward and backward compatible, because users can have mixed versions:

    - A new version of the CLI needs to cope with old transpilers that are already installed. (New CLI, old transpilers.)
    - An old version of the CLI needs to cope with new transpilers. (Old CLI, new transpilers.)

 - Transpilers can be used without being installed.

 - At some point we probably need to rename the repository directory as well because it uses the old `remorph` name. If the opportunity comes up we should deal with this.

These are solvable, but we'll face a transition period where we also need to make sure we have additional tests to cover forward/backward compatibility.

m-abulazm: @asnare what was the reasoning to add product name in the first place? Or why would cli-specific behavior be part of the public interface? And it is not clear in the `TranspilerRepository` interface which name is actually intended here

Bladebridge uses the same transpiler and product name and Morpheus does not which provides an easier path to removing product name. 

we should also think about including the version as part of the install dir as well when a transpiler is installed

> Transpilers can be used without being installed.

This extra state does complicate things. I dont get it fully because I see `install-transpile` handles `artifact` argument and I am not sure how this propagates and affects `transpile`
my gut feeling is that we should have another implementation of `TranspilerRepository` and if `artifact` is used then that repository is used

tests/unit/transpiler/test_repository.py
```diff
@@ -133,8 +133,8 @@ def test_all_transpiler_configurations() -> None:
 
     # Note: keyed by transpiler name, not product name.
```
m-abulazm: This comment is outdated and it does show how confusing this is getting

tests/unit/transpiler/test_repository.py
m-abulazm: I meant for this specific scenario
```
a_transpiler = mock_transpiler("a_transpiler", transpiler_config("A Transpiler", dialects=("foo",)))
    b_transpiler = mock_transpiler("b_transpiler", transpiler_config("B Transpiler", dialects=("bar",)))
    another_b = mock_transpiler("another_b", transpiler_config("B Transpiler", dialects=("baz",)))
```

tests/unit/transpiler/test_repository.py
m-abulazm: the method expects a product name not a transpiler name, so reading this the first time, the comment and the code intentions were misleading.

m-abulazm (changes requested):

src/databricks/labs/lakebridge/transpiler/repository.py
m-abulazm: as an intermediate step, it would be much better using transpiler id (it feels this is what product name is trying to solve) and transpiler name and that change would make the code much more readable

tests/unit/transpiler/test_repository.py
asnare: Ah, now I understand what you were getting at. I've updated that unit test.

PR #2062, opened 2025-09-29
Bump Databricks SDK Version to 0.67.0

goodwillpunning:
### What does this PR do?
Bump Databricks SDK version to 0.67.0

No changes other than updating the project.toml file.

### Linked issues
N/A

### Functionality

- [ ] added new CLI command
- [X] update project.toml file

### Tests

- [X] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/transpiler/installers.py
+++ b/src/databricks/labs/lakebridge/transpiler/installers.py
@@ -8,17 +8,24 @@ import subprocess
 import sys
 import venv
 import xml.etree.ElementTree as ET
-from json import dump, loads
+from json import dump
 from pathlib import Path
 from shutil import rmtree
-from typing import Any, Literal
-from urllib import request
-from urllib.error import HTTPError, URLError
+from typing import Literal
 from zipfile import ZipFile
 
+import requests
+from requests.exceptions import RequestException
+
+from databricks.labs.blueprint.installation import RootJsonValue
 from databricks.labs.lakebridge.transpiler.repository import TranspilerRepository
 
 logger = logging.getLogger(__name__)
+
+
+# This is not the total timeout for a request, but rather a timeout when waiting for the response
+# to start once the request has been sent.
+_DEFAULT_HTTP_TIMEOUT = 60  # seconds
 
 
 class _PathBackup:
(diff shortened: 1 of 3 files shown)


src/databricks/labs/lakebridge/transpiler/installers.py
```diff
@@ -283,8 +283,12 @@ def artifact_url(
     @classmethod
     def get_current_maven_artifact_version(cls, group_id: str, artifact_id: str) -> str | None:
         url = cls.artifact_metadata_url(group_id, artifact_id)
+        headers = {
+            'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Safari/537.36'
+        }
+        req = request.Request(url, headers=headers)
         try:
-            with request.urlopen(url) as server:
+            with request.urlopen(req) as server:
```
sundarshankar89: This was required to add because we ran into HTTP Error: 403

src/databricks/labs/lakebridge/transpiler/installers.py
```diff
@@ -319,6 +323,13 @@ def download_artifact_from_maven(
             return True
         url = cls.artifact_url(group_id, artifact_id, version, classifier, extension)
         try:
+            opener = request.build_opener()
```
sundarshankar89: This was required to add because we ran into HTTP Error: 403

src/databricks/labs/lakebridge/helpers/metastore.py
```diff
@@ -152,7 +152,7 @@ def has_privileges(
 
     @functools.lru_cache(maxsize=1024)
     def _get_user_privileges(self, user: str, securable_type: SecurableType, full_name: str) -> set[Privilege]:
-        permissions = self._ws.grants.get_effective(securable_type, full_name, principal=user)
+        permissions = self._ws.grants.get_effective(str(securable_type), full_name, principal=user)
```
sundarshankar89, line 155: This was a change in the securabletype typing in SDK.

src/databricks/labs/lakebridge/transpiler/installers.py
```diff
@@ -283,8 +283,12 @@ def artifact_url(
     @classmethod
     def get_current_maven_artifact_version(cls, group_id: str, artifact_id: str) -> str | None:
         url = cls.artifact_metadata_url(group_id, artifact_id)
+        headers = {
+            'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Safari/537.36'
+        }
```
asnare: No, we should not be spoofing the user-agent. Let's dig into this to find out more about what's going on. (If we're doing it the wrong way, we need to figure out what the right way is.)

(Sidebar: why is this only happening after we update the SDK? Or this a problem on `main` as well? Or is it something completely different?)

sundarshankar89: something else is broken

https://github.com/databrickslabs/lakebridge/actions/runs/18122655305/job/51570534469?pr=2063

sundarshankar89: it is problem in main, too, I will investigate this

asnare (approved):

src/databricks/labs/lakebridge/transpiler/installers.py
sundarshankar89: Thanks for this fix

PR #2066, opened 2025-09-30
Add Switch transpiler with `--include-llm-transpiler` flag

hiroyukinakazato-db:
## Changes
This PR adds Switch, an LLM-powered transpiler, as an optional component in Lakebridge's `install-transpile` workflow. Switch installation is controlled by the new `--include-llm-transpiler` flag.

### What does this PR do?
Implements complete Switch transpiler integration with idempotent installation, workspace deployment, job management, and resource configuration including Unity Catalog Volume for Switch.

**CLI Integration**:
- Add `--include-llm-transpiler` flag to `install-transpile` command (default: `false`)
- Switch installation is opt-in, allowing users to choose between LSP-only or LLM-powered transpilation
- Bladebridge and Morpheus continue to install by default

**SwitchInstaller Implementation**:
- Extends `TranspilerInstaller` with unified constructor signature
- WheelInstaller pattern for PyPI package management (`databricks-switch-plugin`)
- Workspace deployment: uploads Switch package from site-packages to `/Users/{user}/.lakebridge/switch/`
- Job creation: creates `LAKEBRIDGE_Switch` job with NotebookTask for parallel LLM processing
- Resource configuration: prompts for catalog, schema, and volume for Switch
- Idempotent behavior: supports reinstallation after uninstall with full recovery

**Uninstall Integration**:
- Removes Switch job from workspace via InstallState
- Logs manual cleanup instructions for validation schema and Switch resources (catalog, schema, volume)
- Integrated into `databricks labs uninstall lakebridge` workflow

- **Opt-in by default**: Switch is NOT installed by default. Users must explicitly specify `--include-llm-transpiler` flag to install Switch
- **Telemetry tracking**: Added `include-llm-transpiler` flag to user agent telemetry for usage tracking
- **Pylint configuration**: Updated `max-args` from 12 to 13 in `pyproject.toml` to accommodate new CLI parameter
- **Resource lifecycle**: Catalog, schema, and volume for Switch are prompted during install but require manual cleanup after uninstall
- **Job management**: Switch job is tracked in InstallState and managed independently from Reconciliation jobs
- **Package dependency**: Requires `databricks-switch-plugin>=0.1.0` from PyPI
- 
### Linked issues
Resolves #2048 

### Functionality

- [ ] added new CLI command
- [x] modified existing command: `databricks labs lakebridge install-transpile` (adds `--include-llm-transpiler` flag and Switch installation support)
- [x] modified existing command: `databricks labs uninstall lakebridge` (adds Switch job and resource cleanup)

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/install.py
+++ b/src/databricks/labs/lakebridge/install.py
@@ -2,7 +2,7 @@ import dataclasses
 import logging
 import os
 import webbrowser
-from collections.abc import Set, Callable, Sequence
+from collections.abc import Set, Sequence
 from pathlib import Path
 from typing import Any, cast
 
@@ -28,6 +28,7 @@ from databricks.labs.lakebridge.reconcile.constants import ReconReportType, Reco
 from databricks.labs.lakebridge.transpiler.installers import (
     BladebridgeInstaller,
     MorpheusInstaller,
+    SwitchInstaller,
     TranspilerInstaller,
 )
 from databricks.labs.lakebridge.transpiler.repository import TranspilerRepository
@@ -50,10 +51,12 @@ class WorkspaceInstaller:
         environ: dict[str, str] | None = None,
         *,
         is_interactive: bool = True,
+        include_llm_transpiler: bool = False,
         transpiler_repository: TranspilerRepository = TranspilerRepository.user_home(),
-        transpiler_installers: Sequence[Callable[[TranspilerRepository], TranspilerInstaller]] = (
+        transpiler_installers: Sequence[type[TranspilerInstaller]] = (
             BladebridgeInstaller,
             MorpheusInstaller,
+            SwitchInstaller,
         ),
     ):
         self._ws = ws
@@ -65,6 +68,7 @@ class WorkspaceInstaller:
         self._ws_installation = workspace_installation
         # TODO: Refactor the 'prompts' property in preference to using this flag, which should be redundant.
         self._is_interactive = is_interactive
+        self._include_llm_transpiler = include_llm_transpiler
         self._transpiler_repository = transpiler_repository
         self._transpiler_installer_factories = transpiler_installers
 
@@ -77,7 +81,12 @@ class WorkspaceInstaller:
 
     @property
     def _transpiler_installers(self) -> Set[TranspilerInstaller]:
-        return frozenset(factory(self._transpiler_repository) for factory in self._transpiler_installer_factories)
+        factories = self._transpiler_installer_factories
+        if not self._include_llm_transpiler:
+            if SwitchInstaller in factories:
+                logger.info("Skipping Switch installation (LLM transpiler not requested)")
+            factories = tuple(f for f in factories if f != SwitchInstaller)
+        return frozenset(factory(self._transpiler_repository, self._ws, self._installation) for factory in factories)
 
     def run(
         self,
@@ -386,6 +395,7 @@ def installer(
     transpiler_repository: TranspilerRepository,
     *,
     is_interactive: bool,
+    include_llm_transpiler: bool = False,
 ) -> WorkspaceInstaller:
     app_context = ApplicationContext(_verify_workspace_client(ws))
     return WorkspaceInstaller(
@@ -398,6 +408,7 @@ def installer(
         app_context.workspace_installation,
         transpiler_repository=transpiler_repository,
         is_interactive=is_interactive,
+        include_llm_transpiler=include_llm_transpiler,
     )
 
 
--- a/src/databricks/labs/lakebridge/transpiler/installers.py
+++ b/src/databricks/labs/lakebridge/transpiler/installers.py
@@ -16,7 +16,16 @@ from urllib import request
 from urllib.error import HTTPError, URLError
 from zipfile import ZipFile
 
+from databricks.labs.blueprint.installation import Installation
+from databricks.labs.blueprint.installer import InstallState
+from databricks.labs.blueprint.tui import Prompts
+from databricks.labs.blueprint.wheels import ProductInfo
+from databricks.labs.lakebridge.deployment.configurator import ResourceConfigurator
+from databricks.labs.lakebridge.helpers.metastore import CatalogOperations
 from databricks.labs.lakebridge.transpiler.repository import TranspilerRepository
+from databricks.sdk import WorkspaceClient
+from databricks.sdk.errors import InvalidParameterValue, NotFound
+from databricks.sdk.service.jobs import JobSettings, NotebookTask, Source, Task
 
 logger = logging.getLogger(__name__)
 
@@ -176,6 +185,9 @@ class WheelInstaller(ArtifactInstaller):
 
     def install(self) -> Path | None:
         return self._install_checking_versions()
+
+    def get_site_packages(self) -> Path:
+        return self._site_packages
 
     def _install_checking_versions(self) -> Path | None:
         latest_version = (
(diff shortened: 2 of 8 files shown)


src/databricks/labs/lakebridge/transpiler/installers.py
```diff
@@ -549,3 +564,252 @@ def _parse_java_version(cls, version: str) -> tuple[int, int, int, int] | None:
...
+
+    @property
+    def name(self) -> str:
+        return self._INSTALL_STATE_KEY
+
+    def can_install(self, artifact: Path) -> bool:
+        wheel_name = self._PYPI_PACKAGE_NAME.replace("-", "_")
+        return wheel_name in artifact.name and artifact.suffix == ".whl"
+
+    def install(self, artifact: Path | None = None) -> bool:
```
sundarshankar89: This is how we have separated things out: the installer installs things locally, deployer deploys to the workspace. You need something similar to recon deployer that way you don't need to have workspace client inside `TranspilerInstaller`

src/databricks/labs/lakebridge/install.py
sundarshankar89, line 98:
```suggestion
            config = self.configure(module, include-llm)
```

and use configure to implement switch-related prompts and have a deployer similar to recon deployment for doing workspace-related interaction.

src/databricks/labs/lakebridge/transpiler/installers.py
hiroyukinakazato-db: Thanks for the feedback. Updated in 84393142 - created `SwitchDeployment` for workspace operations (following `ReconDeployment` pattern) and removed workspace dependencies from `TranspilerInstaller`.

src/databricks/labs/lakebridge/install.py
hiroyukinakazato-db: Thanks for the suggestion. Updated in 84393142 to use the `configure(module, include_llm)` pattern. Switch resources are now prompted during configuration, stored in `TranspileConfig`, and passed to `SwitchDeployment` for workspace setup.

src/databricks/labs/lakebridge/deployment/switch.py
```diff
@@ -0,0 +1,182 @@
...
+
+        task = Task(
+            task_key="run_transpilation",
+            notebook_task=NotebookTask(notebook_path=str(notebook_path), source=Source.WORKSPACE),
+            disable_auto_optimization=True,  # To disable retries on failure
+        )
+
+        return {
+            "name": job_name,
+            "tags": {"created_by": self._ws.current_user.me().id, "switch_version": f"v{switch_version}"},
```
hiroyukinakazato-db: @sundarshankar89 By using `.id`, the `created_by` tag value will be the user's ID (e.g. `5099015744649857`) rather than the user's email address. While this is clearer from a system perspective, it may be less intuitive for users themselves.

src/databricks/labs/lakebridge/deployment/configurator.py
```diff
@@ -103,6 +108,35 @@ def warehouse_type(_):
...
+        if foundational_model_names is None:
+            raise DatabricksError("No Foundation Model serving endpoints found. Aborting the installation.")
+        # This logic is implemented to make the default choice always to appear first in the list
+        other_models = sorted(set(foundational_model_names) - {default_choice})
+        choices = [f"[DEFAULT] {default_choice}", *other_models]
+        selected = self._prompts.choice("Select a Foundation Model serving endpoint:", choices, sort=True)
+
+        if selected.startswith("[DEFAULT]"):
+            selected = default_choice
+        return selected
```
hiroyukinakazato-db, line 137: There are cases where users can only use [external models](https://docs.databricks.com/aws/en/generative-ai/external-models/#gsc.tab=0) instead of the FM API due to corporate policies or other restrictions. It would be better to consider the user journey for selecting external models.

src/databricks/labs/lakebridge/deployment/switch.py
sundarshankar89: Done

src/databricks/labs/lakebridge/deployment/configurator.py
```diff
@@ -103,6 +108,35 @@ def warehouse_type(_):
...
+            if ep.name
+            and ep.config
+            and ep.config.served_entities
+            and any(getattr(se, "foundation_model", None) is not None for se in ep.config.served_entities)
+        ]
+
+        foundational_model_names = [ep.name for ep in model_endpoints if ep.name]
+
+        if foundational_model_names is None:
+            raise DatabricksError("No Foundation Model serving endpoints found. Aborting the installation.")
```
asnare: This will never trigger, because `foundational_model_names` is always a list even if empty.
```suggestion
        if not foundational_model_names:
            raise DatabricksError("No Foundation Model serving endpoints found. Aborting the installation.")
```

src/databricks/labs/lakebridge/deployment/switch.py
```diff
@@ -0,0 +1,187 @@
...
+        existing_job_id = self._get_existing_job_id()
+        logger.debug("Setting up Switch job in workspace...")
+        try:
+            job_id = self._create_or_update_switch_job(existing_job_id)
+            self._install_state.jobs[self._INSTALL_STATE_KEY] = job_id
+            self._install_state.save()
+            job_url = f"{self._ws.config.host}/jobs/{job_id}"
+            logger.info(f"Switch job created/updated: {job_url}")
+        except (RuntimeError, ValueError, InvalidParameterValue) as e:
+            logger.error(f"Failed to create/update Switch job: {e}")
```
asnare: This path will lead to confusion, because we return normally. Subsequent logs include:

 - `DEBUG: Switch deployment completed.`
 - `INFO: Installation completed successfully! Please refer to the documentation for the next steps.`

Neither are true. The easiest fix right now is to return a boolean indicating success. (Alternatively, we could re-raise the exception.)

Note: we also currently lose the stacktrace on failure. Maybe `logger.exception()` instead.

src/databricks/labs/lakebridge/deployment/switch.py
```diff
@@ -0,0 +1,187 @@
...
+        installation: Installation,
+        install_state: InstallState,
+        product_info: ProductInfo,
+        job_deployer: JobDeployment,
+    ):
+        self._ws = ws
+        self._installation = installation
+        self._install_state = install_state
+        self._product_info = product_info
+        self._job_deployer = job_deployer
```
asnare: I don't think these are used at the moment?

src/databricks/labs/lakebridge/deployment/configurator.py
```diff
@@ -103,6 +108,35 @@ def warehouse_type(_):
...
+        # This logic is implemented to make the default choice always to appear first in the list
+        other_models = sorted(set(foundational_model_names) - {default_choice})
+        choices = [f"[Recommended] {default_choice}", *other_models]
+        selected = self._prompts.choice("Select a Foundation Model serving endpoint:", choices, sort=True)
```
asnare: Nit: `sort=True` here is redundant.
```suggestion
        selected = self._prompts.choice("Select a Foundation Model serving endpoint:", choices, sort=False)
```

src/databricks/labs/lakebridge/deployment/switch.py
sundarshankar89: Now I introduced a SystemExit if switch deployment fails.

asnare (approved):

PR #2071, opened 2025-10-02
Minor pip improvements during `install-transpile`

asnare:
## Changes

This PR updates the way we invoke `pip` during `install-transpile`:

 - It now verifies that it's running within a virtual environment.
 - It only installs binary packages, meaning we won't accidentally try to compile a source package (if found as a dependency).

### Functionality

- affects existing command: `databricks labs lakebridge install-transpile`

### Tests

- manually tested


sundarshankar89 (approved):
LGTM

PR #2075, opened 2025-10-03
remove old unused code

m-abulazm:
### What does this PR do?
The code is unused and uses `DatabaseConnector` so slows down future development on the class in case the API changes which is in progress [here](https://github.com/databrickslabs/lakebridge/pull/1937)


sundarshankar89:
Before I approve, this was the initial step toward automated config generation for reconcile and potentially useful for lineage or other useful features. 
So I understand this hampers the current usage or improvements required for that particular Database Manager.

sundarshankar89 (approved):
LGTM but please see my previous comments.

PR #2083, opened 2025-10-07
Add java 25 to the list of versions we verify we can check

asnare:
## Changes

Java 25 changes slightly the format of the `-version` line. This PR updates the test that checks we can parse the versions to include this variant. (We could, no changes to the parser required.)

### Tests

- updated unit test


m-abulazm (approved):
LGTM

PR #2084, opened 2025-10-07
Remove warning if `config.yml` missing from workspace during `transpile`

asnare:
## Changes

With the UI in place, non-interactive installations of the transpilers will result in there being no `config.yml` in the user workspace. As such we shouldn't warn about this with instructions to the user.

### Functionality

### Tests

- updated unit test

### DIFF
--- a/src/databricks/labs/lakebridge/cli.py
+++ b/src/databricks/labs/lakebridge/cli.py
@@ -184,9 +184,7 @@ class _TranspileConfigChecker:
         transpiler_repository: TranspilerRepository,
     ) -> None:
         if config is None:
-            logger.warning(
-                "No workspace transpile configuration, use 'install-transpile' to (re)install and configure; using defaults for now."
-            )
+            logger.debug("No workspace transpile configuration, using defaults.")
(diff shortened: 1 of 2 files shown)


src/databricks/labs/lakebridge/cli.py
```diff
@@ -184,9 +184,7 @@ def __init__(
         transpiler_repository: TranspilerRepository,
     ) -> None:
         if config is None:
-            logger.warning(
-                "No workspace transpile configuration, use 'install-transpile' to (re)install and configure; using defaults for now."
-            )
+            logger.debug("No workspace transpile configuration, using defaults.")
```
sundarshankar89: nit: 
```suggestion
            logger.debug("No workspace transpile configuration, using command line inputs.")
```

sundarshankar89 (approved):
small nit

src/databricks/labs/lakebridge/cli.py
asnare: I see your point. I've altered it to "[…] _starting from_ defaults."

PR #2092, opened 2025-10-10
rename transpiler product_name to transpiler_id

m-abulazm:
### What does this PR do?
Use transpiler_id instead of product_name within the transpiler repository and installers module. This helps making that code clearer and consistent

the installers use `artifact_id` instead of `transpiler_id` which fits them better

### DIFF
--- a/src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
+++ b/src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
@@ -26,13 +26,13 @@ def _get_schema_query(catalog: str, schema: str, table: str):
         return f"describe table {catalog}.{schema}.{table}"
 
     query = f"""select
-                            lower(column_name) as column_name,
+                            lower(column_name) as col_name,
                              full_data_type as data_type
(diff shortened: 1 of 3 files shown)


src/databricks/labs/lakebridge/transpiler/installers.py
```diff
@@ -225,7 +225,7 @@ def _create_venv(self) -> None:
         else:
             lib_path = venv_path / "Lib" / "site-packages"
         # Handle installing pip ourselves, to help with diagnostics if something goes wrong.
-        builder = venv.EnvBuilder(with_pip=False, prompt=f"{self._product_name}", symlinks=use_symlinks)
+        builder = venv.EnvBuilder(with_pip=False, prompt=f"{self._artifact_id}", symlinks=use_symlinks)
```
asnare: Nit: might as well tidy that up while you're here:
```suggestion
        builder = venv.EnvBuilder(with_pip=False, prompt=self._artifact_id, symlinks=use_symlinks)
```

Issue #1008, opened 2024-10-16
[FEATURE]: Configurable Secrets


jordan-databricks:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Reconcile

### Problem statement

Currently required Reconcile secret keys are hard-coded.  For example, Oracle requires the secret keys user, password, host, port, database stored in some secret scope.  

This is not flexible enough to adhere to all user security policies.  In particular, some users require that all secret keys adhere to a specified pattern that gives information as to their use.  It is not sufficient enough to allow flexibility at the scope level, as is currently done, as scopes might have their own different policies.

Thinking ahead, one might wish to allow other secret providers different than Databricks, for example a customer may have a policy to keep all credentials for system X in provider Y.  For example, AWS secrets manager might be the location for all Snowflake credentials.

### Proposed Solution

Manage this similar to MLFlow, in that allow a user to override the defaults by creating a list of dictionaries similar to the following:

```
auth_config = [
  {
    "provider": "databricks",  (Optional, if this is all that will ever be supported)
    "connector": "oracle",
    "scope": "<scope_name>"
    "user_key": "<oracle_user_key_name>",
    "password_key": "<oracle_password_key_name>",
  },
{
   "provider": "databricks",
   "connector": "snowflake",
   "scope": "<scope name>"
   "sfUser_key": "<sfUser_key_name>"
]

```

m-abulazm:
will be replaced instead by [Unity Catalog connections](https://docs.databricks.com/aws/en/connect/jdbc-connection)

PR #2123, opened 2025-10-28
Implement databricks creds manager

m-abulazm:
### What does this PR do?
* Support databricks credentials
* Standardize usages of credentials manager

### Linked issues

Progresses https://github.com/databrickslabs/lakebridge/issues/1008

### Functionality

### Tests

- [X] manually tested
- [X] added unit tests
- [ ] added integration tests

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -8,8 +8,9 @@ from databricks.labs.lakebridge.__about__ import __version__
 from databricks.labs.lakebridge.connections.database_manager import DatabaseManager
 from tests.integration.debug_envgetter import TestEnvGetter
 
-logging.getLogger("tests").setLevel("DEBUG")
-logging.getLogger("databricks.labs.lakebridge").setLevel("DEBUG")
+logging.getLogger("tests").setLevel(logging.DEBUG)
+logging.getLogger("databricks.labs.lakebridge").setLevel(logging.DEBUG)
+logging.getLogger("databricks.labs.pytester").setLevel(logging.DEBUG)
 
 logger = logging.getLogger(__name__)
 


src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +38,53 @@ def get_secret(self, key: str) -> str:
...
+
+    def get_databricks_secret(self, scope: str, key: str) -> str:
+        return self.get_secret(f"{scope}/{key}")
+
     def get_secret(self, key: str) -> str:
-        raise NotImplementedError("Databricks secret vault not implemented")
+        """Get the secret value given a secret scope & secret key.
+
+        :param key: key in the format 'scope/secret'
+        :return: The decoded UTF-8 secret value.
```
asnare: Across our projects we've settled on [Google's style for docstrings](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods) rather than the reST-style here.


```suggestion
        Args:
            key: key in the format 'scope/secret'
        Returns:
            The decoded UTF-8 secret value.
```

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +38,53 @@ def get_secret(self, key: str) -> str:
...
+            return base64.b64decode(secret.value).decode("utf-8")
+        except NotFound as e:
+            raise KeyError(f'Secret does not exist with scope: {scope} and key: {key_only} : {e}') from e
+        except UnicodeDecodeError as e:
+            raise UnicodeDecodeError(
+                "utf-8",
+                key_only.encode(),
+                0,
+                1,
+                f"Secret {key} has Base64 bytes that cannot be decoded to utf-8 string: {e}.",
```
asnare: This should probably be `ValueError` (from `e`): we're signalling that there's a problem with a user-supplied argument (due to an underlying unicode issue).

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +38,53 @@ def get_secret(self, key: str) -> str:
...
 
-class DatabricksSecretProvider:
+class DatabricksSecretProvider(SecretProvider):
+    def __init__(self, ws: WorkspaceClient):
+        self._ws = ws
+
+    def get_databricks_secret(self, scope: str, key: str) -> str:
+        return self.get_secret(f"{scope}/{key}")
+
     def get_secret(self, key: str) -> str:
```
asnare: Something is a bit confusing here, because we have two arguments named `key` but they mean different things: one is within a scope, and one includes the scope embedded.

It might also be cleaner with respect to argument handling to swap these so that `get_secret()` is implemented by calling `get_databricks_secret()` instead of what we have here.

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +38,53 @@ def get_secret(self, key: str) -> str:
             return key
 
 
-class DatabricksSecretProvider:
+class DatabricksSecretProvider(SecretProvider):
+    def __init__(self, ws: WorkspaceClient):
+        self._ws = ws
+
+    def get_databricks_secret(self, scope: str, key: str) -> str:
+        return self.get_secret(f"{scope}/{key}")
```
asnare: We should raise `ValueError` here if either `scope` or `key` contain a `/` character.

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +38,53 @@ def get_secret(self, key: str) -> str:
...
+        :param key: key in the format 'scope/secret'
+        :return: The decoded UTF-8 secret value.
+
+        Raises:
+          NotFound: The secret could not be found.
+          UnicodeDecodeError: The secret value was not Base64-encoded UTF-8.
+        """
+        key_parts = key.split(sep="/")
+        assert len(key_parts) == 2, "Secret name must be in the format 'scope/secret'"
+        scope, key_only = key_parts[0], key_parts[1]
```
asnare: Outside tests I'm pretty sure we limit assertions to things that are demonstrably true but necessary for the type-checker to properly understand the situation. We shouldn't rely on them for validation. (Looking around I see a few places where this is being done incorrectly, but we shouldn't make it worse.)

That leads to something like this:
```suggestion
        match key.split(sep="/", maxsplit=3):
            case scope, key_only:
                scope = scope
                key_only = key_only
            case _:
                msg = f"Secret key must be in the format 'scope/secret': {key}"
                raise ValueError(msg)
```

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -76,14 +117,23 @@ def _load_credentials(path: Path) -> dict:
         raise FileNotFoundError(f"Credentials file not found at {path}") from e
 
 
-def create_credential_manager(product_name: str, env_getter: EnvGetter) -> CredentialManager:
-    creds_path = cred_file(product_name)
-    creds = _load_credentials(creds_path)
+def create_databricks_secret_provider(ws) -> DatabricksSecretProvider:
```
asnare: Nit:
```suggestion
def create_databricks_secret_provider(ws: WorkspaceClient) -> DatabricksSecretProvider:
```

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -76,14 +117,23 @@ def _load_credentials(path: Path) -> dict:
...
+        creds = creds_or_path
+
+    # Lazily initialize secret providers
+    secret_providers: dict[str, CredentialManager.SecretProviderFactory] = {
+        'local': LocalSecretProvider,
+        'env': partial(EnvSecretProvider, EnvGetter()),
     }
 
+    if ws:
+        secret_providers['databricks'] = partial(create_databricks_secret_provider, ws)
```
asnare: Is the complexity of adding in lazy initialisation here worth the indirection tradeoff?

They're all trivial initialisers, so I'm not sure it's worth it…

I think we can just add in `DatabricksSecretProvider(ws)` if the workspace is provided.

src/databricks/labs/lakebridge/deployment/recon.py
```diff
@@ -61,7 +61,7 @@ def uninstall(self, recon_config: ReconcileConfig | None):
             f"from catalog `{recon_config.metadata_config.catalog}`. Please remove it and the tables inside manually."
         )
         logging.info(
-            f"Won't remove configured reconcile secret scope `{recon_config.secret_scope}`. "
+            f"Won't remove configured reconcile creds from `{recon_config.creds.vault_type}`. "
```
asnare:
```suggestion
            f"Won't remove configured reconcile credentials from `{recon_config.creds.vault_type}`. "
```

tests/unit/connections/test_credential_manager.py
```diff
@@ -45,46 +43,43 @@ def env_credentials():
...
     }
 
 
-@patch('databricks.labs.lakebridge.connections.credential_manager._load_credentials')
-@patch('databricks.labs.lakebridge.connections.credential_manager._get_home')
-def test_local_credentials(mock_get_home, mock_load_credentials, local_credentials, env_getter):
-    mock_load_credentials.return_value = local_credentials
-    mock_get_home.return_value = Path("/fake/home")
-    credentials = create_credential_manager(product_name, env_getter)
+def test_local_credentials(local_credentials):
```
asnare: Can we add type hints to this?

tests/unit/connections/test_credential_manager.py
```diff
@@ -45,46 +43,43 @@ def env_credentials():
...
 
-@patch('databricks.labs.lakebridge.connections.credential_manager._load_credentials')
-@patch('databricks.labs.lakebridge.connections.credential_manager._get_home')
 @patch.dict('os.environ', {'MSSQL_USER_ENV': 'env_user', 'MSSQL_PASSWORD_ENV': 'env_password'})
-def test_env_credentials(mock_get_home, mock_load_credentials, env_credentials, env_getter):
-    mock_load_credentials.return_value = env_credentials
-    mock_get_home.return_value = Path("/fake/home")
-    env_getter.get.side_effect = lambda key: os.environ[key]
-    credentials = create_credential_manager(product_name, env_getter)
+def test_env_credentials(env_credentials):
```
asnare: Can we add type hints to this?

tests/unit/connections/test_credential_manager.py
```diff
@@ -45,46 +43,43 @@ def env_credentials():
...
 
-@patch('databricks.labs.lakebridge.connections.credential_manager._load_credentials')
-@patch('databricks.labs.lakebridge.connections.credential_manager._get_home')
-def test_databricks_credentials(mock_get_home, mock_load_credentials, databricks_credentials, env_getter):
-    mock_load_credentials.return_value = databricks_credentials
-    mock_get_home.return_value = Path("/fake/home")
-    credentials = create_credential_manager(product_name, env_getter)
-    with pytest.raises(NotImplementedError):
-        credentials.get_credentials('mssql')
+def test_databricks_credentials(databricks_credentials, mock_workspace_client):
```
asnare: Can we add type hints to this?

tests/unit/connections/test_credential_manager.py
```diff
@@ -45,46 +43,43 @@ def env_credentials():
...
+    mock_workspace_client.secrets.get_secret.return_value = GetSecretResponse(
+        key='some_key', value=base64.b64encode(bytes('some_secret', 'utf-8')).decode('utf-8')
+    )
+    credentials = create_credential_manager(databricks_credentials, mock_workspace_client)
+    creds = credentials.get_credentials('mssql')
+    assert creds['user'] == 'some_secret'
+    assert creds['password'] == 'some_secret'
+
+
+def test_databricks_credentials_not_found(databricks_credentials, mock_workspace_client):
```
asnare: Can we add type hints to this?

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -76,14 +117,23 @@ def _load_credentials(path: Path) -> dict:
...
 
-    secret_providers = {
-        'local': LocalSecretProvider(),
-        'env': EnvSecretProvider(env_getter),
-        'databricks': DatabricksSecretProvider(),
+def create_credential_manager(creds_or_path: dict | Path, ws: WorkspaceClient | None = None) -> CredentialManager:
+    if isinstance(creds_or_path, Path):
+        creds = _load_credentials(creds_or_path)
+    else:
+        creds = creds_or_path
```
asnare: I like that we've improved things here to avoid the monkey-patching from the unit tests. (Thanks!)
I think we can do this a little cleaner, however, with the following:

Splitting this into two functions: one for the product name, and a different one for the unit tests.
Moving it to a classmethod on the class itself.

That means something like:
class CredentialManager:

    @classmethod
    def create(cls, product_name: str, ws: WorkspaceClient | None = None) -> CredentialManager:
        # Derive path and load...
        cls.from_credentials(credentials, ws)

    @classmethod
    def from_credentials(cls, credentials: RootJsonValue, ws: WorkspaceClient | None = None) -> CredentialManager:
        # Set up providers as before.
        return cls(credentials, secret_providers)
What do you think?

src/databricks/labs/lakebridge/resources/assessments/synapse/dedicated_sqlpool_extract.py
```diff
@@ -23,7 +24,7 @@ def execute():
 
     db_path, creds_file = arguments_loader(desc="Synapse Synapse Dedicated SQL Pool Extract Script")
 
-    cred_manager = create_credential_manager(PRODUCT_NAME, creds_file)
```
asnare, line 26: I'm quite confused by this, because the second argument previously was supposed to be an `EnvGetter` but here it seems to be a path? Maybe you've dug into this and know what this is all about?

From inspection it looks like it might not have been used?

src/databricks/labs/lakebridge/config.py
```diff
@@ -259,19 +259,32 @@ class ReconcileMetadataConfig:
...
     report_type: str
-    secret_scope: str
+    creds: ReconcileCredentialConfig
     database_config: DatabaseConfig
     metadata_config: ReconcileMetadataConfig
     job_id: str | None = None
     tables: ReconcileTablesConfig | None = None
 
+    @classmethod
+    def v1_migrate(cls, raw: dict[str, Any]) -> dict[str, Any]:
```
asnare: Possibly?
```suggestion
    def v1_migrate(cls, raw: dict[str, JsonValue]) -> dict[str, JsonValue]:
```

src/databricks/labs/lakebridge/config.py
```diff
@@ -259,19 +259,32 @@ class ReconcileMetadataConfig:
...
+        secret_scope = raw.pop("secret_scope")
+        raw["version"] = 2
+        raw["creds"] = {"vault_type": "databricks", "source_creds": {"__secret_scope": secret_scope}}
```
asnare: Why `__secret_scope` and not just `secret_scope`?

src/databricks/labs/lakebridge/config.py
```diff
@@ -259,19 +259,32 @@ class ReconcileMetadataConfig:
     volume: str = "reconcile_volume"
 
 
+@dataclass
+class ReconcileCredentialConfig:
+    vault_type: str  # supports local, env, databricks creds.
+    source_creds: dict[str, str]
```
asnare: This sort of thing has been a source of issues in the past, partly because there's no documentation or past record of what's gone in there. To counter this, it might be useful to attach some methods to this class so that it's clear what's going into and out of it.

For example:
```python
    def get_databricks_scope(self) -> str:
        if self.vault_type != "databricks":
            raise ...
        return self._source_creds["secret_scope"]

    @classmethod
    def from_databricks_scope(self, secret_scope: str) -> str:
        return ReconcileCredentialConfig("databricks", {"secret_scope": secret_scope}
```

src/databricks/labs/lakebridge/config.py
```diff
@@ -259,19 +259,32 @@ class ReconcileMetadataConfig:
     volume: str = "reconcile_volume"
 
 
+@dataclass
+class ReconcileCredentialConfig:
+    vault_type: str  # supports local, env, databricks creds.
```
asnare: In these configuration classes we support `Enum`. I think it might be useful to constrain the values here to things that we expect and support. What do you think?

tests/integration/reconcile/query_builder/test_execute.py
```diff
@@ -735,7 +736,7 @@ def mock_for_report_type_data(
     reconcile_config_data = ReconcileConfig(
         data_source="databricks",
         report_type="data",
-        secret_scope="remorph_databricks",
+        creds=ReconcileCredentialConfig(vault_type="databricks", source_creds={"__secret_scope": "fake"}),
```
asnare: Given these examples, it look like `creds` should be an optional field?

src/databricks/labs/lakebridge/config.py
```diff
@@ -259,19 +259,32 @@ class ReconcileMetadataConfig:
...
+    source_creds: dict[str, str]
+
+
 @dataclass
 class ReconcileConfig:
     __file__ = "reconcile.yml"
-    __version__ = 1
```
asnare, line 263: For version bumps we really need a specific unit test to cover the version upgrades. (It's unfortunately a very error-prone path, and bugs in `blueprint` don't help.)

Although I can see some tests that include upgrades as part of their fixture, we really need to narrowly target just the migration of this config. (Let me know if you need any help setting this up.)

src/databricks/labs/lakebridge/config.py
asnare: I think we need a better name for this: it doesn't hold the credentials… it's more of a vault configuration?

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -76,14 +117,23 @@ def _load_credentials(path: Path) -> dict:
...
-    creds = _load_credentials(creds_path)
+def create_databricks_secret_provider(ws) -> DatabricksSecretProvider:
+    return DatabricksSecretProvider(ws)
+
 
-    secret_providers = {
-        'local': LocalSecretProvider(),
-        'env': EnvSecretProvider(env_getter),
-        'databricks': DatabricksSecretProvider(),
+def create_credential_manager(creds_or_path: dict | Path, ws: WorkspaceClient | None = None) -> CredentialManager:
```
asnare: Although I can see some tests, I can't see where the `ws` argument is provided in any of the non-test call-sites. Aside from the tests, is it actually being used?

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
```diff
@@ -74,7 +74,7 @@ def create_recon_dependencies(
             engine=reconcile_config.data_source,
             spark=spark,
             ws=ws_client,
-            secret_scope=reconcile_config.secret_scope,
+            secret_scope=reconcile_config.creds.source_creds["__secret_scope"],
```
asnare: Should this now be using the `CredentialManager` mechanism?

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +38,53 @@ def get_secret(self, key: str) -> str:
             return key
 
 
-class DatabricksSecretProvider:
+class DatabricksSecretProvider(SecretProvider):
+    def __init__(self, ws: WorkspaceClient):
+        self._ws = ws
+
+    def get_databricks_secret(self, scope: str, key: str) -> str:
```
asnare: Is this actually used? I can't find any callers.

If we update the implementation so that `.get_secret()`  calls this method that's fine, but as things stand I can't find where it's called from.

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +38,53 @@ def get_secret(self, key: str) -> str:
...
+        key_parts = key.split(sep="/")
+        assert len(key_parts) == 2, "Secret name must be in the format 'scope/secret'"
+        scope, key_only = key_parts[0], key_parts[1]
+
+        try:
+            secret = self._ws.secrets.get_secret(scope, key_only)
+            assert secret.value is not None
+            return base64.b64decode(secret.value).decode("utf-8")
+        except NotFound as e:
+            raise KeyError(f'Secret does not exist with scope: {scope} and key: {key_only} : {e}') from e
```
asnare: I think this is different to the other providers: they just return the `key` if the secret cannot be found, whereas here we raise an exception instead.

What do you think the providers should do? I think they need to be consistent.

asnare (commented):
I've highlighted some style and design issues that I think need to be resolved, but appreciate that this is the start of what is needed for #1008. On the testing side I really like that we've eliminated some monkey-patching during tests. (Some integration tests would be nice.)

One big concern I have is that I don't see where we're using the new provider because we don't pass the `WorkspaceClient` in anywhere that I can see. Can you elaborate a bit on the situation there?

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
m-abulazm: yes and it is implemented in https://github.com/databrickslabs/lakebridge/pull/2159 to make the reviews more manageable.
And https://github.com/databrickslabs/lakebridge/pull/2157 adds the prompts to configure the creds.

this current PR can go to main first since it is backwards-compatible without the other two.

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: using credential manager in reconcile and supplying `ws` is done in https://github.com/databrickslabs/lakebridge/pull/2159.

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: we should not raise an error. the return type should be optional and it is up to the caller how to handle missing secrets.

I did not want to change lots of things in one go so the 
the implementation you see here of `DatabricksSecretProvider` is copied from src/databricks/labs/lakebridge/reconcile/connectors/secrets.py without changing the way it works which led to some inconsistency. 

I would address your comment here in a later PR if you dont mind

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: I dont think this is on the user. databricks returned a malformed response and not the utf8 base64 value it should return

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: agree, previous impl was that `create_databricks_secret_provider` initialized `ws` without it being passed so the lazy init was needed

src/databricks/labs/lakebridge/resources/assessments/synapse/dedicated_sqlpool_extract.py
m-abulazm: yes this was a bug, I have no idea how it passed the formatter. there were one or two more and I fixed all of them

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: I dont want to support primitives on the factory. it led to bugs as you saw on some of the call sites. `product_name` as well is a very proprietary concept and should be handled on a different level.

I would prefer
```
@classmethod
def from_file(cls, creds_file: Path) -> CredentialManager:
    # load...
    cls.create(credentials, None)

@classmethod
def create(cls, credentials_mappings: RootJsonValue, ws: WorkspaceClient | None = None) -> CredentialManager:
    # Set up providers as before.
    return cls(credentials, secret_providers)
```

we can add `from_product_name` if you deem it a must. what do you think?

src/databricks/labs/lakebridge/resources/assessments/synapse/dedicated_sqlpool_extract.py
m-abulazm: cc @sundarshankar89

src/databricks/labs/lakebridge/config.py
m-abulazm: it is documented in `ReconcileConfig#v1_migrate()`. I would keep the new class clean without migration tech debt

src/databricks/labs/lakebridge/config.py
m-abulazm: it documents that it is a system property and not the same as normal properties that are expected here e,g, `{url: zx@s, user: xx, password: xx, ...}`

src/databricks/labs/lakebridge/config.py
m-abulazm: agree but renaming this will require changing across three PRs. I would rename it after reviewing all PRs

src/databricks/labs/lakebridge/config.py
m-abulazm: I am not a fan of python. Enum equality does not work as expected. I wanted to use `VaultType = Literal['local', 'env', 'databricks']` but Literal did not work with un/marshalling logic in the sdk so added a `post_init` validate

tests/integration/reconcile/query_builder/test_execute.py
m-abulazm: databricks does not use the creds as the workspace is already authenticated

src/databricks/labs/lakebridge/config.py
m-abulazm: it is implicitly tested in some of the tests. if someone removes those tests, we would see it in the coverage

src/databricks/labs/lakebridge/connections/credential_manager.py
asnare: I'll rephrase then: it should be `ValueError` because there's an issue with the _caller_-supplied argument: we can't handle the requested `key`.

As things stand:

 - we lose the details of the actual unicode error;
 - the `UnicodeDecodeError` we're throwing doesn't provide the correct information in its arguments: the `object` isn't what we were decoding, and the offsets are arbitrary.

If this error _does_ occur there won't be anything useful to help us understand what actually went wrong.

For this reason it really should be:
```python
try:
   ...
except UnicodeDecodeError as e:
   msg = f"Secret {key} has Base64 bytes that cannot be decoded to UTF-8 string"
   raise ValueError(msg) from e
```

m-abulazm: changed as requested

src/databricks/labs/lakebridge/connections/credential_manager.py
asnare: I don't fully understand this…

> I dont want to support primitives on the factory. it led to bugs as you saw on some of the call sites

I think for now a factory using `product_name` is fine: it's enforcing the intent that there be a single file per application. Happy to give it a better name though, as you suggest, so we have something like:

```python
@classmethod
def from_product_name(cls, product_name: str, ws: WorkspaceClient | None = None) -> CredentialManager:
    # Derive path and load...
    return cls.from_credentials(credentials, ws)

@classmethod
def from_credentials(cls, credentials: RootJsonValue, ws: WorkspaceClient | None = None) -> CredentialManager:
    # Set up providers as before.
    return cls(credentials, secret_providers)
```

m-abulazm: > I don't fully understand this…

> > I dont want to support primitives on the factory. it led to bugs as you saw on some of the call sites

the bug -> https://github.com/databrickslabs/lakebridge/pull/2123#discussion_r2572114874

src/databricks/labs/lakebridge/config.py
asnare: I'm afraid that's not a compelling argument, even though I appreciate the frustration. The situation is:

 - Once it's merged to `main`, we need to treat it as live: it could be released in that state without the other PRs. We can't just say "`main` is not releasable because these other PRs need to be merged": that would hamstring us in terms of when we release, particularly if something urgent pops up.
 - Even if we don't release, end users can install `main` in advance of a release. This is something we often do when asking them to test a bug-fix or check if something has been fixed in advance of a release being made.
 - Once it's 'live' we have to either live with it or  those other PRs will then need to implement migration. That's even more work.

asnare: Ah, yes that's true in terms of the attributes and how they've changed with different versions of this class.

However the missing piece is that here we have:

 - `vault_type: str`: When we encounter this we don't know what values are valid? (Has this ever changed?)
 - `source_creds: dict[str, str]`: Again, we don't know anything about what should be in here. What are the valid keys and values? Do they depend on the vault type?

My suggestion here isn't aimed at helping understand the schema changes, but rather understanding what is allowed in those fields right now.

The challenge here is that the schema here doesn't help us understand what's valid data: rather than being a model, this is effectively what is often called a meta-model and the challenge is that while it's flexible enough to allow anything, because it allows anything it's very difficult through inspection to know how it's supposed to work (and whether it's working as intended).

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: added factory methods

src/databricks/labs/lakebridge/config.py
sundarshankar89: Enumerating the Vault type as ENV, LOCAL, or DATABRICKS can be a safe approach.

m-abulazm: we do need a meta model here. it can be anything depending on the data source and the credentials it implements. the `__secret_scope` is implemented just to support older config when only `secret_scope` was allowed. 

I added `get_databricks_secret_scope` although it is used in just one place - something that we talked about before that we dont want to do.

in PR https://github.com/databrickslabs/lakebridge/pull/2159, the data sources use the meta model plus it is documented in reconcile docs PR https://github.com/databrickslabs/lakebridge/pull/2157

only databricks vault is supported for now and it is enforced through a post_init

m-abulazm: renamed

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +36,69 @@ def get_secret(self, key: str) -> str:
...
+
+        Args:
+            key: key in the format 'scope/secret'
+        Returns:
+            The decoded UTF-8 secret value.
+
+        Raises:
+          ValueError: The secret key must be in the format 'scope/secret'.
+          KeyError: The secret could not be found.
+          UnicodeDecodeError: The secret value was not Base64-encoded UTF-8.
```
asnare: I think this can go now.
```suggestion
```

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +36,69 @@ def get_secret(self, key: str) -> str:
...
+        self._ws = ws
+
     def get_secret(self, key: str) -> str:
-        raise NotImplementedError("Databricks secret vault not implemented")
+        """Get the secret value given a secret scope & secret key.
+
+        Args:
+            key: key in the format 'scope/secret'
+        Returns:
+            The decoded UTF-8 secret value.
```
asnare: Nit:
```suggestion
            The secret value.
```

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -32,18 +36,69 @@ def get_secret(self, key: str) -> str:
...
+        credentials = _load_credentials(path)
+        return cls.from_credentials(credentials, ws)
+
+    @classmethod
+    def from_file(cls, path: Path, ws: WorkspaceClient | None = None) -> "CredentialManager":
+        credentials = _load_credentials(path)
+        return cls.from_credentials(credentials, ws)
+
+    @classmethod
+    def from_credentials(cls, credentials: dict, ws: WorkspaceClient | None = None) -> "CredentialManager":
```
asnare, line 93: Can we type-hint the credentials properly?

It looks to me like it's `RootJsonValue`.

src/databricks/labs/lakebridge/config.py
asnare: I'm aware it's implicitly tested, but we still need tests that specific target the migration: as I mentioned it's a very error-prone path that has historically caused us a lot of trouble.

src/databricks/labs/lakebridge/connections/credential_manager.py
asnare: Okay, add a `TODO` marker please.

src/databricks/labs/lakebridge/config.py
asnare: Okay, the main problem with this is that a meta-model is useful mainly when passing something from A to B without any understanding or access of the data without any coupling between the data and the type. (eg. A document database taking JSON, a message queuing system, etc.)

However that's not the situation here: it's a closed configuration, and our code is tightly coupled to the data because we have to explicitly support (and have implementations for) anything in here.

With this in mind, I want us to switch to:

```python
from enum import Enum, unique

@unique
class VaultType(Enum):
    DATABRICKS = "databricks"
    SOMETHING_ELSE = "something-else"

@dataclass
class DatabricksVaultConfiguration:
    secret_scope: str

@dataclass
class SomethingElseVaultConfiguration
    ...

@dataclass
class ReconcileCredentialsConfig:
    vault_type: VaultType
    vault_configuration: DatabricksVaultConfiguration | SomethingElseVaultConfiguration

    def __post_init__(self) -> None:
        # Enforce that vault_type is consistent with the type of vault_configuration.
        ...
```

src/databricks/labs/lakebridge/connections/credential_manager.py
m-abulazm: there is no nesting or arrays in our dict. `RootJsonValue` allows for this which is different from what we need; just a mapping from string to string

asnare:
> Can we keep that as an open dictionary

To be clear, I believe that the parameters (=options) for the connection itself needs to be a `dict[str,str]`. That's the `N` bit in my description above.

However this PR is about `M` and I although it's worth clarifying the surrounding context, I don't want to lose sight of the focus here being on the vault configuration where an open dictionary is neither necessary nor desirable.

m-abulazm:
@asnare @sundarshankar89 

I think we’re conflating two separate pieces, which is why I’m getting lost:
* The connection definition (N): this PR is trying to model the `CREATE CONNECTION … OPTIONS (...)` object from Databricks, including which options exist and which of them are backed by secrets. That needs to stay flexible (`dict[str, str]`‑like) because different JDBC drivers and CUJs require different options, and we want to stay close to UC semantics.​
* The vault backend (M): how we resolve a given key to a secret value (Databricks Secrets, env, local). That can be a small, strongly‑typed config, and I’m fine tightening that separately. `CredentialManager` and `SecretProvider` is already strongly-typed

My goal in this PR is specifically:
* Define the connection/options structure (N) in a way that mirrors `CREATE CONNECTION`
* Ensure that for a given connection we only read the specific secrets referenced by those options, not whole scopes.
* Support legacy config that defined one secret scope only

I agree the vault configuration (M) should not be an open dict, but that feels like a separate, follow‑up refactor focused on `CredentialManager` and the vault types. Right now I’d like to close on the connection shape first (N), since that’s what Recon users will actually configure and what needs to align with the new JDBC connection model.

src/databricks/labs/lakebridge/config.py
m-abulazm: added test https://github.com/databrickslabs/lakebridge/pull/2123/changes#diff-217a439192bc495c66132a58a1bf7a92ee70f0eb9702c6b00f60644e726479b5R106

m-abulazm:
@asnare @sundarshankar89  
 
The goal of the dict is to contain the **connection properties**, for example:  

```json
{
  "user": "scope/key",
  "password": "scope/key2",
  "host": "scope3/key4",
  "port": "scope4/key5",
  "database": "scope/key3",
  "encrypt": "scope/key",
  "trustServerCertificate": "scope/key"
}
```

i.e. mapping logical connection options to `(scope, key)` references so that, at runtime, we only read those specific secrets via the vault, not entire scopes. The use of the `__secret_scope` key is just to support older configs that only specified a single scope and didn’t use this more dynamic mapping.

In other words, my intent in this PR was always to model the **connection properties / OPTIONS(...) side (N)** and how they reference secrets, not to introduce a general vault configuration model. The stacked PRs already use this dict in the connection resolution path; the vault backend shape (M) can still be tightened separately if we want.

src/databricks/labs/lakebridge/connections/credential_manager.py
asnare: Let's go with this then:

```suggestion
    def from_credentials(cls, credentials: dict[str, str], ws: WorkspaceClient | None = None) -> "CredentialManager":
```

src/databricks/labs/lakebridge/reconcile/connectors/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+
+from databricks.labs.lakebridge.connections.credential_manager import build_credentials, CredentialManager
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ReconcileCredentialsConfig:
+    vault_type: str
+    vault_secret_names: dict[str, str]
```
asnare, line 14: We seem to have lost some conversations here[^1][^2], but continuing on…

In what way does enum equality not work as expected? (It seems fine to me.)

Let's go with:
```python
from enum import Enum, unique

@unique
class VaultType(Enum):
    DATABRICKS = "databricks"

@dataclass
class ReconcileCredentialsConfig:
    vault_type: VaultType = VaultType.DATABRICKS
    vault_secret_names: dict[str, str]
```

[^1]: https://github.com/databrickslabs/lakebridge/pull/2123#discussion_r2619707057
[^2]: https://github.com/databrickslabs/lakebridge/pull/2123#discussion_r2585384865

src/databricks/labs/lakebridge/config.py
asnare: This conversation has disappeared from the review screen, so I've continued it [here](https://github.com/databrickslabs/lakebridge/pull/2123/changes#r2668540685).

src/databricks/labs/lakebridge/config.py
asnare: This conversation has disappeared from the review screen, so I've continued it [here](https://github.com/databrickslabs/lakebridge/pull/2123/changes#r2668540685).

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -60,6 +115,23 @@ def _get_secret_value(self, key: str) -> str:
         return self._provider.get_secret(key)
 
 
+def build_credentials(vault_type: str, source: str, credentials: dict) -> dict:
```
asnare, line 118: Type arguments on those `dict` hints.

src/databricks/labs/lakebridge/reconcile/connectors/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+_SOURCE_CREDENTIALS_MAP = {
+    "databricks": [],
+    "snowflake": _SNOWFLAKE_REQUIRED_CREDS,
+    "oracle": _ORACLE_REQUIRED_CREDS,
+    "mssql": _TSQL_REQUIRED_CREDS,
+    "synapse": _TSQL_REQUIRED_CREDS,
+}
+
+
+def build_recon_creds(source: str, secret_scope: str) -> ReconcileCredentialsConfig | None:
```
asnare, line 52: This looks to me like a factory `@classmethod` that should be on `ReconcileCredentialsConfig`?

src/databricks/labs/lakebridge/reconcile/connectors/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+        raise ValueError(f"Unsupported source system: {source}")
+
+    missing = [k for k in required_keys if not creds.vault_secret_names.get(k)]
+    if missing:
+        raise ValueError(
+            f"Missing mandatory {source} credentials. " f"Please configure all of {required_keys}. Missing: {missing}"
+        )
+
+
+def load_and_validate_credentials(
```
asnare, line 81: This looks to me like it should be a method on `ReconcileCredentialsConfig`.

src/databricks/labs/lakebridge/reconcile/connectors/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+
+    if source == "snowflake":
+        logger.warning("Please specify the Snowflake authentication method in the credentials config.")
+        parsed["pem_private_key"] = f"{secret_scope}/pem_private_key"
+        parsed["sfPassword"] = f"{secret_scope}/sfPassword"
+
+    return ReconcileCredentialsConfig("databricks", parsed)
+
+
+def validate_creds(creds: ReconcileCredentialsConfig, source: str) -> None:
```
asnare, line 69: Ah, now this is interesting: why can't we validate the keys when loading/initialisation the data class instead of deferring until here? This reveals a design issue, and creates a problem for users:

 - The set of keys required for this to be valid depends on the `source`;
 - The source is not specified in the configuration;
 - How can a user know if the configuration is "right" or not?
 - What if it's configured incorrectly?

Essentially these are different ways of saying the file is source-specific but the design makes that very difficult.

How can we deal with this?

There's a related problem: with this design I think we now _require_ that the user always provide a full set of keys for a specific source, even though previously we documented they just had to have a specific set of keys in the vault. (Can't we therefore default to those if the user hasn't provided a value for the key to look up?)

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -51,33 +52,55 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
...
+                f"Please update the creds config and include the necessary keys. Docs: {self._DOCS_URL}."
+            )
+            creds.vault_secret_names.pop("pem_private_key")
+
+        self._creds_or_empty = load_and_validate_credentials(creds, self._ws, "snowflake")
+
+        # Ensure at least one authentication method is provided
+        assert any(
+            self._creds.get(k) for k in ("sfPassword", "pem_private_key")
+        ), "Missing Snowflake credentials. Please configure any of [sfPassword, pem_private_key]."
```
asnare, line 82: We only use `assert` in production code as a hint for the type-checker; this should be an explicit check and raise an exception.

src/databricks/labs/lakebridge/reconcile/connectors/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+
+    keys = _SOURCE_CREDENTIALS_MAP.get(source)
+    if not keys:
+        raise ValueError(f"Unsupported source system: {source}")
+    parsed = {key: f"{secret_scope}/{key}" for key in keys}
+
+    if source == "snowflake":
+        logger.warning("Please specify the Snowflake authentication method in the credentials config.")
+        parsed["pem_private_key"] = f"{secret_scope}/pem_private_key"
+        parsed["sfPassword"] = f"{secret_scope}/sfPassword"
```
asnare, line 64: This doesn't look right: we need one or the other, and it looks like users go from a valid configuration before migration to an invalid one after migration that requires them to manually edit the file to remove one of the entries?

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -51,33 +52,55 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
...
+    @property
+    def _creds(self):
+        if self._creds_or_empty:
+            return self._creds_or_empty
+        raise RuntimeError("Snowflake credentials have not been loaded. Please call load_credentials() first.")
+
+    def load_credentials(self, creds: ReconcileCredentialsConfig) -> "SnowflakeDataSource":
+        password = creds.vault_secret_names.get("sfPassword")
+        pem_key = creds.vault_secret_names.get("pem_private_key")
+        if password and pem_key:  # user did not specify auth method after migrating from secret scope
```
asnare, line 70: We can't have this: the user's configuration was valid before this PR, and now after migration it's broken and needs to be edited? I'm afraid we need to do better than that.

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -51,33 +52,55 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
...
-    ):
+    def __init__(self, engine: Dialect, spark: SparkSession, ws: WorkspaceClient):
         self._engine = engine
         self._spark = spark
         self._ws = ws
-        self._secret_scope = secret_scope
+        self._creds_or_empty: dict[str, str] = {}
+
+    @property
+    def _creds(self):
```
asnare, line 62: Type hint the return-type, please.

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -51,33 +52,55 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
...
+        assert any(
+            self._creds.get(k) for k in ("sfPassword", "pem_private_key")
+        ), "Missing Snowflake credentials. Please configure any of [sfPassword, pem_private_key]."
+
+        # Process PEM private key if provided
+        if self._creds.get("pem_private_key"):
+            self._creds["pem_private_key"] = SnowflakeDataSource._get_private_key(
+                self._creds["pem_private_key"],
+                self._creds.get("pem_private_key_password"),
+            )
```
asnare, line 89: This doesn't look right:

 - We don't currently support authentication with private keys.
 - Some of these vault keys aren't mentioned (or validated) when loading.
 - Storing the extracted private key over the top of the encrypted version feels a bit weird.

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -51,33 +52,55 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
...
+        password = creds.vault_secret_names.get("sfPassword")
+        pem_key = creds.vault_secret_names.get("pem_private_key")
+        if password and pem_key:  # user did not specify auth method after migrating from secret scope
+            logger.warning(
+                f"Snowflake auth not specified after migrating from secret scope so defaulting to sfPassword. "
+                f"Please update the creds config and include the necessary keys. Docs: {self._DOCS_URL}."
+            )
+            creds.vault_secret_names.pop("pem_private_key")
+
+        self._creds_or_empty = load_and_validate_credentials(creds, self._ws, "snowflake")
```
asnare, line 77: The instance state shouldn't be updated until after it's been fully processed and validated, some of which is happening below. Suggestion:
```python
        loaded_secrets = load_and_validate_credentials(creds, self._ws, "snowflake")
        # ...
        self._creds_or_empty = loaded_secrets
        return self
```

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -51,33 +52,55 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
...
     @property
     def get_jdbc_url(self) -> str:
-        try:
-            sf_password = self._get_secret('sfPassword')
-        except (NotFound, KeyError) as e:
-            message = "sfPassword is mandatory for jdbc connectivity with Snowflake."
-            logger.error(message)
-            raise NotFound(message) from e
+        if not self._creds:
+            raise RuntimeError("Credentials not loaded. Please call `load_credentials(ReconcileCredentialsConfig)`.")
```
asnare, line 96: Isn't this redundant? I think just using `self._creds` (below) will already trigger the same exception.

src/databricks/labs/lakebridge/reconcile/utils.py
```diff
@@ -17,10 +17,13 @@ def initialise_data_source(
...
-    secret_scope: str,
+    creds: ReconcileCredentialsConfig | None,
 ):
-    source = create_adapter(engine=get_dialect(engine), spark=spark, ws=ws, secret_scope=secret_scope)
-    target = create_adapter(engine=get_dialect("databricks"), spark=spark, ws=ws, secret_scope=secret_scope)
+    source = create_adapter(engine=get_dialect(engine), spark=spark, ws=ws)
+    target = create_adapter(engine=get_dialect("databricks"), spark=spark, ws=ws)
+    if engine != "databricks":
+        assert creds, "Credentials must be provided for non-Databricks sources"
+        source.load_credentials(creds)
```
asnare, line 26: This should be at the top, and raise `ValueError` rather than use an assertion.

src/databricks/labs/lakebridge/reconcile/connectors/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+
+from databricks.sdk import WorkspaceClient
+
+from databricks.labs.lakebridge.connections.credential_manager import build_credentials, CredentialManager
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ReconcileCredentialsConfig:
```
asnare, line 12: This needs to be moved back to `config.py` I think: it's not just a part of our configuration, but being persisted any changes need to be handled carefully for compatibility. Migration code might be needed.

If we leave it here I suspect people will modify it without realising how sensitive it is to changes.

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -51,33 +52,55 @@ class SnowflakeDataSource(DataSource, SecretsMixin, JDBCReaderMixin):
...
-            f"jdbc:{SnowflakeDataSource._DRIVER}://{self._get_secret('sfAccount')}.snowflakecomputing.com"
-            f"/?user={self._get_secret('sfUser')}&password={sf_password}"
-            f"&db={self._get_secret('sfDatabase')}&schema={self._get_secret('sfSchema')}"
-            f"&warehouse={self._get_secret('sfWarehouse')}&role={self._get_secret('sfRole')}"
-        )
+            f"jdbc:{SnowflakeDataSource._DRIVER}://{self._creds['sfUrl']}"
+            f"/?user={self._creds['sfUser']}&password={self._creds['sfPassword']}"
+            f"&db={self._creds['sfDatabase']}&schema={self._creds['sfSchema']}"
+            f"&warehouse={self._creds['sfWarehouse']}&role={self._creds['sfRole']}"
+        )  # TODO Support PEM key auth
```
asnare, line 103: This has piqued my curiosity, can you elaborate?

m-abulazm: when data is read with JDBC, only password authentication is supported.
PEM gets used only when querying the tables information schema because JDBC is not used. JDBC is used in the next steps afterwards 

it leads to an inconsistent UX hence the TODO to support pem as well given the docs mention PEM and password auth is supported.  currently, the user gets a warning log that password is required.

sundarshankar89: No, we support Pem as well. Our current labs infra setup is using a PEM File.

m-abulazm: @sundarshankar89

I see this code on main branch `databricks/labs/lakebridge/reconcile/connectors/snowflake.py:67`
this code gets used when reading data

```
    def get_jdbc_url(self) -> str:
        try:
            sf_password = self._get_secret('sfPassword')
        except (NotFound, KeyError) as e:
            message = "sfPassword is mandatory for jdbc connectivity with Snowflake."
```

PR #2145, opened 2025-11-13
add e2e reconcile test for databricks source

m-abulazm:
### What does this PR do?
adds an end-to-end (e2e) test for reconcile to validate the setup of testing reconcile e2e

* adds first e2e test
* moves older ignored reconcile specs that needed sandbox infra access

* this does not e2e test all sources and more tests will be added in following PRs 

### Tests
- [X] added acceptance tests


src/databricks/labs/lakebridge/cli.py
```diff
@@ -652,9 +653,11 @@ def reconcile(*, w: WorkspaceClient) -> None:
         ctx.workspace_client,
         ctx.installation,
         ctx.install_state,
-        ctx.prompts,
     )
-    recon_runner.run(operation_name=RECONCILE_OPERATION_NAME)
+
+    _, job_run_url = recon_runner.run(operation_name=RECONCILE_OPERATION_NAME)
+    if ctx.prompts.confirm(f"Would you like to open the job run URL `{job_run_url}` in the browser?"):
```
sundarshankar89, line 660: I see where you are going which this, but let us not do this given our overall goal of going away from user prompting to UI.

src/databricks/labs/lakebridge/cli.py
```diff
@@ -668,10 +671,11 @@ def aggregates_reconcile(*, w: WorkspaceClient) -> None:
         ctx.workspace_client,
         ctx.installation,
         ctx.install_state,
-        ctx.prompts,
     )
 
-    recon_runner.run(operation_name=AGG_RECONCILE_OPERATION_NAME)
+    _, job_run_url = recon_runner.run(operation_name=AGG_RECONCILE_OPERATION_NAME)
+    if ctx.prompts.confirm(f"Would you like to open the job run URL `{job_run_url}` in the browser?"):
```
sundarshankar89, line 679: Same as above

src/databricks/labs/lakebridge/cli.py
sundarshankar89: Check the implementation in `llm_transpile`

tests/unit/test_cli_other.py
```diff
@@ -67,12 +67,34 @@ def test_cli_configure_secrets_config(mock_workspace_client):
...
-    with patch("databricks.labs.lakebridge.reconcile.runner.ReconcileRunner.run", return_value=True):
+    ctx_mock = MagicMock()
+    prompts = MockPrompts(
+        {
+            r"Would you like to open the job run URL .*": "no",
+        }
+    )
+    ctx_mock.prompts = prompts
+
+    with (
```
sundarshankar89: use `ctx.replace` instead, however with my comment on going away from prompts do you thign we will needs this?

pyproject.toml
```diff
@@ -97,18 +97,18 @@ reconcile = "databricks.labs.lakebridge.reconcile.execute:main"
...
-                   "pylint --output-format=colorized -j 0 src tests"]
-verify       = ["black --check .",
-                   "ruff check .",
-                   "mypy --disable-error-code 'annotation-unchecked' .",
-                   "pylint --output-format=colorized -j 0 src tests"]
+test           = "pytest --cov src --cov-report=xml tests/unit"
+test-install   = "pytest --cov src --cov-report=xml tests/integration/install"
+test-reconcile = "pytest --cov src tests/integration/reconcile --ignore=tests/integration/reconcile/system_tests --durations 20"
+coverage       = "pytest --cov src tests --cov-report=html --ignore=tests/integration/install --ignore=tests/integration/connections --ignore=tests/integration/assessments"
+fmt            = ["black .",
```
sundarshankar89: What is causing these new empty spaces?

src/databricks/labs/lakebridge/cli.py
m-abulazm: then my refactoring makes sense, the prompting is where it should be in the cli and not in the runner that should not know anything about it.

pyproject.toml
m-abulazm: i added them because `test-reconcile` is longer than `integration` in order to keep the formatting. that is one of the downsides of aligning all assignments because if one changes, all have to change

tests/unit/test_cli_other.py
m-abulazm: the cli only takes input workspace client. I have to patch

src/databricks/labs/lakebridge/cli.py
asnare: This looks like a move: hoisting it from inside the `.run()` call to outside. I'm neutral on that; even though I agree with @sundarshankar89 about the original code, that's not the point of this PR.

Overall I'd prefer to just drop the prompt and leave the existing logging of the URL: every terminal allows the job to be opened by just clicking on it.

That's not the point or focus of this PR, however, which is why I'm neutral on leaving the UX as-is for now.

I assume this was moved to make some of the tests easier in some way?

tests/unit/test_cli_other.py
asnare: I think this will disappear if the prompts are dropped, but for future reference the pattern to use when needing to mock the `ApplicationContext` is:

 - Add a `ctx: ApplicationContext | None = None` argument to the command.
 - At the start of the command:
   ```python
   if ctx is None:
       ctx = ApplicationContext(w)
   ```

This lets tests inject a stub or mock, which we do in several places.

We generally try to avoid monkey-patching throughout our code-base: it's normally a big smell and indicates design flaws. I know it's in place in various places throughout the codebase but we shouldn't be making the situation worse.

Makefile
```diff
@@ -36,8 +36,8 @@ test:
 test-install:
 	hatch run test-install
 
-integration: setup_spark_remote
-	hatch run integration
+test-reconcile: setup_spark_remote
+	hatch run test-reconcile
```
asnare: Let's not do this. I'll provide more context in the review summary.

asnare (changes requested):
I like that we've now got an integration test covering reconciliation, although normally I'd also like something to cover the CLI entry-point. That's an omission from the original code and although it would also be an improvement though, it not being present is not a problem with this PR.

That said, we need to change back to the existing structure with two sets of tests:

 - `unit`: fast tests, typically focused on a narrow area of the code, mocks and so forth for convenience. No dependencies on testing infrastructure. Anyone that clones the repo should be able to run these.
 - `integration`: slower, typically covering larger interactions between components in our code and aligned with user workflows. These typically use actual components rather than mocks or stubs. Although they _may_ rely on testing infrastructure and external resources, its absence should result in skipped tests rather than failures. Although for convenience these can be run directly via pytest, the reference runner is our `labs test` command: this is also what CI/CD does.

I know there's an existing `test-install` entry-point but that's a mistake and we should be trying to eliminate it as a special category. (I understand that its presence is confusing and potentially misleading.)

src/databricks/labs/lakebridge/cli.py
m-abulazm: `Prompts` are part of the cli - they wait till a user interacts. Other entry points like here (an integration test) do not require any user interaction nor the UI. 

Removing it made the tests easier that I dont have to mock it since the real implementation suspends the execution waiting for a user input. I moved it to the cli because it only belongs there

tests/unit/test_cli_other.py
m-abulazm: fixed. the cli functions now accept an `ApplicationContext` factory 
```
application_ctx_factory: Callable[[WorkspaceClient], ApplicationContext] = ApplicationContext
```

Makefile
m-abulazm: I need this convenience in my day to day. `labs test` errors out on my machine and I didnt really get how it works

src/databricks/labs/lakebridge/deployment/job.py
```diff
@@ -35,6 +36,8 @@ def __init__(
         self._installation = installation
         self._install_state = install_state
         self._product_info = product_info
+        if self._is_testing():
+            self._test_env = TestEnvGetter()
```
asnare: I think we can do better than this. It's a pity the existing code mixes test-specific logic into the production code… that's an indicator that we've got a design flaw.

Let's not make it more pervasive though. Instead I think we can pass the cluster configuration (optionally) as an argument when initialising the instance? Tests can then pass in the cluster they want to be used and it can default to the reconciliation cluster. (Ideally the cluster would come from the `ApplicationContext` but that's for another day.)

I think that would allow us to get rid of this optional `_test_env` attribute?

src/databricks/labs/lakebridge/connections/debug_envgetter.py
asnare, line 1: `TestEnvGetter` is intended only for use during integration tests, I'd really prefer to avoid including it in our production code. (As noted: we should be able to use dependency injection from the tests rather than have production code consult this directly.)

tests/unit/deployment/test_job.py
asnare, line 1: I'm a bit curious about why this unit test was removed? (I'm not sure I see an obvious replacement, so it's not clear to me if it was moved?)

tests/integration/reconcile_system_tests/test_recon_databricks.py
```diff
@@ -0,0 +1,60 @@
...
+source_catalog_or_schema = (
+    recon_config.database_config.source_catalog
+    if recon_config.database_config.source_catalog
+    else recon_config.database_config.source_schema
+)
+filename = f"recon_config_{recon_config.data_source}_{source_catalog_or_schema}_{recon_config.report_type}.json"
+
+
+def test_recon_databricks(ws):
+    ctx = ApplicationContext(ws)
```
asnare: Don't forget to type-hint, and I think we need a better name for this test: what we're testing is that the job can be submitted.

Although we wait for the outcome of the job, we don't:

 - Check whether the job succeeded or not.
 - Check the reconciliation results.

tests/integration/reconcile/test_recon_databricks.py
```diff
@@ -0,0 +1,60 @@
...
+)
+filename = f"recon_config_{recon_config.data_source}_{source_catalog_or_schema}_{recon_config.report_type}.json"
+
+
+def test_recon_databricks(ws):
+    ctx = ApplicationContext(ws)
+    ctx.replace(product_info=ProductInfo.for_testing(LakebridgeConfiguration))
+    ctx.installation.save(recon_config)
+    ctx.installation.upload(filename, TABLE_RECON_JSON.encode())
+    ctx.workspace_installation.install(config)
```
asnare: Where do we clean up the resources that this test creates?

In general integration tests need to clean up after themselves, but I must confess the codebase doesn't really seem to be set up to make this easy to do.

tests/unit/test_cli_other.py
```diff
@@ -66,14 +66,25 @@ def test_cli_configure_secrets_config(mock_workspace_client):
...
+        }
+    )
+    ctx_mock.prompts = prompts
+    return ctx_mock
+
+
 def test_cli_reconcile(mock_workspace_client):
-    with patch("databricks.labs.lakebridge.reconcile.runner.ReconcileRunner.run", return_value=True):
-        cli.reconcile(w=mock_workspace_client)
+    with patch("databricks.labs.lakebridge.reconcile.runner.ReconcileRunner.run", return_value=(MagicMock(), True)):
```
asnare: The mocked return value here doesn't match the type hint of that function (`tuple[Wait[Run], str]`).

tests/unit/test_cli_other.py
```diff
@@ -66,14 +66,25 @@ def test_cli_configure_secrets_config(mock_workspace_client):
...
+        cli.reconcile(w=mock_workspace_client, application_ctx_factory=app_factory)
 
 
 def test_cli_aggregates_reconcile(mock_workspace_client):
-    with patch("databricks.labs.lakebridge.reconcile.runner.ReconcileRunner.run", return_value=True):
-        cli.aggregates_reconcile(w=mock_workspace_client)
+    with patch("databricks.labs.lakebridge.reconcile.runner.ReconcileRunner.run", return_value=(MagicMock(), True)):
```
asnare: This tuple returned by the mock doesn't match the type hint of that function (`tuple[Wait[Run], str]`).

tests/integration/deployment/test_job.py
asnare, line 1: Ahh… here they are. It looks like we've accidentally moved these unit tests into the integration suite?

src/databricks/labs/lakebridge/connections/debug_envgetter.py
sundarshankar89: +1

tests/integration/reconcile/test_recon_databricks.py
m-abulazm: watch dog deletes the resources and they are created with remove_after

tests/unit/deployment/test_job.py
m-abulazm: changed back

tests/integration/deployment/test_job.py
m-abulazm: reverted

src/databricks/labs/lakebridge/cli.py
```diff
@@ -642,34 +643,41 @@ def _override_workspace_client_config(ctx: ApplicationContext, overrides: dict[s
 
 
 @lakebridge.command
-def reconcile(*, w: WorkspaceClient) -> None:
+def reconcile(
+    *, w: WorkspaceClient, application_ctx_factory: Callable[[WorkspaceClient], ApplicationContext] = ApplicationContext
```
asnare: If you're going to go with a factory, let's just go with:
```python
ctx_factory: Callable[[WorkspaceClient], ApplicationContext]
```

src/databricks/labs/lakebridge/cli.py
```diff
@@ -642,34 +643,41 @@ def _override_workspace_client_config(ctx: ApplicationContext, overrides: dict[s
...
+
+    _, job_run_url = recon_runner.run(operation_name=RECONCILE_OPERATION_NAME)
+    if ctx.prompts.confirm(f"Would you like to open the job run URL `{job_run_url}` in the browser?"):
+        webbrowser.open(job_run_url)
 
 
 @lakebridge.command
-def aggregates_reconcile(*, w: WorkspaceClient) -> None:
+def aggregates_reconcile(
+    *, w: WorkspaceClient, application_ctx_factory: Callable[[WorkspaceClient], ApplicationContext] = ApplicationContext
```
asnare: (Also `ctx_factory` here.)

tests/unit/test_cli_other.py
```diff
@@ -66,14 +66,25 @@ def test_cli_configure_secrets_config(mock_workspace_client):
         mock_recon_config.assert_called_once_with(mock_workspace_client)
 
 
+def app_factory(_):
+    ctx_mock = MagicMock()
```
asnare: Hmm. Maybe this:
```suggestion
def app_factory(w: WorkspaceClient) -> ApplicationContext:
    ctx_mock = create_autospec(spec=ApplicationContext, spec_set=True)
    type(ctx_mock).workspace_client = PropertyMock(return_value=w)
```

tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -101,6 +100,13 @@ def test_databricks_read_schema_happy(mock_spark):
     assert columns
 
 
+def test_databricks_read_schema_happy_sandbox(spark, ws):
```
asnare: Type hints, please.

src/databricks/labs/lakebridge/config.py
```diff
@@ -251,6 +251,12 @@ class ReconcileMetadataConfig:
     volume: str = "reconcile_volume"
 
 
+@dataclass
+class DeployReconcileConfig:
```
asnare: This is aimed at configuration for the job that we deploy.

Maybe: 

```suggestion
class ReconcileJobConfig:
```

src/databricks/labs/lakebridge/config.py
```diff
@@ -261,6 +267,7 @@ class ReconcileConfig:
     secret_scope: str
     database_config: DatabaseConfig
     metadata_config: ReconcileMetadataConfig
+    deployment_overrides: DeployReconcileConfig | None = None
```
asnare: Should this be `job_overrides`?

tests/integration/reconcile/test_recon_databricks.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    filename = f"recon_config_{recon_config.data_source}_{source_catalog_or_schema}_{recon_config.report_type}.json"
+    return filename
+
+
+def test_recon_databricks_job_succeeds(
+    ws: WorkspaceClient, recon_config: ReconcileConfig, recon_config_filename: str
+) -> None:
+    ctx = ApplicationContext(ws)
+    ctx.replace(product_info=ProductInfo.for_testing(LakebridgeConfiguration))
```
asnare: Nit:
```suggestion
    ctx = ApplicationContext(ws).replace(
        product_info=ProductInfo.for_testing(LakebridgeConfiguration),
    )
```

tests/integration/reconcile/test_recon_databricks.py
```diff
@@ -0,0 +1,95 @@
...
+from databricks.labs.lakebridge.reconcile.runner import ReconcileRunner
+from databricks.labs.blueprint.wheels import ProductInfo
+
+from tests.integration.debug_envgetter import TestEnvGetter
+
+TEST_JOBS_PURGE_TIMEOUT = timedelta(hours=1, minutes=15)
+
+
+def get_test_purge_time() -> str:
+    return (datetime.now(timezone.utc) + TEST_JOBS_PURGE_TIMEOUT).strftime("%Y%m%d%H")
```
asnare: There's a pytest fixture to use instead of this: `watchdog_remove_after`

tests/integration/reconcile/test_recon_databricks.py
```diff
@@ -0,0 +1,95 @@
...
+        ctx.workspace_client,
+        ctx.install_state,
+    )
+    run, _ = recon_runner.run(operation_name=RECONCILE_OPERATION_NAME)
+    result = run.result()
+
+    assert result.status
+    assert result.status.termination_details
+    assert result.status.termination_details.type
+    assert result.status.termination_details.type.value == TerminationTypeType.SUCCESS.value
```
asnare: Somewhere my earlier comments seem to have disappeared, and I'm not sure where they've gone because I think this code moved. To summarise:

1. I asked about where this is cleaned up;
2. @m-abulazm replied that the watchdog handles this.

There are some issues with this test and that approach, firstly being that the watchdog is intended to be a last resort: tests are supposed to clean up after themselves. I understand we have a lot of tests that don't do this, but we need to be better at this so let's do it properly, which I'll get back to further below.

Aside from the cleanup issue, there's also a flakiness and reliability issue because this test assumes exclusive access to the testing infrastructure and resources. This assumption is incorrect:

 - CI/CD can have concurrent test runs;
 - Developers can have this test running concurrently;
 - Mixtures of the above.

Although some of the test resources are deconflicted (install directory, job name) to allow concurrent runs, there seems to be an issue with the catalog: concurrent tests will all use the same `sandbox` catalog and schemas.

This is a pattern that we've seen before, and solved for. For this test it means the following pattern:

1. Somehow the `.uninstall()` method needs to be invoked to clean up that installation. This is preferably _after_ the test result is known (to distinguish it from test failure), and the following options are available:
    - A yielding fixture is often used, although here I'm not sure if that works because it needs to be configured for the specific test. But this is a common path, and we make `factory()` for this.
    - `request.addfinalizer()` is another path.
    - There's always `try/finally` if nothing else seems appropriate.
2. Use the `make_catalog` fixture to create a unique catalog within which the test can work. (The fixture will ensure cleanup.)

If you'd like any help with this, please let me know.

asnare (changes requested):

src/databricks/labs/lakebridge/cli.py
m-abulazm: blueprint will not be able to pass that parameter. `ApplicationContext` is part of lakebridge hence the default value

tests/integration/reconcile/test_recon_databricks.py
m-abulazm: @asnare 

This test now explicitly cleans up the installation and no longer relies on the watchdog as the primary mechanism. The installation is managed by a fixture and calls `ctx.workspace_installation.uninstall(config)` in the teardown section after the test finishes (whether it passes or fails). This makes the lifecycle of the installation clear and deterministic.

For isolation, the test no longer assumes global schemas like `test_source` / `test_target`. Instead, it uses `make_schema` to create a per‑test schema in the shared `sandbox` catalog, and `make_table` to create source and target tables inside that schema. Both tables are populated via a small SQL insert.

So all test artifacts live in an isolated namespace while still respecting the constraint that we must use `sandbox`.

If you see any further opportunities to tighten the isolation or simplify the fixture setup, happy to iterate.

src/databricks/labs/lakebridge/cli.py
asnare: I meant:

```suggestion
    *, w: WorkspaceClient, ctx_factory: Callable[[WorkspaceClient], ApplicationContext] = ApplicationContext
```

tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -80,32 +80,41 @@ def _get_snowflake_options(self):
...
+def test_sql_server_read_schema_happy(mock_spark: SparkSession) -> None:
     mock_ws = create_autospec(WorkspaceClient)
     connector = TSQLServerDataSourceUnderTest(mock_spark, mock_ws)
 
     columns = connector.get_schema("labs_azure_sandbox_remorph", "dbo", "reconcile_in")
     assert columns
 
 
-def test_databricks_read_schema_happy(mock_spark):
+def test_databricks_read_schema_happy(mock_spark: SparkSession) -> None:
```
asnare, line 92: Hmm, it's not a change that's been introduced by this PR, but I'm a bit confused about the `mock_spark` fixture: it doesn't seem to be a mock at all but rather a real spark session.

What is this connected to? Is it connected to the acceptance Databricks workspace?

tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -80,32 +80,41 @@ def _get_snowflake_options(self):
...
     mock_ws = create_autospec(WorkspaceClient)
     connector = DatabricksDataSource(get_dialect("databricks"), mock_spark, mock_ws, "my_secret")
 
     mock_spark.sql("CREATE DATABASE IF NOT EXISTS my_test_db")
     mock_spark.sql("CREATE TABLE IF NOT EXISTS my_test_db.my_test_table (id INT, name STRING) USING parquet")
     df = mock_spark.sql("SELECT * FROM my_test_db.my_test_table")
-    df.createGlobalTempView("my_global_test_view")
-    columns = connector.get_schema(None, "global_temp", "my_global_test_view")
+    random_view = f"test_view_{uuid.uuid4().hex}"
```
asnare, line 95: This has triggered me. :) I see that it's trying to do, but there's more going on here. Those `IF NOT EXISTS` clauses above hint at the problem: this test doesn't seem to clean up after itself.

The pattern should be:

 - Use a randomised database/schema name.
 - Everything else with a fixed name is fine, because it's scoped.
 - Delete the database when done.

(It's unclear to me if the spark session is connected to a real workspace: if it is it should use the `make_schema` fixture.)

tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -80,32 +80,41 @@ def _get_snowflake_options(self):
...
+    columns = connector.get_schema(None, "global_temp", random_view)
+
+    assert columns
+    assert mock_spark.catalog.dropGlobalTempView(random_view)
+
+
+def test_databricks_read_schema_happy_sandbox(spark: SparkSession, ws: WorkspaceClient) -> None:
+    connector = DatabricksDataSource(get_dialect("databricks"), spark, ws, "my_secret")
+
+    columns = connector.get_schema("main", "lakebridge", "diamonds")
```
asnare: This test needs to create any resources it's going to use. That particular table is currently present, but all resources (with a few exceptions) the acceptance environment are ephemeral and regularly wiped.

(Use the `make_table` fixture to do this.)

tests/integration/reconcile/test_recon_databricks.py
```diff
@@ -0,0 +1,165 @@
...
+          "join_columns": ["color", "clarity"]
+        }
+      ]
+    }
+    """
+    )
+
+    return template.substitute(
+        schema=recon_schema.name, catalog=recon_schema.catalog_name, src=src_table.name, tgt=tgt_table.name
+    )
```
asnare: Is there a reason we can't just use an f-string here instead of the template mechanism?

tests/integration/reconcile/test_recon_databricks.py
```diff
@@ -0,0 +1,165 @@
...
+    (0.23, 'Good', 'E', 'VS1'),
+    (0.29, 'Premium', 'I', 'VS2'),
+    (0.29, 'Gold', 'Invariant', 'VS22'),
+    (0.31, 'Good', 'J', 'SI2');
+"""
+
+
+@pytest.fixture
+def recon_schema(make_schema) -> SchemaInfo:
+    catalog = "sandbox"
```
asnare: The "sandbox" catalog is ephemeral, and regularly wiped. If this test needs a specific catalog it needs to create (and clean up afterwards). The `make_catalog` fixture can be used for this.

asnare (changes requested):

tests/integration/reconcile/test_recon_databricks.py
m-abulazm: yes because of the curly brackets of the json objects will trigger interpolation and we dont want that. I prefer template than escaping all curly bracktes

tests/integration/reconcile/connectors/test_read_schema.py
asnare: > This test needs to create any resources it's going to use. That particular table is currently present, but all resources (with a few exceptions) the acceptance environment are ephemeral and regularly wiped.
> 
> (Use the `make_table` fixture to do this.)

It turns out the `main` and `sandbox` catalogs aren't being regularly wiped, but they're also not supposed to be used by tests: tests like this should be bringing along any resources they need.

tests/integration/reconcile/test_recon_databricks.py
asnare: > The "sandbox" catalog is ephemeral, and regularly wiped. If this test needs a specific catalog it needs to create (and clean up afterwards). The `make_catalog` fixture can be used for this.

I just had a longer look, and the "sandbox" catalog is not ephemeral. However we shouldn't be using it for tests: the `make_catalog` fixture should be used.

tests/integration/reconcile/connectors/test_read_schema.py
m-abulazm: sandbox is used because of user permissions. I dont have permissions to create new catalogs

tests/integration/reconcile/test_recon_databricks.py
asnare: Ahh, of-course.

I think here we can create a dictionary literal and then use `json.dumps`:

```suggestion
    config = {
      "source_schema": recon_schema.name,
      "target_catalog": recon_schema.catalog_name,
      "target_schema": recon_schema.name,
      "tables": [
        {
          "source_name": src_table.name,
          "target_name": tgt_table.name,
          "join_columns": ["color", "clarity"],
        }
      ]
    }
    return json.dumps(config)
```

tests/integration/reconcile/connectors/test_read_schema.py
m-abulazm: that is in-memory spark "mock spark". the temp view is dropped after the test runs. 

we can migrate it to sandbox spark and use pytester but that is not within the scope of this PR.

tests/integration/reconcile/connectors/test_read_schema.py
m-abulazm: done

tests/integration/reconcile/connectors/test_read_schema.py
m-abulazm: we have a local spark that we run for reconcile/integration specs. before this PR, we never used databricks spark. 

I agree the name is confusing and needs to change. local spark is managed using a shell script in our repo - afaik, you worked on that script

tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -80,32 +81,48 @@ def _get_snowflake_options(self):
...
     mock_spark.sql("CREATE TABLE IF NOT EXISTS my_test_db.my_test_table (id INT, name STRING) USING parquet")
     df = mock_spark.sql("SELECT * FROM my_test_db.my_test_table")
-    df.createGlobalTempView("my_global_test_view")
-    columns = connector.get_schema(None, "global_temp", "my_global_test_view")
+    random_view = f"test_view_{uuid.uuid4().hex}"
+    df.createGlobalTempView(random_view)
+    columns = connector.get_schema(None, "global_temp", random_view)
+
+    assert columns
+    assert mock_spark.catalog.dropGlobalTempView(random_view)
```
asnare: This should be in a finally block to ensure cleanup.

m-abulazm: done

asnare (approved):

src/databricks/labs/lakebridge/deployment/installation.py
```diff
@@ -105,11 +102,6 @@ def install(self, config: LakebridgeConfiguration):
 
     def uninstall(self, config: LakebridgeConfiguration):
         # This will remove all the Lakebridge modules
-        if not self._prompts.confirm(
```
sundarshankar89, line 108: this logic is moved inside uninstall.

PR #2150, opened 2025-11-17
remove redundant reconcile config params

### LINKED ISSUE #518: [Feature]: Enhancement of interface for  experimental function for reconcile

### Category of Bug / Issue

ReconcileError

### Current Behavior

In `reconcile/execute.py` file `reconcile` function these both `table_recon: TableRecon`, `reconcile_config: ReconcileConfig` params has below redundant variables 
    source_schema: str
    target_catalog: str
    target_schema: str
    source_catalog: str | None = None
    
Due to this, the user has to create redundant information.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

_No response_

### Sample Query

_No response_

### Operating System

macOS

### Version

latest via Databricks CLI

m-abulazm:
### What does this PR do?
Remove redundant reconcile config

Resolves #518 

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/docs/lakebridge/docusaurus.config.ts
+++ b/docs/lakebridge/docusaurus.config.ts
@@ -21,7 +21,6 @@ const config: Config = {
   projectName: 'Lakebridge', // Usually your repo name.
 
   onBrokenLinks: 'throw',
-  onBrokenMarkdownLinks: 'throw',
   onDuplicateRoutes: 'throw',
   onBrokenAnchors: 'throw',
 
@@ -32,7 +31,12 @@ const config: Config = {
     defaultLocale: 'en',
     locales: ['en'],
   },
-  markdown: { mermaid: true },
+  markdown: {
+    mermaid: true,
+    hooks: {
+      onBrokenMarkdownLinks: 'throw',
+    },
+  },
   themes: ['@docusaurus/theme-mermaid'],
 
   plugins: [
--- a/docs/lakebridge/package.json
+++ b/docs/lakebridge/package.json
@@ -15,32 +15,35 @@
     "typecheck": "tsc"
   },
   "dependencies": {
-    "@docusaurus/core": "3.8.0",
-    "@docusaurus/preset-classic": "3.8.0",
-    "@docusaurus/theme-mermaid": "3.8.0",
-    "@mdx-js/react": "^3.0.0",
-    "@radix-ui/react-slot": "^1.1.1",
+    "@docusaurus/core": "3.9.2",
+    "@docusaurus/preset-classic": "3.9.2",
+    "@docusaurus/theme-classic": "3.9.2",
+    "@docusaurus/theme-mermaid": "3.9.2",
+    "@mdx-js/react": "^3.1.1",
+    "@radix-ui/react-slot": "^1.2.4",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "docusaurus-lunr-search": "3.6.1",
-    "docusaurus-plugin-image-zoom": "^2.0.0",
-    "lucide-react": "^0.471.1",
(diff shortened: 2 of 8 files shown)


gueniai (approved):
LGTM

docs/lakebridge/static/lakebridge_reconcile/lakebridge_recon_main.html
asnare, line 1: I'm not sure what the changes in here are, this looks like static documentation setup rather than content-related?

src/databricks/labs/lakebridge/config.py
```diff
@@ -212,17 +212,7 @@ class TableRecon:
     __file__ = "recon_config.yml"
     __version__ = 1
 
-    source_schema: str
-    target_catalog: str
-    target_schema: str
     tables: list[Table]
-    source_catalog: str | None = None
```
asnare, line 219: If we're changing the schema of something we store in the user's workspace, I think we need:

 - To bump the version number.
 - Tests for the migration/loading path.

Even if loading ignores (now unknown) fields, an older version of the code will choke if it tries to load an instance persisted with this new schema.

src/databricks/labs/lakebridge/config.py
```diff
@@ -269,8 +259,6 @@ class ReconcileConfig:
     secret_scope: str
     database_config: DatabaseConfig
     metadata_config: ReconcileMetadataConfig
-    job_id: str | None = None
-    tables: ReconcileTablesConfig | None = None
```
asnare, line 273: Also here, regarding version/schema change.

asnare (changes requested):
We're modifying a versioned configuration file that we store in the user's workspace, so we really need to deal with compatibility and ensure there is test coverage for compatibility.

docs/lakebridge/static/lakebridge_reconcile/lakebridge_recon_main.html
m-abulazm: this is reconcile notebook exported as html from databricks workspace. the change in the notebook is deleting the removed config

src/databricks/labs/lakebridge/config.py
m-abulazm: we can bump the version but I would not (here at least).
* this was never set in the old code and has no impact on the existing installations
* https://github.com/databrickslabs/lakebridge/pull/2123 bumps the version so no need to do it wice
* these are optional values so the older code will continue working

src/databricks/labs/lakebridge/config.py
m-abulazm: since this is a backwards compatible change, I didn't bump the version. So I added a test for testing that.

> , an older version of the code will choke if it tries to load an instance persisted with this new schema.

And I am not aware of any forward-compatibility support in blueprint. or did I misunderstood something?

asnare: Here's the issue:
> an older version of the code will choke if it tries to load an instance persisted with this new schema.

The scenario is:
1. Someone runs with this version of the code.
2. The file is created.
3. For whatever reason (downgrade, cut'n'paste to another team-member, different environment), someone runs an older version of the code.

This leads to:
 - If the version is _not_ bumped: the older version will choke because it expects fields that are missing.
 - If the version _is_ bumped: the older version will detect that it can't be loaded and raise an `IllegalState` error with something like "expected version 1, got 2".

Both are errors, but the second one is controlled and expected, and also much easier to diagnose.

m-abulazm: got it. just pushed version 2

PR #2157, opened 2025-11-20
prompt for recon secrets

m-abulazm:
### What does this PR do?
Add prompts during `configure-reconcile` for the data sources' credentials. this allows users to use local, env or databricks secrets.
this is more flexible and secure than the current approach that requires users to create a secret specifically for reconcile and now users can reuse their existing secrets. 

Also allows reconcile to support new vault types e.g. azure, google secrets as we implement more supported backends in the future

### Linked issues

Progresses https://github.com/databrickslabs/lakebridge/issues/1008, https://github.com/databrickslabs/lakebridge/pull/2123, https://github.com/databrickslabs/lakebridge/pull/2159

### Functionality

- [X] modified existing command: `databricks labs lakebridge configure-reconcile`

### Tests

- [X] manually tested
- [X] added unit tests
- [ ] added integration tests

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v6
         with:
-          fetch-depth: 1
+          fetch-depth: 0
 
       - name: Install Python
         uses: actions/setup-python@v6
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -8,8 +8,9 @@ from databricks.labs.lakebridge.__about__ import __version__
 from databricks.labs.lakebridge.connections.database_manager import DatabaseManager
 from tests.integration.debug_envgetter import TestEnvGetter
 
-logging.getLogger("tests").setLevel("DEBUG")
-logging.getLogger("databricks.labs.lakebridge").setLevel("DEBUG")
+logging.getLogger("tests").setLevel(logging.DEBUG)
+logging.getLogger("databricks.labs.lakebridge").setLevel(logging.DEBUG)
+logging.getLogger("databricks.labs.pytester").setLevel(logging.DEBUG)
 
 logger = logging.getLogger(__name__)
 


gueniai (approved):
LGTM

src/databricks/labs/lakebridge/helpers/recon_config_utils.py
```diff
@@ -103,74 +21,108 @@ def _prompt_snowflake_connection_details(self) -> tuple[str, dict[str, str]]:
...
         """
         logger.info(
             f"Please answer a couple of questions to configure `{ReconSourceType.ORACLE.value}` Connection profile"
         )
-        user = self._prompts.question("Enter User")
-        password = self._prompts.question("Enter Password")
-        host = self._prompts.question("Enter host")
-        port = self._prompts.question("Enter port")
-        database = self._prompts.question("Enter database/SID")
+        user = self._prompts.question("Enter User Secret")
```
sundarshankar89:
```suggestion
        user = self._prompts.question("Enter User Secret Key")
```

sundarshankar89: I would rather name end as `key` or `name`

src/databricks/labs/lakebridge/helpers/recon_config_utils.py
```diff
@@ -103,74 +21,108 @@ def _prompt_snowflake_connection_details(self) -> tuple[str, dict[str, str]]:
...
-        connection_details = self._connection_details()
-        logger.debug(f"Storing `{self._source}` connection details as Secrets in Databricks Workspace...")
-        self.store_connection_secrets(scope_name, connection_details)
+    def prompt_recon_creds(self, source: str) -> tuple[str, dict[str, str]]:
+        logger.info(
+            "\nChoose vault type (local | env | databricks) \nlocal means values are read as plain text \nenv means values are read "
+            "from environment variables fall back to plain text if not variable is not found\ndatabricks means values are read from Databricks Secrets\n",
+        )
+        secret_vault_type = str(
+            self._prompts.choice("Enter secret vault type (local | env | databricks)", ["local", "env", "databricks"])
```
sundarshankar89: for reconcile we will always need to select `databricks` secret never use local or env mode.

m-abulazm: @sundarshankar89 `databricks` is the most secure option. we can turn off the others. reasons to keep them:
* env vars allows end-users to use cluster Init scripts or .env files
* useful in CI
* local can be useful for experimenting with reconcile and enable fast prototyping

sundarshankar89: No, those are not safe options recommended by Databricks. 
Depdency Injection can help us during tests.

docs/lakebridge/docs/reconcile/recon_notebook.mdx
```diff
@@ -104,20 +105,31 @@ class ReconcileMetadataConfig:
 ```
 If not set the default values will be used to store the metadata. The default resources are created during the installation
 of Lakebridge.
+- `creds_or_secret_scope`: The credentials to use to connect to the data source. Made optional for backwards compatibility.
```
sundarshankar89: Do we want to give this flexibility ?
I would rather force them to use secret scope for reconcile do not open up. 

if this is required for testing then we can explore other options.

m-abulazm: changed back to secret scope

PR #2163, opened 2025-11-24
Improve handling of output from LSP servers

### LINKED ISSUE #2164: [BUG]: Using `transpile` hangs on large logging output from LSP plugins
### Category of Bug / Issue

Converter bug

### Current Behavior

Using `transpile` can sometimes hang rather than complete normally. This typically happens when using version 0.1.20 or later of the BB plugin. This can happen when the underlying `dbconv` process completes and the BB plugin is sending the conversion results back to the CLI. The logs from the BB plugin, written to stderr, are never visible in the CLI.

### Expected Behavior

The CLI should not hang, and the logs from the BB plugin should be processed normally irrespective of their size.

### Steps To Reproduce

Install the following versions:

 - Lakebridge 0.11.0
 - BB 0.20.0

During the configuration questionnaire choose a SQL dialect such as `mssql` and ensure that the BB backend is configured.

Generate a SQL file that is larger than 64KiB in size (e.g. 100KiB), then try to convert it via `databricks labs lakebridge transpile`. The SQL file will convert, but the CLI will hang because the stderr processing fails if more than 64KiB is present on a single line.

asnare:
## Changes

LSP servers use stdin/stdout for the LSP calls, and stderr as a general logging facility. Currently we mirror the stderr output as logs, but due to the implementation details this fails when an individual line received from stderr is longer than 64KiB in size.

Although #2160 contains a hot fix for this problem, this PR improves the handling in a more robust manner:

 - Long lines from the LSP server are now detected and broken up into chunks, as several log entries. (This is preferable to a large limit because it means we have an upper bound on memory usage rather than potentially buffering forever and running out of memory.)
 - The cause of the error when a long line hits has been addressed, however if we hit some other error during stderr processing we now log the (critical!) error so that it's at least clear something has gone wrong. (The nature of this is that the CLI might hang, but it will be noisy about it.)

The primary change here is to handle line-breaking and decoding ourselves rather than relying on the `StreamReader.readline()` implementation. (The behaviour of the latter cannot be reliably controlled when a line exceeds the configured `limit`, which defaults to 64K.)

Further changes are needed to the test LSP server (`lsp_server.py`) that we use for testing, they are out of scope for this PR.

### Linked issues

Supersedes #2160, closes #2164.

This PR is the basis for databrickslabs/blueprint#320: once that's available the changes in here can be simplified/refactored.

### Functionality

### Tests

- manually tested
- added unit tests

### DIFF
--- a/src/databricks/labs/lakebridge/install.py
+++ b/src/databricks/labs/lakebridge/install.py
@@ -345,6 +345,10 @@ class WorkspaceInstaller:
         source_catalog = None
         if source == ReconSourceType.SNOWFLAKE.value:
             source_catalog = self._prompts.question(f"Enter source catalog name for `{source.capitalize()}`")
+        if source == ReconSourceType.DATABRICKS.value:
+            source_catalog = self._prompts.question(
+                f"Enter source catalog name for `{source.capitalize()}`", default="hive_metastore"
+            )
 
         schema_prompt = f"Enter source schema name for `{source.capitalize()}`"
         if source in {ReconSourceType.ORACLE.value}:
--- a/tests/unit/test_install.py
+++ b/tests/unit/test_install.py
@@ -778,6 +778,87 @@ def test_configure_reconcile_no_existing_installation(ws: WorkspaceClient) -> No
     )
 
 
+@patch("webbrowser.open")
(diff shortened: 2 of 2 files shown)


m-abulazm (approved):
LGTM

sundarshankar89 (approved):
LGTM

PR #2184, opened 2025-12-09
Improve launch of python-based LSP server within IDEs

asnare:
## Changes

This PR tweaks the way we launch python-based LSP launchers, mainly to help with developer workflows in the latest IntelliJ release. The issue here is that when running individual tests via the IDE, as of the latest release the (hatch-created) virtual environment is not activated: this causes many tests to run the stub LSP server against the system python, or fail because the system only provides `python3` (and not `python`).

The key change here is that if the LSP server is python-based but there's no dedicated `.venv` for it, the LSP server is now always launched with the current python interpreter. 

As a related minor improvement, we also raise an error earlier if the executable we're about to launch can't be found.

### Tests

- manually tested
- existing unit tests
- existing integration tests

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -45,39 +45,6 @@ jobs:
           codecov_yml_path: codecov.yml
           token: ${{ secrets.CODECOV_TOKEN }}
 
-  integration-python:
-    runs-on:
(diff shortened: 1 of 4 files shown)


gueniai (approved):
LGTM

PR #2185, opened 2025-12-09
[TECH DEBT]: Refactor `@lsp_feature` mechanism for marking LSP client callbacks

asnare:
### What does this PR do?

This PR refactors the `@lsp_feature` decorator mechanism that was implemented internally to mark LSP client callbacks. We're going to have a lot more callbacks, and the existing mechanism had some drawbacks:

 - It relied on mutable global state within the module. This has been eliminated.
 - It didn't support asynchronous callbacks.

The existing mechanism has been replaced with a self-contained mechanism encapsulated within a base class. (For convenience the decorator remains a module function so it can be easily invoked.) The new mechanism also supports asynchronous callbacks.

The PyGLS library doesn't allow for easy subclassing of `LanguageClient` with methods on the subclass as callbacks: callbacks can only be set up _after_ the client has been instantiated, which poses a problem for subclassing like this. The decorator mechanism here works as such:

 - The decorator itself merely marks the callbacks, for later registration. (We use the same mechanism to do this as PyTest uses with its decorators.)
 - During initialisation registration is handled by locating marked methods on the current instance and registering them.

### Tests

- existing unit tests
- existing integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,66 @@
 # Version changelog
 
+## 0.11.2
+
+# Analyzer
+
+- Normalized complexity categories in the analyzer from “COMPLEX/VERY_COMPLEX” to “HIGH/VERY_HIGH” for clearer reports.
+# Converters
+
+## Morpheus
+### Snowflake
+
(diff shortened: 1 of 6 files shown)


sundarshankar89 (approved):
LGTM

gueniai (approved):
LGTM

PR #2186, opened 2025-12-11
Profiler Test Connection

sundarshankar89:
### What does this PR do?
Profiler Test Connection

`databricks labs lakebridge test-profiler-connection`

### Linked issues

Resolves #..

### Functionality

- [x] added new CLI command: `databricks labs lakebridge test-profiler-connection`
- [ ] ... +add your own

### Tests

- [x] manually tested
- [ ] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/__init__.py
+++ b/src/databricks/labs/lakebridge/__init__.py
@@ -4,6 +4,7 @@ from databricks.sdk.core import with_user_agent_extra, with_product
 from databricks.labs.blueprint.entrypoint import is_in_debug
 from databricks.labs.blueprint.logger import install_logger
 from databricks.labs.lakebridge.__about__ import __version__
+from databricks.labs.lakebridge.helpers.telemetry_utils import get_entrypoint_from_env
 
 # Ensure that anything that imports this (or lower) submodules triggers setup of the blueprint logging.
 install_logger()
@@ -26,3 +27,7 @@ with_user_agent_extra("lakebridge", __version__)
 
 # Add lakebridge/<version> for re-packaging of lakebridge, where product name is omitted
 with_product("lakebridge", __version__)
+
+# Detect and track execution entrypoint (CLI vs desktop app vs Databricks app)
+_entrypoint = get_entrypoint_from_env()
+with_user_agent_extra("entrypoint", _entrypoint)
--- a/src/databricks/labs/lakebridge/helpers/telemetry_utils.py
+++ b/src/databricks/labs/lakebridge/helpers/telemetry_utils.py
@@ -1,4 +1,11 @@
+import os
+from collections.abc import Mapping
+
 from databricks.sdk.useragent import alphanum_pattern, semver_pattern
+
+# Entrypoint constants
+DEFAULT_ENTRYPOINT = "cli"
+VALID_ENTRYPOINTS = {DEFAULT_ENTRYPOINT, "desktop-app", "databricks-app"}
 
 
 def make_alphanum_or_semver(value: str) -> str:
@@ -11,3 +18,24 @@ def make_alphanum_or_semver(value: str) -> str:
             char = '_'
         result.append(char)
     return "".join(result)
+
+
+def get_entrypoint_from_env(environ: Mapping[str, str] | None = None) -> str:
+    """Detect execution entrypoint from LAKEBRIDGE_ENTRYPOINT env var.
+
+    Args:
+        environ: Optional environment dict. If None, uses os.environ.
+
+    Returns:
+        One of "cli", "desktop-app", or "databricks-app". Defaults to "cli" if
+        the environment variable is missing or contains an invalid value.
+    """
+    if environ is None:
+        environ = os.environ
+    entrypoint = environ.get("LAKEBRIDGE_ENTRYPOINT")
+    if not entrypoint:
+        return DEFAULT_ENTRYPOINT
+
+    normalized = entrypoint.lower().strip()
+
+    return normalized if normalized in VALID_ENTRYPOINTS else DEFAULT_ENTRYPOINT
--- a/tests/unit/helpers/test_telemetry_utils.py
+++ b/tests/unit/helpers/test_telemetry_utils.py
@@ -1,7 +1,7 @@
 import pytest
 from databricks.sdk.useragent import alphanum_pattern, semver_pattern
 
-from databricks.labs.lakebridge.helpers.telemetry_utils import make_alphanum_or_semver
+from databricks.labs.lakebridge.helpers.telemetry_utils import make_alphanum_or_semver, get_entrypoint_from_env
 
 
 @pytest.mark.parametrize(
@@ -21,6 +21,28 @@ from databricks.labs.lakebridge.helpers.telemetry_utils import make_alphanum_or_
         "@&x2",
     ],
 )
-def test_make_alphanum_or_semver(value: str):
+def test_make_alphanum_or_semver(value: str) -> None:
     value = make_alphanum_or_semver(value)
     assert alphanum_pattern.match(value) or semver_pattern.match(value)
+
+
+@pytest.mark.parametrize(
+    "env,expected",
+    [
+        ({"LAKEBRIDGE_ENTRYPOINT": "cli"}, "cli"),
+        ({"LAKEBRIDGE_ENTRYPOINT": "desktop-app"}, "desktop-app"),
+        ({"LAKEBRIDGE_ENTRYPOINT": "databricks-app"}, "databricks-app"),
+        ({"LAKEBRIDGE_ENTRYPOINT": "  CLI  "}, "cli"),
+        ({"LAKEBRIDGE_ENTRYPOINT": "Desktop-App"}, "desktop-app"),
+        ({"LAKEBRIDGE_ENTRYPOINT": "DATABRICKS-APP"}, "databricks-app"),
+        ({}, "cli"),
+        ({"LAKEBRIDGE_ENTRYPOINT": "invalid"}, "cli"),
+    ],
+)
+def test_get_entrypoint_from_env(env: dict[str, str], expected: str) -> None:
+    assert get_entrypoint_from_env(env) == expected
+
+
+def test_get_entrypoint_uses_os_environ_by_default(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setenv("LAKEBRIDGE_ENTRYPOINT", "desktop-app")
+    assert get_entrypoint_from_env() == "desktop-app"


gueniai (approved):
LGTM

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -105,16 +106,49 @@ def _connect(self) -> Engine:
         return create_engine(connection_string)
 
 
+class SynapseConnector(MSSQLConnector):
+    """
```
asnare: I'm not convinced we need a new class for this:

 - I don't think we need to support the alternatives for the username and password properties: that's just going to lead to confusion about the property that needs to be set.
 - With the exception of the endpoint-to-server resolution, every other property is identical to the underlying connector.

Maybe it would be better to create a factory method on `MSSQLConnector` to replace this and `get_sqlpool_reader()` where there's a lot of duplication/overlap/redundancy. In particular the logic for `server` is confusing.

Something like this:
```python
class MSSQLConnector(…):
    …

    @classmethod
    def create_sqlpool_connector(class, config: dict[str, Any]) -> "MSSQLConnector":
        if "server" not in config:
            # Figure out value of server property.
            config = {**config, server=...}
        return class(config)
```

(This can be registered as a factory for the `synapse` dialect.)

src/databricks/labs/lakebridge/resources/assessments/synapse/common/connector.py
```diff
@@ -8,16 +8,22 @@ def get_sqlpool_reader(
...
+    """
     config = {
+        "endpoint_key": endpoint_key,
         "driver": input_cred['driver'],
         "server": input_cred[endpoint_key],
         "database": db_name,
+        "sql_user": input_cred['sql_user'],
+        "sql_password": input_cred['sql_password'],
         "user": input_cred['sql_user'],
         "password": input_cred['sql_password'],
```
asnare: Why do we need to specify both `sql_user` and `user` here? (And likewise for `sql_password`/`password`?)

src/databricks/labs/lakebridge/resources/assessments/synapse/common/connector.py
```diff
@@ -8,16 +8,22 @@ def get_sqlpool_reader(
     endpoint_key: str = 'dedicated_sql_endpoint',
     auth_type: str = 'sql_authentication',
 ) -> DatabaseManager:
+    """
+    Create Synapse SQL pool reader.
+    """
     config = {
+        "endpoint_key": endpoint_key,
         "driver": input_cred['driver'],
         "server": input_cred[endpoint_key],
```
asnare: Here's an example of my confusion (mentioned elsewhere): once we specify `server` then I believe `endpoint_key` is ignored?

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1037,5 +1040,82 @@ def create_profiler_dashboard(
...
+
+    cred_manager = create_credential_manager(PRODUCT_NAME, EnvGetter())
+
+    try:
+        raw_config = cred_manager.get_credentials(source_tech)
+    except KeyError as e:
+        logger.error(f"Credential configuration error: {e}")
+        raise SystemExit(
+            f"Invalid credentials for {source_tech}. Please run `databricks labs lakebridge configure-database-profiler`. Exiting..."
+        ) from e
```
asnare: Hmm, if we can it's best to avoid raising `SystemExit`. (I understand it's in use in various places.)

Here, maybe something like this?

```suggestion
        logger.fatal(f"Invalid credentials for {source_tech}. Please run `databricks labs lakebridge configure-database-profiler`.")
        return
```

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1037,5 +1040,82 @@ def create_profiler_dashboard(
     ctx.dashboard_manager.create_profiler_summary_dashboard(source_tech, catalog_name, schema_name)
 
 
+def _transform_profiler_credentials(source_tech: str, raw_config: dict) -> dict:
```
asnare: Do those type hints spark joy? :)

This strikes me as a place where mistakes in configuration files are likely to trigger confusing errors.

src/databricks/labs/lakebridge/connections/synapse_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+        ...     'profiler': {'exclude_serverless_sql_pool': False},
+        ... }
+        >>> validate_synapse_pools(config)  # Tests both pools
+    """
+    workspace_config = raw_config.get("workspace", {})
+    jdbc_config = raw_config.get("jdbc", {})
+    profiler_config = raw_config.get("profiler", {})
+
+    auth_type = jdbc_config.get("auth_type", "sql_authentication")
+
```
asnare: I've seen this default in several places. Instead of having it repeated everywhere, we can do something like:

```python
base_config = {
    **workspace_config,
    "database": "master",
}
if (auth_type := jdbc_config.get("auth_type") is not None:
    base_config["auth_type"] = auth_type
```

src/databricks/labs/lakebridge/connections/synapse_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+    if test_serverless:
+        success, error_msg = _test_pool_connection("serverless", base_config, "serverless_sql_endpoint")
+        results["serverless"] = success
+        if error_msg:
+            error_messages["serverless"] = error_msg
+
+    # Check if any pools failed
+    if not all(results.values()):
+        failed_pools = [pool for pool, success in results.items() if not success]
```
asnare: I think this is clearer:
```suggestion
    failed_pools = [pool for pool, success in results.items() if not success]
    if failed_pools:
```

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1037,5 +1040,82 @@ def create_profiler_dashboard(
...
+    config = _transform_profiler_credentials(source_tech, raw_config)
+    db_manager = DatabaseManager(source_tech, config)
+
+    try:
+        # Handle synapse-specific validation
+        if source_tech == "synapse":
+            validate_synapse_pools(raw_config)
+            logger.info("Connection to the source system successful")
+            return
+        response = db_manager.check_connection()
```
asnare: This is a little confusing to me, because:

1. We only support synapse, so we always take that branch.
2. If we took that branch, did we need to transform the credentials and create `db_manager` above? They're not used, but I can't tell if there's a side-effect of some sort that we're expecting?

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1037,5 +1040,82 @@ def create_profiler_dashboard(
...
+        logger.debug(f"Connection response: {response}")
+        logger.info("Connection to the source system successful")
+    except ConnectionError as e:
+        logger.error(f"Failed to connect to the source system: {e}")
+        if "IM002" in str(e) or "ODBC driver not found" in str(e):
+            raise SystemExit("Missing ODBC driver, Please install pre-req. Exiting...") from e
+        raise SystemExit("Connection validation failed. Exiting...") from e
+    except Exception as e:
+        logger.error(f"Unexpected error during connection test: {e}")
+        raise SystemExit("Connection test failed. Exiting...") from e
```
asnare: See earlier comments re. `SystemExit` from CLI commands.

tests/integration/connections/test_synapse_connector.py
asnare, line 1: Let's get some type hints on the functions in here, to ensure the bodies are being linted.

labs.yml
```diff
@@ -121,3 +121,9 @@ commands:
     flags:
       - name: source-tech
         description: (Optional) The technology/platform of the sources to Profile
+
+  - name: test-profiler-connection
+    description: (Internal) Test the connection to the source database for profiling
+    flags:
+      - name: source-tech
+        description: The technology/platform of the source to test connection
```
asnare:
```suggestion
        description: Database platform/technology of source connection
```

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1037,5 +1040,82 @@ def create_profiler_dashboard(
...
+        return {
+            **workspace_config,
+            "database": "master",  # Use master database for connection test
+            "auth_type": jdbc_config.get("auth_type", "sql_authentication"),
+        }
+    return raw_config
+
+
+@lakebridge.command()
+def test_profiler_connection(w: WorkspaceClient, source_tech: str | None = None) -> None:
```
asnare: All arguments for CLI commands are provided by keyword:
```suggestion
def test_profiler_connection(*, w: WorkspaceClient, source_tech: str | None = None) -> None:
```

asnare: For entry points we need integration tests.

(This one has sufficient code present that to test it some things might need to be extracted to allow for unit testing.)

src/databricks/labs/lakebridge/connections/synapse_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        # Catch all exceptions to gracefully handle any connection failure (network, auth, config, etc.)
+        error_msg = f"Failed to connect to {pool_name} SQL pool: {e}"
+        logger.error(f"✗ {error_msg}")
+        return False, error_msg
+    finally:
+        # Clean up database engine resources
+        if db_manager and hasattr(db_manager, 'connector') and hasattr(db_manager.connector, 'engine'):
+            db_manager.connector.engine.dispose()
+            logger.debug(f"Disposed engine for {pool_name} SQL pool")
```
asnare: There's a design issue here, essentially related to separation of responsibilities: `DatabaseManager` (and/or `.check_connection()` should be responsible for cleaning up their own resources. Instead they've leaked and this `finally` block has to poke around the internals to (maybe!) clean up things, depending on the implementation.

What needs to happen here:

 - If initialising `DatabaseManager` allocates resources that need to be released, it needs to become a context manager and manage that internally.
 - Similarly, `.check_connection()` should do the same internally if it's allocating resources.

(None of these `hasattr()` calls should be necessary.)

src/databricks/labs/lakebridge/connections/database_manager.py
sundarshankar89: Thanks this definitely reduces handling in multiple places.

src/databricks/labs/lakebridge/resources/assessments/synapse/common/connector.py
sundarshankar89: This is a plumbing because synapse is not just a sql with direct one type of connections. it is a platform which has multiple ways to connect, but I see how this can be confusing, let me see how to simplify this. 


Note: Connection manager itself uses MSSQL patterns but synapse can have `dedicated`, `serverless` so the connection url keeps changing based on what we query.

src/databricks/labs/lakebridge/resources/assessments/synapse/common/connector.py
sundarshankar89: there might be a good approach to conslidate this, originally the connections was handled within the profiler extraction now we have moved it centrally so this file might not be needed.

sundarshankar89: and all the plumbing should go away.

tests/integration/connections/test_synapse_connector.py
sundarshankar89: Done

tests/integration/cli/test_profiler_connection_cli.py
```diff
@@ -0,0 +1,110 @@
...
+from typing import Any
+
+import pytest
+import yaml
+from databricks.sdk import WorkspaceClient
+
+from databricks.labs.lakebridge.cli import test_profiler_connection as check_connection
+
+
+def _create_credentials_file(base_config: dict[str, Any], tmp_path: Path, **modifications: Any) -> Path:
```
asnare: The set of modifications is small and simpler to deal with as arguments:
```suggestion
def _create_credentials_file(
    base_config: JsonObject,
    tmp_path: Path,
    *,
    exclude_serverless: bool | None = None,
    invalid_server: bool = False,
    invalid_driver: bool = False,
    missing_workspace: bool = False,
    use_same_serverless_endpoint: bool = False,
) -> Path:
```

tests/integration/conftest.py
```diff
@@ -1,4 +1,5 @@
 import logging
+from typing import Any
```
asnare: Let's not use `Any` but instead use `JsonObject` (from `blueprint`) instead of `dict[str,Any]`.

In tests this is easy to cope with: use `assert` to narrow the type of the value.

(In production code, it needs to be narrowed carefully.)

src/databricks/labs/lakebridge/connections/synapse_connection_helpers.py
```diff
@@ -0,0 +1,133 @@
...
+        "port": workspace_config.get('port', 1433),
+        "auth_type": auth_type,
+    }
+
+    return DatabaseManager(db_type="synapse", config=config)
+
+
+def _test_pool_connection(
+    pool_name: str,
+    workspace_config: dict,
```
asnare, line 41: What's this dictionary contain?

tests/integration/cli/test_profiler_connection_cli.py
```diff
@@ -0,0 +1,110 @@
+import copy
+from pathlib import Path
+from typing import Any
```
asnare: Use `JsonObject` instead of `dict[str, Any]`.

tests/integration/connections/test_synapse_connector.py
```diff
@@ -0,0 +1,67 @@
+from typing import Any
```
asnare: Use `JsonObject` instead of `dict[str, Any]`.

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -109,12 +110,23 @@ class DatabaseManager:
...
+    def __enter__(self):
+        """Support context manager protocol for resource management."""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Clean up connector resources when exiting context."""
+        if hasattr(self.connector, '__exit__'):
+            self.connector.__exit__(exc_type, exc_val, exc_tb)
+        return False
+
```
asnare: These need to be implemented.

It looks to me like this means:
1. `DatabaseConnector` should extend `contextlib.AbstractContextManager`.
2. `_BaseConnector()` should be updated.

The general design/contract for context managers is:

1. If resources are initialised during __init__(), then:

   a) A `.close()` or other method (e.g. `dispose()`) is needed to allow clean up for folk not using the `with` statement;
   b) `__enter__()` is a no-op;
   c) `__exit__()` invokes the method above in a).

   Files follow this pattern: `open()` initialises, and `.close()` is allowed but it's also compatible with the `with` statement.

2. Alternatively, resources can be initialised during `__enter__()`:

   a) Callers are forced to use the `with` statement.
   b) `__enter__()` initialises the resource;
   c) `__exit__()` releases it.

   Using the resource before `__enter__()` or after `__exit__()` should crash: the caller isn't allowed to do that.

We shouldn't need `hasattr` or any other form of introspection.

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1037,5 +1040,83 @@ def create_profiler_dashboard(
...
+    """Test connection to the source database with appropriate error handling."""
+    # Handle synapse-specific validation using dedicated helper
+    if source_tech == "synapse":
+        validate_synapse_pools(raw_config)
+        logger.info("Connection to the source system successful")
+        return
+
+    # For other source technologies, use DatabaseManager directly
+    db_manager = DatabaseManager(source_tech, raw_config)
+    response = db_manager.check_connection()
```
asnare: This needs to be:
```suggestion
    with DatabaseManager(source_tech, raw_config) as db_manager:
        response = db_manager.check_connection()
```

src/databricks/labs/lakebridge/connections/database_manager.py
sundarshankar89: Done, implemented them.

src/databricks/labs/lakebridge/cli.py
sundarshankar89: Done

tests/integration/cli/test_profiler_connection_cli.py
sundarshankar89: Done

src/databricks/labs/lakebridge/connections/synapse_connection_helpers.py
sundarshankar89: example of the config workspace_config: dict is actually there in line#80

src/databricks/labs/lakebridge/connections/synapse_connection_helpers.py
```diff
@@ -0,0 +1,133 @@
...
+        return False, error_msg
+
+
+def validate_synapse_pools(raw_config: dict) -> None:
+    """
+    Validate connections to enabled Synapse SQL pools based on profiler configuration.
+    Each connection is properly cleaned up after testing to prevent resource leaks.
+
+    Example:
+        >>> config = {
```
sundarshankar89, line 80: @asnare example of workspace_config

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -46,6 +51,13 @@ def __init__(self, config: dict[str, Any]):
     def _connect(self) -> Engine:
         raise NotImplementedError("Subclasses should implement this method")
 
+    def close(self) -> None:
+        self.engine.dispose()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
asnare: I think this is the signature:
```suggestion
    def __exit__(self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None) -> bool:
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -109,12 +122,22 @@ class DatabaseManager:
     def __init__(self, db_type: str, config: dict[str, Any]):
         self.connector = _create_connector(db_type, config)
 
+    def __enter__(self):
+        """Support context manager protocol for resource management."""
+        return self
```
asnare: I would leave this out…

1. This is the default implementation.
2. It's difficult to type-hint properly (because we don't have `Self` available until 3.11).

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -109,12 +122,22 @@ class DatabaseManager:
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
asnare: I think the signature here is:
```suggestion
    def __exit__(self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None) -> bool:
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -109,12 +122,22 @@ class DatabaseManager:
...
+        """Clean up connector resources when exiting context."""
+        self.connector.__exit__(exc_type, exc_val, exc_tb)
+        return False
```
asnare:
```suggestion
        return self.connector.__exit__(exc_type, exc_val, exc_tb)
```

tests/integration/cli/test_profiler_connection_cli.py
```diff
@@ -0,0 +1,104 @@
...
+        yaml.dump(credentials, f)
+
+    return cred_path
+
+
+def test_profiler_connection_synapse_success(
+    sandbox_synapse_cred_config: JsonObject,
+    tmp_path: Path,
+    ws: WorkspaceClient,
+    caplog,
```
asnare:
```suggestion
    caplog: pytest.LogCaptureFixture,
```
(I think?)

tests/integration/cli/test_profiler_connection_cli.py
```diff
@@ -0,0 +1,104 @@
...
+) -> None:
+    """Test successful connection to Synapse dedicated SQL pool."""
+    cred_path = _create_credentials_file(sandbox_synapse_cred_config, tmp_path, exclude_serverless=True)
+
+    check_connection(w=ws, source_tech="synapse", cred_file_path=str(cred_path))
```
asnare, line 59: Just a small thing, but this is why I insisted on the `*` in the CLI method: it forces our test to use keyword arguments, which means that if someone renames an argument it's clear it's a breaking change.

(The names of the arguments have to match those in `labs.yml`, so I want friction here.)

tests/integration/conftest.py
```diff
@@ -65,6 +67,73 @@ def sandbox_sqlserver(sandbox_sqlserver_config) -> DatabaseManager:
...
+                "exclude_monitoring_metrics": False,
+                "redact_sql_pools_sql_text": False,
+                "databases": sandbox_sqlserver_config["database"],
+            },
+        },
+    }
+
+
+@pytest.fixture()
+def sandbox_synapse(sandbox_synapse_config: dict[str, Any]) -> DatabaseManager:
```
asnare:
```suggestion
def sandbox_synapse(sandbox_synapse_config: JsonObject) -> DatabaseManager:
```

tests/integration/conftest.py
```diff
@@ -65,6 +67,73 @@ def sandbox_sqlserver(sandbox_sqlserver_config) -> DatabaseManager:
     return DatabaseManager("mssql", sandbox_sqlserver_config)
 
 
+@pytest.fixture()
+def sandbox_synapse_config(sandbox_sqlserver_config: dict[str, Any]) -> dict[str, Any]:
```
asnare:
```suggestion
def sandbox_synapse_config(sandbox_sqlserver_config: JsonObject) -> JsonObject:
```

asnare (commented):
This is much better… there's a few type-hinting things remaining but it's almost good to go.

PR #2187, opened 2025-12-11
feature/oracle_profiler

lolo115:
## Changes
Creating the oracle extractor.
Unit tests passed for:
- dialog to configure the extractor
- run scripts to gather data
- local database population with extracted data from the Oracle Database.
### What does this PR do?
Oracle profiler data extraction
The data analysis is still WIP.

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/docs/lakebridge/docs/assessment/profiler/oracle.mdx
+++ b/docs/lakebridge/docs/assessment/profiler/oracle.mdx
@@ -1,0 +1,203 @@
+---
+sidebar_position: 4
+title: Oracle Profiler Details
+---
+import Admonition from '@theme/Admonition';
+
+# Oracle Profiler Details
+
+- [Prerequisites](#prerequisites)
+- [Profiled Tables and Views](#profiled-tables-and-views)
+- [Configure Connection to Oracle](#configure-connection-to-oracle)
+- [Execute the Profiler](#execute-the-profiler)
+
+:::info
+The Oracle profiler targets **multitenant container databases (CDB)** and reads from `CDB_*`, `GV$*`, and AWR views.
+It must be run against `CDB$ROOT`, not an individual PDB.
+:::
+
+## Prerequisites
+
+### 1. Download
+
+No local Oracle client install is required.
+
+### 2. Network and Authentication
+
+:::warning Attention:
+The profiler connects with **username and password over TCP** only. The following are **not supported**:
+
+- TCPS / TLS-encrypted listener connections
+- Oracle Wallet (mTLS, externally identified users)
+- Kerberos / OS authentication
+- Multi-factor authentication
+
+The Oracle listener must be reachable from the host running `databricks labs lakebridge` on the configured TNS port.
+:::
+
+### 3. Required Database Permissions
+
+The Oracle user configured for the profiler must be able to read the data dictionary and AWR views listed in
+[Profiled Tables and Views](#profiled-tables-and-views).
+
+Connect to `CDB$ROOT` as `SYS` (or a DBA-privileged user) and run:
+
+```sql
+-- Create a common profiler user (skip if reusing an existing account)
+CREATE USER c##lakebridge_profiler IDENTIFIED BY <password> CONTAINER=ALL;
+GRANT CREATE SESSION TO c##lakebridge_profiler CONTAINER=ALL;
+GRANT SET CONTAINER TO c##lakebridge_profiler CONTAINER=ALL;
+
+-- Catalog and AWR access
+GRANT SELECT_CATALOG_ROLE TO c##lakebridge_profiler CONTAINER=ALL;
+```
+
+:::warning Diagnostic Pack license
+The performance queries read from AWR (`CDB_HIST_*`), which requires an Oracle **Diagnostic Pack** license. AWR
+views return data on any Enterprise Edition database whether or not the pack is licensed — querying them without
+a license is an audit-time violation, not a runtime error.
+
+By running the profiler against an Oracle source you are asserting that your database is appropriately licensed.
+:::
+
+## Profiled Tables and Views
+
+The Oracle profiler executes queries against the following Oracle data dictionary and AWR views. The results are
+organized into two extraction steps: **configuration metadata** and **performance metrics**.
+
+### Configuration Metadata
+
+<table>
+  <thead>
+    <tr>
+      <th>Query</th>
+      <th>Source View(s)</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Containers</td>
+      <td><code>gv$containers</code></td>
+      <td>CDB and PDB inventory across all RAC instances.</td>
+    </tr>
+    <tr>
+      <td>DB Features</td>
+      <td><code>product_component_version</code>, <code>gv$instance</code>, <code>gv$pdbs</code>, <code>gv$osstat</code></td>
+      <td>Database version, instance count, PDB list, and CPU/core/socket counts.</td>
+    </tr>
+    <tr>
+      <td>Instance</td>
+      <td><code>gv$instance</code></td>
+      <td>Instance id, name, version, and database type.</td>
+    </tr>
+    <tr>
+      <td>Memory Evolution</td>
+      <td><code>cdb_hist_parameter</code>, <code>cdb_hist_snapshot</code></td>
+      <td>Historical SGA/PGA/buffer-cache parameter values across AWR snapshots.</td>
+    </tr>
+    <tr>
+      <td>PDB Objects</td>
+      <td><code>cdb_objects</code>, <code>cdb_users</code></td>
+      <td>Object counts per PDB, owner, and object type (Oracle-maintained schemas excluded).</td>
+    </tr>
+    <tr>
+      <td>PDB Partitions</td>
+      <td><code>cdb_tables</code>, <code>cdb_indexes</code>, <code>cdb_lobs</code></td>
+      <td>Partitioned vs. non-partitioned counts for tables, indexes, and LOBs per PDB and owner.</td>
+    </tr>
+    <tr>
+      <td>Storage</td>
+      <td><code>cdb_data_files</code>, <code>cdb_free_space</code>, <code>cdb_tablespaces</code></td>
+      <td>Allocated, free, and max sizes per container, grouped into <code>SYSTEM</code>, <code>UNDO</code>, and <code>USER_DATA</code>.</td>
+    </tr>
+  </tbody>
+</table>
+
+### Performance Metrics
+
+<table>
+  <thead>
+    <tr>
+      <th>Query</th>
+      <th>Source View(s)</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>CPU &amp; Waits</td>
+      <td><code>cdb_hist_active_sess_history</code></td>
+      <td>Hourly wait-event and CPU-time breakdown per PDB and instance from ASH samples.</td>
+    </tr>
+    <tr>
+      <td>Foreground Session Evolution</td>
+      <td><code>cdb_hist_active_sess_history</code>, <code>cdb_users</code></td>
+      <td>Per-minute distinct foreground session counts by PDB, instance, and end-user.</td>
+    </tr>
+    <tr>
+      <td>CPU Heatmap</td>
+      <td><code>cdb_hist_active_sess_history</code>, <code>gv$osstat</code></td>
+      <td>Hourly Average Active Sessions normalized by CPU core count.</td>
+    </tr>
+    <tr>
+      <td>SQL Text</td>
+      <td><code>cdb_hist_sqlstat</code>, <code>cdb_hist_sqltext</code>, <code>audit_actions</code></td>
+      <td>SQL statements from AWR, classified into workload categories (BI/QUERY, ETL, DML, DDL, PL/SQL) with execution counts and total elapsed time per schema and PDB.</td>
+    </tr>
+  </tbody>
+</table>
+
+## Configure Connection to Oracle
+
+Run the following command to configure the profiler connection to your Oracle CDB:
+
+```console
+databricks labs lakebridge configure-database-profiler
+
+Please select the source system you want to configure
+[0] synapse
+[1] mssql
+[2] oracle
+Enter a number between 0 and 2: 2
+
+(local | env)
+local means values are read as plain text
+env means values are read from environment variables, and fall back to plain text if not variable is not found
+
+Enter secret vault type (local | env)
+[0] env
+[1] local
+Enter a number between 0 and 1: 1
+Enter the host details (Server name, IP address, SCAN Name): oracle-host.example.com
+Enter the host port number (default: 1521):
+Enter the service name (default: orcl): ORCLCDB
+Enter user with privileges:
+Enter user password:
+```
+
+### Configuration Parameters
+
+| Parameter | Description | Default |
+|:----------|:------------|:--------|
+| **Secret vault type** | `local` for plain text values, `env` to read from environment variables | — |
+| **Host** | Oracle server hostname, IP, or RAC SCAN name | — |
+| **Port** | Port the Oracle listener is bound to | `1521` |
+| **Service name** | Service name registered in the Oracle listener (the CDB service) | `orcl` |
+| **User** | Database user with the grants from [Required Database Permissions](#3-required-database-permissions). `SYSTEM` works out of the box but is over-privileged; the dedicated user above is recommended. | — |
+| **Password** | Password for the database user (hidden input) | — |
+
+## Execute the Profiler
+
+Once configured, run the profiler to extract metadata and performance metrics from your Oracle CDB:
+
+```bash
+databricks labs lakebridge execute-database-profiler --source-tech oracle
+```
+
+The profiler will:
+1. Connect to your Oracle CDB using the configured credentials
+2. Execute the configuration and performance metric extraction queries
+3. Store the results in a local DuckDB extract file
+
+[Back to Configure Profiler](../#configure-profiler)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dependencies = [
   "SQLAlchemy~=2.0.40",
   "pygls~=2.1.1",
   "duckdb>=1.2.2,<1.5.0",
+  "oracledb~=3.4.2",
   "databricks-switch-plugin~=0.1.7",  # Temporary, until Switch is migrated to be a transpiler (LSP) plugin.
   "requests>=2.28.1,<3",  # Matches databricks-sdk (and 'types-requests' below), to avoid conflicts.
   "pandas~=2.3.1", # Required for new configure assessment
--- a/src/databricks/labs/lakebridge/assessments/configure_assessment.py
+++ b/src/databricks/labs/lakebridge/assessments/configure_assessment.py
@@ -73,6 +73,35 @@ class AssessmentConfigurator(ABC):
         logger.info(f"{source.capitalize()} Assessment Configuration Completed")
 
 
+class ConfigureOracleAssessment(AssessmentConfigurator):
+    """Oracle specific assessment configuration."""
+
+    def _configure_credentials(self) -> str:
+        cred_file = self._credential_file
+        source = self._source_name
+
+        logger.info(
+            "\n(local | env) \nlocal means values are read as plain text \nenv means values are read "
+            "from environment variables fall back to plain text if not variable is not found\n",
+        )
+        secret_vault_type = str(self.prompts.choice("Enter secret vault type (local | env)", ["local", "env"])).lower()
+
+        credential = {
+            "secret_vault_type": secret_vault_type,
+            source: {
+                "host": self.prompts.question("Enter the host details (Server name, IP address, SCAN Name)"),
+                "port": int(self.prompts.question("Enter the host port number", default=str(1521), valid_number=True)),
+                "service_name": self.prompts.question("Enter the service name", default="orcl"),
+                "user": self.prompts.question("Enter user with privileges"),
+                "password": self.prompts.password("Enter user password"),
+            },
+        }
+
+        _save_to_disk(credential, cred_file)
+        logger.info(f"Credential template created for {source}.")
+        return source
+
+
 class ConfigureSqlServerAssessment(AssessmentConfigurator):
     """SQL Server-family assessment configuration.
 
@@ -199,6 +228,7 @@ def create_assessment_configurator(
         "mssql": ConfigureSqlServerAssessment,
         "synapse": ConfigureSynapseAssessment,
         "legacy_synapse": ConfigureSqlServerAssessment,
+        "oracle": ConfigureOracleAssessment,
     }
 
     if source_system not in configurators:
--- a/src/databricks/labs/lakebridge/connections/database_manager.py
+++ b/src/databricks/labs/lakebridge/connections/database_manager.py
@@ -119,6 +119,25 @@ class MSSQLConnector(_BaseConnector):
         return create_engine(connection_string)
 
 
+class OracleConnector(_BaseConnector):
+    def _connect(self) -> Engine:
+        connection_string = URL.create(
+            drivername="oracle+oracledb",
+            username=str(self.config['user']),
+            password=str(self.config['password']),
+            host=str(self.config['host']),
+            port=int(str(self.config.get('port', 1521))),
+            database=str(self.config.get('service_name')),
+        )
+
+        return create_engine(connection_string)
+
+    def health_check(self) -> bool:
+        query = "SELECT 101 AS test_column FROM dual"
+        result = self.fetch(query)
+        return result.rows[0][0] == 101
+
+
 class RedshiftConnector(DatabaseConnector):
     def __init__(self, config: JsonObject):
         self.config = config
@@ -187,6 +206,7 @@ def _create_connector(db_type: str, config: JsonObject) -> DatabaseConnector:
         "synapse": MSSQLConnector,  # Synapse uses MSSQL protocol
         "legacy_synapse": MSSQLConnector,
         "redshift": RedshiftConnector,
+        "oracle": OracleConnector,
     }
 
     connector_class = connectors.get(db_type.lower())
--- a/src/databricks/labs/lakebridge/resources/assessments/oracle/config_containers.sql
+++ b/src/databricks/labs/lakebridge/resources/assessments/oracle/config_containers.sql
@@ -1,0 +1,1 @@
+select listagg(inst_id, ',') within group (order by name) as inst_ids,name,open_mode,pdb_count from gv$containers group by name,open_mode,pdb_count
--- a/src/databricks/labs/lakebridge/resources/assessments/oracle/config_db_features.sql
+++ b/src/databricks/labs/lakebridge/resources/assessments/oracle/config_db_features.sql
@@ -1,0 +1,27 @@
+with t as (select version from product_component_version where product like 'Oracle Database%'),
+	  inst_cnt as (select count(*) as cnt from gv$instance),
+	  pdb_cnt as (select count(distinct con_id ) as cnt from gv$pdbs where name!='PDB$SEED'),
+	  pdb_namelist as (select listagg(name,',') within group (order by name) as pdbnames from (select distinct name from gv$pdbs where name !='PDB$SEED')),
+	  cpu_cores_global as (select 'CLUSTER' as scope, null , stat_name,'CPU GLOBAL (Cluster): '||stat_name as detailed_stat_name,sum(to_number(value)) as value
+			from gv$osstat
+			where stat_name in ('NUM_CPUS','NUM_CPU_CORES','NUM_CPU_SOCKETS')
+			group by stat_name),
+	  cpu_cores_details as (select 'INSTANCE' as scope,inst_id, stat_name,'CPU per Instance Id: '||inst_id||' - '||stat_name as detailed_stat_name, to_number(value) as value
+				from gv$osstat
+				where stat_name in ('NUM_CPUS','NUM_CPU_CORES','NUM_CPU_SOCKETS')
+				order by inst_id,stat_name)
+select * from
+(
+select null as scope, null as inst_id, null as stat_name, 'VERSION' as name,to_char(t.version) as value from t
+union
+select null as scope, null as inst_id, null as stat_name, 'INSTANCE COUNT', to_char(inst_cnt.cnt) from inst_cnt
+union
+select null as scope, null as inst_id, null as stat_name, 'PDB COUNT',to_char(pdb_cnt.cnt) from pdb_cnt
+union
+select null as scope, null as inst_id, null as stat_name, 'PDB LIST', to_char(pdb_namelist.pdbnames) from pdb_namelist
+union
+select scope, null as inst_id, stat_name,detailed_stat_name, to_char(value) from cpu_cores_global
+union
+select scope, inst_id, stat_name, detailed_stat_name, to_char(value) from cpu_cores_details
+)
+order by name
--- a/src/databricks/labs/lakebridge/resources/assessments/oracle/config_instance.sql
+++ b/src/databricks/labs/lakebridge/resources/assessments/oracle/config_instance.sql
@@ -1,0 +1,1 @@
+select inst_id,instance_name,version,database_type from gv$instance
--- a/src/databricks/labs/lakebridge/resources/assessments/oracle/config_pdb_objects.sql
+++ b/src/databricks/labs/lakebridge/resources/assessments/oracle/config_pdb_objects.sql
@@ -1,0 +1,7 @@
+SELECT cont.NAME as PDB_NAME,OWNER,OBJECT_TYPE,COUNT(*) as CNT
+FROM CDB_OBJECTS o,
+     (select distinct con_id as con_id, name from gv$containers) cont
+WHERE o.CON_ID=cont.con_id
+AND OWNER in (select username from cdb_users where oracle_maintained='N' and cont.con_id=con_id)
+GROUP BY cont.name,OWNER,OBJECT_TYPE
+ORDER BY 1,2
--- a/src/databricks/labs/lakebridge/resources/assessments/oracle/config_pdb_partitions.sql
+++ b/src/databricks/labs/lakebridge/resources/assessments/oracle/config_pdb_partitions.sql
@@ -1,0 +1,21 @@
+-- Has to be executed on CDB
+SELECT cont.NAME as PDB_NAME,OWNER,OBJECT_TYPE,CNT
+from
+(
+    select owner,con_id,'TABLE (NON PARTITIONED)' as OBJECT_TYPE,count(*) as cnt from cdb_tables where partitioned='NO' group by owner,con_id
+    union
+    select owner,con_id,'TABLE (PARTITIONED)',count(*) as cnt from cdb_tables where partitioned='YES' group by owner,con_id
+    union
+    select owner,con_id,'INDEX (NON PARTITIONED)',count(*) as cnt from cdb_indexes where partitioned='NO' group by owner,con_id
+    union
+    select owner,con_id,'INDEX (PARTITIONED)',count(*) as cnt from cdb_indexes where partitioned='YES' group by owner,con_id
+    union
+    select owner,con_id,'LOBS (NON PARTITIONED)',count(*) as cnt from cdb_lobs where partitioned='NO' group by owner,con_id
+    union
+    select owner,con_id,'LOBS (PARTITIONED)',count(*) as cnt from cdb_lobs where partitioned='YES' group by owner,con_id
+) u,
+(select distinct con_id as con_id, name from gv$containers) cont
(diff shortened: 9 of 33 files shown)


src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -152,6 +169,8 @@ def fetch(self, query: str) -> FetchResult:
 
     def check_connection(self) -> bool:
         query = "SELECT 101 AS test_column"
+        if self._db_type.lower() == "oracle":
```
m-abulazm: move this method to the connectors

tests/integration/assessments/test_profiler.py
```diff
@@ -8,6 +8,37 @@
 from databricks.labs.lakebridge.assessments.profiler import Profiler
 
 
+# Tests for Oracle profiler section
+def test_Oracle_as_supported_source_technologies() -> None:
```
m-abulazm: missing tests

src/databricks/labs/lakebridge/resources/assessments/validation/oracle_extract_schema.yml
```diff
@@ -0,0 +1,138 @@
+source_tech: oracle
```
m-abulazm: investigate usage

tests/test_blueprint_oracle.py
```diff
@@ -0,0 +1,11 @@
...
+from databricks.labs.lakebridge.assessments.profiler import Profiler
+
+pipeline_config_file = "../src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml"
+config = PipelineClass.load_config_from_yaml(pipeline_config_file)
+print(config)
+
+profiler = Profiler("oracle", None)
+# supported_platforms = profiler.supported_platforms()
+# print(supported_platforms)
+profiler.profile(pipeline_config=config)
```
m-abulazm: clean

tests/resources/assessments/pipeline_oracle_config_main.yml
```diff
@@ -0,0 +1,11 @@
+name: oracle_assessment
```
m-abulazm: investigate

m-abulazm (changes requested):
missing docs

m-abulazm (approved):

uv.lock
dgomez04, line 1: Can we check with @asnare if there are changes required here? I think we shouldn't be updating `uv.lock` based on some work on the CI.

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
   "SQLAlchemy~=2.0.40",
   "pygls~=2.1.1",
   "duckdb>=1.2.2,<1.5.0",
+   "oracledb==3.4.0",
```
dgomez04: The **exact pin** is inconsistent with the rest of the file. Every other dependency uses `~=` or `>=,<` ranges.

.gitignore
dgomez04, line 1: Are these needed here? Weren't they being ignored beforehand?

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -72,6 +73,43 @@ def run(self):
...
+        secret_vault_name = None
+
+        logger.info("Please refer to the documentation to understand the difference between local and env.")
+
+        credential = {
+            "secret_vault_type": secret_vault_type,
+            "secret_vault_name": secret_vault_name,
+            source: {
+                "host": self.prompts.question("Enter the host details (Server name, IP address, SCAN Name)"),
+                "tnsPort": int(
```
dgomez04: `tnsPort` and `tnsService` uses camelCase where the existing configurators use snake_case. This is incosnistent and we should standardize.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -72,6 +73,43 @@ def run(self):
...
+                    self.prompts.question("Enter the TNS Listener port number", default=str(1521), valid_number=True)
+                ),
+                "tnsService": self.prompts.question(
+                    "Enter the TNS service name as registered in the Oracle listener", default="orcl"
+                ),
+                "user": self.prompts.question("Enter user name with system privileges", default="SYSTEM"),
```
dgomez04: The documentation tells users to create a common profiler user `c##lakebridge_profiler` but the default prompt undermines this and defaults to `SYSTEM` - a priviledged Oracle account. Default should either empty or `c##lakebridge_profiler`.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -72,6 +73,43 @@ def run(self):
...
+        source = self._source_name
+
+        logger.info(
+            "\n(local | env) \nlocal means values are read as plain text \nenv means values are read "
+            "from environment variables fall back to plain text if not variable is not found\n",
+        )
+        secret_vault_type = str(self.prompts.choice("Enter secret vault type (local | env)", ["local", "env"])).lower()
+        secret_vault_name = None
+
+        logger.info("Please refer to the documentation to understand the difference between local and env.")
```
dgomez04: This is redundant, lines 83-86 explain this already.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -72,6 +73,43 @@ def run(self):
...
+
+        credential = {
+            "secret_vault_type": secret_vault_type,
+            "secret_vault_name": secret_vault_name,
+            source: {
```
dgomez04, line 91: The `credential` dict here jumps straight from `secret_vault_*` keys into Oracle connection parameters with no `auth_type` field. All other configurations set `auth_type` at the top of the `source` block. This matters because Oracle's enterprise authentication surface is meaningfully richer than SQL Server's, and several of these might be common asks for the actual customers we're targeting: 
- **Oracle Wallet** (TNS_ADMIN + sqlnet.ora + cwallet.sso) — stored encrypted credentials, very common in regulated environments where security teams ban plaintext DB passwords in YAML files.
- **External / Kerberos / OS authentication** — for SSO and AD integration.
- **Token-based auth for Oracle Cloud** (OCI IAM tokens) and Autonomous Database.
- **mTLS / TCPS with client certs** — required for Autonomous Database "TLS-only" mode.

No need to actually implement these types today - just reserve the field and document the one currently-supportede value, like `MSSQLConnector`does.

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -79,7 +79,7 @@ def _create_connector(db_type: str, config: JsonObject) -> DatabaseConnector:
         "snowflake": SnowflakeConnector,
         "mssql": MSSQLConnector,
         "tsql": MSSQLConnector,
-        "synapse": MSSQLConnector,  # Synapse uses MSSQL protocol
```
dgomez04: Why is `synapse` deleted here? Is `snowflake` already implemented?

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -125,9 +125,26 @@ def _connect(self) -> Engine:
         return create_engine(connection_string)
 
 
+class OracleConnector(_BaseConnector):
```
dgomez04, line 122: - No support for other authentication methods, such as TCPS (TLS), Oracle Wallet, or RAC easy-connect strings beyond host:port/service.
- Does not pass `mode=oracledb.SYSDBA` etc., which is irrelevant for SELECT-only profilers but worth noting.
- No `loginTimeout` equivalent (the MSSQL connector at least sets `"loginTimeout": "30"`).

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -152,6 +169,8 @@ def fetch(self, query: str) -> FetchResult:
 
     def check_connection(self) -> bool:
```
dgomez04, line 245: `DatabaseManager` is supposed to be the engine-agnostic layer that dispatches to a connector via `_create_connector`; once that dispatch happens, no code in `DatabaseManager` should be re-deriving the engine via string checks on `db_type`.

**Concrete suggestion:** have each connector own its own smoke-test query as a class attribute, and let `DatabaseManager` just read it.

```python
class _BaseConnector(DatabaseConnector):
    # Subclasses override this with a dialect-appropriate one-row SELECT.
    smoke_test_query: str = "SELECT 101 AS test_column"
    ...

class OracleConnector(_BaseConnector):
    smoke_test_query = "SELECT 101 AS test_column FROM dual"
    ...

class DatabaseManager:
    def __init__(self, db_type: str, config: JsonObject):
        # self._db_type = db_type   ← can be deleted

    def check_connection(self) -> bool:
        result = self.fetch(self.connector.smoke_test_query)
        if result is None:
            return False
        return result.rows[0][0] == 101
```

src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml
```diff
@@ -0,0 +1,136 @@
...
+version: "1.0"
+extract_folder: "/tmp/data/oracle_assessment"
+steps:
+  - name: config_containers
+    type: ddl
+    extract_source: src/databricks/labs/lakebridge/resources/assessments/oracle/config_containers_ddl.sql
+    mode: overwrite
+    frequency: once
+    flag: active
+  - name: config_containers
```
dgomez04, line 11: Every logical extractor here is split into two YAML steps that share the same name: — one with `type: ddl`, one with `type: sql`. That works today because `pipeline.py` happens to dispatch on `step.type` first and uses `step.name` as the destination DuckDB table name in both branches.

But the duplicate names cause the following problems: 
- `PipelineClass.execute()` aggregates `StepExecutionResult` objects keyed by `step.name`, then logs `failed_steps = [r for r in execution_results if r.status == StepExecutionStatus.ERROR]`. If the SQL half of `config_containers` fails, the user sees Pipeline execution failed due to errors in `steps: config_containers` with no way to tell whether the DDL or the SQL step failed. This makes debugging a headache.
- The DDL step does if not `self._table_exists(conn, step.name): conn.execute(ddl)` — i.e., it implicitly assumes the `CREATE TABLE` in the DDL file uses a table named exactly `step.name`. There's nothing enforcing that; if a future edit changes the name in one place but not the other, the pipeline silently no-ops the DDL or fails the insert.
- Anyone refactoring `pipeline.py` to dedupe steps by name (perfectly reasonable assumption) breaks this YAML silently.

I think we need to discuss how to handle it, one alternative could be giving DDL steps a different suffix:
```yaml
- name: config_containers_schema
  type: ddl
  extract_source: .../config_containers_ddl.sql
- name: config_containers
  type: sql
  extract_source: .../config_containers.sql
```

The `_execute_ddl_step` check would still need to know the target table name (which is currently inferred from `step.name`); the cleanest version of that is to either parse the table name out of the CREATE statement, or add a `target_table: field` to Step so name (the identifier in logs/errors) and target (the DuckDB table the step writes to) are decoupled. The second is a small change in `profiler_config.py` and `pipeline.py` and is worth doing now while the file count is small.

**Longer-term thought:** it might be cleaner to fold DDL into the SQL step entirely — i.e., a single step with `setup_ddl: ...` and `extract_sql: ... fields`, since DDL+SQL are conceptually one extractor — but that's a bigger change and probably belongs in a follow-up. **We should probably file an issue for this.**

src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml
```diff
@@ -0,0 +1,136 @@
+name: oracle_assessment
+version: "1.0"
+extract_folder: "/tmp/data/oracle_assessment"
```
dgomez04, line 3: Three things bundled here:
- `/tmp/...` isn't a real Windows path — `Path("/tmp/...").expanduser()` in `pipeline.py:42` resolves it to `<current drive>:\tmp\data\oracle_assessment`, which "works" but isn't where any Windows user expects an extract to land. The Synapse profiler already solves this with `~/.databricks/labs/lakebridge_profilers/synapse_assessment`, which `expanduser()` resolves correctly on every OS.
- Two of the three existing profilers (`mssql`, this `oracle` one) write to `/tmp/data/...`, while `synapse` writes to `~/.databricks/labs/lakebridge_profilers/...`. A user running multiple profilers ends up needing two different `--extract-file` paths in their `create-profiler-dashboard` command for no good reason.
- macOS purges files in `/tmp` periodically; AWR-based extracts that take 30+ minutes to produce shouldn't live there.

Could we standardize on the Synapse-style path here, e.g.
```bash
extract_folder: "~/.databricks/labs/lakebridge_profilers/oracle_assessment"
```
_If that's done, the integration test (tests/integration/assessments/test_profiler.py) needs its hardcoded Path("/tmp/data/oracle_assessment/profiler_extract.db") assertion updated to match._

src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml
dgomez04, line 1: Two pairs of files exist under `oracle/` that are not referenced anywhere in this `pipeline_config.yml`:
- `test_nulloutput.sql` + `test_nulloutput_ddl.sql` — a `SELECT 'value' as X from dual where 1=2` query and a matching `CREATE TABLE`. Looks like author-local debugging that wasn't cleaned up before opening the pull request. Delete.
- `perf_heatmap_raw.sql` — this one is more concerning. It's a SQL*Plus driver script:
```sql
declare
  cursor c is select inst_id from gv$instance order by 1;
begin
  for i in c loop
    dbms_output.put_line('@./scripts/main/perf_addon/perf_hm_raw.sql '||i.inst_id);
    ...
  end loop;
end;
/
spool off
start ./scripts/main/perf_addon/script.sql
exit
```

A few problems with this file:
- The `declare ... begin ... end`; `/`, `spool off`, `start`, `exit` directives are SQL*Plus-specific. None of them are valid through SQLAlchemy / oracledb — they'd just error.
- It references `./scripts/main/perf_addon/perf_hm_raw.sql` and `./scripts/main/perf_addon/script.sql`, neither of which exists in the pull request.
- `dbms_output.put_line` would silently do nothing without `dbms_output.enable()` and an output handler attached on the client side.
- Even if all of the above were fixed, the file isn't referenced in `pipeline_config.yml`, so it would never be invoked.

**Please review these.**

src/databricks/labs/lakebridge/resources/assessments/oracle/config_db_features.sql
dgomez04, line 1: Three nits in one file:
- **Stray commented-out alternative** on line 6 (-- `select listagg(distinct name,',')...`) sitting inside a `WITH` clause. Either delete it or, if it's the "right" version, swap it in. Leaving it there as an out-of-tree alternative will confuse future readers.
- **Implicit-join syntax** in the inner subqueries (`FROM gv$osstat WHERE stat_name in (...)` etc.) and ANSI-style UNION between heterogeneous shapes is hard to read. The query produces six column shapes via null as ... placeholders to make them union-compatible — that's clever but signals the data is really three different result sets glued together. Consider returning them as separate steps (`config_db_version`, `config_db_pdbs`, `config_db_cpu`) — it's a nicer separation, and downstream dashboarding gets cleaner table shapes.
- **Implicit join in the outer query** could become explicit `JOIN` syntax. Same correctness, much easier to grep when something breaks.

src/databricks/labs/lakebridge/resources/assessments/oracle/config_storage.sql
dgomez04, line 1: There are two commented-out branches that would have classified TEMP tablespaces (lines 5 and ~27 in the case and the matching group by case). Either:
- **Implement it:** uncomment the branch in both places, add 'TEMP' to the doc's tablespace-type list, and confirm `cdb_data_files` actually contains TEMP rows on your target Oracle versions (TEMP files live in `dba_temp_files` / `cdb_temp_files`, not `cdb_data_files`, so the commented-out logic might not even work as written).
- **Delete it:** drop both `-- when ...` lines and document that TEMP tablespaces are out of scope for the storage extract.

src/databricks/labs/lakebridge/resources/assessments/oracle/config_pdb_objects.sql
dgomez04, line 1: No change needed in this file — the `oracle_maintained='N'` filter against cdb_users is the right way to exclude Oracle-internal schemas on 12.2+. Calling it out only because `perf_fgd_session_evolution.sql` does the same job by hardcoding a list of ~30 schema names; the two should be consistent. See my comment on that file.

src/databricks/labs/lakebridge/resources/assessments/oracle/config_pdb_partitions.sql
dgomez04, line 1: No change needed in this file — the `oracle_maintained='N'` filter against `cdb_users` is the right way to exclude Oracle-internal schemas on 12.2+. Calling it out only because `perf_fgd_session_evolution.sql` does the same job by hardcoding a list of ~30 schema names; the two should be consistent. See my comment on that file.

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_cpu_waits.sql
dgomez04, line 1: The inner subquery `TO_CHARs` sample_time to `'YYYY-MM-DD HH24'` (so it's a string truncated to the hour), and the outer query then `TO_DATEs` that string back to a timestamp. Two issues:
1. **Information loss.** `sample_time` from `cdb_hist_active_sess_history` is a high-precision timestamp; round-tripping through an hour-grained string drops minute/second/fractional precision unrecoverably. If you want hourly aggregates, that's fine — but please use `TRUNC(sample_time, 'HH24')` directly, which keeps it as a `DATE` and is also faster (no string conversion).
2. **DDL/data mismatch.** `perf_cpu_waits_ddl.sql` declares `mtime TIMESTAMP`, which suggests sub-second fidelity. After the round-trip, `mtime` only carries hour-grained values. Either change the DDL to `mtime DATE` (more honest) or change the SQL to keep the `original sample_time` precision.
3. **NLS dependency.** `TO_DATE(string, format)` is sensitive to `NLS_DATE_LANGUAGE / NLS_DATE_FORMAT` — one customer's session settings can break this. Direct `TRUNC(sample_time, 'HH24')` avoids that entirely.

Suggested rewrite:
```sql
SELECT
    instance_number,
    con_id,
    CON_DBID,
    TRUNC(sample_time, 'HH24') AS mtime,
    NVL(event, 'ON CPU') AS event,
    NVL(wait_class, 'ON CPU') AS wait_class,
    COUNT(*) * 10 AS total_wait_time
FROM cdb_hist_active_sess_history
GROUP BY instance_number, con_id, CON_DBID, TRUNC(sample_time, 'HH24'), event, wait_class
```

_Then keep `mtime DATE` (or `TIMESTAMP` if you intentionally want sub-hour granularity, in which case drop the `TRUNC`)._

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_fgd_session_evolution.sql
dgomez04, line 1: 1. **Copy-paste bug.** The `WHERE username NOT IN (...)` clause lists the same ~30 Oracle system schemas twice within the same parenthesised list — `'SYS','SYSTEM','XS$NULL', ..., 'SYSDG'`then immediately again. Functionally harmless (Oracle dedupes IN-list literals), but a clear signal the file wasn't proofread, and it makes future edits dangerous: if someone adds a new schema to one half but not the other, future maintainers will see asymmetry and waste time figuring out whether it's intentional.
2. **The whole approach should match `cdb_objects` / `cdb_users` filtering.** The earlier `config_pdb_objects.sql` and `config_pdb_partitions.sql` queries use the much cleaner `WHERE oracle_maintained='N'` against `cdb_users`. That handles every Oracle-shipped schema correctly across versions and is one extra column instead of a 30-line hardcoded literal.

Suggested rewrite:
```sql
... ,
(
  SELECT user_id, username
  FROM cdb_users
  WHERE oracle_maintained = 'N'
) u
```
_Same shape, ~28 fewer lines, robust to Oracle adding new internal schemas in future versions._

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_heatmap.sql
dgomez04, line 1: The active query ends at ~line 38; the remaining ~87 lines are commented out. Two distinct dead blocks:
1. **SQL*Plus column-formatting directives** (`-- col "00-01_ " for 90.99...`) — only useful when running against `sqlplus` interactively, irrelevant under `SQLAlchemy`/`oracledb`.
2. A complete alternative implementation using 1WITH t AS (...) SELECT ... FROM t pivot(SUM(value) ...)` — this is the "wide-format" version of the heatmap.

_Both should be deleted. If the `PIVOT` version is the intended endgame (looks like it's the better shape for downstream dashboarding), please file a follow-up issue capturing the design decision and delete the commented block here. Shipping it commented-out makes it look like a feature flag and will create confusion the moment someone goes to "tweak the heatmap query."_

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_heatmap_raw.sql
dgomez04, line 1: This file is a **SQL*Plus** driver script and won't run via SQLAlchemy/oracledb:
- `declare ... begin ... end; /`, `spool off`, `start ./scripts/...`, and exit are SQL*Plus client commands, not server-side SQL. The pipeline runner would error on the very first statement.
- It references `./scripts/main/perf_addon/perf_hm_raw.sql` and `./scripts/main/perf_addon/script.sql`, neither of which exists in the repo.
- `dbms_output.put_line` would silently produce nothing without `dbms_output.enable()` and a client-side output handler.
- It isn't referenced in `pipeline_config.yml`, so even if it worked, nothing would invoke it.

_Looks like leftover from a prior manual SQL*Plus assessment workflow that got committed by mistake. Please delete the file. If "raw heatmap" was intended to be a real feature, please describe it's intent on a follow-up issue. We shouldn't ship it on this state._

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_sqltext.sql
dgomez04, line 1: **1. 2KB SQL sampling.** Every `dbms_lob.substr(t.SQL_TEXT, 2000)` call truncates the SQL text to the first 2KB before doing the pattern match. That's reasonable — long stored procedures aren't classified usefully — but it means the extract intentionally does not exfiltrate full SQL text. The user-facing doc (`oracle.mdx`) is currently silent on this; please add a one-liner there saying "SQL text is sampled to the first 2KB for classification only; full statements are not exported." That sets expectations and answers a security-review question before it gets asked.
**2. ETL heuristic.** The classifier calls an `INSERT/UPDATE/DELETE/CREATE TABLE "ETL"` if its leading 2KB contains the substring `SELECT`. That's a defensible proxy (`CTAS` / `INSERT ... SELECT` / merge-style updates) but it's not obvious from the SQL itself. A header comment explaining the design would help the next maintainer:
```md
-- Workload classification heuristic:
--   INSERT/UPDATE/DELETE/CREATE TABLE statements that contain a SELECT
--   (within the first 2KB of SQL text) are bucketed as ETL; otherwise
--   they fall into their literal command type. This intentionally
--   prefers false-positive ETL over false-negative — adjust per customer.
```

src/databricks/labs/lakebridge/resources/assessments/oracle/config_containers.sql
dgomez04, line 1: - **-- Has to be executed on CDB** is now enforced by the doc and the connector setup ("must run against CDB$ROOT"), so the inline reminder is redundant. Same comment appears in several other files; if it's kept anywhere, please make it accurate (e.g. "Requires SET CONTAINER to CDB$ROOT" — but honestly, deleting all of them is fine since the doc owns this contract).

Feel free to delete this file if unused.

src/databricks/labs/lakebridge/resources/assessments/oracle/config_instance.sql
dgomez04, line 1: - **-- Has to be executed on CDB** is now enforced by the doc and the connector setup ("must run against CDB$ROOT"), so the inline reminder is redundant. Same comment appears in several other files; if it's kept anywhere, please make it accurate (e.g. "Requires SET CONTAINER to CDB$ROOT" — but honestly, deleting all of them is fine since the doc owns this contract).
- **-- spool results/config_instance.csv**  is a SQL*Plus output directive that has no effect under `oracledb`. 

Feel free to delete this file if unused.

src/databricks/labs/lakebridge/resources/assessments/validation/oracle_extract_schema.yml
dgomez04, line 1: What is this file for? It's not wired in anywhere in the codebase.

tests/integration/assessments/test_profiler.py
dgomez04, line 1: - `test_Oracle_*` naming violates **PEP 8** and the surrounding nake_case test function convention. 
- No unit tests for the new `ConfigureOracleAssessment` class, the `OracleConnector` dispatch, or the Oracle branch in `check_connection`. Compare to `legacy_sqldw`, which has all three.

tests/integration/assessments/test_profiler.py
```diff
@@ -8,6 +8,37 @@
...
+
+
+def test_Oracle_profile_missing_platform_config() -> None:
+    """Test that profiling an unsupported platform raises ValueError"""
+    with pytest.raises(ValueError, match="Cannot Proceed without a valid pipeline configuration for oracle"):
+        profiler = Profiler("oracle", None)
+        profiler.profile()
+
+
+def test_Oracle_profile_execution() -> None:
```
dgomez04: There's no `@pytest.mark.skipif(not env_var)`, no mock or fixture. CI will need an Oracle instance configured or this will fail. Previous pull requests' analogous tests (`tests/unit/...`) used `MockPrompts` and patched the connector; this one doesn't.

tests/integration/assessments/test_profiler.py
```diff
@@ -8,6 +8,37 @@
...
+    """Test successful profiling execution using actual pipeline configuration"""
+    profiler = Profiler("oracle")
+    path_prefix = Path(__file__).parent / "../../../"
+    config_file = path_prefix / "src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml"
+    config = profiler.path_modifier(config_file=config_file, path_prefix=path_prefix)
+    profiler.profile(pipeline_config=config)
+    assert Path(
```
dgomez04: Hardcoded `/tmp/...` path for the assertion — same Windows-portability issue. Also: the test will leave a DuckDB file in `/tmp` after running, and there's no cleanup.

docs/lakebridge/docs/assessment/profiler/oracle.mdx
```diff
@@ -0,0 +1,188 @@
+---
+sidebar_position: 3
```
dgomez04: The new `legacy_sqldw.mdx` (#2441) also uses `sidebar_position: 3`, and `mssql.mdx` already uses `sidebar_position: 2`. If both PRs land, two pages have position 3 and ordering becomes undefined. Whichever lands second needs to bump.

docs/lakebridge/docs/assessment/profiler/oracle.mdx
```diff
@@ -0,0 +1,188 @@
...
+- [Execute the Profiler](#execute-the-profiler)
+
+:::info
+The Oracle profiler targets **multitenant container databases (CDB)** and reads from `CDB_*`, `GV$*`, and AWR views.
+It must be run against `CDB$ROOT`, not an individual PDB.
+:::
+
+## Prerequisites
+
+### 1. Required Database Permissions
```
dgomez04: **Numbering** inconsistencies across the section.

docs/lakebridge/docs/assessment/profiler/oracle.mdx
dgomez04, line 1: **No mention of other authentication mechanisms such as TLS / wallet** — Oracle is virtually always behind some authentication boundary in enterprise environments; this gives no hint that wallet/TCPS aren't supported.

**No prerequisite section about the `oracledb` driver.** The other profiler docs explicitly call out the ODBC driver + VC++ runtime; this doc says nothing about Python driver requirements (luckily `oracledb` thin mode needs no native client, so this is informational only — but it's worth saying so).

docs/lakebridge/docs/assessment/profiler/oracle.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+### Configuration Parameters
+
+| Parameter | Description | Default |
+|:----------|:------------|:--------|
+| **Secret vault type** | `local` for plain text values, `env` to read from environment variables | — |
+| **Host** | Oracle server hostname, IP, or RAC SCAN name | — |
+| **TNS Listener port** | Port the Oracle listener is bound to | `1521` |
+| **TNS service name** | Service name registered in the Oracle listener (the CDB service) | `orcl` |
+| **User** | Database user with the grants from [Required Database Permissions](#2-required-database-permissions) | `SYSTEM` |
```
dgomez04: The actual heading is `### 1. Required Database Permissions`, so the anchor is `#1-required-database-permissions`. The `#2 ...` anchor is broken.

docs/lakebridge/docs/assessment/profiler/oracle.mdx
```diff
@@ -0,0 +1,188 @@
...
+CREATE USER c##lakebridge_profiler IDENTIFIED BY <password> CONTAINER=ALL;
+GRANT CREATE SESSION TO c##lakebridge_profiler CONTAINER=ALL;
+GRANT SET CONTAINER TO c##lakebridge_profiler CONTAINER=ALL;
+
+-- Catalog and AWR access
+GRANT SELECT_CATALOG_ROLE TO c##lakebridge_profiler CONTAINER=ALL;
+GRANT SELECT ANY DICTIONARY TO c##lakebridge_profiler CONTAINER=ALL;
+```
+
+:::tip
```
dgomez04: **Diagnostic Pack callout is good** — flagging that AWR (`CDB_HIST_*`) requires a paid Oracle add-on. But the YAML's `flag: active` for all `perf_*` steps means you'll silently fail on databases without the license. Should be `flag: inactive` by default for those steps, with the documentation explaining how to flip them on after confirming Diagnostic Pack is licensed.

docs/lakebridge/docs/assessment/profiler/oracle.mdx
```diff
@@ -0,0 +1,188 @@
...
+```
+
+The profiler will:
+1. Connect to your Oracle CDB using the configured credentials
+2. Execute the configuration and performance metric extraction queries
+3. Store the results in a local DuckDB extract file
+
+### Profiler Summary Dashboard
+
+Coming Soon
```
dgomez04: Either add or remove this section.

dgomez04 (changes requested):
Left my comments - let me know if there are any questions or follow-ups I need to address.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
m-abulazm: we can do a followup to support different auth methods

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: lets iterate on the connection in a followup

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: this is fixed on https://github.com/databrickslabs/lakebridge/pull/2421. please review

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: https://github.com/databrickslabs/lakebridge/pull/2421

src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml
m-abulazm: cleaned up

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_heatmap.sql
m-abulazm: cleaned

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_heatmap_raw.sql
m-abulazm: cleaned

src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml
m-abulazm: lets clean this up in a separate PR and investigate how to standardize it

src/databricks/labs/lakebridge/resources/assessments/oracle/config_db_features.sql
m-abulazm: done

src/databricks/labs/lakebridge/resources/assessments/oracle/config_storage.sql
m-abulazm: done

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_fgd_session_evolution.sql
m-abulazm: done

src/databricks/labs/lakebridge/resources/assessments/oracle/config_pdb_objects.sql
m-abulazm: done

src/databricks/labs/lakebridge/resources/assessments/oracle/perf_cpu_waits.sql
m-abulazm: done

docs/lakebridge/docs/assessment/profiler/oracle.mdx
m-abulazm: I dont expect end users to edit src code. it is called out in the docs and if they fail, that will not break profiler run and querying other tables

src/databricks/labs/lakebridge/assessments/configure_assessment.py
m-abulazm: I would still default to SYSTEM. 
the docs explain what to do for least privileged access. or we can remove the default completely

src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml
m-abulazm: Agreed. lets have a followup for that

.gitignore
dgomez04, line 1: Aren't these just duplicates of the above? Non-blocking but can be cleaned up.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
dgomez04: Honestly - I'd remove the default to make sure users are completely aware of what they're selecting (considering they'll need to select a password afterwards). Let me know if you agree.

src/databricks/labs/lakebridge/connections/database_manager.py
dgomez04: When this PR rebases on that, please confirm: (a) the db_type == "oracle" branch in check_connection is gone, (b) OracleConnector.health_check() returns SELECT 101 FROM dual, (c) self._db_type attribute is removed from __init__, and (d) a unit test covers the new OracleConnector.health_check() path.

tests/integration/assessments/test_profiler_validator.py
```diff
@@ -115,7 +115,7 @@ def test_validate_invalid_source_tech(mock_synapse_profiler_extract: Path, test_
         schema_check = ExtractSchemaValidationCheck(
             "main",
             "dedicated_routines",
-            source_tech="oracle",
+            source_tech="bogus",
```
dgomez04, line 118: Are we not adding the tests with this pull request?

dgomez04 (changes requested):
Left small comments - please review. 

**Additionally**, can you provide more context on what `/oracle_extract_schema does?`

src/databricks/labs/lakebridge/resources/assessments/validation/oracle_extract_schema.yml
m-abulazm: These are post-run validation schemas for the profiler-dashboard upload step. `create-profiler-dashboard` command runs `_validate_profiler_extract`, which loads the YAML and runs two checks per declared table:                                     
  1. EmptyTableValidationCheck — table contains at least one row.                                                                                                                                                
  2. ExtractSchemaValidationCheck — actual DuckDB column types match the YAML spec.

dgomez04 (approved):
LGTM - let's keep track of the follow-up issues we opened.

PR #2196, opened 2025-12-16
Fix reconciliation hash mismatch for due to column ordering

### LINKED ISSUE #2195: [BUG]: Reconcile Hash Calculation Uses Different Column Ordering for T-SQL/Synapse Source and Databricks Target

### Category of Bug / Issue

Reconcile bug

### Current Behavior

When reconciling tables, the column ordering used for string concatenation during the hash calculation is different between the T-SQL/Synapse source query and the Databricks target query. This difference is likely due to how the query builder handles or sorts column names. 

Discrepancies arise when tables contain columns that are subsets of other column names (e.g., **month** and **month_num**), leading to non-identical sort results between the two dialects.

This inconsistent ordering results in the generation of completely different hash values for the same row data, causing **false mismatches for all rows** in the affected tables.

Observed Ordering Discrepancy (Example from logs):
Image

### Expected Behavior

The column order used for string concatenation in the hash calculation **must be identical** for both the T-SQL/Synapse source query and the Databricks target query to ensure comparable hash values.

The generator logic should implement a deterministic sorting mechanism (e.g., a simple alphabetical sort of the unquoted column names) before applying dialect-specific quoting (like square brackets), thus guaranteeing hash consistency across all supported environments.

### Steps To Reproduce

1. Create a table in both the T-SQL/Synapse source and the Databricks target environments.
2. Ensure the table contains columns whose names differ only by suffixes (e.g., product and product_name, or month and month_num).
3. Load identical data into both tables.
4. Run the Lakebridge reconcile process using T-SQL/Synapse as the source.
5. Observe that **all rows are reported as mismatched** due to the hash ordering inconsistency.

```shell
No exception is thrown. The symptom is a 100% mismatch rate for the affected tables.
```

### Logs Confirmation

### Sample Query

```shell

```

### Operating System

Windows

### Version

latest via Databricks CLI

sundarshankar89:
### What does this PR do?
  Fixes a bug in reconciliation hash query generation where columns with names that are substrings of other column names (e.g., `month` vs `month_num`, `product` vs `product_name`) were sorted inconsistently across SQL dialects, causing hash mismatches and false reconciliation failures.

 - **`hash_query.py`**: 
    - Add `sort_key` field containing unnormalized column name
    - Change sorting logic from `column["alias"]` to `column["sort_key"].lower()`
    - Add detailed comments explaining the fix and referencing issue #2195

  ### Before
  ```sql
  -- T-SQL might generate: [month] < [month_num] (correct)
  -- Databricks might generate: `month_num` < `month` (WRONG due to delimiter sort)
  -- Result: Hash mismatch for identical data

  After

  -- T-SQL: [id], [month], [month_num], [revenue], [year] (alphabetical)
  -- Databricks: `id`, `month`, `month_num`, `revenue`, `year` (alphabetical)
  -- Result: Consistent hashes across all dialects ✅
```

### Linked issues

Fixes #2195 

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/reconcile/query_builder/hash_query.py
+++ b/src/databricks/labs/lakebridge/reconcile/query_builder/hash_query.py
@@ -13,7 +13,6 @@ from databricks.labs.lakebridge.reconcile.query_builder.expression_generator imp
     transform_expression,
     build_column_no_alias,
 )
-from databricks.labs.lakebridge.reconcile.connectors.dialect_utils import DialectUtils
 
 logger = logging.getLogger(__name__)
 
@@ -52,7 +51,7 @@ class HashQueryBuilder(QueryBuilder):
             {
                 "this": self._build_column_name_source_normalized(col),
                 "alias": self._build_alias_source_normalized(col),
-                "sort_key": DialectUtils.unnormalize_identifier(self._build_alias_source_normalized(col)),
+                "sort_key": self._unnormalize_identifier(col),
             }
             for col in hash_cols
         ]

### CODE REVIEW

File: src/databricks/labs/lakebridge/reconcile/query_builder/hash_query.py [hunk 1]
```diff
@@ -44,12 +45,19 @@ def build_query(self, report_type: str) -> str:
...
-        # same hash value
+        # same hash value. We sort by the unnormalized column name (without delimiters) to ensure consistent
+        # ordering across all dialects, regardless of their delimiter characters ([], "", ``, etc.)
+        # Fix for https://github.com/databrickslabs/lakebridge/issues/2195
         hash_cols_with_alias = [
-            {"this": self._build_column_name_source_normalized(col), "alias": self._build_alias_source_normalized(col)}
+            {
+                "this": self._build_column_name_source_normalized(col),
+                "alias": self._build_alias_source_normalized(col),
+                "sort_key": DialectUtils.unnormalize_identifier(self._build_alias_source_normalized(col)),
```
m-abulazm:
```suggestion
                "sort_key": self._unnormalize_identifier(col),
```

this method is safer.  `DialectUtils.unnormalize_identifier` works only on ansi identifiers (so it worked only for the target layer). not sure why backticks affect sorting opposed to tsql delimiter. just to be safe lets use the other method which will always return unnormalized identifiers

sundarshankar89: addressed

PR #2200, opened 2025-12-19
Revert back to use MD5 with oracle

### LINKED ISSUE #2199: [FEATURE]: Improve Oracle checksum calculation to support Oracle 11g hash limitations

Reconcile

### Problem statement
Oracle 11g’s `DBMS_CRYPTO` package only supports MD5, MD4, and SHA-1 hash algorithms and does not support SHA-256, so using `DBMS_CRYPTO.hash` with SHA-256 raises `ORA-28827: invalid cipher type passed`, causing reconcile jobs to fail for 11g sources.  This breaks Lakebridge Reconcile for customers still running Oracle 11g whenever the checksum logic relies on SHA-256 in `DBMS_CRYPTO.hash`.

### Proposed Solution
Introduce a backwards-compatible mechanism that allows Oracle 11g databases to continue using a supported hash algorithm (for example MD5) while keeping SHA-256 for newer Oracle versions that support it.  This could be implemented as:

- Reconcile first checks whether SHA-256 is supported (for example via a lightweight capability/version check or a guarded test call), and if not, automatically falls back to MD5.  
- This preserves a more secure default (SHA-256) for modern Oracle versions while restoring functionality for legacy Oracle 11g deployments, without requiring users to manually fork or pin to older Lakebridge versions.

### Additional Context

_No response_

m-abulazm:
### What does this PR do?
Revert back to MD5 hashing for oracle in reconcile

Although the feature request wanted to support both sh256 and MD5. the bug why MD5 didnt work originally was found and fixed  

As a quick win, using MD5 for oracle as it was originally

### Linked issues

Resolves #2199 

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [X] manually tested through integration test
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -26,7 +26,7 @@ jobs:
       labels: larger
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -20,7 +20,7 @@ jobs:
       run:
         working-directory: docs/lakebridge
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - uses: actions/setup-node@v6
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -17,7 +17,7 @@ jobs:
       run:
         working-directory: docs/lakebridge
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - uses: actions/setup-node@v6
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -24,7 +24,7 @@ jobs:
       labels: linux-ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install Python
         uses: actions/setup-python@v6
@@ -51,7 +51,7 @@ jobs:
       labels: linux-ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Install Python
         uses: actions/setup-python@v6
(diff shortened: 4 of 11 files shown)


src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
```diff
@@ -281,10 +281,7 @@ def _get_is_string(column_types_dict: dict[str, DataType], column_name: str) ->
     ),
     get_dialect("oracle"): HashAlgoMapping(
         source=partial(
-            anonymous,
-            func="DBMS_CRYPTO.HASH(UTL_I18N.STRING_TO_RAW({}, 'AL32UTF8'), 4)",
-            is_expr=True,
-            dialect=get_dialect("oracle"),
+            anonymous, func="DBMS_CRYPTO.HASH(RAWTOHEX({}), 2)", is_expr=True, dialect=get_dialect("oracle")
```
sundarshankar89: Self note: @sundarshankar89  to check if source algo changes target algo needs to map.

sundarshankar89 (commented):
@m-abulazm I assumed you were going to auto-check for function availability before determining the algorithm? 
Is it because we don't have the current mechanism to do that?

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
m-abulazm: target is MD5. I fixed the bug that target algorithm was not used

m-abulazm:
> @m-abulazm I assumed you were going to auto-check for function availability before determining the algorithm? Is it because we don't have the current mechanism to do that?

I went for a quick win. I also tested that it works with new oracle versions using the integration test

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
sundarshankar89: I will need more time to check, given that if you use MD5, Databricks should also use MD5; otherwise, the comparison will break. I need to check the test.

asnare:
```suggestion
            # Need to support Oracle 11; modern hashes aren't available.
            # (This hashing function does not serve a security purpose: MD5 is fine in this situation.)
            anonymous, func="DBMS_CRYPTO.HASH(RAWTOHEX({}), DBMS_CRYPTO.HASH_MD5)", is_expr=True, dialect=get_dialect("oracle")
```

(If the constant isn't available, please add a comment that '2' selects the MD5 hashing algorithm.)

m-abulazm: constant is not available. comment added

asnare (approved):

PR #2210, opened 2025-12-30
Eliminate `Path(__file__)` to locate resource files

asnare:
### What does this PR do?

This PR eliminates the use of `Path(__file__)` to locate file-based resources. In general:

 - In the main code, this is buggy: modern Python should use `importlib.resources` instead.
 - In tests, it's brittle and difficult to read because it's normally based on relative paths and traversal of many parent directories. It's also unnecessary, because pytest provides a mechanism for locating the project root. (Tests can also use `importlib.resources`, and probably should, but that's a bigger refactoring: this is a first step.)

Some incidental changes that are introduced:

 - Type hints in a few places.
 - Functional tests against Snowflake, Oracle and Presto have been consolidated into a single module. (The Snowflake functional tests have been moved into `test_snowflake.py`: `test_databricks.py` was misleading.)
 - The `PipelineConfig` (and `Step`) classes have been made immutable. (An unnecessary `__post__init()` method has also been eliminated.)

### Tests

- existing unit tests
- existing integration tests

### DIFF
--- a/.codegen.json
+++ b/.codegen.json
@@ -5,6 +5,7 @@
   "toolchain": {
     "required": ["hatch"],
     "pre_setup": ["hatch env create"],
-    "prepend_path": ".venv/bin"
+    "prepend_path": ".venv/bin",
+    "acceptance_path": "tests/integration"
   }
 }
--- a/.github/scripts/setup_spark_remote.sh
+++ b/.github/scripts/setup_spark_remote.sh
@@ -91,6 +91,8 @@ else
   fi
 fi
 
+rm -rf "${HOME}"/spark/"${spark}"/spark-warehouse
+echo "Cleared old spark warehouse default directory"
 
 cd "${spark}" || exit 1
 ## check spark remote is running,if not start the spark remote
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -45,6 +45,12 @@ jobs:
           chmod +x $GITHUB_WORKSPACE/.github/scripts/setup_mssql_odbc.sh
           $GITHUB_WORKSPACE/.github/scripts/setup_mssql_odbc.sh
 
+      # TODO: Migrate tests to use Databricks clusters instead of Spark local mode
+      - name: Setup spark
+        run: |
+          chmod +x $GITHUB_WORKSPACE/.github/scripts/setup_spark_remote.sh
+          $GITHUB_WORKSPACE/.github/scripts/setup_spark_remote.sh
+
       - name: Run integration tests
         uses: databrickslabs/sandbox/acceptance@acceptance/v0.4.4
         with:
--- a/src/databricks/labs/lakebridge/reconcile/recon_capture.py
+++ b/src/databricks/labs/lakebridge/reconcile/recon_capture.py
@@ -95,7 +95,7 @@ def generate_final_reconcile_output(
     metadata_config: ReconcileMetadataConfig = ReconcileMetadataConfig(),
     local_test_run: bool = False,
 ) -> ReconcileOutput:
-    _db_prefix = "default" if local_test_run else f"{metadata_config.catalog}.{metadata_config.schema}"
+    _db_prefix = metadata_config.schema if local_test_run else f"{metadata_config.catalog}.{metadata_config.schema}"
     recon_df = spark.sql(
         f"""
     SELECT
@@ -237,7 +237,9 @@ class ReconCapture:
         self.source_dialect = source_dialect
         self.ws = ws
         self.spark = spark
-        self._db_prefix = "default" if local_test_run else f"{metadata_config.catalog}.{metadata_config.schema}"
+        self._db_prefix = (
+            metadata_config.schema if local_test_run else f"{metadata_config.catalog}.{metadata_config.schema}"
+        )
 
     def _generate_recon_main_id(
         self,
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,4 +1,3 @@
-import os
 import logging
 from urllib.parse import urlparse
 
@@ -9,8 +8,9 @@ from databricks.labs.lakebridge.__about__ import __version__
 from databricks.labs.lakebridge.connections.database_manager import DatabaseManager
 from tests.integration.debug_envgetter import TestEnvGetter
 
-logging.getLogger("tests").setLevel("DEBUG")
-logging.getLogger("databricks.labs.lakebridge").setLevel("DEBUG")
+logging.getLogger("tests").setLevel(logging.DEBUG)
+logging.getLogger("databricks.labs.lakebridge").setLevel(logging.DEBUG)
+logging.getLogger("databricks.labs.pytester").setLevel(logging.DEBUG)
 
 logger = logging.getLogger(__name__)
 
@@ -28,22 +28,6 @@ def product_info() -> tuple[str, str]:
 @pytest.fixture
 def get_logger():
     return logger
-
-
-def pytest_collection_modifyitems(config, items):
-    if os.getenv('TEST_ENV') != 'ACCEPTANCE':
-        return
-    selected_items = []
-    deselected_items = []
-    # Add only specific tests to run from acceptance.yml
-    inclusions = {'assessments', 'connections', 'config', 'discovery', 'helpers', 'transpile'}
-    for item in items:
-        if any(f"tests/integration/{inclusion}" in str(item.fspath) for inclusion in inclusions):
-            selected_items.append(item)
-        else:
-            deselected_items.append(item)
-    items[:] = selected_items
-    config.hook.pytest_deselected(items=deselected_items)
 
 
 @pytest.fixture(scope="session")
--- a/tests/integration/reconcile/conftest.py
+++ b/tests/integration/reconcile/conftest.py
@@ -1,14 +1,24 @@
 import logging
+import uuid
+from collections.abc import Generator
 
 import pytest
 
 from databricks.sdk import WorkspaceClient
 from databricks.sdk.errors.platform import PermissionDenied
 from databricks.sdk.service.catalog import TableInfo, SchemaInfo
+
+from databricks.labs.lakebridge.config import ReconcileMetadataConfig
 from tests.integration.debug_envgetter import TestEnvGetter
 
 logger = logging.getLogger(__name__)
 
+DIAMONDS_COLUMNS = [
+    ("carat", "DOUBLE"),
+    ("cut", "STRING"),
+    ("color", "STRING"),
+    ("clarity", "STRING"),
+]
 DIAMONDS_ROWS_SQL = """
                     INSERT INTO {catalog}.{schema}.{table} (carat, cut, color, clarity) VALUES
                         (0.23, 'Ideal', 'E', 'SI2'),
@@ -24,6 +34,7 @@ DIAMONDS_ROWS_SQL = """
 def recon_catalog(make_catalog) -> str:
     try:
         catalog = make_catalog().name
+        logger.info(f"Created catalog {catalog} for recon tests")
     except PermissionDenied as e:
         logger.warning("Could not create catalog for recon tests, using 'sandbox' instead", exc_info=e)
         catalog = "sandbox"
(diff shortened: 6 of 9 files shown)


src/databricks/labs/lakebridge/assessments/dashboards/execute.py
```diff
@@ -64,10 +68,11 @@ def _validate_profiler_extract(
 ) -> bool:
     logger.info("Validating the profiler extract file.")
     validation_checks: list[EmptyTableValidationCheck | ExtractSchemaValidationCheck] = []
-    schema_def_path = f"{Path(__file__).parent}/../../resources/assessments/{source_tech}_schema_def.yml"
-    tables = _get_extract_tables(schema_def_path)
+    # TODO: Verify this, I don't think it works? (These files are part of the test resources.)
+    schema_def = resources.files(assessment_resources).joinpath(f"{source_tech}_schema_def.yml")
```
asnare, line 72: I don't think this worked prior to this PR (or after): this is a pure refactoring.

But it does raise an interesting point: how is this tested?

@sundarshankar89: How can I check this?

tests/integration/assessments/test_pipeline.py
```diff
@@ -1,52 +1,42 @@
...
+from databricks.labs.lakebridge.connections.database_manager import DatabaseManager
 
 
-@pytest.fixture(scope="module")
-def pipeline_config():
-    prefix = Path(__file__).parent
-    config_path = f"{prefix}/../../resources/assessments/pipeline_config.yml"
-    config = PipelineClass.load_config_from_yaml(config_path)
+@pytest.fixture
+def pipeline_configuration_loader(test_resources: Path, project_path: Path) -> Callable[[Path], PipelineConfig]:
```
m-abulazm:
```suggestion
type Loader = Callable[[Path], PipelineConfig]

def pipeline_configuration_loader(test_resources: Path, project_path: Path) -> Loader:
```

tests/integration/assessments/test_pipeline.py
```diff
@@ -1,52 +1,42 @@
...
+    return _load
 
 
-@pytest.fixture(scope="module")
-def pipeline_dep_failure_config():
-    prefix = Path(__file__).parent
-    config_path = f"{prefix}/../../resources/assessments/pipeline_config_failure_dependency.yml"
-    config = PipelineClass.load_config_from_yaml(config_path)
+@pytest.fixture
+def pipeline_config(pipeline_configuration_loader) -> PipelineConfig:
```
m-abulazm:
```suggestion
def pipeline_config(pipeline_configuration_loader: Loader) -> PipelineConfig:
```

tests/integration/assessments/test_pipeline.py
```diff
@@ -1,52 +1,42 @@
...
+    return pipeline_configuration_loader(Path("pipeline_config.yml"))
 
-    for step in config.steps:
-        step.extract_source = f"{prefix}/../../{step.extract_source}"
-    return config
 
+@pytest.fixture
+def pipeline_dep_failure_config(pipeline_configuration_loader) -> PipelineConfig:
```
m-abulazm:
```suggestion
def pipeline_dep_failure_config(pipeline_configuration_loader: Loader) -> PipelineConfig:
```

tests/integration/assessments/test_pipeline.py
```diff
@@ -1,52 +1,42 @@
...
-def sql_failure_config():
-    prefix = Path(__file__).parent
-    config_path = f"{prefix}/../../resources/assessments/pipeline_config_sql_failure.yml"
-    config = PipelineClass.load_config_from_yaml(config_path)
-    for step in config.steps:
-        step.extract_source = f"{prefix}/../../{step.extract_source}"
-    return config
 
+@pytest.fixture
+def sql_failure_config(pipeline_configuration_loader) -> PipelineConfig:
```
m-abulazm:
```suggestion
def sql_failure_config(pipeline_configuration_loader: Loader) -> PipelineConfig:
```

tests/integration/assessments/test_pipeline.py
```diff
@@ -1,52 +1,42 @@
...
-def python_failure_config():
-    prefix = Path(__file__).parent
-    config_path = f"{prefix}/../../resources/assessments/pipeline_config_python_failure.yml"
-    config = PipelineClass.load_config_from_yaml(config_path)
-    for step in config.steps:
-        step.extract_source = f"{prefix}/../../{step.extract_source}"
-    return config
+
+@pytest.fixture
+def python_failure_config(pipeline_configuration_loader) -> PipelineConfig:
```
m-abulazm:
```suggestion
def python_failure_config(pipeline_configuration_loader: Loader) -> PipelineConfig:
```

src/databricks/labs/lakebridge/assessments/dashboards/execute.py
```diff
@@ -79,7 +84,7 @@ def _validate_profiler_extract(
                     table_info[1],
                     source_tech=source_tech,
                     extract_path=extract_location,
-                    schema_path=schema_def_path,
+                    schema_path=str(schema_def_path),
```
m-abulazm, line 87: `schema_path` should be a `Path`

src/databricks/labs/lakebridge/assessments/profiler.py
```diff
@@ -41,9 +41,8 @@ def supported_platforms(cls) -> list[str]:
     def path_modifier(*, config_file: str | Path, path_prefix: Path = PRODUCT_PATH_PREFIX) -> PipelineConfig:
         # TODO: Make this work install during developer mode
         config = PipelineClass.load_config_from_yaml(config_file)
-        for step in config.steps:
-            step.extract_source = f"{path_prefix}/{step.extract_source}"
-        return config
+        new_steps = [step.copy(extract_source=str(path_prefix / step.extract_source)) for step in config.steps]
```
m-abulazm, line 44: `extract_source` should be a `Path` as well

tests/integration/assessments/test_profiler_validator.py
```diff
@@ -87,18 +83,17 @@ def test_validate_mixed_checks(mock_synapse_profiler_extract):
...
-        prefix = Path(__file__).parent
-        schema_def_path = f"{prefix}/../../resources/assessments/synapse_scheme_def_nonexists.yml"
+        schema_def_path = test_resources / "assessments" / "synapse_scheme_def_nonexists.yml"
         schema_check = ExtractSchemaValidationCheck(
             "main",
             "dedicated_routines",
             source_tech="synapse",
             extract_path=mock_synapse_profiler_extract,
-            schema_path=schema_def_path,
+            schema_path=str(schema_def_path),
```
m-abulazm, line 98: all the `schema_path` changes will not be needed when we change the property type to `Path`

src/databricks/labs/lakebridge/assessments/profiler.py
```diff
@@ -41,9 +41,8 @@ def supported_platforms(cls) -> list[str]:
     def path_modifier(*, config_file: str | Path, path_prefix: Path = PRODUCT_PATH_PREFIX) -> PipelineConfig:
```
m-abulazm, line 41: this method name is a bit confusing, it should be `load_config`. can we add a TODO here? or do the changes directly - as you prefer

tests/unit/conftest.py
```diff
@@ -214,6 +218,7 @@ def expr():
 
 
 def path_to_resource(*args: str) -> str:
+    # TODO: Figure out how to extract this.
```
m-abulazm: I don't get it. isn't this `test_resources` as well?

m-abulazm (changes requested):

tests/integration/assessments/test_pipeline.py
asnare: We can't use `type` (Python 3.12+) but can indeed make an alias for easier hinting.

tests/unit/conftest.py
asnare: Yes, but `test_resources` is a fixture, this is not, and there are 30 calls to it that all need to pass it in as an argument.

I've eliminated this method entirely and updated the call-sites to use `test_resources / *args` directly.

src/databricks/labs/lakebridge/assessments/profiler.py
asnare: I agree it's a confusing name.

I was already concerned about scope creep: the primary goal of this PR is to eliminate the usage of `Path(__file__)` to find resources, and I've already needed to make more adjustments in adjacent areas that I really want to. As such I've updated the TODO marker to indicate a less-confusing name would be welcome.

src/databricks/labs/lakebridge/assessments/dashboards/execute.py
asnare: At the moment the `schema_path` field is a string.

Yes, I could change it but I've been trying to limit the scope of the PR and avoid unrelated changes, even though a few adjacent areas have already been pulled in. I'd prefer to defer this to a separate PR.

src/databricks/labs/lakebridge/assessments/profiler.py
asnare: I agree it should be, but at the moment it's not and it wasn't introduced in this PR.

Yes, I could change it but I've been trying to limit the scope of the PR and avoid unrelated changes, even though a few adjacent areas have already been pulled in. I'd prefer to defer this to a separate PR.

tests/integration/assessments/test_profiler_validator.py
asnare: That may be, and I agree it would be better as `Path`. However it's not a field introduced in this PR. Although I could change it but I've been trying to limit the scope of the PR and avoid unrelated changes and need to draw the line somewhere. Even though a few adjacent areas have already been pulled in I'd prefer to defer this to a separate PR.

src/databricks/labs/lakebridge/assessments/dashboards/execute.py
sundarshankar89: These should be the source resources directory, not part of tests. 

@goodwillpunning correct me if I'm wrong

tests/unit/transpiler/test_databricks.py
```diff
@@ -1,15 +0,0 @@
...
-import pytest
-
-from ..conftest import FunctionalTestFile, get_functional_test_files_from_directory
-
-path = Path(__file__).parent / Path('../../resources/functional/snowflake/')
-functional_tests = get_functional_test_files_from_directory(path, "snowflake", "databricks", False)
-test_names = [f.test_name for f in functional_tests]
-
-
-@pytest.mark.parametrize("sample", functional_tests, ids=test_names)
```
sundarshankar89, line 12: @asnare are we not running these functional tests any more?

tests/unit/transpiler/test_snowflake.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+from ..conftest import FunctionalTestFile, FunctionalTestFileWithExpectedException, get_functional_test_files
+
+
+def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
```
sundarshankar89, line 6: I see you moved test_databricks to test_snowflake

asnare: Indeed, the pattern for the other ones is to name these after the source system. So the changes here are:

 - The snowflake ones were moved from `test_databricks.py` to `test_snowflake.py`.
 - There were 2 sets of functional tests for each source system, in separate files: I've moved them into the same file.

tests/unit/transpiler/test_databricks.py
asnare: They're still running. As you noticed further in the PR, these were moved into `test_snowflake.py`.

src/databricks/labs/lakebridge/assessments/dashboards/execute.py
asnare: @sundarshankar89, @goodwillpunning: The only resource I can find is: `tests/resources/assessments/synapse_schema_def.yml`

If this is not working, I propose creating an issue for this and fixing it independently of this PR.

Issue #2167, opened 2025-11-25
[BUG]: CLI ignores `--log-level=warn`


asnare:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Category of Bug / Issue

Other

### Current Behavior

The Databricks CLI has a protocol for handing off its log-level to the Lakebridge CLI. Currently this doesn't seem to work properly. (Possibly since #2117 ?)

For example:
```
databricks labs lakebridge labs --log-level=warn
```
...produces `INFO`-level logging.

Note that `--debug` is an alias for `--log-level=debug`.

### Expected Behavior

When running via the Databricks CLI, for the `databricks.*` logging categories:

 - When a log-level is provided via `--log-level=` or `--debug`, the provided log-level should be used.
 - If no explicit log-level is requested, `INFO` should be used.

The root log-level should be `DEBUG` irrespective of this.

Note that during install/uninstall (`base_install.py` and `uninstall.py`) there is no log-level handoff from the Databricks CLI: in these situations `databricks.*` should be `INFO`.

### Steps To Reproduce

Install:

 - Databricks CLI v0.278.0
 - Lakebridge 0.11.0

Install the transpilers:
```shell
databricks labs lakebridge install-transpile --log-level=warn
```

`INFO`-level logging will be produced.

### Relevant log output or Exception details

```shell
% databricks labs lakebridge install-transpile --log-level=warn
12:17:29  WARNING [d.labs.lakebridge] Working around DATABRICKS_HOST normalization issue: adb-841570839244994.14.azuredatabricks.net -> https://adb-841570839244994.14.azuredatabricks.net
12:17:31     INFO [d.l.l.transpiler.installers] databricks-bb-plugin v0.1.21 already installed
12:17:31     INFO [d.l.l.transpiler.installers] Databricks databricks-morph-plugin transpiler v0.6.11 already installed
12:17:31     INFO [d.l.lakebridge.install] Configuring lakebridge `transpile`.
12:17:34     INFO [d.l.lakebridge.install] Lakebridge `transpile` is already installed on this workspace.
Do you want to override the existing installation? (default: no): 
12:17:41     INFO [d.l.lakebridge.install] Installation completed successfully! Please refer to the documentation for the next steps
```

### Logs Confirmation

- [ ] I ran the command line with `--debug`
- [ ] I have attached the `lsp-server.log` under USER_HOME/.databricks/labs/remorph-transpilers/<converter_name>/lib/lsp-server.log

[...]

PR #2216, opened 2026-01-06
Survey and fix logger initialization

### LINKED ISSUE #2211: [BUG]: CLI mishandles `--log-level=` and `--debug`
### Category of Bug / Issue

Other

### Current Behavior

The Databricks CLI has a protocol for handing off its log-level to the Lakebridge CLI. Since #2117 this is largely ignored, with `databricks.labs.lakebridge` logging only being emitted at `INFO`-level.

For example, with:
```shell
databricks labs lakebridge install-transpile --log-level=warn
databricks labs lakebridge install-transpile --log-level=debug
databricks labs lakebridge install-transpile --debug
```
...we see:

 - The given log-level[^1] is used for `databricks.*`;
 - BUT `databricks.labs.lakebridge.*` is limited to only `INFO` (or higher) logging.

(Note that `--debug` is an alias for `--log-level=debug`.)

[^1]: A separate bug, #2167, affects `--log-level=warn` separately in a different manner.

### Expected Behavior

All of `databricks.*` should be set to the same log-level as specified by the `--log-level=` option.

### Steps To Reproduce

Install:

 - Databricks CLI v0.281.0
 - Lakebridge 0.11.3

Install the transpilers:
```shell
databricks labs lakebridge install-transpile --log-level=debug
```

Under `databricks.labs.lakebridge` there won't any `DEBUG` logging produced, even though there should be for things like the Java check and metadata lookups.

```shell
11:50:36     INFO [d.l.l.transpiler.installers] Databricks databricks-morph-plugin transpiler v0.6.14 already installed             
11:50:36     INFO [d.l.l.transpiler.installers] databricks-bb-plugin v0.1.24 already installed                                      
11:50:36     INFO [d.l.lakebridge.install] Configuring lakebridge `transpile`.
```

### Operating System

macOS

### Version

latest via Databricks CLI

asnare:
### What does this PR do?

This PR makes the following changes related to logging within the project:

 - Log-levels are only configured at entry points, not within code and not as a side-effect of importing anything or running tests.
 - Logging initialisation during the install/uninstall hooks is now consistent.
 - A temporary workaround (needed due to #2167, resolves #2211) for the CLI is introduced so that we can once again get `--debug` logs if we need them.

The `base_install.py` entry point for installation has been moved into `install.py`: this where it was logging from, and where it's normally located in blueprint-based applications.

### Linked issues

Resolves: #2211
Relates to: #2167

### Functionality

Modified existing commands:

 -  `databricks labs lakebridge *` (all subcommands)
 -  `databricks labs install lakebridge`
 -  `databricks labs uninstall lakebridge`

### Tests

- manually tested
- existing unit tests
- existing integration tests

#### Manual Tests

 - `databricks labs install lakebridge`
 - `databricks labs install lakebridge@consistent-log-initialization`
 - `rm -fr ~/.databricks/labs`
 - `databricks labs install lakebridge@consistent-log-initialization`
 - `databricks labs lakebridge install-transpile --debug`
 - `databricks labs install lakebridge@consistent-log-initialization`
 - `databricks labs uninstall lakebridge`


src/databricks/labs/lakebridge/cli.py
```diff
@@ -57,15 +55,25 @@ def create_workspace_client(self) -> WorkspaceClient:
         self._patch_databricks_host()
         return self._workspace_client()
 
-    def get_logger(self) -> logging.Logger:
-        if self._logger_instance is None:
-            self._logger_instance = self._logger
-            self._logger_instance.setLevel(logging.INFO)
-        return self._logger_instance
+    def _log_level(self, raw: str) -> int:
```
m-abulazm, line 59: are we doing this change here because it is faster than releasing blueprint with that change?

src/databricks/labs/lakebridge/__init__.py
```diff
@@ -2,6 +2,8 @@
 from databricks.labs.blueprint.logger import install_logger
 from databricks.labs.lakebridge.__about__ import __version__
 
+# Ensure that anything that imports this (or lower) submodules triggers setup of the blueprint
+# logging.
 install_logger()
```
m-abulazm, line 9: we should move this to the entry points similar to bladebridge

src/databricks/labs/lakebridge/install.py
```diff
@@ -430,3 +434,33 @@ def _verify_workspace_client(ws: WorkspaceClient) -> WorkspaceClient:
...
+
+def initialize_logging() -> None:
+    """Common logging initialisation during install and uninstall."""
+    # This is intended to emulate the behaviour of the blueprint App() initialisation, except that we don't have
+    # handoff from the Databricks CLI. As such the policy is:
+    #   - The root (and logging system in general) is left alone.
+    #   - If running in the IDE debugger, databricks.* will be set to DEBUG.
+    #   - Otherwise, databricks.* will be set to INFO.
+    databricks_log_level = logging.DEBUG if is_in_debug() else logging.INFO
+    logging.getLogger("databricks").setLevel(databricks_log_level)
```
m-abulazm:
```suggestion
    log_level = logging.DEBUG if is_in_debug() else logging.INFO
    install_logger(log_level)
```

src/databricks/labs/lakebridge/cli.py
asnare: Yes, and I'm hoping that it's a temporary. (My plan is to implement this workaround in `blueprint` next week if I don't have a better idea about when it will fixed upstream.)

src/databricks/labs/lakebridge/resources/assessments/synapse/common/functions.py
```diff
@@ -2,24 +2,13 @@
 import json
 import sys
 import logging
+
 from azure.identity import DefaultAzureCredential
 from azure.monitor.query import MetricsQueryClient
 from azure.synapse.artifacts import ArtifactsClient
 
 
-def set_logger(name: str = __name__) -> logging.Logger:
```
sundarshankar89, line 10: @asnare, I need to test this was done because the pipeline creates a separate venv. I'm not fully sure the log levels are passed through without explicitly setting them.

asnare: @sundarshankar89: Thank-you, please let me know. :)

Irrespective of what we find, this wasn't the right way to do it… depending on what the tests show I'll figure out where this should go.

sundarshankar89: it fails 

``` console
09:40:23     INFO [d.l.l.assessments.pipeline] ImportError: cannot import name 'set_logger' from 'databricks.labs.lakebridge.resources.assessments.synapse.common.functions' (/Users/xxxxxxx/.databricks/labs/lakebridge/lib/src/databricks/labs/lakebridge/resources/assessments/synapse/common/functions.py)
09:40:24     INFO [d.l.l.assessments.pipeline] Step 'workspace_info' completed with status: StepExecutionStatus.ERROR
```

How to run `databricks labs lakebridge configure-database-profiler` 
and then `databricks labs lakebridge execute-database-profiler`

asnare: @sundarshankar89: I've had a look at how this works, and tweaked things a bit. Can you let me know if it's looking better now?

src/databricks/labs/lakebridge/__init__.py
asnare: Hmm… why?

src/databricks/labs/lakebridge/install.py
asnare: Can you elaborate on this a bit more? I don't understand why you're suggesting this change.

Here we're setting the filtering log-level on the `databricks` logger (which `databricks.*` delegate to by default)

This is different to the log-level that we pass to `install_logger()`: that one configures the minimum level that will be written out to the console: it defaults to `DEBUG` so that anything that reaches it will be written out, which is the conventional way to configure the handlers.

src/databricks/labs/lakebridge/resources/assessments/synapse/common/functions.py
asnare: (Manual testing showed that it now works.)

gueniai (approved):
LGTM

src/databricks/labs/lakebridge/install.py
m-abulazm: blueprint configures databricks logger depending on the requested log level from the user. why do we need to set it again?

we should also configure the root logger with the right log level. if we do not set it and only set databricks logger, the user might get the wrong `DEBUG` logs if they run with info as root logger allows `DEBUG`

src/databricks/labs/lakebridge/__init__.py
m-abulazm: because this means we keep reinitializing the logger over and over again instead of just once. it feels odd and will break some handlers e.g. file handlers. I know we dont use it and will be blocked in the future to use if we want to

asnare: @m-abulazm: That's not how imports work in python: modules are effectively imported once and cached. The way `import` works is:

1. Python looks inside `sys.modules` for the module. If it's there, import just uses the already-initialised module.
2. If not, locate via `sys.path` and initialise, which "runs" the module. (This is where the above `install_logger()` call happens.
3. Add the initialised module to `sys.modules`, after which step 1 will always find it.

(Without this approach things like using local imports to break cycles would not work.)

This means that the `install_logger()` above is only invoked once, not over and over again.

src/databricks/labs/lakebridge/install.py
asnare: @m-abulazm: Blueprint does not configure the logger level, it only configures the level of the handler it attaches. Handlers and loggers have their own (independent) levels. This is described in the docstring as such:

```
    The root logger will be modified:

     - Its logging level will be left as-is.
     - All existing handlers will be removed.
     - A new handler will be installed with our custom formatter. It will be configured to emit logs at the given level
       (default: DEBUG) or higher, to the specified stream (default: sys.stderr).
```

Note that the level is set on the handler, we leave the logging level as it was.

> we should also configure the root logger with the right log level. if we do not set it and only set databricks logger, the user might get the wrong `DEBUG` logs if they run with info as root logger allows `DEBUG`

As documented [here](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel) Python's logging system defaults to `WARNING` and we don't touch that. When the user requests `--debug` the intent is that this _only_ applies to the `databricks.*` modules/loggers.

src/databricks/labs/lakebridge/__init__.py
m-abulazm: ok got it. I also googled a bit more and looks like `import mypkg` vs `import src.mypkg` will cause a module to run twice. ideally this should not happen otherwise it would be a bug

src/databricks/labs/lakebridge/install.py
m-abulazm: `App#__call__()` configures the databricks logger which gets called in `cli.py`

m-abulazm: meaning that we could leverage that instead of adding our own method

Issue #1438, opened 2025-02-04
[BUG]: Remorph Reconcile fails to run on serverless compute


ApoorvaD13:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Category of Bug / Issue

ReconcileError

### Current Behavior

<img width="984" alt="Image" src="https://github.com/user-attachments/assets/209185da-4458-4ef9-814e-e8017e4eb925" />

<img width="1019" alt="Image" src="https://github.com/user-attachments/assets/44166a7c-e456-4ee4-88b3-6c4adf884c94" />

### Expected Behavior

It gives the error as persist is not allowed on serverless compute

### Steps To Reproduce

Trigger reconcile using serverless compute 

### Relevant log output or Exception details

```shell
[NOT_SUPPORTED_WITH_SERVERLESS] PERSIST TABLE is not supported on serverless compute. SQLSTATE: 0A000
File <command-971429592608259>, line 2
      1 ws= WorkspaceClient(product="remorph", product_version=1)
----> 2 result = recon(ws, spark, table_recon, reconcile_config)
      3 print(f" Success : {result.recon_id}")
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-bf43dd66-28fd-401e-8b50-105249c81cb0/lib/python3.10/site-packages/databricks/labs/remorph/reconcile/execute.py:294, in recon(ws, spark, table_recon, reconcile_config, local_test_run)
    291         logger.warning("Schema comparison is completed.")
    293     if report_type in {"data", "row", "all"}:
--> 294         data_reconcile_output = _run_reconcile_data(
    295             reconciler=reconciler, table_conf=table_conf, src_schema=src_schema, tgt_schema=tgt_schema
    296         )
    297         logger.warning(f"Reconciliation for '{report_type}' report completed.")
    299 recon_process_duration.end_ts = str(datetime.now())
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-bf43dd66-28fd-401e-8b50-105249c81cb0/lib/python3.10/site-packages/databricks/labs/remorph/reconcile/execute.py:933, in _run_reconcile_data(reconciler, table_conf, src_schema, tgt_schema)
    926 def _run_reconcile_data(
    927         reconciler: Reconciliation,
    928         table_conf: Table,
    929         src_schema: list[Schema],
    930         tgt_schema: list[Schema],
    931 ) -> DataReconcileOutput:
    932     try:
[...]

PR #2218, opened 2026-01-07
Fix serverless compatibility by implementing conditional DataFrame caching

BesikiML:
## 🐛 Problem

Reconciliation fails on Databricks serverless compute with:
[NOT_SUPPORTED_WITH_SERVERLESS] PERSIST TABLE is not supported on serverless compute

## ✅ Solution

Implemented capability-based caching detection that tests if caching actually works:

### Changes Made

1. **Added `is_serverless` property** - Tests caching by attempting to cache an empty DataFrame
2. **Conditional caching** - Only caches if `is_serverless` is `False`
    - Classic clusters| Spark connect: → DataFrame is cached
    - Serverless: → write to delta

Fixed issue: https://github.com/databrickslabs/lakebridge/issues/1438

### Tests

- [x] manually tested on Databricks serverless compute
- [x] manually tested on classic cluster
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/docs/lakebridge/docs/reconcile/index.mdx
+++ b/docs/lakebridge/docs/reconcile/index.mdx
@@ -7,10 +7,7 @@ import TabItem from '@theme/TabItem';
 
 # Reconcile Guide
 
-Reconcile is an automated tool designed to streamline the reconciliation process between source data and target data
-residing on Databricks. Currently, the platform exclusively offers support for Snowflake, Oracle and other Databricks
-tables as the primary data source. This tool empowers users to efficiently identify discrepancies and variations in data
-when comparing the source with the Databricks target.
+This tool empowers users to efficiently identify discrepancies and variations in data when comparing the source with the Databricks target.
 
 
 ### Execution Pre-Set Up
--- a/src/databricks/labs/lakebridge/reconcile/execute.py
+++ b/src/databricks/labs/lakebridge/reconcile/execute.py
@@ -8,7 +8,6 @@ from databricks.sdk import WorkspaceClient
 
 from databricks.labs.lakebridge import initialize_logging
 from databricks.labs.lakebridge.config import ReconcileConfig, TableRecon
-from databricks.labs.lakebridge.reconcile.exception import ReconciliationException
 from databricks.labs.lakebridge.reconcile.recon_config import AGG_RECONCILE_OPERATION_NAME, RECONCILE_OPERATION_NAME
 from databricks.labs.lakebridge.reconcile.trigger_recon_aggregate_service import TriggerReconAggregateService
 from databricks.labs.lakebridge.reconcile.trigger_recon_service import TriggerReconService
@@ -58,18 +57,13 @@ def _trigger_recon(
     table_recon: TableRecon,
     reconcile_config: ReconcileConfig,
 ):
-    try:
-        recon_output = TriggerReconService.trigger_recon(
-            ws=w,
-            spark=DatabricksSession.builder.getOrCreate(),
-            table_recon=table_recon,
-            reconcile_config=reconcile_config,
-        )
-        logger.info(f"recon_output: {recon_output}")
-        logger.info(f"recon_id: {recon_output.recon_id}")
-    except ReconciliationException as e:
-        logger.error(f"Error while running recon: {e.reconcile_output}")
-        raise e
+    recon_output = TriggerReconService.trigger_recon(
+        ws=w,
+        spark=DatabricksSession.builder.getOrCreate(),
+        table_recon=table_recon,
+        reconcile_config=reconcile_config,
+    )
+    logger.info(f"Output: {recon_output}")
 
 
 def _trigger_reconcile_aggregates(
@@ -94,19 +88,14 @@ def _trigger_reconcile_aggregates(
     - ReconciliationException: If an error occurs during the reconciliation process, it is caught and re-raised
       after logging the error details.
     """
-    try:
-        reconcile_config.report_type = "aggregate"
-        recon_output = TriggerReconAggregateService.trigger_recon_aggregates(
-            ws=ws,
-            spark=DatabricksSession.builder.getOrCreate(),
-            table_recon=table_recon,
-            reconcile_config=reconcile_config,
-        )
-        logger.info(f"recon_output: {recon_output}")
-        logger.info(f"recon_id: {recon_output.recon_id}")
-    except ReconciliationException as e:
-        logger.error(f"Error while running aggregate reconcile: {str(e)}")
-        raise e
+    reconcile_config.report_type = "aggregate"
+    recon_output = TriggerReconAggregateService.trigger_recon_aggregates(
+        ws=ws,
+        spark=DatabricksSession.builder.getOrCreate(),
+        table_recon=table_recon,
+        reconcile_config=reconcile_config,
+    )
+    logger.info(f"Output: {recon_output}")
 
 
 if __name__ == "__main__":
--- a/src/databricks/labs/lakebridge/reconcile/trigger_recon_aggregate_service.py
+++ b/src/databricks/labs/lakebridge/reconcile/trigger_recon_aggregate_service.py
@@ -11,11 +11,12 @@ from databricks.labs.lakebridge.reconcile.recon_capture import (
     generate_final_reconcile_aggregate_output,
     ReconCapture,
 )
-from databricks.labs.lakebridge.reconcile.recon_config import AGG_RECONCILE_OPERATION_NAME, Table
+from databricks.labs.lakebridge.reconcile.recon_config import Table
 from databricks.labs.lakebridge.reconcile.recon_output_config import (
     ReconcileProcessDuration,
     AggregateQueryOutput,
     DataReconcileOutput,
+    ReconcileOutput,
 )
 from databricks.labs.lakebridge.reconcile.reconciliation import Reconciliation
 from databricks.labs.lakebridge.reconcile.trigger_recon_service import TriggerReconService
@@ -31,7 +32,7 @@ class TriggerReconAggregateService:
         table_recon: TableRecon,
         reconcile_config: ReconcileConfig,
         local_test_run: bool = False,
-    ):
+    ) -> ReconcileOutput:
         reconciler, recon_capture = TriggerReconService.create_recon_dependencies(
             ws, spark, reconcile_config, local_test_run
         )
@@ -49,7 +50,7 @@ class TriggerReconAggregateService:
                     metadata_config=reconcile_config.metadata_config,
                     local_test_run=local_test_run,
                 ),
-                operation_name=AGG_RECONCILE_OPERATION_NAME,
+                report_type="aggregate",
             )
         finally:
             try:
@@ -60,7 +61,7 @@ class TriggerReconAggregateService:
     @staticmethod
     def recon_aggregate_one(
         reconciler: Reconciliation, table_conf: Table, reconcile_config: ReconcileConfig, recon_capture: ReconCapture
-    ):
+    ) -> None:
         normalized_table_conf = NormalizeReconConfigService(
             reconciler.source, reconciler.target
         ).normalize_recon_table_config(table_conf)
--- a/src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
+++ b/src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
@@ -22,6 +22,7 @@ from databricks.labs.lakebridge.reconcile.recon_output_config import (
     ReconcileProcessDuration,
     SchemaReconcileOutput,
     DataReconcileOutput,
+    ReconcileTableOutput,
 )
 from databricks.labs.lakebridge.reconcile.reconciliation import Reconciliation
 from databricks.labs.lakebridge.reconcile.schema_compare import SchemaCompare
@@ -57,7 +58,8 @@ class TriggerReconService:
                     spark=spark,
                     metadata_config=reconcile_config.metadata_config,
                     local_test_run=local_test_run,
-                )
+                ),
+                reconcile_config.report_type,
             )
         finally:
             try:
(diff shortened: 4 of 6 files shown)


src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -69,6 +71,62 @@ def __init__(
         self._source_engine = source_engine
         self._spark = spark
         self._metadata_config = metadata_config
+        self._use_serverless = self._is_serverless()
+
+    def _is_serverless(self) -> bool:
+        """Detect if running on serverless compute"""
+        try:
+            # Try to get compute type from Spark conf
+            compute_type = self._spark.conf.get("spark.databricks.clusterUsageTags.clusterType", "")
```
m-abulazm: can you link the documentation of this property?

BesikiML: `spark.databricks.clusterUsageTags.clusterType` is an internal Databricks metadata tag used to identify the compute type. It's not officially documented in public Databricks docs.

m-abulazm:
```suggestion
            compute_type = self._spark.conf.get("spark.databricks.clusterUsageTags.clusterNodeType", "")
```

It is possible to find out what is there by running `spark.conf.getAll` in a databricks notebook

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -69,6 +71,62 @@ def __init__(
...
+            if compute_type is None:
+                compute_type = ""
+            return "serverless" in compute_type.lower()
+        except (AnalysisException, AttributeError, KeyError, RuntimeError):
+            # If detection fails (Spark config unavailable or invalid), assume serverless for safety
+            logger.warning("Unable to detect compute type, assuming serverless mode")
+            return True
```
m-abulazm:
```suggestion
        try:
            compute_type = self._spark.conf.get("spark.databricks.clusterUsageTags.clusterNodeType")
        except (AnalysisException, SparkNoSuchElementException):
            compute_type = "unknown"
        return "standard" not in compute_type.lower()
```

m-abulazm: I investigated this a bit on a standard cluster:
`spark.conf.get("spark.databricks.clusterUsageTags.clusterNodeType")` returns 'Standard_D8ads_v6'

on a serverless cluster it does not work, that is why on errors we can assume serverless

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: For serveless:
```
Total configs: 3

spark.databricks.execution.timeout = 9000
spark.sql.ansi.enabled = true
spark.sql.shuffle.partitions = auto
```

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -69,6 +70,31 @@ def __init__(
...
+            logger.debug(f"Detected classic cluster (node type: {node_type})")
+            return False
+        except AnalysisException:
+            # CONFIG_NOT_AVAILABLE on serverless
+            logger.debug("Detected serverless compute (clusterNodeType not available)")
+            return True
+        except AttributeError as e:
+            # Spark context not available - assume serverless for safety
+            logger.warning(f"Unable to detect compute type: {e}. Defaulting to serverless mode")
+            return True
```
m-abulazm:
```suggestion
        except (AnalysisException, SparkNoSuchElementException, AttributeError) as e:
            logger.warning(f"Unable to detect compute type. Assuming serverless cluster", exc_info=e)
            return True
```

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -69,6 +70,31 @@ def __init__(
         self._source_engine = source_engine
         self._spark = spark
         self._metadata_config = metadata_config
+        self._use_serverless = self._is_serverless()
```
m-abulazm: I would remove this and just use `@cached_property` on the function directly. it will achieve the same result

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -69,6 +70,31 @@ def __init__(
...
+
+    def _is_serverless(self) -> bool:
```
m-abulazm:
```suggestion
    @cached_property
    def _is_serverless(self) -> bool:
```

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -370,7 +397,12 @@ def _get_mismatch_data(
 
         # Uses pre-calculated `mismatch_count` from `reconcile_output.mismatch_count` to avoid from recomputing `mismatch` for RandomSampler.
         mismatch_sampler = SamplerFactory.get_sampler(sampling_options)
-        df = mismatch_sampler.sample(mismatch, mismatch_count, key_columns, sampling_model_target).cache()
+        sampled_df = mismatch_sampler.sample(mismatch, mismatch_count, key_columns, sampling_model_target)
+
+        if self._use_serverless:
+            df = sampled_df
+        else:
+            df = sampled_df.cache()
```
m-abulazm:
```suggestion
        if not self._is_serverless:
            sampled_df = sampled_df.cache()
```

src/databricks/labs/lakebridge/reconcile/reconciliation.py
m-abulazm: we should add a integration test for this method. we use https://github.com/databrickslabs/pytester to test against our sandbox environment

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: Done

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: done

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: done

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: let's discuss about an integration test

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -79,6 +79,22 @@ def write_and_read_unmatched_df_with_volumes(
             raise ReadAndWriteWithVolumeException(message) from e
 
 
+def classify_spark_runtime(spark):
```
m-abulazm: please add type annotation


```suggestion
SparkRuntimeType = Literal["DATABRICKS_SERVERLESS",  "CLASSIC" , .....]

def classify_spark_runtime(spark: SparkSession) -> SparkRuntimeType: 
```

tests/integration/reconcile/test_recon_capture.py
```diff
@@ -1011,3 +1012,7 @@ def test_apply_threshold_for_only_threshold_mismatch_with_true_absolute(
     remorph_recon_metrics_df = spark.sql(f"select * from {recon_metadata.schema}.metrics")
     row = remorph_recon_metrics_df.collect()[0]
     assert row.run_metrics.status is True
+
+
+def test_classify_spark_runtime(spark):
+    assert classify_spark_runtime(spark) != "DATABRICKS_SERVERLESS"
```
m-abulazm: please add a comment that our sandbox environment uses spark connect so in case this changes in the future, the comment explains why it was done in this way originally

src/databricks/labs/lakebridge/reconcile/recon_capture.py
BesikiML: done

tests/integration/reconcile/test_recon_capture.py
BesikiML: done

src/databricks/labs/lakebridge/reconcile/compare.py
```diff
@@ -452,7 +453,10 @@ def join_aggregate_data(
     joined_df = df.select(*normalized_joined_cols)
 
     # Write the joined df to volume path
-    joined_volume_df = ReconIntermediatePersist(spark, path).write_and_read_unmatched_df_with_volumes(joined_df).cache()
+    joined_volume_df = ReconIntermediatePersist(spark, path).write_and_read_unmatched_df_with_volumes(joined_df)
+    cluster_type = classify_spark_runtime(spark)
```
m-abulazm: this method is used inside a loop and we should not calculate this every time. I would either cache this value or calculate at the caller before any loops and pass it here

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -79,6 +80,25 @@ def write_and_read_unmatched_df_with_volumes(
             raise ReadAndWriteWithVolumeException(message) from e
 
 
+SparkRuntimeType = Literal["DATABRICKS_SERVERLESS", "CLASSIC", "SPARK_CONNECT", "NO_JVM_UNKNOWN"]
+
+
+def classify_spark_runtime(spark) -> SparkRuntimeType:
```
m-abulazm:
```suggestion
def classify_spark_runtime(spark: SparkSession) -> SparkRuntimeType:
```

src/databricks/labs/lakebridge/reconcile/compare.py
BesikiML: We can create `ReconIntermediatePersist(spark, path) `before the loop and pass it in, but since joined_df changes with the loop parameters, how can we cache it ahead of time?

m-abulazm: I do think a `ReconIntermediatePersist#cache_if_supported()` method is a good idea

m-abulazm: another question though; why are we caching here after writing to storage? what is the improvement? otherwise we can just remove it from here

src/databricks/labs/lakebridge/reconcile/recon_capture.py
BesikiML: done

src/databricks/labs/lakebridge/reconcile/compare.py
BesikiML: This (cache) approach was already there; I didn’t add it.

BesikiML: done

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -79,6 +80,33 @@ def write_and_read_unmatched_df_with_volumes(
             raise ReadAndWriteWithVolumeException(message) from e
 
 
+SparkRuntimeType = Literal["DATABRICKS_SERVERLESS", "CLASSIC", "SPARK_CONNECT", "NO_JVM_UNKNOWN"]
+
+
+@lru_cache(maxsize=1)
+def _classify_spark_runtime(spark: SparkSession) -> SparkRuntimeType:
```
m-abulazm: didn't we have a test for this?

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -79,6 +80,33 @@ def write_and_read_unmatched_df_with_volumes(
...
+        if "serverless" in msg:
+            return "DATABRICKS_SERVERLESS"
+
+        if "spark connect" in msg:
+            return "SPARK_CONNECT"
+
+        return "NO_JVM_UNKNOWN"
+
+
+def cache_df_if_supported(spark: SparkSession, df: DataFrame) -> DataFrame:
```
m-abulazm: please also add a test for this. you can use `assert df.is_cached`

src/databricks/labs/lakebridge/reconcile/recon_capture.py
BesikiML: it is a private function

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -69,6 +75,10 @@ def _is_databricks(self) -> bool:
     def base_dir(self) -> Path:
         return Path(self._base_dir)
 
+    @property
+    def can_cache(self) -> bool:
+        return self._can_cache
```
m-abulazm:
```suggestion
    @cached_property
    def is_cache_supported(self) -> bool:
        try:
            schema = StructType([StructField("col1", StringType(), True)])
            df = self._spark.createDataFrame([], schema)
            df.cache()
            return True
        except (AnalysisException, PySparkAttributeError):
            return False
```

m-abulazm: it is a common practice to name functions that return bool with `is_`. I would recommend sticking with known practices. and I would remove all the indirections by just implementing the function here.

tests/integration/reconcile/test_recon_capture.py
```diff
@@ -1003,6 +1003,10 @@ def is_databricks(self) -> bool:
     def format(self):
         return self._format
 
+    @property
+    def try_cache(self) -> bool:
+        return self._try_cache()
```
m-abulazm: you already have a public member. you do not need to do this. 

the other methods are a workaround for testing private members.

BesikiML: done

src/databricks/labs/lakebridge/reconcile/recon_capture.py
BesikiML: done

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -368,7 +370,8 @@ def _get_mismatch_data(
         # Uses pre-calculated `mismatch_count` from `reconcile_output.mismatch_count` to avoid from recomputing `mismatch` for RandomSampler.
         mismatch_sampler = SamplerFactory.get_sampler(sampling_options)
         df = mismatch_sampler.sample(mismatch, mismatch_count, key_columns, sampling_model_target)
-        # TODO write `df` to delta
+        if self.intermediate_persist.is_cache_supported:
+            df = df.cache()
```
m-abulazm: we need an else here which the original TODO was about.

```
else:
    df = intermediate_persist.write_and_read_df_with_volumes(df)
```

BesikiML: done

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -69,6 +74,16 @@ def _is_databricks(self) -> bool:
     def base_dir(self) -> Path:
         return Path(self._base_dir)
 
+    @cached_property
+    def is_cache_supported(self) -> bool:
+        try:
```
sundarshankar89: checking internally if there is a simpler and safer way to check for this. will get back soon.

PR #2223, opened 2026-01-09
Debug and fix acceptance runs

m-abulazm:
* Fix recon tests after `sqlglot` upgrade
   * pipe operator is used in hash queries so test fixtures were updated to expect the same
* use different spark schemas for different `reconcile` tests to get rid of spark write errors about existing schema
* fix e2e recon test; we didnt see the issues as it did not run in the CI env before
* delete `pytest` modifier that excluded `reconcile` tests and run all integration tests in acceptance CI

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -45,6 +45,7 @@ jobs:
           chmod +x $GITHUB_WORKSPACE/.github/scripts/setup_mssql_odbc.sh
           $GITHUB_WORKSPACE/.github/scripts/setup_mssql_odbc.sh
 
+      # TODO: Migrate tests to use Databricks clusters instead of Spark local mode
       - name: Setup spark
         run: |
           chmod +x $GITHUB_WORKSPACE/.github/scripts/setup_spark_remote.sh


tests/integration/reconcile/test_recon_databricks.py
```diff
@@ -88,7 +90,7 @@ def application_context(
 ):
     logger.info("Setting up application context for recon tests")
     config = LakebridgeConfiguration(None, recon_config)
-    ctx = ApplicationContext(ws).replace(product_info=ProductInfo.for_testing(type(config)))
+    ctx = ApplicationContext(ws)
```
m-abulazm, line 93: reconcile does not support other product names than `lakebridge`

.github/workflows/acceptance.yml
```diff
@@ -45,6 +45,11 @@ jobs:
           chmod +x $GITHUB_WORKSPACE/.github/scripts/setup_mssql_odbc.sh
           $GITHUB_WORKSPACE/.github/scripts/setup_mssql_odbc.sh
 
+      - name: Setup spark
```
sundarshankar89, line 49: we can iterate on this, now that we are running within acceptance, we can use remote spark connect to actual workspace.

sundarshankar89: Can you file a feature request to iterate on this.

.codegen.json
```diff
@@ -5,6 +5,7 @@
   "toolchain": {
     "required": ["hatch"],
     "pre_setup": ["hatch env create"],
-    "prepend_path": ".venv/bin"
+    "prepend_path": ".venv/bin",
+    "acceptance_path": "tests/integration"
```
sundarshankar89, line 9: Do we need this now?

m-abulazm: yes to run the tests under `integration` only not everything. previously the pytest modifier took care of that

.github/workflows/acceptance.yml
m-abulazm: done https://github.com/databrickslabs/lakebridge/issues/2229

gueniai (approved):
LGTM

Issue #1905, opened 2025-08-06
Refactor Utils inside reconcile package


sundarshankar89:
Can we reimagine this. utils like this will become a place holder for undefined scope with static functions. which is difficult to manage in future.

_Originally posted by @sundarshankar89 in https://github.com/databrickslabs/lakebridge/pull/1892#discussion_r2253105044_

PR #2230, opened 2026-01-15
Refactor recon intermediate persistence

### LINKED ISSUE #1056: [TODO] for now we are overwriting the intermediate cache path. We should delete the volume in future
See: https://github.com/databrickslabs/remorph/blob/661b68471a043d68b673fbca50b31684fa447310/src/databricks/labs/remorph/reconcile/recon_capture.py#L57-L62

m-abulazm:
### What does this PR do?
* Clean checkpoint volume after reconcile runs
* Remove usage of `overwrite` write mode 
* use delta volumes if running on databricks instead of parquet
* Refactor interface to hide implementation details and be more generic
* add TODO to mark where we need to persist to delta instead of hitting source system

On Serverless we cannot use cache/persist, so we use Delta writes acting as materialization boundaries

### Linked issues

Resolves #1056
Advances #1905, #1438 

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [X] added integration tests


src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -38,44 +39,66 @@
...
+        self._spark = spark
+        self._metadata_config = metadata_config
+        self._format = "delta" if self._is_databricks else "parquet"
+        self._base_dir = self._get_uc_volume_path if self._is_databricks else tempfile.gettempdir()
+
+    @cached_property
+    def _is_databricks(self) -> bool:
+        is_db = any(k.startswith("spark.databricks") for k in self._spark.conf.getAll.keys())
+        logger.info(f"Running on Databricks check completed with result: {is_db}")
+        return is_db
```
BesikiML: I would create this method outside the class.

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -38,44 +39,66 @@
...
```
BesikiML, line 66: I would create this method outside the class.
```suggestion
@lru_cache(maxsize=1)
def is_databricks(spark: SparkSession) -> bool:
   is_db = any(k.startswith("spark.databricks") for k in spark.conf.getAll.keys())
   logger.info(f"Running on Databricks check completed with result: {is_db}")
   return is_db
```

sundarshankar89: you use env variable for this in all `DATABRICKS_RUNTIME_VERSION` this variable will be set on all databricks instance both serverless as well as classic

sundarshankar89: reference implementation https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/runtime.py

m-abulazm: I prefer `cached_property`. we can move this method if we need it in more places. no need for premature optimization

BesikiML (approved):
LGTM

PR #2235, opened 2026-01-20
Adjust project metadata for supported python versions

### LINKED ISSUE #2234: [BUG]: Transpilers fail to install with python 3.10.0
### Category of Bug / Issue

Application crashed

### Current Behavior

Installing the transpilers using python 3.10.0 will fail because the installer relies on a python attribute that was introduced in python 3.10.1.

### Expected Behavior

Our documentation (and `databricks labs` metadata) lists Python 3.10 as supported, so this should work.

The easiest fix is to update our documentation and 

### Steps To Reproduce

Analysis was based on python sources. This hasn't been tested, but to reproduce the bug (and verify the solution works) this should work:

1. Install Python 3.10.0.
2. Modify `PATH` to ensure that no higher version of Python is available.
3. Run `databricks labs install lakebridge --debug` and verify in the logs that it chose to create the venv with Python 3.10.0.
4. Run `databricks labs lakebridge install-transpile --debug`

```shell
File "C:\Users\xxx\.databricks\labs\lakebridge\lib\src\databricks\labs\lakebridge\transpiler\installers.py", line 238, in _ensure_pip
    install_pip_cmd = [venv_context.env_exec_cmd, "-m", "ensurepip", "--upgrade"]
AttributeError: 'types.SimpleNamespace' object has no attribute 'env_exec_cmd'
```

### Operating System

Windows

### Version

latest via Databricks CLI

asnare:
### What does this PR do?

This PR updates the project metadata and documentation for the versions of python that are supported, having discovered (see #2234) that version 3.10.0 is missing some functionality that we use: it was introduced in version 3.10.1. With this in mind:

 - Previously: metadata expressed that 3.10+ was supported and the documentation matched, although the documentation noted that 3.14 isn't supported.
 - Now: metadata now expresses a minimum version of 3.10.1 and a maximum of 3.13.x, to the extent that we can.

The documentation has been adjusted accordingly.

During the normal installation flow with the Databricks CLI:

1. The Databricks CLI searches `PATH` for interpreters to use when setting up the virtual environment. Interpreters that do not meet the minimum version in `labs.yml` are excluded. (Full semantic versioning is supported in the comparison.) There is no way to specify a maximum version.
2. `pip` is used to install the package; this checks the required version in `pyproject.yml` and will halt with an error if the interpreter is 3.14+.

When setting up the virtual environment for BB, we rely on some functionality that introduced in python 3.10.1. Although not mentioned in the python documentation, the `env_exec_cmd` attribute isn't available in the virtual environment metadata in python 3.10.0. The relevant change between python 3.10.0 and 3.10.1 is [here](https://github.com/python/cpython/commit/eabca6e593269301a0b7a36c4dc3525f04f5bb36).

### Linked issues

Resolves #2234: Most installations of 3.10 will be 3.10.1 or later, so this isn't really worth working around or the accompanying maintenance/testing burden. As such these changes are really about guarding against the odd edge case.

### Functionality

- modified existing command: `databricks labs install lakebridge`

### Tests

- manually tested:

   - Failure of `install-transpile` prior to this PR with 3.10.0:
     [install-transpile-3.10.0.scrub.log](https://github.com/user-attachments/files/24763836/install-transpile-3.10.0.scrub.log)
   - New: Refusal of fresh `labs install` with Python 3.10.0:
     [install-lakebridge-3.10.0.scrub.log](https://github.com/user-attachments/files/24763842/install-lakebridge-3.10.0.scrub.log)
   - New: Refusal of upgrade via `labs install` with existing version using Python 3.10.0:
     [upgrade-lakebridge-3.10.0.scrub.log](https://github.com/user-attachments/files/24763848/upgrade-lakebridge-3.10.0.scrub.log)
   - New: Success of `labs install` and `install-transpile` with Python 3.10.1:
     [install-transpile-3.10.1.scrub.log](https://github.com/user-attachments/files/24763853/install-transpile-3.10.1.scrub.log)
   - New: Error (expected) of fresh `labs install` with Python 3.14.
     [install-lakebridge-3.14.0.scrub.log](https://github.com/user-attachments/files/24763868/install-lakebridge-3.14.0.scrub.log)

### DIFF
--- a/docs/lakebridge/docs/installation.mdx
+++ b/docs/lakebridge/docs/installation.mdx
@@ -87,7 +87,7 @@ databricks configure --host <host> --profile <profile_name>
 
 **Version Required:** Python between 3.10.1 and 3.13.x (inclusive).
 
-- **Windows** - Install python from [here](https://www.python.org/downloads/). Your Windows computer will need a shell environment ([GitBash](https://www.git-scm.com/downloads) or [WSL](https://learn.microsoft.com/en-us/windows/wsl/about))
+- **Windows** - Install python from [here](https://www.python.org/downloads/).
 - **MacOS/Unix** - Use [brew](https://formulae.brew.sh/formula/python@3.10) to install python in macOS/Unix machines
 
 **Check Python version on Windows, macOS, and Unix:**
--- a/docs/lakebridge/docs/reconcile/recon_notebook.mdx
+++ b/docs/lakebridge/docs/reconcile/recon_notebook.mdx
@@ -75,7 +75,7 @@ class ReconcileConfig:
 ```
 Parameters:
 
-- `data_source`: The data source to be reconciled. Supported values: `snowflake`, `teradata`, `oracle`, `mssql`, `synapse`, `databricks`.
+- `data_source`: The data source to be reconciled. Supported values: `snowflake`, `oracle`, `mssql`, `synapse`, `databricks`.
 - `report_type`: The type of report to be generated. Available report types are `schema`, `row`, `data` or `all`. For details check [here](./dataflow_example.mdx).
 - `secret_scope`: The secret scope name used to store the connection credentials for the source database system.
 - `database_config`: The database configuration for connecting to the source database. expects a `DatabaseConfig` object.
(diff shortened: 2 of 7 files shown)


sundarshankar89 (approved):
LGTM

PR #2237, opened 2026-01-21
Standardize Profiler Path Extract 

sundarshankar89:
### What does this PR do?
Standardize Profiler Path Extract 

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   "SQLAlchemy~=2.0.40",
   "pygls~=2.0.0",
   "duckdb~=1.2.2",
-  "databricks-switch-plugin~=0.1.5",  # Temporary, until Switch is migrated to be a transpiler (LSP) plugin.
+  "databricks-switch-plugin~=0.1.6",  # Temporary, until Switch is migrated to be a transpiler (LSP) plugin.
   "requests>=2.28.1,<3"  # Matches databricks-sdk (and 'types-requests' below), to avoid conflicts.
 ]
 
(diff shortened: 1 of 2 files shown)


src/databricks/labs/lakebridge/resources/assessments/synapse/pipeline_config.yml
```diff
@@ -1,7 +1,6 @@
 name: synapse_assessment
 version: "1.0"
-# TODO: This needs to be removed.
-extract_folder: "/tmp/data/synapse_assessment"
+extract_folder: "~/.databricks/labs/lakebridge_profilers/synapse_assessment"
```
goodwillpunning, line 3: FYI we'll need to add the run Id as a path prefix in a follow-up PR.

sundarshankar89: why at the folder level? can we not do it at table level inside the database?

gueniai: Why don't we make the profiler output configurable by the user?

sundarshankar89: having it central will help us share state more easily across component, That was the principle behind a common central location.

gueniai (approved):

src/databricks/labs/lakebridge/resources/assessments/synapse/pipeline_config.yml
gueniai: I can see the value of that. Maybe we need to think about how to manage the state for all modules and then standardize the optionality of outputs for all.

PR #2240, opened 2026-01-21
Removed WSL as part to the installation pre-reqs.

### LINKED ISSUE #2233: [DOCS]: Confirm Windows prerequisites
### Problem statement

Currently the docs state the prerequisites for Windows as:
> Windows - Install python from [here](https://www.python.org/downloads/). Your Windows computer will need a shell environment ([GitBash](https://www.git-scm.com/downloads) or [WSL](https://learn.microsoft.com/en-us/windows/wsl/about))

Although the Python instructions are fine, it's not clear the GitBash/WSL thing is correct: we don't test either of those, and the preferred shell is probably just _Command Prompt_.

We should instead look to the [documentation for the Databricks CLI](https://docs.databricks.com/aws/en/dev-tools/cli/install#install) as inspiration.

### Additional Context

The issues with suggesting GitBash or WSL are:

 - GitBash: this seems counterproductive. There are no advantages to this compared to Command Prompt, and unnecessarily suggesting that users install git seems like the start of a "now you have $x + 1$ problems" situation; 
 - WSL: I'm sceptical about whether this actually works, in particular due to path translation issues. Running under WSL means there's a good chance that WSL-only paths will be passed to Windows (non-WSL) binaries (`java`, `analyze`, `dbxconv`) and these won't be recognised.

andresgarciaf:
### What does this PR do?
Removes WSL as an option to install LakeBridge.

### Related Issues

Resolves #2233

### Functionality

- [x] updated relevant user documentation

### Tests

- [x] manually tested

### DIFF
--- a/docs/lakebridge/docs/installation.mdx
+++ b/docs/lakebridge/docs/installation.mdx
@@ -87,8 +87,7 @@ databricks configure --host <host> --profile <profile_name>
 
 **Version Required:** Python 3.10 or newer; verify that the installed version is in the supported versions (3.10–3.13).
 
-- **Windows** - Install python from [here](https://www.python.org/downloads/). Your Windows computer will need a shell environment (CMD, PowerShell or [GitBash](https://www.git-scm.com/downloads)).
-- **Linux** - Use your package manager to install python (e.g.
+- **Windows** - Install python from [here](https://www.python.org/downloads/).
 - **MacOS/Unix** - Use [brew](https://formulae.brew.sh/formula/python@3.10) to install python in macOS/Unix machines
 
 **Check Python version on Windows, macOS, and Unix:**


docs/lakebridge/docs/installation.mdx
```diff
@@ -87,7 +87,8 @@ databricks configure --host <host> --profile <profile_name>
 
 **Version Required:** Python 3.10 or newer; verify that the installed version is in the supported versions (3.10–3.13).
 
-- **Windows** - Install python from [here](https://www.python.org/downloads/). Your Windows computer will need a shell environment ([GitBash](https://www.git-scm.com/downloads) or [WSL](https://learn.microsoft.com/en-us/windows/wsl/about))
+- **Windows** - Install python from [here](https://www.python.org/downloads/). Your Windows computer will need a shell environment (CMD, PowerShell or [GitBash](https://www.git-scm.com/downloads)).
```
asnare: I've just linked #2233 to this PR. Have a quick look, I think we should also drop the reference to GitBash.

andresgarciaf: @asnare, removing it now.

andresgarciaf: Keeping Just - **Windows** - Install python from [here](https://www.python.org/downloads/). I think is redundant to say a terminal is needed if on the previous steps show how to install the CLI from a terminal.

gueniai (approved):
LGTM

PR #2244, opened 2026-01-22
Remove Pylint and Configure Ruff

sundarshankar89:
### What does this PR do?
Remove Pylint and Configure Ruff to run pylint rules as documented here: https://docs.astral.sh/ruff/rules/#pylint-pl
https://docs.astral.sh/ruff/rules/#pylint-pl

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/assessments/dashboards/dashboard_manager.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: PLR0911
```
sundarshankar89, line 1: too-many-return-statements.

src/databricks/labs/lakebridge/assessments/dashboards/execute.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: PLR2004
```
sundarshankar89, line 1: magic-value-comparison suppression

sundarshankar89:
> I would fix the problems bit by bit although it will take time instead of suppressing

Objective was to retain exisitng supression.

sundarshankar89:
closing this will raise a new cleaner one to implement this change in much more cleaner way.

PR #2252, opened 2026-01-28
Implement retargeting `workspace/applyEdit` for future conversion output

asnare:
### What does this PR do?

This PR follows on from #2201 by implementing the retargeting that we need for `workspace/applyEdit` so converted files are written to a separate location to the input.

This has been implemented by setting up an editor chain, where each modifies the edits before passing them on:

1. `SandboxEditor`, which rejects edits outside the input directory;
2. `RetargetingEditor`, which modifies the edits so that they apply to the destination directory instead of the input;
3. `LakebridgeEditor`, which implements the subset we need for modifying files.

These can be tested independently. We also test the exact combinations of events that are required by the converters.

### Linked issues

Follows #2201.

### Tests

- added unit tests


sundarshankar89 (approved):
LGTM

gueniai (approved):
LGTM

PR #2254, opened 2026-01-28
Display real-time transpiler debug output in LSP client

simone-dbx-labs:
## Summary
- Register handler for `window/logMessage` notifications from bladebridge LSP server
- Set `PYTHONUNBUFFERED=1` to ensure subprocess output streams immediately
- Automatically configure `CONVERTER_KEY_FILE` environment variable for bladebridge

## Test plan
- [x] Verified real-time progress messages appear during transpilation
- [x] Tested with and without --debug flag
- [x] Confirmed messages stream line-by-line instead of buffering


simone-dbx-labs:
the real-time output works when providing --debug option.  without that we don't get any output at all.
```
databricks labs lakebridge transpile \              
  --source-dialect "informatica (desktop edition)" \
  --input-source wf_demo_large_mapping_load.XML \
  --output-folder ./out \     
  --target-technology SPARKSQL \
  --debug
```

PR #2259, opened 2026-02-02
Improve reconciliation result handling and logging

m-abulazm:
### What does this PR do?
Raise `ReconciliationException` when any table reconciliation produced an exception message. Otherwise log a warning when there are reconciliation mismatches and an info message when reconciliation passes according to the configured checks.
The log message includes the `report_type` to make it easier to trace which reconciliation check completed.

### Functionality

- [ ] added new CLI command
- [X] modified existing command: `databricks labs lakebridge ...`
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests

### DIFF
--- a/tests/unit/reconcile/test_verify_reconcile.py
+++ b/tests/unit/reconcile/test_verify_reconcile.py
@@ -60,6 +60,6 @@ def test_raises_on_exception_message() -> None:
     reconcile_output = ReconcileOutput(recon_id="mock-id", results=results)
 
     with pytest.raises(ReconciliationException) as excinfo:
-        TriggerReconService.verify_successful_reconciliation(reconcile_output, report_type="daily")
+        TriggerReconService.verify_successful_reconciliation(reconcile_output, report_type="all")
 
-    assert "Reconciliation failed with exceptions for" in str(excinfo.value)
+    assert "Reconciliation **all** with id: mock-id failed with exceptions for" in str(excinfo.value)
--- a/src/databricks/labs/lakebridge/reconcile/trigger_recon_aggregate_service.py
+++ b/src/databricks/labs/lakebridge/reconcile/trigger_recon_aggregate_service.py
@@ -61,7 +61,7 @@ class TriggerReconAggregateService:
     @staticmethod
     def recon_aggregate_one(
         reconciler: Reconciliation, table_conf: Table, reconcile_config: ReconcileConfig, recon_capture: ReconCapture
-    ):
+    ) -> None:
         normalized_table_conf = NormalizeReconConfigService(
             reconciler.source, reconciler.target
         ).normalize_recon_table_config(table_conf)


src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
```diff
@@ -218,21 +220,32 @@ def _run_reconcile_data(
...
+            raise ReconciliationException(
+                f"Reconciliation failed for one or more tables during **{report_type}** check. "
+                "Please check the recon metrics for details.",
+                reconcile_output=reconcile_output,
+            )
+
+        mismatched = [r for r in reconcile_output.results if is_table_recon_mismatch(r)]
+        if mismatched:
+            logger.warning(
+                f"Reconciliation found mismatches in {len(mismatched)} table(s). "
```
sundarshankar89: Why is a mismatch not considered an exception or error?

m-abulazm: Because it looks like a crash and confuses the user. Hence we need to distinguish between “system is down” and “system is working and found issues”

A mismatch is treated as a warning instead of an exception because it represents a valid outcome of `reconcile`, not a technical failure of the system itself. And we let the reviewer decide whether they’re acceptable, expected, or require action

asnare: Trailing whitespace:
```suggestion
                f"Reconciliation found mismatches in {len(mismatched)} table(s)."
```

asnare: I've had more time to think about this, and I believe that `warning` is the right thing to do here: nothing has gone wrong, the purpose of reconciliation is to check for differences. The fact that it found differences is an _expected_ result, even if it signals something that the user needs to address.

I'm also okay if we log it at `error` so that it's more visible, but it's not an exception.

tests/unit/reconcile/test_verify_reconcile.py
asnare, line 1: Please ensure that all tests have test have type hints.

For example:
```python
def test_xxx() -> None:
```

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
sundarshankar89: let us settle on logging it as error not as exception

m-abulazm: changed and also added more logs to enhance user experience

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
```diff
@@ -218,21 +220,46 @@ def _run_reconcile_data(
...
                 )
 
-        logger.info("Reconciliation completed successfully.")
+            return is_mismatch
+
+        exceptions = [r for r in reconcile_output.results if r.exception_message]
+        mismatched = [r for r in reconcile_output.results if is_table_recon_mismatch(r)]
+
+        (total_count, exc_count, mismatched_count) = (len(reconcile_output.results), len(exceptions), len(mismatched))
+        success_count = min(total_count, exc_count + mismatched_count - total_count)
```
sundarshankar89: is this correct? 

  - total_count = 10
  - exc_count = 2
  - mismatched_count = 3

then it becomes min(10, -5)

sundarshankar89 (commented):
smalll nit on the formula for success_count rest look ok to me

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
m-abulazm: fixed nice catch

m-abulazm:
> one more comment I missed is we should also add recon_id to the logs and exception so that we can go and trace from the recon metadata tables.

the id and a summary are logged before any other error or info logs so the user will be able to see that

src/databricks/labs/lakebridge/reconcile/execute.py
```diff
@@ -94,19 +88,14 @@ def _trigger_reconcile_aggregates(
...
-        logger.error(f"Error while running aggregate reconcile: {str(e)}")
-        raise e
+    reconcile_config.report_type = "aggregate"
+    recon_output = TriggerReconAggregateService.trigger_recon_aggregates(
+        ws=ws,
+        spark=DatabricksSession.builder.getOrCreate(),
+        table_recon=table_recon,
+        reconcile_config=reconcile_config,
+    )
+    logger.info(f"Output: {recon_output}")
```
sundarshankar89, line 98: missing explicity recon_id log to redirect to metrics table.

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
```diff
@@ -218,21 +220,46 @@ def _run_reconcile_data(
...
+
+        logger.info(
+            f"Ran **{report_type}** reconcile with id: {reconcile_output.recon_id} for total {total_count} source tables and their targets."
+            f" {success_count} tables succeeded, {exc_count} tables failed with exceptions and {mismatched_count} tables mismatched."
+        )
+
+        if exceptions:
+            raise ReconciliationException(
+                f"Reconciliation failed with exceptions for {exc_count} table(s). Please check recon metrics for details.",
+                reconcile_output=reconcile_output,
```
sundarshankar89, line 253: I will add explicit recon id here as well.

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
```diff
@@ -218,21 +220,46 @@ def _run_reconcile_data(
...
+            )
+
+        if mismatched:
+            logger.error(
+                f"Reconciliation found mismatches in {mismatched_count} table(s). Please check recon metrics for details."
```
sundarshankar89: I will add explicit recon id here as well.

src/databricks/labs/lakebridge/reconcile/trigger_recon_service.py
```diff
@@ -218,21 +220,46 @@ def _run_reconcile_data(
...
+            )
+        else:
+            logger.info(
+                f"Reconciliation type **{report_type}** completed successfully. Please check recon metrics for details."
```
sundarshankar89: I will add explicit recon id here as well.

sundarshankar89 (commented):
nits:

src/databricks/labs/lakebridge/reconcile/execute.py
m-abulazm: it is `recon_output.recon_id`

sundarshankar89 (approved):
LGTM: Non blocking but even for succesfull run we can iterate on redirect them to dashboard link with recon id filter as enhanced user experience.

PR #2263, opened 2026-02-05
Env Secret Provider Nested Credentials

### LINKED ISSUE #2262: [FEATURE]: Improve Synapse Profiler Error Message for Incorrect Secret Vault Type

Profiler

### Problem statement

If a user misconfigures the secret vault type (e.g., sets it to `env` instead of `local` for local runs), the pipeline fails with a generic and misleading error message:

```
ERROR [d.l.l.assessments.pipeline] Step dedicated_sqlpool_extract failed with error: Script reported error: Could not parse script output as JSON, manually validate the logs.
```

This message does not indicate the root cause and sends users down the wrong debugging path.

### Proposed Solution

Update the error handling to detect this misconfiguration and return a more actionable message that explicitly instructs users to verify the configured secret vault type (`env` vs. `local`) for their execution context, rather than suggesting a generic script output issue.

### Additional Context

This issue is limited to the Synapse profiler run using the `execute-database-profiler` command

sundarshankar89:
## Summary

 Enhances the credential manager to support nested credential structures (e.g., Synapse workspaces with multiple configuration levels) while maintaining backward compatibility with existing flat credential formats. This addresses #2262 by improving credential
  resolution robustness.

## Changes
  Core Implementation `credential_manager.py`:
  - Replaced simple dictionary comprehension in get_credentials() with a new recursive _resolve_credentials() method
  - Added type-aware processing:
    - Dictionaries: Recursively processes nested key-value pairs
    - Lists: Recursively processes each list item
    - Strings: Applies secret provider resolution (env vars, local values, etc.)
    - Primitives (int, bool, float, None): Preserves values unchanged
  - Updated _get_secret_value() docstring for clarity

### What does this PR do?

 - Synapse and other data sources can use complex nested credential structures (workspace config, API access, JDBC settings, profiler options)
  - Previous implementation only handled flat key-value pairs at a single level
  - Env secret provider now correctly resolves environment variables within nested structures and lists

### Test Coverage Improvements

  Enhaced the test suite added `test_credential_manager.py`:
  - Session-scoped environment setup
  - Real secret providers: Tests use actual LocalSecretProvider, EnvSecretProvider, DatabricksSecretProvider instances (no mocks)
  

### Linked issues

Resolves #..
Fixes #2262 

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [X] manually tested
- [X] added unit tests
- [X] added integration tests


tests/unit/connections/test_credential_manager.py
```diff
@@ -40,51 +80,298 @@ def env_credentials():
...
-    env_getter.get.side_effect = lambda key: os.environ[key]
-    credentials = create_credential_manager(product_name, env_getter)
-    creds = credentials.get_credentials('mssql')
-    assert creds['user'] == 'env_user'
-    assert creds['password'] == 'env_password'
+def test_nested_dict_credentials_local_vault():
+    """Test Synapse-style nested dictionaries with local secret vault ."""
+    credentials = {
+        'secret_vault_type': 'local',
+        'synapse': {
```
m-abulazm, line 143: what is workspace, azure_api_access, jdbc, profiler? is this standard synapse config that is document on azure? are we using a client that handles those parameters?

sundarshankar89: this is just an example of nested strucutre, but these are the some of them are the necessary inputs needed for actually connecting and extracting information.

src/databricks/labs/lakebridge/connections/credential_manager.py
```diff
@@ -53,9 +53,30 @@ def get_credentials(self, source: str) -> dict:
         if not isinstance(value, dict):
             raise KeyError(f"Invalid credential format for source: {source}")
 
-        return {k: self._get_secret_value(v) for k, v in value.items()}
+        return self._resolve_credentials(value)
+
+    def _resolve_credentials(self, value: Any) -> Any:
```
m-abulazm: we can use `JsonValue` from blueprint

tests/unit/connections/test_credential_manager.py
```diff
@@ -1,35 +1,75 @@
...
+    creds = manager.get_credentials('mssql')
 
-@pytest.fixture
-def env_credentials():
-    return {
+    assert creds['user'] == 'local_user'
+    assert creds['password'] == 'local_password'
+    assert creds['database'] == 'DB_NAME'
+    assert creds['port'] == 1433
+    assert isinstance(creds['port'], int)
```
m-abulazm: isnt the first line just enough?

sundarshankar89: I have adjusted

src/databricks/labs/lakebridge/connections/credential_manager.py
sundarshankar89: this recursive credentials sure we can't given the way we handle the nested structure,  I have instead used `dict[str, Any]` 
Let me know if suits.

m-abulazm (approved):
LGTM

PR #2264, opened 2026-02-05
added type hints and reinstated linter for mypy on resources folder

sundarshankar89:
### What does this PR do?

This PR introduces type hints and enables type checking (mypy) for the `resources/assessments/synapse` folder while keeping pylint disabled for that folder to avoid code complexity warnings that would require significant refactoring.

**Bug Fixes:**
- Fixed incorrect Azure SDK class name: `ArtifactClient` → `ArtifactsClient`
- Made type hints more flexible to handle Azure SDK overloaded functions (using `Any` for overloaded function parameters)

**Code Quality Improvements:**
- Prefixed intentionally unused function parameters with underscore
- Renamed variables to comply with naming conventions:
  - `dedicated_sql_pools_profiling_list` → `dedicated_pools_list`
  - `serverless_database_groups_in_scope` → `serverless_db_groups_in_scope`
  - `db` → `database`

**Configuration Changes:**
- Added `pyproject.toml` optional dependency group with Azure SDK packages:
  - `azure-identity~=1.19.0`
  - `azure-monitor-query~=1.4.0`
  - `azure-synapse-artifacts~=0.20.0`
- Removed `src/databricks/labs/lakebridge/resources/assessments` from mypy exclude list
- Added `src/databricks/labs/lakebridge/resources/assessments` to pylint ignore-paths

- The `resources/assessments` folder is now type-checked by mypy but excluded from pylint
- Remaining pylint warnings in that folder (too many variables, too complex, etc.) would require significant refactoring and are intentionally left for future work
- Azure SDK optional dependencies are only needed for linting, not runtime

### Linked issues

N/A - Proactive code quality improvement

### Functionality

- [x] No functional changes - type hints and linting only
- [x] All existing tests pass

### Tests

- [x] manually tested - `make fmt` now passes with 10.00/10 score
- [x] mypy type checking passes on resources folder
- [x] no new tests needed (type hints only)


src/databricks/labs/lakebridge/resources/assessments/synapse/common/profiler_classes.py
```diff
@@ -1,40 +1,64 @@
+from abc import ABC
+from collections.abc import Callable, Generator, Iterable, MutableMapping
+from datetime import date, datetime, time as timex, timedelta, tzinfo
```
asnare: Protip:
```suggestion
import datetime as dt
```

src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
```diff
@@ -59,9 +59,9 @@ def list_columns(pool_name: str) -> str:
                """
 
     @staticmethod
-    def list_views(pool_name, redact_sql_text: bool = False) -> str:
+    def list_views(pool_name, _redact_sql_text: bool = False) -> str:
```
asnare: Maybe just drop the argument? If it's not being used, then the fact that it's present is misleading to callers?

asnare: Missed the type hint on `pool_name`.

src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
```diff
@@ -59,9 +59,9 @@ def list_columns(pool_name: str) -> str:
...
         """Get list of views"""
-        return """
+        return f"""
```
asnare, line 64: Oh wow. Does this mean it never worked? (Corollary: this has never been tested?)

src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
```diff
@@ -74,7 +74,7 @@ def list_views(pool_name, redact_sql_text: bool = False) -> str:
                """
 
     @staticmethod
-    def list_routines(pool_name, redact_sql_text: bool = False) -> str:
+    def list_routines(pool_name, _redact_sql_text: bool = False) -> str:
```
asnare: Same here.

src/databricks/labs/lakebridge/resources/assessments/synapse/serverless_sqlpool_extract.py
```diff
@@ -42,13 +42,13 @@ def get_serverless_database_groups(
     inclusion_list = inclusion_list or []
     exclusion_list = exclusion_list or []
 
-    serverless_database_groups_in_scope = {}
+    serverless_db_groups_in_scope = {}
```
asnare, line 46: I'm mildly surprised that a type hint isn't needed here…

Ah. The method definition doesn't have any hints, so this body is being ignored.

.gitignore
```diff
@@ -23,3 +23,4 @@ remorph_transpile/
 .databricks-login.json
 .mypy_cache
 .env
+/.venv-lint/
```
asnare: I don't think this is needed?

pyproject.toml
```diff
@@ -45,6 +45,14 @@ dependencies = [
   "pandas~=2.3.1", # Required for new configure assessment
 ]
 
+[project.optional-dependencies]
+# Optional dependencies for linting resources folder (assessments, profilers)
+resources-lint = [
+  "azure-identity~=1.19.0",
+  "azure-monitor-query~=1.4.0",
+  "azure-synapse-artifacts~=0.20.0",
+]
```
asnare: Shouldn't these go instead in the normal dev environment, alongside `pytest` and friends?

Then a lot of these project changes aren't needed and things remain (relatively) simple.

sundarshankar89: this bloat, never needed during regular cli, we mange them usign custom `venv` when pipeline.py runs. 

and I don't want this to be part of core rather actu as pluggable scripts pipeline.py will be ale to use

src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
sundarshankar89: we didn't run into multiple pools yet.

pyproject.toml
asnare: `pytest` et al. are not part of the regular CLI: they're part of the developer setup, which is also what this is.

(The CLI itself uses the _package_ dependencies, not the Hatch `default` environment which is for development.)

src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
sundarshankar89: also we didn't attribute rigth pools but overall object counts and estimate goal was not incorrect, so the TCO number where right but not attribution at low level during the dashboard build.

asnare: That's a partial answer. ;)

pyproject.toml
sundarshankar89: Removed

src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
sundarshankar89: Done

sundarshankar89: Done

src/databricks/labs/lakebridge/resources/assessments/synapse/common/profiler_classes.py
sundarshankar89: Done

PR #2268, opened 2026-02-06
Update file/path handling for `databricks labs lakebridge analyze`

asnare:
### What does this PR do?

This PR updates file/path handling for the `analyze` command to both simplify things and make them more consistent. Changes include:

 - The `--report-file` argument is now much simpler:
 
    - It is the name of the Excel file that will be written: what you specify is what you get.
    - If a relative path, we now use normal semantics: it is relative to the current working directory. (Previously this differed if specified via the command-line versus the prompt.)
    - An undocumented feature has been removed: previously if the target file already existed the existing file would be left as-is, and instead a timestamped directory would be created into which the new report was placed. Logging did not reflect this, leading to confusion because the target file did not contain the results of analyser run.
    - The prompt has been updated.

 - The `--source-directory` argument is also simpler:

    - If a relative path, we now use normal semantics: it is relative to the current working directory.
    - The prompt has been updated.

 - Documentation has been updated to match the actual implementation, with a lot of the caveats/qualifications removed. The goal here is that it should be clearer to the reader what is expected, and what they should expect to happen for a given argument.

The `move_tmp_file()` method has been removed: this method was an unnecessary complication and served no real benefit. The names of various arguments has also been updated; previously some of the names indicated things were directories when they were not.

### Linked issues

Blocks #2011.

### Functionality

- [x] updated relevant user documentation

   - [x] updated screencap

- [x] modified existing command: `databricks labs lakebridge analyze`

### Tests

- [x] manually tested
- [x] added unit tests
- [x] existing integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
+++ b/src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
@@ -76,16 +76,18 @@ class AnalyzerRunner:
             raise ValueError(f"Invalid source directory, not writable: {source_dir}")
         if not check_path(results_file_path):
             raise ValueError(f"Invalid result path, not writable: {results_file_path}")
-        _run_imp: Callable[[Path, Path, str], AnalyzerResult]
+        _runnable: Callable[[Path, Path, str, bool], None]
         if results_file_path.suffix == ".xlsx":
-            _run_imp = self._run
+            _runnable = self._runnable
         else:
             # Bladespector currently fails if the path doesn't have a .xlsx extension.
             logger.warning(f"Excel report will be written without .xlsx extension: {results_file_path}")
-            _run_imp = self._run_with_staged_report
-        return _run_imp(source_dir, results_file_path, platform)
+            _runnable = self._run_with_staged_report
+        _runnable(source_dir, results_file_path, platform, self._is_debug)
+        logger.info(f"Analyzed {platform} files in {source_dir}; report saved to: {results_file_path}")
+        return AnalyzerResult(source_dir, results_file_path, platform)
 
-    def _run_with_staged_report(self, source_dir: Path, results_file_path: Path, platform: str) -> AnalyzerResult:
+    def _run_with_staged_report(self, source_dir: Path, results_file_path: Path, platform: str, is_debug: bool) -> None:
         """Run the analyzer, staging the results first to a temporary directory.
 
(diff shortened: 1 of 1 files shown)


src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -26,61 +25,61 @@ def __init__(self, prompts: Prompts):
...
+        logger.debug(f"Starting analyzer execution for {platform}: {source_dir}")
 
-        if not check_path(source_dir) or not check_path(results_dir):
-            raise ValueError(f"Invalid path(s) provided: source_dir={source_dir}, results_dir={results_dir}")
+        if not source_dir.is_absolute():
+            source_dir = source_dir.resolve()
+            logger.debug(f"Relative path provided for source directory, will use: {source_dir}")
+        if not results_file_path.is_absolute():
+            results_file_path = results_file_path.resolve()
+            logger.debug(f"Relative path provided for results file, will use: {results_file_path}")
```
sundarshankar89, line 73: nice enhanced support for relative paths.

src/databricks/labs/lakebridge/analyzer/lakebridge_analyzer.py
```diff
@@ -26,61 +25,61 @@ def __init__(self, prompts: Prompts):
...
-        self._runnable(source_dir, tmp_dir, platform, self._is_debug)
-        self._move_file(tmp_dir, Path(results_dir))
-        logger.info(f"Successfully Analyzed files in {source_dir} for {platform} and saved report to {results_dir}")
-        return AnalyzerResult(Path(source_dir), Path(results_dir), platform)
+        if not check_path(source_dir):
+            raise ValueError(f"Invalid source directory, not writable: {source_dir}")
+        if not check_path(results_file_path):
+            raise ValueError(f"Invalid result path, not writable: {results_file_path}")
+        if results_file_path.suffix != ".xlsx":
+            logger.warning(f"Excel report will be written without .xlsx extension: {results_file_path}")
```
sundarshankar89, line 84: should we check and warn or just add `.xlsx` suffix by default as a normlisation exercise?

asnare: I thought about that, and prefer to not modify what they provided: especially for something CLI-driven, that's not the way things are normally done.

gueniai (approved):
LGTM

PR #2277, opened 2026-02-10
Add DDL Step Support for Typed Table Schemas

sundarshankar89:
## Changes                                                                                                                                                                                                                                                                                                                             

  Adds support for defining explicit table schemas using DDL files in pipeline configurations. This enables creating tables with proper data types (INTEGER, TIMESTAMP, VARCHAR, etc.) instead of defaulting all columns to strings.

  Motivation

  The current pipeline implementation casts all columns to strings due to limitations in inferring types from SQLAlchemy result objects. This causes:
  - Loss of type information and precision
  - Inefficient storage in DuckDB
  - Poor query performance on typed columns
  - Inability to enforce data type constraints

  Changes

  Core Functionality

  - New DDL step type: Pipelines can now include type: ddl steps that execute DDL from external .sql files
  - Two-phase execution: DDL steps execute first to create schemas, then data extraction/processing steps populate tables
  - Schema preservation: Modified _save_to_db() to preserve existing table schemas when appending/overwriting data

  Implementation Details

  - DDL steps support overwrite mode (DROP + CREATE) and append mode (CREATE IF NOT EXISTS)
  - Table existence checks using information_schema before data insertion
  - Backward compatible: non-DDL steps continue to work as before (cast to string)
  - Explicit commits added for transaction safety

  Follow-up Work

  - SQL injection protection: Added TODO comments at 6 locations where table names need proper quoting/parameterization (to be addressed in subsequent PR) #2278 


### Linked issues
<!-- DOC: Link issue with a keyword: close, closes, closed, fix, fixes, fixed, resolve, resolves, resolved. See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword -->

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests
<!-- How is this tested? Please see the checklist below and also describe any other relevant tests -->

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/tests/integration/assessments/test_pipeline.py
+++ b/tests/integration/assessments/test_pipeline.py
@@ -265,38 +265,6 @@ def test_run_pipeline_with_ddl(
         assert usage_result is not None and usage_result[0] > 0, "Usage table should have data"
 
 
-def test_ddl_overwrite_mode(
-    sandbox_sqlserver: DatabaseManager,
-    pipeline_config_with_ddl: PipelineConfig,
-    get_logger: Logger,
-) -> None:
-    """Test that DDL steps in overwrite mode properly drop and recreate tables."""
-    pipeline = PipelineClass(config=pipeline_config_with_ddl, executor=sandbox_sqlserver)
-
-    # Run pipeline first time
-    pipeline.execute()
-
-    db_path = str(Path(pipeline_config_with_ddl.extract_folder)) + "/" + DB_NAME
-
-    # Insert additional data directly
-    with duckdb.connect(db_path) as conn:
-        conn.execute("INSERT INTO inventory VALUES (999, 'test', 'test_collation', NOW(), NOW())")
-        count_before_result = conn.execute("SELECT COUNT(*) FROM inventory").fetchone()
-        assert count_before_result is not None
-        count_before = count_before_result[0]
-
-    # Run pipeline again with overwrite mode
-    pipeline.execute()
-
-    # Verify table was recreated (additional data should be gone)
-    with duckdb.connect(db_path) as conn:
-        count_after_result = conn.execute("SELECT COUNT(*) FROM inventory").fetchone()
-        assert count_after_result is not None
-        count_after = count_after_result[0]
-        # Count should be back to original (without the extra row we added)
-        assert count_after < count_before, "Overwrite mode should recreate table"
-
-
 def test_run_pipeline_with_combined_ddl(
     sandbox_sqlserver: DatabaseManager,
     pipeline_config_combined_ddl: PipelineConfig,


src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -38,14 +38,39 @@ class PipelineClass:
...
+        self._db_path = str(self._db_path_prefix / DB_NAME)
 
     def execute(self) -> list[StepExecutionResult]:
         logging.info(f"Pipeline initialized with config: {self.config.name}, version: {self.config.version}")
         execution_results: list[StepExecutionResult] = []
         error_flag = False
-        for step in self.config.steps:
+
+        # Separate DDL steps from other steps
+        ddl_steps = [step for step in self.config.steps if step.type == "ddl"]
```
goodwillpunning: Should a DDL step be enforced as mandatory?

goodwillpunning: In this design, each pipeline step would need to have its own complimentary DDL step to properly ingest the extracted data. This would 2x the number of pipeline steps. The Oracle profiler, for example has 13 SQL steps, which would bring the total steps to 26. I think there is a more concise way to express the DDL statement for a SQL task. Perhaps we can have a lookup key as a part of the SQL step, e.g. `ddl_key` and a global-level attribute that would serve as a pointer to the DDL definitions?

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -38,14 +38,39 @@ class PipelineClass:
...
+        other_steps = [step for step in self.config.steps if step.type != "ddl"]
+
+        # Execute all DDL steps first
+        logging.info("Executing DDL steps first to create table schemas")
+        for step in ddl_steps:
```
goodwillpunning: Isn't this identical code/logic as lines 73-94? Should be able to condense this.

src/databricks/labs/lakebridge/assessments/pipeline.py
sundarshankar89: we can iterate on that to be mandatory with checks once this code is in other wise it becomes large refactor of existing profilers, for all new profilers we can make ddl step as mandatory.

sundarshankar89: and also the current design allows you to have a single file with all ddls together.

sundarshankar89:
@goodwillpunning  ready for review

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -38,59 +38,74 @@ class PipelineClass:
...
-            logging.info(f"Executing SQL step {step.name}")
-            self._execute_sql_step(step)
-            return StepExecutionStatus.COMPLETE
-        if step.type == "python":
-            logging.info(f"Executing Python step {step.name}")
-            self._execute_python_step(step)
-            return StepExecutionStatus.COMPLETE
-        logging.error(f"Unsupported step type: {step.type}")
-        raise RuntimeError(f"Unsupported step type: {step.type}")
+    def _log_step_result(self, result: StepExecutionResult):
```
goodwillpunning, line 97: Is this really needed? The result is already being logged in places like L#64.

tests/resources/assessments/pipeline_config_with_combined_ddl.yml
```diff
@@ -0,0 +1,25 @@
+name: PipelineWithCombinedDDL
+version: "1.0"
+# Value replaced prior to actual use:
+extract_folder: /replaced/after/loading/
+steps:
+  # Single DDL step with multiple CREATE TABLE statements
```
goodwillpunning, line 6: +1

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -112,10 +122,48 @@ def _execute_sql_step(self, step: Step):
...
+
+        try:
+            with duckdb.connect(self._db_path) as conn:
+                if step.mode == 'overwrite':
+                    # Overwrite mode: drop existing table first
+                    # TODO: SQL injection vulnerability - use quote_identifier(step.name)
+                    conn.execute(f"DROP TABLE IF EXISTS {step.name}")
+                    logging.debug(f"Dropped existing table '{step.name}' for overwrite mode")
+                    # Execute the DDL statement
+                    conn.execute(ddl)
```
gueniai, line 138: Is there a chance that the ddl file contains multiple queries? Like in the case of combined_ddl?

sundarshankar89: good point, I can restrict. it is not recommended option from the pipeline behaviour.

sundarshankar89: I can build a small no cheat check when there are more than one statement in ddl.

sundarshankar89: I made few changes to seperate the DDL changes, now we can safely have more than one ddl per step and the way it is handled is based on existance, schema evolution needs to be subsequently tackled when we encouter one.
 
  
 - Fix overwrite mode in _save_to_db: replace CREATE OR REPLACE TABLE                                                                                                                                                                                                                                                                                    
    with TRUNCATE + INSERT INTO within a transaction to preserve DDL schema                                                                                                                                                                                                                                                                             
  - Simplify _execute_ddl_step: remove overwrite branch, table creation
    is now idempotent (skip if exists)

some minor other chores.

PR #2334, opened 2026-03-12
Eliminate test dependency: `numpy`

asnare:
### What does this PR do?

This PR eliminates `numpy` as a (direct) project dependency: we're only using it in one specific location in a test, and it can be trivially replaced with code from the standard library.

It's worth removing: compatibility with other dependencies restricts us to a really old version, and for some of the supported platforms (and python versions) there's no prebuilt package available. This leads to unnecessarily long test/CI because numpy needs to be compiled from source during (virtual) environment setup.

The integration test that runs the updated pipeline is: `test_profile_execution_config_override`

### Tests

- modified integration test


sundarshankar89 (approved):
LGTM

PR #2336, opened 2026-03-16
Add Entra ID auth (SPN, AD password, Interactive) across MSSQL connector and configurators

sundarshankar89:
## Changes

  Refactors the MSSQL-family connection layer to support the full ODBC Entra ID auth surface: SPN, AD password, AD interactive alongside SQL authentication. 
                                                                                                                                                                                                                 
  ### Auth strategies
   
  | Class | ODBC `Authentication=` | Credentials source | MFA-capable |                                                                                                                                          
  |---|---|---|---|                                                                                                                                                                                              
  | `SqlPassword` | `SqlPassword` | config (`user`, `password`) | No |
  | `ActiveDirectoryPassword` | `ActiveDirectoryPassword` | config (`user`, `password`) | No |                                                                                                                   
  | `ActiveDirectoryServicePrincipal` | `ActiveDirectoryServicePrincipal` | `AZURE_CLIENT_ID` / `AZURE_CLIENT_SECRET` env vars | No |
  | `DefaultAzureCredential` | _(token injection, not yet impl)_ | future: Azure SDK + `SQL_COPT_SS_ACCESS_TOKEN` | N/A |                                                          
                                                                                                                                                                                                                    
  ### Caveats / things to watch out for when reviewing                 
  - Pre-existing credentials.yml files containing `sql_authentication`will be rejected with `ConnectionError`. Users must re-run `configure-database-profiler` (which now writes the canonical class names).
  - **Synapse workspace credential field names**. `sql_user` → `user`, `sql_password` → `password` in the synapse credentials YAML. Brings the workspace schema in line with what `MSSQLConnector` consumes;    
  eliminates the field-name remap that `synapse_connection_helpers` used to do.                                                                                                                                                                                                                                                                                                                              
  - **No `AZURE_TENANT_ID` in the SPN ODBC path** — only `AZURE_CLIENT_ID`/`SECRET` are required, per the ODBC driver contract. The `AZURE_TENANT_ID` references that remain in `synapse.mdx` apply to the 
  *separate* DefaultAzureCredential management-API path (workspace metadata fetch), unrelated to ODBC SPN.                                                                                                       
                                                                                                                                                                                                                 
  ### Linked issues                                                    

Progresses #2462 (remaining `DefaultAzureCredential` token-injection path is wired but stubbed).
                                                                                                                                                                                                                 
  ### Functionality
                                                                                                                                                                                                                 
  ### Tests                                                                                                                                                                                                      
                                                                                                                                                                                                                 
  - [x] added unit tests
  - [ ] added integration tests
  - [ ] manually tested


src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -120,32 +120,40 @@ def _configure_credentials(self) -> str:
...
             "tz_info": self.prompts.question("Enter timezone (e.g. America/New_York)", default="UTC"),
             "driver": self.prompts.question(
                 "Enter the ODBC driver installed locally", default="ODBC Driver 18 for SQL Server"
             ),
         }
 
+        if auth_type == "spn_authentication":
+            logger.info(
+                "SPN authentication selected. "
+                "Ensure AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, and AZURE_TENANT_ID "
```
goodwillpunning: Inconsistent messaging - `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, and `AZURE_TENANT_ID` are listed as required, but L#118 of database_manager.py only validates `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET`. Please update.

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -105,18 +106,29 @@ def _connect(self) -> Engine:
...
+        password: str | None = None
+
         if auth_type == "ad_passwd_authentication":
             query_params = {
                 **query_params,
                 "authentication": "ActiveDirectoryPassword",
             }
         elif auth_type == "spn_authentication":
-            raise NotImplementedError("SPN Authentication not implemented yet")
+            missing = [v for v in ("AZURE_CLIENT_ID", "AZURE_CLIENT_SECRET") if not os.environ.get(v)]
```
goodwillpunning: `AZURE_TENANT_ID` is a required environment variable when using the Azure Identity SDK. A single check for all 3 enviornment variables should be done before the database manager is called.

tests/integration/connections/test_mssql_connector.py
```diff
@@ -14,3 +16,11 @@ def test_mssql_connector_execute_query(sandbox_sqlserver):
 
 def test_connection_test(sandbox_sqlserver):
     assert sandbox_sqlserver.check_connection()
+
+
+@pytest.mark.skip(
```
goodwillpunning: This is permanently skipped, meaning that there is absolutely no automated test coverage for this.

goodwillpunning: This can be done via a mock `create_engine` and using monkeypatch to control environment variables

tests/integration/connections/test_mssql_connector.py
```diff
@@ -14,3 +16,11 @@ def test_mssql_connector_execute_query(sandbox_sqlserver):
 
 def test_connection_test(sandbox_sqlserver):
     assert sandbox_sqlserver.check_connection()
+
```
goodwillpunning, line 18: There is no unit test to test that the `spn_authentication` branch sets authentication to `ActiveDirectoryServicePrincipal`

goodwillpunning (dismissed):
Inconsistent messaging on `AZURE_TENANT_ID`, which _is_ required for Azure Identity SDK but not for ODBC connection. 

There is effectively **zero** test coverage for SPN logic. The unit tests mock past the critical portions of the logic, and the integration test is permanently skipped! `test_configure_synapse_credentials` hardcodes the auth type to `sql_authentication`, and this PR does not introduce tests for the `spn_authentication` path. There are no tests at all for `synapse_connection_helpers`. Please address.

tests/integration/connections/test_mssql_connector.py
sundarshankar89: This is an existing tests which tests user and password authentication flow.

tests/integration/connections/test_mssql_connector.py
sundarshankar89: I want this to be an integration tests, as mentioned the infra PR is in flight once the infra is ready we will be able to test this branch.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
sundarshankar89: I understand why this can lead to confusion. When a user logs in, he logs in to a tenant by default, so we can skip that message. I adjust the `configure_assessment.py` accordingly. 

We will not need `AZURE_TENANT_ID`

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -120,32 +120,40 @@ def _configure_credentials(self) -> str:
...
 
+        if auth_type == "spn_authentication":
+            logger.info(
+                "SPN authentication selected. "
+                "Ensure AZURE_CLIENT_ID and AZURE_CLIENT_SECRET "
+                "are set as environment variables before running the profiler."
+            )
+        else:
+            synapse_workspace["sql_user"] = self.prompts.question("Enter SQL user")
+            synapse_workspace["sql_password"] = self.prompts.password("Enter SQL password")
```
m-abulazm: lets keep auth config together so the UX is consistent. we break the flow by asking for auth type then prompt for other config then once again auth

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -120,32 +120,40 @@ def _configure_credentials(self) -> str:
...
             "driver": self.prompts.question(
                 "Enter the ODBC driver installed locally", default="ODBC Driver 18 for SQL Server"
             ),
         }
 
+        if auth_type == "spn_authentication":
+            logger.info(
+                "SPN authentication selected. "
+                "Ensure AZURE_CLIENT_ID and AZURE_CLIENT_SECRET "
+                "are set as environment variables before running the profiler."
```
m-abulazm: why hardcoded env variables and not managed like all other creds? lets keep a consistent UX

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -120,32 +120,40 @@ def _configure_credentials(self) -> str:
         secret_vault_type = str(self.prompts.choice("Enter secret vault type (local | env)", ["local", "env"])).lower()
         secret_vault_name = None
 
+        # JDBC Settings
+        logger.info("Please select JDBC authentication type:")
```
```suggestion
        logger.info("Please select Synapse authentication mode:")
```

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -120,32 +120,40 @@ def _configure_credentials(self) -> str:
...
+        auth_type = self.prompts.choice(
+            "Select authentication type", ["sql_authentication", "ad_passwd_authentication", "spn_authentication"]
```
```suggestion
            "Select authentication type", ["SQL User", "ActiveDirectoryPassword", "ActiveDirectoryServicePrincipal"]
```

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: this default value is unused, shoudl be removed

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -105,18 +106,29 @@ def _connect(self) -> Engine:
...
+            if missing:
+                raise EnvironmentError(f"SPN authentication requires env vars: {', '.join(missing)}")
+            query_params = {
+                **query_params,
+                "authentication": "ActiveDirectoryServicePrincipal",
+            }
+            username = os.environ["AZURE_CLIENT_ID"]
+            password = os.environ["AZURE_CLIENT_SECRET"]
```
m-abulazm: this parsing code gets easier once the other prompt changes are done.
* the value of auth type will have the needed value without extra parsing
* no magic env lookups, username and pasword will contain the values correctly.

and lets keep related code together. we read auth type at the beginning of the function then read other values then get back to using auth type again

src/databricks/labs/lakebridge/connections/synapse_connection_helpers.py
m-abulazm, line 7: why is this not part of `DatabaseManager` module? please move for easier maintenance

src/databricks/labs/lakebridge/connections/synapse_connection_helpers.py
```diff
@@ -23,16 +23,18 @@ def create_synapse_connection(
...
 
-    config = {
+    config: dict = {
         "driver": workspace_config['driver'],
         "server": server,
         "database": database,
-        "user": workspace_config['sql_user'],
-        "password": workspace_config['sql_password'],
         "port": workspace_config.get('port', 1433),
         "auth_type": auth_type,
```
m-abulazm, line 32: I see that synapse and mssql are not consistent which complicates which values to use. the mssql values are better readable

m-abulazm (changes requested):
I guess we need to support the configuration of mssql profiler. I am fine doing that in another PR with the refactoring of using `DatabaseManager` there but we need an issue to track. let me know how you prefer to move forward

src/databricks/labs/lakebridge/assessments/configure_assessment.py
sundarshankar89: There 2 authentication in Synapse one is the API authentication which we reuse the azlogin properties. 
and the other is the actual database. 

Now when they use SPN to authenticate az, we can reuse the same creds for our databse authentication as well. That was the idea of this implementation.

m-abulazm: we are using pyodbc protocol here not az login.

sundarshankar89: there are 2 parts to synapse authentication, so we use az login default crednetails to extract metrics, there the user defines those standard env variables it in our docs onhow to set this up, we are going to reuse the same env variables for database authentication as well. 

I know we use pyodbc ;)

m-abulazm (approved):

docs/lakebridge/docs/assessment/profiler/mssql.mdx
dgomez04, line 198: Don't we need this section to be updated to include all authentication methods - as `/synapse.mdx` was?

src/databricks/labs/lakebridge/connections/mssql_auth.py
```diff
@@ -0,0 +1,136 @@
...
+    def resolve_credentials(cls, config: JsonObject) -> ResolvedCredentials: ...
+
+
+def _require_user_password(auth_name: str, config: JsonObject) -> tuple[str, str]:
+    if not config.get("user") or not config.get("password"):
+        raise KeyError(f"{auth_name} requires non-empty 'user' and 'password' in config")
+    return str(config["user"]), str(config["password"])
+
+
+class SqlPassword:
```
dgomez04, line 48: After checking Microsoft's [documentation](https://learn.microsoft.com/en-us/sql/connect/odbc/using-azure-active-directory?view=sql-server-ver17), ODBC Driver 18 recognizes `Authentication=SqlPassword` as a literal. Either set `authentication_param="SqlPassword"` so all four classes behave uniformly, or add a one-line comment explaining the omission.

src/databricks/labs/lakebridge/connections/mssql_auth.py
```diff
@@ -0,0 +1,136 @@
...
+    def resolve_credentials(cls, config: JsonObject) -> ResolvedCredentials:
+        user, password = _require_user_password(cls.__name__, config)
+        return ResolvedCredentials(
+            username=user,
+            password=password,
+            authentication_param="ActiveDirectoryPassword",
+        )
+
+
+class ActiveDirectoryServicePrincipal:
```
dgomez04, line 73: This is the only strategy that reads from environment variables, there is no driver-level reason it can't come from the same `config`. The asymmetry also defeats the existing `secret_vault: env` mechanism: a user who wants their `SqlPassword` to be sourced from env vars goes through `secret_vault_type=env` + `EnvGetter`, but for service principal they have to use different hard-coded `env`var names instead.

I would have every strategy resolve from `config[...]` first, with `EnvGetter` providing the env-var fallback (the same path `secret_vault: env` already takes).

src/databricks/labs/lakebridge/connections/mssql_auth.py
```diff
@@ -0,0 +1,136 @@
...
+    @classmethod
+    def resolve_credentials(cls, config: JsonObject) -> ResolvedCredentials:
+        user_value = config.get("user")
+        return ResolvedCredentials(
+            username=str(user_value) if user_value else None,
+            authentication_param="ActiveDirectoryInteractive",
+        )
+
+
+class DefaultAzureCredential:
```
dgomez04, line 91: Shadows `azure.identity.DefaultAzureCredential` which the codebase imports in `resources/assessments/synapse/common/functions.py`. Any future code in this module that needs the real Azure SDK class (the implementation of this strategy) will need to call `azure.identity.DefaultAzureCredential` to acquire the token - having to alias on one of them to import. 

Either: 
- Rename to something distinct that describes the role `DefaultCredentialTokenAuth`, `AzureSdkTokenInjection.
- Keep the name but add a one-liner `"# Distinct from azure.identity.DefaultAzureCredential — that's the SDK chain; this class is the ODBC token-injection strategy that uses it."`

src/databricks/labs/lakebridge/connections/mssql_auth.py
```diff
@@ -0,0 +1,136 @@
...
+# `DefaultAzureCredential` is intentionally excluded until its `resolve_credentials()` is implemented.
+AUTH_CHOICES: list[type[MSSQLAuth]] = [
+    SqlPassword,
+    ActiveDirectoryPassword,
+    ActiveDirectoryServicePrincipal,
+    ActiveDirectoryInteractive,
+]
+
+# All known auth classes — used for runtime dispatch only (includes wired-but-unimplemented).
+_AUTH_REGISTRY: dict[str, type[MSSQLAuth]] = {cls.__name__: cls for cls in (*AUTH_CHOICES, DefaultAzureCredential)}
```
dgomez04: Including DefaultAzureCredential in `_AUTH_REGISTRY` means a user can still hit it by hand-editing `.credentials.yml`, and the failure surfaces as a `NotImplementedError` raised deep in `MSSQLConnector._connect()` rather than at config-load time.

The case is rare, but the fix is a one-line drop: removing it from `_AUTH_REGISTRY` until `resolve_credentials()` is implemented hands the case off to the dispatcher's existing `ConnectionError("Invalid MSSQL auth_type: ...")` branch, which is a clearer message anyway. The test `test_default_azure_credential_is_in_registry_but_not_in_user_choices` would need to be deleted at the same time.

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -86,37 +87,32 @@ def _connect(self) -> Engine:
...
     def _connect(self) -> Engine:
-        auth_type = self.config.get('auth_type', 'sql_authentication')
         db_value = self.config.get('database')
         db_name = str(db_value) if db_value else None
 
+        resolved = resolve_mssql_credentials(self.config)
+
         query_params: dict[str, str] = {
             "driver": str(self.config['driver']),
             "loginTimeout": "30",
```
dgomez04: `loginTimeout` is hardcoded to "30" even though the configurator captures `login_timeout` from the user, so the captured value is unused. Either drop the prompt, or thread it through with `config.get("login_timeout", "30")`.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -13,12 +13,40 @@
...
 )
 from databricks.labs.lakebridge.connections.database_manager import DatabaseManager
+from databricks.labs.lakebridge.connections.mssql_auth import AUTH_CHOICES
 from databricks.labs.lakebridge.connections.env_getter import EnvGetter
 from databricks.labs.lakebridge.assessments import CONNECTOR_REQUIRED
 
 logger = logging.getLogger(__name__)
 
 
+def _prompt_mssql_auth_credentials(prompts: Prompts, auth_type: str) -> dict[str, str]:
```
dgomez04, line 25: Non-blocking, but adding a new authentication strategy can create a drift if the developer forgets adding into the prompt branch. 

A cleaner shape could be adding `prompt_for_credentials(prompts: Prompts) -> dict` classmethod on each `MSSQLAuth` class, and then this collapses into `_AUTH_REGISTRY[auth_type].prompt_for_credentials(prompts)`.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -121,23 +149,28 @@ def _configure_credentials(self) -> str:
         secret_vault_type = str(self.prompts.choice("Enter secret vault type (local | env)", ["local", "env"])).lower()
         secret_vault_name = None
 
+        auth_choices = [cls.__name__ for cls in AUTH_CHOICES]
+        auth_type = self.prompts.choice("Select authentication method", auth_choices)
```
dgomez04: `Prompts.choice` defaults to `sort=True` (alphabetical, case-folded — see `databricks.labs.blueprint.tui.Prompts.choice`). So even though `AUTH_CHOICES` lists `SqlPassword` first, the user sees:

```
[0] ActiveDirectoryInteractive
[1] ActiveDirectoryPassword
[2] ActiveDirectoryServicePrincipal
[3] SqlPassword
```

Most users will pick the first thing they see, which is `ActiveDirectoryInteractive`. Given that `SqlPassword` is the most common case (and our default fallback), consider passing `sort=False` to `prompts.choice` so the visual order matches `AUTH_CHOICES` priority. Mostly a suer experience concern.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -121,23 +149,28 @@ def _configure_credentials(self) -> str:
         secret_vault_type = str(self.prompts.choice("Enter secret vault type (local | env)", ["local", "env"])).lower()
         secret_vault_name = None
```
dgomez04, line 151: I believe `secret_vault_name` might be dead code. Worth flagging.

src/databricks/labs/lakebridge/connections/synapse_connection_helpers.py
dgomez04, line 1: For all three MSSQL-family profilers we acknowledge that pre-existing `.credentials.yml` files will be rejected — but the user just sees `ConnectionError: Invalid MSSQL auth_type: sql_authentication` with no hint on how to recover. We should extend the message to "...; re-run 'configure-database-profiler' to update the credentials format."

src/databricks/labs/lakebridge/resources/assessments/mssql/common/connector.py
dgomez04: `input_cred['user']` and `input_cred['password']` use subscripts that raise `KeyError` when the keys aren't present in the credentials file. That's exactly what happens for two of the four new authentication methods:
- `ActiveDirectoryServicePrincipal` — `_prompt_mssql_auth_credentials` returns `{}`, so the configurator never persists user/password.
- `ActiveDirectoryInteractive` — same thing when the user leaves the optional AAD username blank.

The new `test_spn_authentication_connection` doesn't catch this because it instantiates `DatabaseManager("mssql", config)` directly, bypassing `get_sqlserver_reader`.

Fix is to mirror what `synapse_connection_helpers.create_synapse_connection` already does — use `.get()` and let the connector decide whether the values are required:

```
"user": input_cred.get('user'),
"password": input_cred.get('password'),
```

docs/lakebridge/docs/assessment/profiler/synapse.mdx
dgomez04: This warning is wrong now.

docs/lakebridge/docs/assessment/profiler/mssql.mdx
dgomez04, line 1: The `legacy_synapse.mdx` is out of sync - as it utilizes the same authenitcation menu as `mssql`.

docs/lakebridge/docs/assessment/profiler/synapse.mdx
dgomez04: The pull request description flags that `AZURE_TENANT_ID` is not needed for ODBC SPN, but this step still lists it as part of the requirements. Worth clarifying.

tests/integration/connections/test_mssql_connector.py
```diff
@@ -14,3 +15,8 @@ def test_mssql_connector_execute_query(sandbox_sqlserver):
...
+
+def test_spn_authentication_connection(sandbox_spn_sqlserver_config: JsonObject) -> None:
```
dgomez04, line 20: This test doesn't cover `get_sqlserver_reader`.

dgomez04 (changes requested):
Left a couple of comments - please review and LMK your thoughts.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
m-abulazm: I thought about this: in the database/connection, we should know nothing about cli or prompts. So it is decoupled from cli and reusable in different places.

src/databricks/labs/lakebridge/resources/assessments/mssql/common/connector.py
m-abulazm: this is a bigger refactor to do it correctly

m-abulazm: https://github.com/databrickslabs/lakebridge/pull/2482

tests/integration/connections/test_mssql_connector.py
m-abulazm: refactored https://github.com/databrickslabs/lakebridge/pull/2482

src/databricks/labs/lakebridge/resources/assessments/mssql/common/connector.py
dgomez04: Closed as #2482 removed `get_sqlserver_reader`.

tests/integration/connections/test_mssql_connector.py
dgomez04: Closed as #2482 removed `get_sqlserver_reader`

src/databricks/labs/lakebridge/connections/mssql_auth.py
dgomez04: Addressed since last review, closing.

dgomez04 (approved):
Approved with soft asks worth getting to before merging.

Issue #2247, opened 2026-01-23
[BUG]: Reconciliation fails with AnalysisException when using jdbc_options: partition_column not found in JDBC relation


mikemillerr:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Category of Bug / Issue

Reconcile bug

### Current Behavior

When running a reconciliation using an Oracle source with `jdbc_options` configured (specifically `partition_column`, `lower_bound`, `upper_bound`, and `num_partitions`), the process fails with an `org.apache.spark.sql.AnalysisException`.

Even though the column exists in the source table, Spark reports that the user-defined partition column is not found in the JDBC relation. This appears to be caused by how the `partition_column` string is handled or quoted before being passed to the Spark JDBC reader.

Error Snippet:
```text
AnalysisException: User-defined partition column `id_col_name` not found in the JDBC relation: struct<HASH_VALUE_RECON: string, AMOUNT: decimal(10,0), ..., ID_COL_NAME: decimal(10,0), ...>
```

I found that the issue is related to backticks/quoting. I applied a local workaround in `lakebridge/reconcile/connectors/jdbc_reader.py` at line 18 by stripping the wrapping characters, which resolved the issue:
```python
# Temporary fix/workaround
option_dict["partitionColumn"] = options.partition_column[1:-1]
```

### Expected Behavior

he JDBC reader should correctly accept the `partition_column` defined in the YAML configuration, match it against the JDBC schema, and successfully split the data into the requested number of partitions without throwing a `partition column not found` error.

### Steps To Reproduce

1. Install `databricks-labs-lakebridge==0.11.3.`
2. Configure a reconciliation with an Oracle source.
3. Add jdbc_options to the source configuration:
```yaml
jdbc_options:
    number_partitions: "8"
    partition_column: "ID_COL_NAME"
    lower_bound: "1"
    upper_bound: "10000000"
```
4. Run the reconciliation.
5. The process fails during the data fetching phase with a `ReconciliationException` wrapping a Spark `AnalysisException`.

### Relevant log output or Exception details

```shell
[...]

PR #2362, opened 2026-04-03
Add foreign connections as reconcile source

m-abulazm:
<!-- REMOVE IRRELEVANT COMMENTS BEFORE CREATING A PULL REQUEST -->
## Changes
<!-- Summary of your changes that are easy to understand. Add screenshots when necessary, they're helpful to illustrate the before and after state -->
### What does this PR do?

**Replaced direct JDBC connections with Databricks Unity Catalog remote_query().**                                                                                                                             
                                                                       
  Previously, reconciliation connected to external databases (Oracle, Snowflake, SQL Server) by managing JDBC credentials through Databricks secret scopes — each connector built JDBC URLs, handled             
  authentication (passwords, PEM keys), and executed queries directly.
                                                                                                                                                                                                                 
  Now, connections are managed through **UC Connections** — a centralized Databricks-native way to connect to foreign data sources. The connectors call `remote_query()` via Spark SQL, and Databricks handles   
  authentication and connectivity.
                                                                                                                                                                                                                 
  ### Why                                                              

  - **No more secret management** — credentials are managed in Unity Catalog, not in application code or secret scopes
  - **Simpler configuration** — users provide a UC connection name instead of configuring secret scopes with individual credential keys
  - **Centralized governance** — UC Connections are auditable, permission-controlled, and managed through standard Databricks tooling                                                                            
   
  ### Config change (v1 → v2)                                                                                                                                                                                    
                                                                       
  ```yaml
  # Before (v1)
  data_source: oracle
  secret_scope: remorph_oracle
  database_config:                                                                                                                                                                                               
    source_schema: HR                                                                                                                                                                                            
    target_catalog: main                                               
    target_schema: reconcile_target

  # After (v2)
  source:
    dialect: oracle
    catalog: ORCL
    schema: HR                                                                                                                                                                                                   
    uc_connection_name: my_oracle_connection
  target:                                                                                                                                                                                                        
    catalog: main                                                      
    schema: reconcile_target

  Existing v1 configs are auto-migrated on load.
  ```

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests
<!-- How is this tested? Please see the checklist below and also describe any other relevant tests -->

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/reconcile/connectors/oracle.py
```diff
@@ -91,33 +70,14 @@ def get_schema(
...
+            df = self._reader.read_data(schema_query, catalog, "service_name", "query")
             schema_metadata = df.select([col(c).alias(c.lower()) for c in df.columns]).collect()
             logger.info(f"Schema fetched successfully. Completed at: {datetime.now()}")
             logger.debug(f"schema_metadata: {schema_metadata}")
             return [self._map_meta_column(field, normalize) for field in schema_metadata]
         except (RuntimeError, PySparkException) as e:
             return self.log_and_throw_exception(e, "schema", schema_query)
 
-    @staticmethod
-    def _get_timestamp_options() -> dict[str, str]:
```
m-abulazm, line 103: needs to be evaluated if those oprtions are still needed

src/databricks/labs/lakebridge/reconcile/connectors/tsql.py
```diff
@@ -50,62 +47,36 @@
...
     ) -> DataFrame:
-        table_query = query.replace(":tbl", f"{catalog}.{schema}.{self.normalize_identifier(table).source_normalized}")
-        with_clause_pattern = re.compile(r'WITH\s+.*?\)\s*(?=SELECT)', re.IGNORECASE | re.DOTALL)
-        match = with_clause_pattern.search(table_query)
-        if match:
-            prepare_query_string = match.group(0)
-            query = table_query.replace(match.group(0), '')
-        else:
-            query = table_query
-            prepare_query_string = ""
```
m-abulazm, line 95: which reconcile query needs `prepareQuery` option? dont think it was used in the first place

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
```diff
@@ -119,15 +118,3 @@ def _normalize_col_threshold(self, threshold: ColumnThresholds):
         normalized = dataclasses.replace(threshold)
         normalized.column_name = self.source.normalize_identifier(threshold.column_name).ansi_normalized
         return normalized
-
-    def _normalize_jdbc_options(self, table: Table):
-        if table.jdbc_reader_options:
-            normalized = dataclasses.replace(table.jdbc_reader_options)
-            normalized.partition_column = (
-                self.source.normalize_identifier(normalized.partition_column).ansi_normalized
```
m-abulazm, line 127: this is not used in any sql and should not be escaped

src/databricks/labs/lakebridge/config.py
```diff
@@ -260,15 +281,42 @@ class ReconcileJobConfig:
...
     report_type: str
-    secret_scope: str
-    database_config: DatabaseConfig
+    source: SourceConnectionConfig
+    target: TargetConnectionConfig
     metadata_config: ReconcileMetadataConfig
     job_overrides: ReconcileJobConfig | None = None
 
+    @classmethod
+    def v1_migrate(cls, raw: dict[str, Any]) -> dict[str, Any]:
```
m-abulazm, line 293: v1 cannot be migrated to v2 unless we load the old secret and create a connection. maybe warn the user

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
m-abulazm: https://github.com/databrickslabs/lakebridge/issues/2247

bishwajit-db (commented):
Overall looks good. But I think instead of removing the existing mechanism fully, we should add this as an additional mechanism. This would allow connecting to data sources which are not part of this list https://docs.databricks.com/aws/en/query-federation/database-federation#supported-data-sources. What do you think @m-abulazm @sundarshankar89 ?

BesikiML (changes requested):
### Overall

- Direction looks good: moving reconcile source connectivity from ad-hoc JDBC + secret scopes to Unity Catalog Connections + remote_query() is a big simplification.
- Main risk: the new remote_query() query construction + option encoding has a couple correctness issues.

src/databricks/labs/lakebridge/reconcile/connectors/remote_query_reader.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _build_query(self, query_options: str, source_query: str, source_query_key: str) -> str:
+        escaped = source_query.replace("'", "\'")
+        return (
+            f"SELECT * FROM remote_query('{self._connection_name}', {source_query_key} => '{escaped}', {query_options})"
+        )
+
+    @staticmethod
+    def _build_options(catalog: str, catalog_key: str, options: JdbcReaderOptions | None = None) -> str:
+        def camelcase(underscored):
```
BesikiML, line 35: mismatches expected parameter names:

- `number_partitions` → `numberPartitions` (but Spark JDBC uses `numPartitions`)
- `fetch_size` → `fetchSize` (Spark JDBC uses `fetchsize` / `fetchsize`-like depending on context)

Instead of using “generic camelcase”  conversion, use an explicit mapping dictionary to ensure the parameter names match those expected by `remote_query()`.

src/databricks/labs/lakebridge/reconcile/connectors/remote_query_reader.py
```diff
@@ -0,0 +1,47 @@
...
+        catalog_key: str = "database",
+        source_query_key: str = "query",
+        options: JdbcReaderOptions | None = None,
+    ) -> DataFrame:
+        query_options = self._build_options(catalog, catalog_key, options)
+        query = self._build_query(query_options, source_query, source_query_key)
+        return self._spark.sql(query)
+
+    def _build_query(self, query_options: str, source_query: str, source_query_key: str) -> str:
+        escaped = source_query.replace("'", "\'")
```
BesikiML: In Python, "'' is just a single quote character (there is no special escaping here), so this does not escape anything. Also, in SQL string literals you generally escape ' as '' (double quote), not backslash.
```suggestion
        escaped = source_query.replace("'", "''")
```

m-abulazm: I am following the docs and implemented as stated there https://docs.databricks.com/aws/en/query-federation/remote-queries#basic-query-execution

m-abulazm: Good catch, you're right that `replace("'", "\'")` is a no-op. The fix is `replace("'", r"\'")` . 

On the escaping convention though, databricks actually uses backslash escaping rather than the standard SQL double-quote approach. the docs show `\'` in their examples. https://docs.databricks.com/aws/en/query-federation/remote-queries#basic-query-execution

src/databricks/labs/lakebridge/reconcile/connectors/remote_query_reader.py
m-abulazm: thanks I overlooked that, I prefer to make it simpler by changing the class to follow same naming conventions as databricks and reduce the transformations. I was even contemplating dropping the camel case transformation completely and using camel case on the class directly. I didnt at the end because it is not pythonic.

BesikiML: Sounds good. When you say “align with Databricks naming,” are you planning to use explicit `remote_query` parameter name mapping, or to rename the Python dataclass fields to camelCase?
Either approach works, but explicit mapping helps avoid subtle mismatches (e.g., `numPartitions` vs `fetchsize`) that a generic camelCase conversion might introduce.

m-abulazm: lets keep it like this for now, we might introduce a different jdbc reader and we can evaluate then

src/databricks/labs/lakebridge/config.py
m-abulazm: implemented https://github.com/databrickslabs/lakebridge/pull/2399

BesikiML:
Overall direction looks good (UC connections + `remote_query`, v2 config + migration). Recent commits address the main `remote_query` naming/escaping concerns.
Before I am comfortable approving: please reconcile docs/YAML field names with `JdbcReaderOptions` (e.g. `fetchsize` vs `fetch_size` in docs), and cconfirm whether all dynamically provided option values should be escaped in the same way as the query.

src/databricks/labs/lakebridge/reconcile/connectors/tsql.py
bishwajit-db: The MS SQL Server JDBC driver wraps the dbtable value in a subquery (SELECT * FROM (<dbtable>) tmp), and SQL Server rejects that when the inner statement is a CTE. WITH
  must be the first statement in a batch and can't sit inside parentheses as a derived table. The driver's prepareQuery option runs SQL in the same session before the actual query,
  so the CTE is declared up-front and the wrapped SELECT can reference it. None of the other reconcile dialects (Oracle, Snowflake, Redshift, Databricks) hit this driver quirk, so
  they don't need the split.

m-abulazm: we dont use `dbtable` option anymore. `query` is used

BesikiML (approved):
LGTM

src/databricks/labs/lakebridge/reconcile/connectors/remote_query_reader.py
bishwajit-db: @m-abulazm, IMO, keeping an explicit mapping of JdbcReaderOptions property names to JDBC properties is better here as user will expect those to be applied correctly (right now it won't e.g., number_partitions -> numPartitions not happening). As JdbcReaderOptions has only 5 properties, it is a small change. Later on as a separate change, we can keep it as dict like object so user can pass any JDBC supported properties and responsibility of passing the correct names remains with users.

src/databricks/labs/lakebridge/reconcile/recon_config.py
```diff
@@ -82,12 +82,12 @@ def __post_init__(self):
...
-    partition_column: str | None = None
-    lower_bound: str | None = None
-    upper_bound: str | None = None
-    fetch_size: int = 100
+class JdbcReaderOptions:  # This class follows the same naming as db `remote_query` options
+    partition_column: str  # if used, other props here have to be set as well
+    num_partitions: int
+    lower_bound: str
+    upper_bound: str
+    fetchsize: int = 0  # this uses driver default
```
bishwajit-db, line 90: By driver default, do you mean the fetchsize won't be passed as an JDBC option if zero. If not we should add a check where it is used. Also let's make sure we have per-field None check. Old _get_jdbc_reader_options had if options.number_partitions: guards. If any field slips in as None, the generated SQL will contain numPartitions =>
  'None' and remote_query will error.

m-abulazm: we rely on the type definition to disallow that

src/databricks/labs/lakebridge/reconcile/connectors/oracle.py
m-abulazm: we go with databricks UC standards and remove custom behavior

bishwajit-db (approved):
LGTM.

src/databricks/labs/lakebridge/reconcile/connectors/tsql.py
m-abulazm: got rid of CTE completely so it works

m-abulazm:
> @m-abulazm, IMO, keeping an explicit mapping of JdbcReaderOptions property names to JDBC properties is better here as user will expect those to be applied correctly (right now it won't e.g., number_partitions -> numPartitions not happening). As JdbcReaderOptions has only 5 properties, it is a small change. Later on as a separate change, we can keep it as dict like object so user can pass any JDBC supported properties and responsibility of passing the correct names remains with users.

per the [docs](https://docs.databricks.com/aws/en/query-federation/remote-queries#mysql-postgresql-sql-server-redshift-and-teradata), the options are well known and we model them correctly

sundarshankar89 (commented):
@m-abulazm small nit: once fixed it is good to go from my end.

sundarshankar89 (commented):
forgot to submit the nit

sundarshankar89 (commented):
s

sundarshankar89 (commented):
Small nit: seems like we missed this in snowflake.py

```suggestion
            df = self._reader.read_data(table_query, catalog, "database", "query" , options)
```

m-abulazm:
> Small nit: seems like we missed this in snowflake.py
> 
> ```
>             df = self._reader.read_data(table_query, catalog, "database", "query" , options)
> ```

snowflake does not support jdbc options per the [docs](https://docs.databricks.com/aws/en/query-federation/remote-queries#snowflake).

sundarshankar89 (approved):
LGTM

Issue #933, opened 2024-09-24
[FEATURE]: Reconcile using Databricks Federation as source


ludovicc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Reconcile

### Problem statement

I am trying to use Recon to compare data between a MariaDB database and our Lakehouse. 

MariaDB/MySQL is not supported by Recon, however using Databricks as source should work. I have setup a foreign catalog to MariaDB using Databricks Federation, yet Recon failed with this error message:

Reconciliation failed for one or more tables. Please check the recon metrics for more details. **reconcile** failed.', ReconcileOutput(recon_id='52633b8f-36f3-4738-bb1e-ca0c65674f29', results=[ReconcileTableOutput(target_table_name='prod_ca_regional.landing.sgabar', source_table_name='mariadb_ca.sga.sgabar', status=StatusOutput(row=None, column=None, schema=None, aggregate=None), exception_message='Runtime exception occurred while fetching schema using select lower(column_name) as col_name, full_data_type as data_type from mariadb_ca.information_schema.columns where lower(table_catalog)=mariadb_ca and lower(table_schema)=sga and lower(table_name) =sgabar order by col_name : [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter with name `full_data_type` cannot be resolved.

### Proposed Solution

Recon should be able to query the information schema generated for a foreign catalog, and leverage this to support any data source supported by Databricks Federation.

sundarshankar89:
@ludovicc, we will investigate and get back to you; we haven't added support for the federated catalogue. It might require some enhancements to accommodate this. 

cc: @ganeshdogiparthi-db

vijaypavann-db:
When the federated catalog is used, the Information Schema is specific to that Data Source. 
Here, you are trying Databricks to Databricks reconcile. So, the Schema query is trying to fetch from Databricks Information schema but since it's a federated Catalog, the Information Schema Table definition / Columns are different. Hence the query is failing.

We need to see how we can extend the reconcile to Federated Catalog and the limitations of using it.

PR #2367, opened 2026-04-09
Fix reconcile schema fetch failure on Foreign Catalogs (Lakehouse Federation)

### LINKED ISSUE #2366: [BUG]: Reconciliation fails with UNRESOLVED_COLUMN error on Foreign Catalogs (Lakehouse Federation)
**Category:** Reconcile bug

## Current Behavior

When running Lakebridge Reconcile with a Foreign Catalog (created via Lakehouse Federation) as the source, all report types (`schema`, `data`, `row`, `all`) fail with:

```
[UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter
with name `full_data_type` cannot be resolved.
```

This is because `DatabricksDataSource.get_schema()` queries `information_schema.columns` using the `full_data_type` column, which is Databricks-specific and does not exist in Foreign Catalog's `information_schema`.

## Expected Behavior

Reconcile should work with Foreign Catalogs as source/target, falling back to an alternative schema fetching method (e.g. `DESCRIBE TABLE`) when `full_data_type` is unavailable.

## Steps To Reproduce

1. Create a Foreign Catalog via Lakehouse Federation (e.g. PostgreSQL, Teradata)
2. Configure Lakebridge Reconcile with `data_source="databricks"` using the Foreign Catalog as source
3. Run reconcile with any `report_type`
4. Schema fetch fails with `UNRESOLVED_COLUMN` error

## Relevant log output

```shell
[UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter
with name `full_data_type` cannot be resolved.
Did you mean one of the following? [`data_type`, ...]
```

**Operating System:** macOS
**Version:** latest via Databricks CLI (v0.12.2)

moomindani:
## Summary
- Fixes #2366, #933
- Foreign Catalogs (created via Lakehouse Federation) lack the Databricks-specific `full_data_type` column in `information_schema.columns`, causing `UNRESOLVED_COLUMN` errors for all reconcile report types (`schema`, `data`, `row`, `all`)
- Add `DatabricksNonUnityCatalogDataSource(DatabricksDataSource)` that overrides only `get_schema` to use `DESCRIBE TABLE`, covering hive_metastore, global_temp views, and Foreign Catalogs
- `DatabricksDataSource` is scoped to native Unity Catalog only (`information_schema.columns` with `full_data_type`)
- `source_adapter.create_adapter` dispatches via `is_target`: target → `DatabricksDataSource`, source → `DatabricksNonUnityCatalogDataSource`
- `catalog` parameter on `get_schema` / `read_data` tightened from `str | None` to `str` (now required by `SourceConnectionConfig`)

## Test plan
- [x] Existing unit tests pass (1159)
- [x] New / updated unit tests: `test_get_schema_uses_information_schema`, `test_get_schema_non_uc_uses_describe_table`, `test_get_schema_non_uc_foreign_catalog`, `test_get_schema_information_schema_exception_handling`, `test_get_schema_describe_exception_handling`
- [x] Source adapter tests: `test_create_adapter_for_databricks_dialect_source`, `test_create_adapter_for_databricks_dialect_target`
- [x] Integration test: Reconcile with Foreign Catalog (Lakebase PostgreSQL via Lakehouse Federation) as direct source


sundarshankar89:
@BesikiML  and @bishwajit-db  can you take a look at this when you get a chance

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -94,7 +98,27 @@ def get_schema(
             logger.info(f"Schema fetched successfully. Completed at: {datetime.now()}")
             return [self._map_meta_column(field, normalize) for field in schema_metadata]
         except (RuntimeError, PySparkException) as e:
-            return self.log_and_throw_exception(e, "schema", schema_query)
+            if "full_data_type" not in str(e):
+                return self.log_and_throw_exception(e, "schema", schema_query)
+
+        # Fallback to DESCRIBE TABLE for catalogs that lack the full_data_type column
```
bishwajit-db: Can we add an optional parameter to _get_schema_query?

```python
  def _get_schema_query(catalog: str, schema: str, table: str, force_describe: bool = False):
      if catalog == "hive_metastore" or force_describe:
          return f"describe table {catalog}.{schema}.{table}"
      ...
```

  Then the fallback becomes:

  `describe_query = _get_schema_query(catalog_str, schema, table, force_describe=True)`

bishwajit-db (approved):
Overall LGTM. Please have a look at the comment https://github.com/databrickslabs/lakebridge/pull/2367/changes#r3056652489

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: Thanks for the review! Updated to reuse `_get_schema_query` with a `force_describe` parameter instead of a separate function. Please take another look.

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -17,12 +17,12 @@
 logger = logging.getLogger(__name__)
 
 
-def _get_schema_query(catalog: str, schema: str, table: str):
+def _get_schema_query(catalog: str, schema: str, table: str, force_describe: bool = False):
```
m-abulazm: I would prefer to split this into `_get_schema_query_for_source` that uses describe table exclusively since we know databricks sources can only be Hive, global views or foreign catalog that only supporst `DESCRIBE TABLE`.

and `_get_schema_query_for_target`  that uses the `SELECT` query only as this is the needed behavior always. 

This requires distinguishing source vs target databricks data source. so I would implement a subclass instead of nested conditionals and control through exception handling. 

Also to note this rubs a bit with PR #2362. should not block us on this PR but keep in mind that there might be conflicts

m-abulazm (changes requested):
Thanks for raising this. I documented a different approach as scaling this might require dedicated logic for hive vs foreign catalog vs native catalog. to be open for future changes, I would prefer a subclass instead conditional control

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: Thanks for the detailed feedback! Refactored to split `DatabricksDataSource` into two subclasses:
- `DatabricksSourceDataSource`: always uses `DESCRIBE TABLE` (works for hive_metastore, global_temp, and Foreign Catalogs)
- `DatabricksTargetDataSource`: uses `information_schema` with `full_data_type` (for native UC tables)

The `source_adapter` and `utils` now create the appropriate subclass based on source vs target role. This eliminates the need for exception-based fallback.

Also verified with an integration test — reconcile against a Foreign Catalog (Lakebase PostgreSQL via Lakehouse Federation) succeeded as direct source.

Noted the potential conflict with #2362 — will keep an eye on it.

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -36,6 +35,15 @@ def _get_schema_query(catalog: str, schema: str, table: str):
...
 
 class DatabricksDataSource(DataSource):
+    """Databricks data source backed by Unity Catalog `information_schema`.
+
+    Reads schema metadata from `information_schema.columns` (with `full_data_type`),
+    which is only available for native Unity Catalog catalogs. Use
+    `DatabricksNonUnityCatalogDataSource` for hive_metastore, global_temp views,
+    or Foreign Catalogs (Lakehouse Federation), where `information_schema` is
+    either missing or lacks the `full_data_type` column.
+    """
```
m-abulazm, line 42:
```suggestion
    """Databricks data source backed by Unity Catalog `information_schema`.

    Use `DatabricksNonUnityCatalogDataSource` for hive_metastore, global views,
    or Foreign Catalogs
    """
```

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -99,3 +108,26 @@ def normalize_identifier(self, identifier: str) -> NormalizedIdentifier:
...
+
+
+class DatabricksNonUnityCatalogDataSource(DatabricksDataSource):
+    """Databricks data source for catalogs outside Unity Catalog `information_schema`.
+
+    Uses `DESCRIBE TABLE` for schema fetching, which works for hive_metastore,
+    global_temp views, and Foreign Catalogs (Lakehouse Federation) — none of
+    which expose the Databricks-specific `full_data_type` column on
+    `information_schema.columns`.
+    """
```
m-abulazm:
```suggestion
```

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -50,17 +58,16 @@ def __init__(
...
         table: str,
         query: str,
         options: JdbcReaderOptions | None,
     ) -> DataFrame:
-        namespace_catalog = "hive_metastore" if not catalog else catalog
         if schema == "global_temp":
             namespace_catalog = "global_temp"
         else:
-            namespace_catalog = f"{namespace_catalog}.{schema}"
+            namespace_catalog = f"{catalog}.{schema}"
```
m-abulazm, line 67: this is okay for the time being but we might have to split it in the future if specific behavior is needed

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -50,17 +58,16 @@ def __init__(
 
     def read_data(
         self,
-        catalog: str | None,
+        catalog: str,
```
m-abulazm, line 58: I am suprised this is part of the changes, it is already on main

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -69,27 +76,29 @@ def read_data(
...
+            .distinct()
+            .collect()
+        )
+        logger.info(f"Schema fetched successfully. Completed at: {datetime.now()}")
+        return [self._map_meta_column(field, normalize) for field in schema_metadata]
+
     def get_schema(
         self,
-        catalog: str | None,
+        catalog: str,
```
m-abulazm, line 78: already on main

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
```diff
@@ -69,27 +76,29 @@ def read_data(
         except (RuntimeError, PySparkException) as e:
             return self.log_and_throw_exception(e, "data", table_query)
 
+    def _fetch_schema(self, schema_query: str, normalize: bool) -> list[Schema]:
```
m-abulazm: lets not share methods between the two classes, the code will be better readable if methods are self contained

m-abulazm (changes requested):

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: Applied — thanks.

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: Removed.

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: Acknowledged — will split `read_data` if/when source/target behavior diverges.

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: The abstract `DataSource.read_data` declares `catalog: str` on main, but the concrete `DatabricksDataSource.read_data` was still `catalog: str | None` with a `namespace_catalog = "hive_metastore" if not catalog else catalog` fallback on this exact line in main. Tightening here brings the concrete impl in line with the abstract base, since `SourceConnectionConfig.catalog` is now required. Happy to revert if you'd prefer to keep that change separate.

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: Same as the `read_data` thread — the abstract declares `catalog: str`, but main's concrete `get_schema` was still `catalog: str | None` with a `catalog_str = catalog if catalog else "hive_metastore"` fallback. Aligning with the now-required `SourceConnectionConfig.catalog`.

src/databricks/labs/lakebridge/reconcile/connectors/databricks.py
moomindani: Done — inlined the schema-fetch logic into each `get_schema`. The two classes no longer share `_fetch_schema`.

moomindani:
Reopened as #2422 on an upstream branch to bypass the fork-PR OIDC restriction on JFrog auth (CI cannot run on fork PRs). All review comments here are preserved as history; further discussion will happen on #2422.

PR #2371, opened 2026-04-10
[DRAFT] Refactor PyPI packages version query

ashenm:
### What does this PR do?
Replaces the previous direct `requests.get` call to PyPI for package version lookup with a pip invocation instead. This allows the lookup to automatically respect the existing _pip.conf_ configuration, including custom indexes, mirrors, proxies, and authentication settings.

Package version resolution now uses `pip index versions --json` and parses the structured output instead of calling the package index directly over HTTP. This makes the behaviour consistent with the user’s configured Python packaging environment and avoids maintaining separate logic for custom PyPI endpoint handling.

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [X] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/transpiler/installers.py
+++ b/src/databricks/labs/lakebridge/transpiler/installers.py
@@ -1,5 +1,6 @@
 import abc
 import datetime as dt
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


asnare:
Duplicates #2404

PR #2374, opened 2026-04-15
Modernize project setup, migrate from hatch to uv and pin dependencies

asnare:
## Changes

This PR updates this project to use uv instead of Hatch.

 - All `Makefile` targets have been updated to use `uv` instead of `hatch`.
 - The CI/CD workflows have been updated to work with `uv` instead of `hatch`.
 - Locking information for all third-party dependencies has been added via `uv.lock` and `.build-constraints.txt`.
 - Workflow permissions are now scoped more narrowly.
 - Contribution documentation has been fixed.
 - The release workflow has been removed to prevent confusion.
 - The docusaurus `Makefile` targets have been refactored, and npm/yarn configured to not run any of the scripts associated with packages; we do not need these. Auto-update of the `yarn.lock` should no longer occur.

Aside from the GHA-related changes, some of the project setup can be checked manually while CI/CD isn't available:
```shell
make clean dev fmt test
labs test
```

The documentation setup needs further work, especially if a mirror should be used for packages instead of direct access to the npm/yarn package registries. This will be addressed in a subsequent PR. 

### Tests

- manually tested


pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
   "databricks-labs-blueprint[yaml]>=0.12.0,<0.13.0",
   "databricks-labs-lsql==0.16.0",
   "cryptography>=44.0.2,<46.1.0",
+  "pip~=26.0",
```
sundarshankar89, line 39: This is the only thing I question the need for. I remember we discussed this, can you help me rejig my memory on this.

sundarshankar89 (approved):
LGTM, small clairfications on pip as dep

pyproject.toml
asnare: @sundarshankar89: It's a few things:
 - We directly run `pip` in places, in particular when installing the BB plugin. So we need pip.
 - Normally `pip` is part of the python distribution, _but not always_. In particular the uv-based environments during development don't have it.
 - This means we guarantee it will be available. (And a specific version of it, so we don't end up with hatch-style misery.)

So this dependency was always there, but we didn't explicitly state it in our packaging metadata.

PR #2381, opened 2026-04-15
Documentation: switch from `yarn` to `bun` for building documentation

asnare:
## Changes

This PR updates our documentation building process to be based on [bun](https://bun.comm/) instead of [Yarn 1 (classic)](https://classic.yarnpkg.com/lang/en/).

Aside from being faster, bun gives us better control over the way the JavaScript dependencies are locked including support for a cooldown period.

An incidental change is the checkouts when working with the documentation are now limited to just the documentation.

### Functionality

- Documentation plumbing now uses `bun` instead of `yarn`.
- Workflows are now in terms of `Makefile` targets.
- Contribution documentation has been updated.

### Tests

- manually tested


asnare:
JFrog is currently not compatible with bun. Some of the JSON responses from JSON contain duplicate keys, which bun fails to parse. There is no known workaround for this. (At the time of writing the `typescript` NPM package exhibits this problem.)

PR #2384, opened 2026-04-16
Update documentation infrastructure

asnare:
## Changes

This PR updates the docusaurus plumbing so that CI/CD works with JFrog. The yarn config/setup has also been updated to ensure that the developer environment matches CI/CD.

Incidental changes:

 - Now building with Node 25 instead of 20.
 - Sparse checkouts, and only the current version of the repo instead of the complete history.

For some reason I had to pin the version of `webpack` to get the build to succeed: newer versions aren't compatible with the version of `docusaurus` that we are using.

 - This PR is only about the documentation workflows; the others are still failing and out of scope for this PR.

### Linked issues

Replaces #2381 (bun is incompatible with JFrog)

### Tests

- manually tested
- CI/CD automation

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -19,10 +19,14 @@ jobs:
     runs-on:
       group: databrickslabs-protected-runner-group
       labels: linux-ubuntu-latest
+    permissions:
+      # JFrog OIDC authentication.
+      id-token: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
           sparse-checkout: |
+            .github/actions
             docs
           sparse-checkout-cone-mode: 'true'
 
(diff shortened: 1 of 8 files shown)


.github/workflows/docs-build.yml
```diff
@@ -19,21 +19,28 @@ jobs:
     runs-on:
       group: databrickslabs-protected-runner-group
       labels: linux-ubuntu-latest
-    defaults:
-      run:
-        working-directory: docs/lakebridge
+    permissions:
+      # JFrog OIDC authentication.
+      id-token: write
```
sundarshankar89, line 24: isn't this also needed in push.yml?

sundarshankar89 (approved):
LGTM

.github/workflows/docs-build.yml
asnare: Yup, #2385

PR #2385, opened 2026-04-16
Update CI/CD JFrog integration

asnare:
### What does this PR do?

This PR addresses some issues that were preventing the normal pull-request CI/CD jobs from working properly:

 - The jobs were missing permissions that are needed to access JFrog.
 - The project setup needed some changes to the way the dependencies for build backend (`hatchling`) are resolved.
 - A workaround is also put in place to deal with a packaging error with `databricks-bb-analyzer`: depending on the install order of packages, this error prevents the import of `databricks.*` and `databricks.labs.*` modules from other packages.

Acceptance tests still fail, because we can't set up a Spark service at the moment and some tests depend on this.

The build backend for a project does not fall within uv's normal locking/pinning (`uv.lock`) mechanism, which means uv always attempts to resolve dependencies. When doing so, it will apply the configured cooldown period: 7 days. Unfortunately JFrog does not provide timestamp information for PyPi artefacts, and uv therefore disqualifies everything during resolution. There are two ways of dealing with this: 1) drop the cooldown configuration; 2) exempt those dependencies from the cooldown configuration. Of these, this PR implements the latter so that the rest of the project dependencies are handled properly when locking the dependencies. (The mirrors we're using _also_ apply the cooldown period, but for this sort of thing it's best if both apply it just in case there's a problem with one of them.)

The scope of this PR is addressing issues in the `build` check (`push.yml`): other workflows and jobs still fail but are not in scope.

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -48,7 +48,10 @@ jobs:
           $GITHUB_WORKSPACE/.github/scripts/setup_mssql_odbc.sh
 
       # TODO: Migrate tests to use Databricks clusters instead of Spark local mode
+      # Disabled for now, because access to archives.apache.org is blocked; a new approach will be needed for this.
+      # (The integration tests will still run, but many will fail due to this.)
       - name: Setup spark
+        if: false
         run: |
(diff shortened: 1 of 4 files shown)


sundarshankar89 (approved):
LGTM

PR #2388, opened 2026-04-21
Fix MSSQL profiler extract path, credential manager init, and query/cert handling

eri-adepoju:
### What does this PR do?
Makes fixes so the MSSQL profiler runs cleanly on both on-prem SQL Server and Azure SQL Database:                                                                                                     
  
  1. Profiler output path — pipeline_config.yml now extracts to ~/.databricks/labs/lakebridge_profilers/mssql_assessment, matching the path used by the other profilers instead of /tmp/data/....                                                   
  2. Credential manager init — activity_extract.py and info_extract.py now construct the credential manager via EnvGetter instead of a file path, so it works when secrets come from env vars.
  3. TrustServerCertificate support — added in database_manager.py (SQLAlchemy query params), configure_assessment.py (defaults to "yes" during prompts), and propagated through connector.py so subprocess steps see the flag.                     
  4. Default database = master — in database_manager.py, db_name falls back to 'master' when not provided.                                                                                                                                          
  5. Engine-edition-aware query selection — new get_query_class(connection) helper in mssql/common/connector.py runs AzureSQLDetect.is_azure_sql_db() and returns AzureSQLQueries when EngineEdition == 5, otherwise MSSQLQueries. Adds an AzureSQLQueries set in common/queries.py.   

- Engine-edition detection swallows failures and falls back to MSSQLQueries with a warning log, so existing on-prem flows stay unchanged if detection can't run.    

  - trust_server_certificate defaults to "yes" during interactive prompting in ConfigureSqlServerAssessment. That's a deliberate profiler-UX choice, not a library default — confirm the ergonomics are what reviewers expect.                      
  - pipeline_config.yml path change is a behaviour change for anyone relying on the old /tmp/data/... location.  

### Functionality
- [x] MSSQL profiler now writes to the shared profiler directory                                                                                                                                                                                      
- [x] MSSQL profiler now supports TrustServerCertificate and Azure SQL DB   

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/assessments/configure_assessment.py
+++ b/src/databricks/labs/lakebridge/assessments/configure_assessment.py
@@ -137,6 +137,7 @@ class ConfigureSqlServerAssessment(AssessmentConfigurator):
                 "driver": self.prompts.question(
                     "Enter the ODBC driver installed locally", default="ODBC Driver 18 for SQL Server"
                 ),
+                "trust_server_certificate": "yes",
             },
         }
 
--- a/src/databricks/labs/lakebridge/connections/database_manager.py
+++ b/src/databricks/labs/lakebridge/connections/database_manager.py
@@ -88,11 +88,12 @@ class MSSQLConnector(_BaseConnector):
     def _connect(self) -> Engine:
         auth_type = self.config.get('auth_type', 'sql_authentication')
         db_value = self.config.get('database')
-        db_name = str(db_value) if db_value else None
+        db_name = str(db_value) if db_value else 'master'
 
         query_params: dict[str, str] = {
             "driver": str(self.config['driver']),
             "loginTimeout": "30",
+            "TrustServerCertificate": str(self.config.get('trust_server_certificate', 'no')),
         }
 
         if auth_type == "ad_passwd_authentication":
--- a/src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
+++ b/src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
@@ -8,7 +8,10 @@ from databricks.labs.lakebridge.connections.env_getter import EnvGetter
 from databricks.labs.lakebridge.assessments import PRODUCT_NAME
 from databricks.labs.lakebridge.resources.assessments.common.cli import arguments_loader
 from databricks.labs.lakebridge.resources.assessments.common.duckdb_helpers import save_to_duckdb
-from databricks.labs.lakebridge.resources.assessments.mssql.common.connector import get_sqlserver_reader, get_query_class
+from databricks.labs.lakebridge.resources.assessments.mssql.common.connector import (
+    get_sqlserver_reader,
+    get_query_class,
+)
 from databricks.labs.lakebridge.resources.assessments.mssql.common.schemas import MSSQL_SCHEMAS
 
 logger = get_logger(__file__)
--- a/src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
+++ b/src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
@@ -4,6 +4,7 @@ import sys
 from databricks.labs.blueprint.entrypoint import get_logger
 
 from databricks.labs.lakebridge.connections.credential_manager import create_credential_manager
+from databricks.labs.lakebridge.connections.env_getter import EnvGetter
 from databricks.labs.lakebridge.assessments import PRODUCT_NAME
 from databricks.labs.lakebridge.resources.assessments.common.cli import arguments_loader
 from databricks.labs.lakebridge.resources.assessments.common.duckdb_helpers import save_to_duckdb
@@ -17,7 +18,7 @@ logger = get_logger(__file__)
 def execute():
 
     db_path, creds_file = arguments_loader(desc="MSSQL Server Activity Extract Script")
-    cred_manager = create_credential_manager(PRODUCT_NAME, creds_file)
+    cred_manager = create_credential_manager(PRODUCT_NAME, EnvGetter())
     mssql_settings = cred_manager.get_credentials("mssql")
     auth_type = mssql_settings.get("auth_type", "sql_authentication")
     server_name = mssql_settings.get("server", "")
--- a/src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
+++ b/src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
@@ -8,8 +8,7 @@ from databricks.labs.lakebridge.connections.env_getter import EnvGetter
 from databricks.labs.lakebridge.assessments import PRODUCT_NAME
 from databricks.labs.lakebridge.resources.assessments.common.cli import arguments_loader
 from databricks.labs.lakebridge.resources.assessments.common.duckdb_helpers import save_to_duckdb
-from databricks.labs.lakebridge.resources.assessments.mssql.common.connector import get_sqlserver_reader
-from databricks.labs.lakebridge.resources.assessments.mssql.common.queries import MSSQLQueries
+from databricks.labs.lakebridge.resources.assessments.mssql.common.connector import get_sqlserver_reader, get_query_class
 from databricks.labs.lakebridge.resources.assessments.mssql.common.schemas import MSSQL_SCHEMAS
 
 logger = get_logger(__file__)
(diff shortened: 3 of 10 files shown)


m-abulazm:
@eri-adepoju please update branch with main and drop all will commits as his pr is now merged.

eri-adepoju:
> @eri-adepoju please update branch with main and drop all will commits as his pr is now merged.

Done!

src/databricks/labs/lakebridge/resources/assessments/mssql/common/connector.py
```diff
@@ -1,4 +1,28 @@
...
+from databricks.labs.lakebridge.resources.assessments.mssql.common.queries import (
+    AzureSQLDetect,
+    AzureSQLQueries,
+    MSSQLQueries,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def get_query_class(connection: DatabaseManager):
```
dgomez04, line 13: Discussed with @m-abulazm - I think we need to take this back to the drawing board because it's not a trivial design choice and we want to have proper support and make sure we have a good user experience. We should document this into an issue with broader SQL Server variants support.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -137,6 +137,7 @@ def _configure_credentials(self) -> str:
                 "driver": self.prompts.question(
                     "Enter the ODBC driver installed locally", default="ODBC Driver 18 for SQL Server"
                 ),
+                "trust_server_certificate": "yes",
```
dgomez04, line 140: `legacy_synapse` uses `ConfigureSqlServerAssessment` so it will also pick up `trust_server_certificate`. That should be fine but the pull request description should take that into account.

dgomez04, line 140: The hardcoded `"trust_server_certificate": "yes"` silently disables TLS server-cert validation for every new MSSQL/legacy_synapse profiler install. I think we should convert this to a `self.prompts.question(..., default="yes", valid_regex=r"^(?i)(yes|no)$")` (or confirm if you want the secure default) so a user with strict TLS requirements has a one-keystroke way to opt out, and so the value in `.credentials.yml` reflects an explicit user choice rather than a tool default.

src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
```diff
@@ -17,7 +20,7 @@
 def execute():
 
     db_path, creds_file = arguments_loader(desc="MSSQL Server Activity Extract Script")
-    cred_manager = create_credential_manager(PRODUCT_NAME, creds_file)
+    cred_manager = create_credential_manager(PRODUCT_NAME, EnvGetter())
```
dgomez04, line 23: Both `activity_extract.py` and `info_extract.py` still bind `creds_file` from `arguments_loader(...)` and then never use it. The script reads its credentials from the default path inside `create_credential_manager(PRODUCT_NAME, EnvGetter())`, so the `--credential-config-path` value coming in from `pipeline.py:234` is silently dropped. Synapse already behaves this way, so it isn't new — but it's misleading to a reader and a missed extensibility hook. Two options:

- _**Honor the flag.**_ `cli.py:1106` already wires it up and would let pipeline runs override the credentials path.
- **Acknowledge it in the script.** rename to `db_path, _ = arguments_loader(...)` and add a one-line comment that the flag is preserved for `pipeline.py` compatibility but unused. Don't try to remove the argparse flag itself — `pipeline.py` always passes it.

src/databricks/labs/lakebridge/resources/assessments/mssql/info_extract.py
```diff
@@ -34,9 +37,11 @@ def execute():
             mssql_settings, db_name="master", server_name=server_name, auth_type=auth_type
```
dgomez04, line 37: On Azure SQL Database, `master` is a thin pseudo-database at the server level. Running these queries while connected to `master` will produce empty or near-empty results, the profiler will appear to succeed but the assessment data will be hollow. 

Ideally - we would want to connect to the user database when `EngineEdition` is `5`. I think we should take this back for a proper discussion.

src/databricks/labs/lakebridge/resources/assessments/mssql/common/queries.py
```diff
@@ -383,3 +391,163 @@ def get_table_sizes():
           WHERE o.type = 'U'
           GROUP BY schema_name(o.schema_id), o.name
         """
+
+
+class AzureSQLQueries(MSSQLQueries):
```
dgomez04, line 396: `AzureSQLQueries` overrides only `get_query_stats`, `get_procedure_stats`, `get_sessions`, and `get_cpu_utilization`, all of which are called by `activity_extract.py`. 

The `info_extract` flow calls `get_sys_info`, `get_databases`, `get_tables`, `get_views,` `get_columns`, `get_indexed_views`, `get_routines`, `get_db_sizes`, `get_table_sizes`, none of which are overridden. 

So the `get_query_class(connection)` call in `info_extract.py:40` does an extra round-trip and try/except but always behaves the same.

src/databricks/labs/lakebridge/resources/assessments/mssql/common/queries.py
dgomez04, line 198: `sys.dm_os_sys_info` is a server-scoped DMV requiring `VIEW SERVER STATE`, which is not available in Azure SQL Database (only `VIEW DATABASE STATE` is). If `AzureSQLQueries` doesn't override `get_sys_info`, the Azure SQL Database run will fail on the first table.

dgomez04 (changes requested):
There are genuinely useful fixes like `EnvGetter()`, extract path, and `TrustServerCertificate`. These parts should land. 

That said, adding Azure SQL Database support is not a small change, and I don't think we should merge it as a point-fix. Supporting the full SQL Server family is something I've been thinking about and already discussed with Mohammed / Guenia - there is a big architectural gap and we need to sit down and create a proper design to make sure the experience is user friendly and feature complete.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
m-abulazm: agree with the prompt and default is "no"

src/databricks/labs/lakebridge/resources/assessments/mssql/activity_extract.py
m-abulazm: switching to SQL steps would solve this

src/databricks/labs/lakebridge/resources/assessments/mssql/info_extract.py
m-abulazm: in a followup PR since that is already on main

src/databricks/labs/lakebridge/resources/assessments/mssql/common/connector.py
m-abulazm: our live docs mention support of azure sql database

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -88,11 +88,12 @@ class MSSQLConnector(_BaseConnector):
     def _connect(self) -> Engine:
         auth_type = self.config.get('auth_type', 'sql_authentication')
         db_value = self.config.get('database')
-        db_name = str(db_value) if db_value else None
+        db_name = str(db_value) if db_value else 'master'
```
m-abulazm, line 91:
```suggestion
        db_name = self.config.get('database', 'master')
```

src/databricks/labs/lakebridge/resources/assessments/mssql/common/queries.py
```diff
@@ -1,3 +1,11 @@
+class AzureSQLDetect:
+    """Detect whether the target is Azure SQL Database vs on-prem / VM SQL Server."""
+
+    @staticmethod
+    def is_azure_sql_db() -> str:
```
m-abulazm, line 5: any method that starts with `is_*` have to return a bool. that is a common best practice. this method is `get_engine_edition`

src/databricks/labs/lakebridge/resources/assessments/mssql/common/queries.py
m-abulazm: does the existing `info_extract` work for azure sql db?

m-abulazm:
under design and will be implemented in followups

Issue #997, opened 2024-10-14
[FEATURE]: Create table recon json file (base version) during lakebridge recon configuration


Jyothsnatadi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of feature request

Reconcile

### Problem statement

Creating table_recon_config can be quite challenging because of the intricate naming conventions and the variety of parameters involved. The complexity makes it difficult to get the configurations right on the first attempt, often requiring us to revisit the documentation multiple times to ensure everything is accurate. This process can be time-consuming and prone to errors, especially when handling frequent updates or changes.

### Proposed Solution

Creating a sample table_recon_config based on the provided details would significantly assist users. This approach serves as a reference guide, helping them understand the correct structure and parameters, and reducing the need to frequently consult the documentation for every configuration.

PR #2392, opened 2026-04-24
add auto-discovery of source/target tables for reconcile config

### LINKED ISSUE #209: [FEATURE]: Ability to generate recon config from database and schema

### Problem statement

The ability to create JSON config from a data source, when the amount of tables increases done by hand, will increase the potential risk of errors

We need an auto generator as well as validator.

### Proposed Solution

proposed solution to generate `remorph generate-recon-config` and `remorph validate-recon-config`

### Additional Context

_No response_

m-abulazm:
## Changes
  
  ### What does this PR do?
  **Auto-discover source/target tables and auto-configure column mappings for reconcile via a new CLI command.** Running `databricks labs lakebridge auto-configure-recon-tables` prompts the user for two stages
  (discover tables? auto-configure discovered tables?) and triggers the existing reconcile job with the appropriate operation. 
  The recommended flow is two-stage human-in-the-loop: **discover only** (no configurers) → review and curate the draft → **auto-configure** to fill in column mappings.

building blocks:
  -  `NormalizedMatcher` (case + delimiter + plural/singular);
  - `TableMatcher` (discovers table pairs); `ColumnMappingAutoConfigurer` (the only configurer today).
  - Three workspace-job operation names: `discover-tables` (discover only), `auto-configure-tables` (apply configurers to an existing file), `discover-auto-configure-tables` (both).
  - New `auto-configure-recon-tables` CLI command in `cli.py` and triggers the job via `ReconcileRunner.run(...)`. 
  - End-user docs added to `docs/lakebridge/docs/reconcile/running.mdx`, the recommended human-in-the-loop pattern, and the single-table workflow (call `ColumnMappingAutoConfigurer().configure(...)` directly).
  - Extending the framework with a new configurer is a 2-step recipe documented at the top of `config_generator/execute.py`: write a class implementing `TableAutoConfigurer` and append it to
  `SUPPORTED_AUTO_CONFIGURERS`.

  - Tables and columns that cannot be confidently auto-matched are logged and added to `drop_columns` config part
  - `join_columns`, `column_thresholds`, `table_thresholds`, `transformations`, and `filters` are not auto-discovered — the user must add them manually before running reconcile.

  ### Linked issues
  

  Resolves #209  #997

  ### Functionality
  
  - [x] added new CLI command: `databricks labs lakebridge auto-configure-recon-tables`

  ### Tests
  

  - [ ] manually tested
  - [X] added unit tests
  - [X] added integration tests


BesikiML:
### Overall

Strong direction: running auto-discovery **inside the existing reconcile job** avoids Spark-less installer/CLI issues; `reconcile/config_generator/` with `generate_table_recon` vs `configure_tables`, plus a `ReconcileStrategy` / matcher chain, is a clear extension point (incl. example LLM matcher).

**`stacked PR`:** please confirm **merge order** with the rest of the stack before merge.

**CI:** acceptance **155 passed / 6 skipped** is a good signal. **Codecov** reports **~73% patch** coverage; biggest gaps are new **`list_schemas` / `list_tables`** paths on connectors.

---

### Verify before merge

1. **Omitted unmatched tables** (per PR caveat) — consider making this **very visible** to operators (e.g. explicit log summary counts, or a `skipped_tables` / `unmatched` section in the draft JSON, even when empty) so “silent omission” is not mistaken for “everything matched.”

2. **Connector listing** — add or extend **unit tests per connector** for `list_schemas` / `list_tables` (happy path + empty / error edge where relevant); Codecov shows **low patch %** on those methods.

3. **`execute.py` `configure-tables` dispatch** — Codecov shows **0%** on that small new slice; a **focused unit test** (mock orchestration) would lock the operation contract.

4. **Example LLM matcher** — confirm **no secrets**, **off by default**, and docs for any **optional deps** so default installs are not surprised.

5. **Stack / base** — base moved to `main` per timeline; worth a final **diff vs intended merge base** so no stray commits from an old base remain.

---

### Suggestions (non-blocking)

- Short **user doc** after `configure-recon-tables`: edit joins, reconcile omitted tables from logs, then run `reconcile` as usual.
- Document limitations of **heuristic / singularisation** matching (false pairings possible).
- PR checklist: **“manually tested”** is unchecked — tick or remove for accuracy.

m-abulazm:
> 1. **Omitted unmatched tables** (per PR caveat) — consider making this **very visible** to operators (e.g. explicit log summary counts, or a `skipped_tables` / `unmatched` section in the draft JSON, even when empty) so “silent omission” is not mistaken for “everything matched.”

we deal with thousands of tables sometimes. lets stray off optimizations till we have a solid foundation. we can reiterate on the UX as needed.

> 2. **Connector listing** — add or extend **unit tests per connector** for `list_schemas` / `list_tables` (happy path + empty / error edge where relevant); Codecov shows **low patch %** on those methods.

these are tested as integration tests

> 3. **`execute.py` `configure-tables` dispatch** — Codecov shows **0%** on that small new slice; a **focused unit test** (mock orchestration) would lock the operation contract.

let me check if something is missing from tests

> 4. **Example LLM matcher** — confirm **no secrets**, **off by default**, and docs for any **optional deps** so default installs are not surprised.

this is to be implemented and not used still so no surprises there

> 5. **Stack / base** — base moved to `main` per timeline; worth a final **diff vs intended merge base** so no stray commits from an old base remain.

up to date with main

m-abulazm:
@BesikiML I added an e2e test to test the jobs run correctly and execute logic is tested. I also improved the unit test

src/databricks/labs/lakebridge/cli.py
```diff
@@ -839,6 +890,9 @@ def configure_reconcile(
     reconcile_installer = installer(w, transpiler_repository, is_interactive=True)
     reconcile_installer.run(module="reconcile")
 
+    if ctx.prompts.confirm("Auto-discover source/target tables and pre-fill the table mappings now?"):
+        _run_configure_recon_tables(ApplicationContext(w))
```
bishwajit-db: Can the existing `ctx` be used here?

src/databricks/labs/lakebridge/reconcile/connectors/redshift.py
```diff
@@ -86,6 +86,12 @@ def get_schema(
         except (RuntimeError, PySparkException) as e:
             return self.log_and_throw_exception(e, "schema", schema_query)
 
+    def list_tables(self, catalog: str, schema: str) -> list[str]:
```
bishwajit-db: A Redshift user running configure-recon-tables will get a successful job run that writes an empty TableRecon and a "Could not auto-match 0 source table(s)" log. Prefer to implement it or raise NotImplementedError("Redshift auto-discovery is not yet supported; populate the TableRecon manually") so the job fails loudly, or document the gap in the CLI output before triggering the job.

Same for `list_schemas`

src/databricks/labs/lakebridge/reconcile/config_generator/matchers.py
```diff
@@ -0,0 +1,141 @@
+from __future__ import annotations
+
+import re
+from typing import Protocol
+
+
+class ReconcileStrategy(Protocol):
```
bishwajit-db: Can we use some other name, like `IdentifierMatchingStrategy`? The current name hints at the overall recon strategy, which it is not.

src/databricks/labs/lakebridge/reconcile/config_generator/matchers.py
```diff
@@ -0,0 +1,141 @@
...
+        if word.endswith("ies"):
+            return word[:-3] + "y"
+        if word.endswith("ses"):
+            return word[:-2]
+        if word.endswith("s"):
+            return word[:-1]
+        return word
+
+
+class LlmMatcher:
```
bishwajit-db: If it is a placeholder, we should remove it from the current PR and treat it as a separate change.

src/databricks/labs/lakebridge/reconcile/connectors/snowflake.py
```diff
@@ -98,6 +103,22 @@ def get_schema(
...
+    def list_schemas(self, catalog: str) -> list[str]:
+        query = SnowflakeDataSource._LIST_SCHEMAS_QUERY.format(catalog=catalog)
+        try:
+            df = self._reader.read_data(query, catalog, "database", "query")
+            return [row.schema_name for row in df.select(col("SCHEMA_NAME").alias("schema_name")).collect()]
+        except (RuntimeError, PySparkException) as e:
+            return self.log_and_throw_exception(e, "schemas", query)
+
+    def list_tables(self, catalog: str, schema: str) -> list[str]:
+        query = SnowflakeDataSource._LIST_TABLES_QUERY.format(catalog=catalog, schema=schema.upper())
```
bishwajit-db: The query already filters with lower(table_schema) = lower('{schema}'), so the .upper() is dead code.

src/databricks/labs/lakebridge/reconcile/config_generator/matchers.py
```diff
@@ -0,0 +1,141 @@
...
+        return forms
+
+    @staticmethod
+    def naive_singularize(word: str) -> str:
+        """Best-effort singularisation for English table/column names.
+
+        Rules (applied in order):
+        - `ies` -> `y`  (categories -> category)
+        - `ses` -> `s`  (addresses -> address)
+        - `s`   -> ``   (employees -> employee)
```
bishwajit-db: naive_singularize over-strips singular nouns whose lemma already ends in s (address, class, business, status). Trace address vs addresses. Narrowing the s rule (endswith("s") and not endswith("ss") is a cheap fix, keeps address/class/pass intact; although this may not cover other genuine cases. 

Also, something like Porter Stemmer or Snowball Stemmer may be used here?

src/databricks/labs/lakebridge/reconcile/connectors/redshift.py
m-abulazm: implemented

BesikiML:
**LGTM with small nits:**
- Please keep **log summary counts** for skipped tables/columns at job end (OK to defer richer UX for v1).
- Confirm **CLI behavior** when user picks auto-configure but **no table-recon file exists** (intentional “nothing to do” vs prompt to discover first).
- Quick check that all **three job operations** are covered in tests, not only the combined path.

Happy to approve once those are confirmed.

BesikiML (approved):

src/databricks/labs/lakebridge/reconcile/config_generator/configure.py
```diff
@@ -0,0 +1,216 @@
...
+        """Best-effort singularisation for English table/column names.
+
+        Rules (applied in order):
+        - `ies` -> `y`  (categories -> category)
+        - `ses` -> `s`  (addresses -> address)
+        - `s`   -> ``   (employees -> employee)
+        """
+        if word.endswith("ies"):
+            return word[:-3] + "y"
+        if word.endswith("ses"):
```
bishwajit-db: The `ses` rule fixes `addresses` but regresses the common `-se + s` plurals that the plain `s`-strip used to handle: `houses` -> `hous` (vs `house` -> `house`), same for `cases`, `phases`, `releases`, `expenses`, `warehouses`, `databases`.

Also the original concern is still half-open: singular `address`/`class` still strip to `addres`/`clas` via the final `s` rule, so `address` vs `addresses` still won't match.

Minimal rule set that fixes both:

```python
if word.endswith("ies"):  return word[:-3] + "y"
if word.endswith("sses"): return word[:-2]   # addresses -> address, classes -> class
if word.endswith("ss"):   return word        # address, class stay intact
if word.endswith("s"):    return word[:-1]   # houses -> house, cases -> case
```

src/databricks/labs/lakebridge/reconcile/config_generator/execute.py
```diff
@@ -0,0 +1,180 @@
...
+        table = configurer.configure(table, ctx)
+    return table
+
+
+def _save(installation: Installation, filename: str, table_recon: TableRecon) -> None:
+    installation.upload(filename, json.dumps(asdict(table_recon), indent=2).encode())
+    logger.info(f"Saved table mappings to {filename} ({len(table_recon.tables)} table(s))")
+
+
+def _build_adapters(reconcile_config: ReconcileConfig, spark: SparkSession) -> tuple[DataSource, DataSource]:
```
bishwajit-db, line 167: This duplicates `reconcile.utils.initialise_data_source` and diverges from it in two ways:

1. The target adapter is built without `is_target=True`, so it gets `DatabricksNonUnityCatalogDataSource` instead of the UC `DatabricksDataSource` the recon run path uses. It works today only because that subclass just swaps `get_schema` to DESCRIBE, but it's a landmine if the two classes diverge further.
2. There's no validation that non-databricks sources have a `uc_connection_name`; the empty connection string fails obscurely at query time instead of the clear `InvalidInputException` the helper raises.

Suggest replacing `_build_adapters` with `utils.initialise_data_source(spark, src.dialect, src.uc_connection_name)`.

src/databricks/labs/lakebridge/reconcile/config_generator/execute.py
```diff
@@ -0,0 +1,180 @@
...
+from databricks.labs.lakebridge.reconcile.connectors.source_adapter import create_adapter
+from databricks.labs.lakebridge.reconcile.connectors.data_source import DataSource
+from databricks.labs.lakebridge.reconcile.recon_config import Table
+from databricks.labs.lakebridge.transpiler.sqlglot.dialect_utils import get_dialect
+
+logger = logging.getLogger(__name__)
+
+
+# Register auto-configurers here. Order is the order they run for each Table.
+SUPPORTED_AUTO_CONFIGURERS: tuple[TableAutoConfigurer] = (ColumnMappingAutoConfigurer(),)
```
bishwajit-db: `tuple[TableAutoConfigurer]` types an exactly-one-element tuple. It type-checks today only because there is exactly one configurer; the extension recipe documented above ("append an instance") would break it. Should be `tuple[TableAutoConfigurer, ...]`.

src/databricks/labs/lakebridge/reconcile/config_generator/configure.py
m-abulazm: on it. good call

src/databricks/labs/lakebridge/reconcile/config_generator/execute.py
m-abulazm: good call

src/databricks/labs/lakebridge/reconcile/config_generator/execute.py
m-abulazm: good call

gueniai (approved):
LGTM

PR #2394, opened 2026-04-27
Migrates the reconcile integration tests to Databricks away from local spark

m-abulazm:
### What does this PR do?                                                                                                                                                                                      
  Migrates the reconcile integration tests off local Spark and onto the Databricks `spark` fixture (UC), and removes the no-longer-needed `local_test_run` plumbing that existed only to support that local-Spark
   path.                                                                                                                                                                                                         
                                                                                                                                                                                                                 
  - Replaces the local `mock_spark` fixture (Spark Connect to `sc://localhost`) with the `spark` fixture from `databricks-labs-pytester`, so all tests run against Databricks Spark with UC.
  `DELTA_FAILED_TO_MERGE_FIELDS` from the hand-rolled `report_tables_schema`).                                                                                                                                   
  - Drops the `local_test_run` flag
  - Metadata tables are always addressed as `catalog.schema.<table>`. Tests updated accordingly (every assertion now reads `{catalog}.{schema}.MAIN` and derived from a real `make_schema` + `make_volume`     
  - Skips tests that depend on currently-unavailable infra (Snowflake)                                                                                                                                                                       
   
  - Integration tests now require a Databricks workspace + UC. They will not run against a bare local Spark anymore.
  - snowflake infra is not working and respective integration tests are ignored
                                                                                                                                                                                           
                                                                    
  ### Functionality
  - [ ] added new CLI command
                                                                                                                                                                                                                 
  ### Tests                                                         
  - [x] manually tested
  - [ ] added unit tests                                                                                                                                                                                         
  - [x] added integration tests


tests/integration/assessments/test_pipeline.py
```diff
@@ -19,6 +19,8 @@
 
 _Loader: TypeAlias = Callable[[Path], PipelineConfig]
 
+_skip_sqlserver = pytest.mark.skip(reason="Sandbox SQL Server credentials are stale.")
```
asnare: With this in place we risk leaving these disabled once the infrastructure is fixed; instead we should use `xfail`:
```suggestion
_skip_sqlserver = pytest.mark.xfail(reason="Sandbox SQL Server credentials are stale.", strict=True)
```

tests/integration/assessments/test_profiler.py
```diff
@@ -7,6 +7,8 @@
 from databricks.labs.lakebridge.assessments.pipeline import PipelineClass
 from databricks.labs.lakebridge.assessments.profiler import Profiler
 
+_skip_sqlserver = pytest.mark.skip(reason="Synapse credentials are stale.")
+
```
asnare: The same applies here:
```suggestion
_skip_sqlserver = pytest.mark.xfail(reason="Synapse credentials are stale.", strict=True)
```

tests/integration/connections/test_mssql_connector.py
```diff
@@ -1,5 +1,9 @@
+import pytest
+
 from databricks.labs.lakebridge.connections.database_manager import MSSQLConnector
 
+pytestmark = pytest.mark.skip(reason="Sandbox SQL Server credentials are stale.")
```
asnare: Oh wow, today I learned about package markers. 🤯

As with the others, it should be `xfail`:
```suggestion
pytestmark = pytest.mark.xfail(reason="Sandbox SQL Server credentials are stale.", strict=True)
```

tests/integration/connections/test_synapse_connector.py
```diff
@@ -1,6 +1,10 @@
+import pytest
+
 from databricks.labs.blueprint.installation import JsonObject
 from databricks.labs.lakebridge.connections.database_manager import DatabaseManager, MSSQLConnector
 
+pytestmark = pytest.mark.skip(reason="Sandbox SQL Server credentials are stale.")
```
asnare: As with the others, this should be `xfail`:
```suggestion
pytestmark = pytest.mark.xfail(reason="Sandbox SQL Server credentials are stale.", strict=True)
```

tests/integration/cli/test_profiler_connection_cli.py
```diff
@@ -47,6 +47,7 @@ def _create_credentials_file(
     return cred_path
 
 
+@pytest.mark.skip(reason="Sandbox Synapse credentials are stale; Synapse connection fails.")
```
asnare: As with the others, this should be `xfail`:
```suggestion
@pytest.mark.xfail(reason="Sandbox Synapse credentials are stale; Synapse connection fails.", strict=True)
```

tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -81,29 +81,30 @@ def _get_snowflake_options(self):
         return opts
 
 
-def test_sql_server_read_schema_happy(mock_spark: SparkSession) -> None:
+@pytest.mark.skip(reason="Sandbox SQL Server credentials are stale.")
```
asnare: And here:

```suggestion
@pytest.mark.xfail(reason="Sandbox SQL Server credentials are stale.", strict=True)
```

tests/integration/reconcile/connectors/test_read_schema.py
```diff
@@ -124,17 +125,18 @@ def test_databricks_read_schema_happy_sandbox(
...
     mock_ws = create_autospec(WorkspaceClient)
-    connector = OracleDataSourceUnderTest(mock_spark, mock_ws)
+    connector = OracleDataSourceUnderTest(spark, mock_ws)
 
     columns = connector.get_schema(None, "SYSTEM", "help")
     assert columns
 
 
-def test_snowflake_read_schema_happy(mock_spark: SparkSession) -> None:
+@pytest.mark.skip(reason="Snowflake account suspended")
```
asnare: And here:

```suggestion
@pytest.mark.xfail(reason="Snowflake account unavailable", strict=True)
```

tests/integration/reconcile/test_aggregates_recon_capture.py
```diff
@@ -44,31 +31,18 @@ def agg_data_prep(spark: SparkSession):
...
-        tables = ["main", "aggregate_rules", "aggregate_metrics", "aggregate_details"]
-        for table in tables:
-            remove_directory_recursively(f"{warehouse_location.lstrip('file:')}/{table}")
-
     return agg_reconcile_output, table_conf, reconcile_process_duration
 
 
-def test_aggregates_reconcile_store_aggregate_metrics(mock_workspace_client, mock_spark):
+def test_aggregates_reconcile_store_aggregate_metrics(
+    mock_workspace_client, spark, recon_metadata: ReconcileMetadataConfig
```
asnare, line 38: An observation, but I don't think it needs to be addressed in this PR, is that it seems suspicious to be using a mock `WorkspaceClient` in integration tests: the real thing is available.

tests/integration/reconcile/test_recon_e2e.py
```diff
@@ -12,6 +13,8 @@
 
 logger = logging.getLogger(__name__)
 
+pytestmark = pytest.mark.timeout(1800)
```
asnare, line 16: I think this only applies to a single test at the moment: can we apply the marker there instead of here? (Package markers seem to be an undocumented feature, and I don't think people will know to look here to find it.)

asnare, line 16: Half an hour seems kind of long?

tests/integration/reconcile/test_recon_e2e.py
```diff
@@ -69,13 +72,15 @@ def test_recon_databricks_job_succeeds(
         _run_recon_e2e_spec(app_ctx)
 
 
+@pytest.mark.skip(reason="Sandbox SQL Server credentials are stale.")
```
asnare: Also `xfail` here:
```suggestion
@pytest.mark.xfail(reason="Sandbox SQL Server credentials are stale.", strict=True)
```

tests/integration/reconcile/test_recon_e2e.py
```diff
@@ -69,13 +72,15 @@ def test_recon_databricks_job_succeeds(
...
 def test_recon_sql_server_job_succeeds(
     application_ctx: ApplicationContext, tsql_recon_config: ReconcileConfig, tsql_recon_table_config: TableRecon
 ) -> None:
     with generate_recon_application_context(application_ctx, tsql_recon_config, tsql_recon_table_config) as app_ctx:
         _run_recon_e2e_spec(app_ctx)
 
 
+@pytest.mark.skip(reason="Snowflake account suspended")
```
asnare: And here:
```suggestion
@pytest.mark.xfail(reason="Snowflake account unavailable", strict=True)
```

pyproject.toml
```diff
@@ -480,7 +480,7 @@ max-bool-expr = 5
 max-branches = 20
 
 # Maximum number of locals for function / method body.
-max-locals = 20
+max-locals = 21
```
asnare, line 483: Yikes… unavoidable?

tests/integration/reconcile/query_builder/test_execute.py
```diff
@@ -1592,65 +1601,66 @@ def test_schema_recon_with_data_source_exception(
...
         ),
         pytest.raises(ReconciliationException, match=recon_id.hex),
     ):
-        mock_datetime.now.return_value = datetime(2024, 5, 23, 9, 21, 25, 122185)
-        recon_datetime.now.return_value = datetime(2024, 5, 23, 9, 21, 25, 122185)
-        TriggerReconService.trigger_recon(
-            mock_workspace_client, mock_spark, table_recon, reconcile_config_exception, local_test_run=True
-        )
+        mock_datetime.now.return_value = datetime(2024, 5, 23, 9, 21, 25, 122185, tzinfo=timezone.utc)
+        recon_datetime.now.return_value = datetime(2024, 5, 23, 9, 21, 25, 122185, tzinfo=timezone.utc)
```
asnare: This doesn't look right: `datetime.now()` always returns a naïve timestamp. If this code only works by providing a timestamp-aware value here, something elsewhere isn't working properly.


```suggestion
        mock_datetime.now.return_value = datetime(2024, 5, 23, 9, 21, 25, 122185)
        recon_datetime.now.return_value = datetime(2024, 5, 23, 9, 21, 25, 122185)
```

asnare (changes requested):
This is an improvement we've been wanting for a long time, it's great to see it land.

The main thing I'd like to see addressed is the use of `xfail` instead of `skip` markers. There's also something suspicious about mocking `datetime.now()` that needs looking at.

tests/integration/reconcile/query_builder/test_execute.py
m-abulazm: we were able to get with local timestamps since the spark server and client were in the same timezone. `setup_spark_remote` ran in the same machine as the tests. 

after the migration, databricks follows iso time and my machine is in central europe so locally the test was red as databricks returned a timeaware timestamp and the mock was not. on the CI this probably does not happen as I assume the CI also follows iso time

asnare: This is masking a proper bug then, and the two other tests being skipped will also exhibit the same problem.

When patching, the replacement needs to emulate the original call otherwise we're not actually testing the same thing.

I think part of the fix here is to update the call-sites to use `dt.datetime.now(tz=dt.timezone.utc)` instead.

tests/integration/reconcile/test_recon_e2e.py
m-abulazm: creating a test cluster takes on average 20 minutes

tests/integration/reconcile/query_builder/test_execute.py
m-abulazm: lets have that in subsequent PR. that bug is not caused by this PR

tests/integration/reconcile/test_recon_e2e.py
m-abulazm: acceptance runner also used 30 minutes timeout

pyproject.toml
m-abulazm: yes five tests in `test_execute` crossed that limit as a variable holding the value of the used catalog was introduced. with local spark we had no catalogs

tests/integration/reconcile/query_builder/test_execute.py
m-abulazm: thinking about this more, this was never an issue since reconcile always runs in the workspace. that only affects running integration tests on developer machines

asnare: I think you're missing the primary point here, which is that the change to the mock in this PR is incorrect: mocks need to match the behaviour of what they're replacing, and here we're replacing the use of naïve timestamps with timezone-aware ones. This is a distinction that we know matters and is a frequent gremlin and cause of subtle bugs.

For the integrity of the integration test either:
 - the call-site (and all mocks) should be timezone-aware; or
 - they should all be naïve.
(This applies not just to the test updated here but the 2 siblings as well.)

If the latter fails on dev machines, I suggest going with the former.

m-abulazm:
> This is an improvement we've been wanting for a long time, it's great to see it land.
> 
> The main thing I'd like to see addressed is the use of `xfail` instead of `skip` markers. There's also something suspicious about mocking `datetime.now()` that needs looking at.

addressed, we should merge this asap to unblock the other PRs

src/databricks/labs/lakebridge/reconcile/trigger_recon_aggregate_service.py
sundarshankar89: don't we need the same timezone handling here?

sundarshankar89 (approved):
LGTM with small question.

src/databricks/labs/lakebridge/reconcile/trigger_recon_aggregate_service.py
m-abulazm: good catch, added

asnare (approved):

PR #2399, opened 2026-04-30
Prompt user to reconfigure reconcile on UC connections migration

m-abulazm:
### What does this PR do?
After migrating recon to [Unity Catalog connections](https://docs.databricks.com/aws/en/query-federation/remote-queries), old config with secret scopes are broken and need the config to be recreated with the connection. that scenario is detected during the install and the user is prompted if they want to configure reconcile and create the connection

A deliberate choice of letting the user create the connection is implemented instead of silently creating the connection. As different sources need different create options and instead of owning that complexity, it is deferred to the product itself in databricks and user can follow the official docs

We don't check if the connection exists after the user enters it and that can be implemented in a subsequent PR if the resulting UX improvement is valid

### Functionality

- [ ] added new CLI command
- [X] modified existing command: `databricks labs install/upgrade lakebridge`
- [ ] ... +add your own

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests


tests/unit/test_install.py
```diff
@@ -713,7 +714,7 @@ def test_configure_reconcile_installation_config_error_continue_install(ws: Work
 
 
 @patch("webbrowser.open")
-def test_configure_reconcile_no_existing_installation(ws: WorkspaceClient) -> None:
+def test_configure_reconcile_no_existing_installation(_, ws: WorkspaceClient) -> None:
```
asnare, line 717: `_`?

tests/unit/test_install.py
```diff
@@ -799,7 +800,7 @@ def test_configure_reconcile_no_existing_installation(ws: WorkspaceClient) -> No
 
 
 @patch("webbrowser.open")
-def test_configure_reconcile_databricks_no_existing_installation(ws: WorkspaceClient) -> None:
+def test_configure_reconcile_databricks_no_existing_installation(_, ws: WorkspaceClient) -> None:
```
asnare, line 803: … `_` here as well?

asnare (commented):
I've had a quick skim, but my focus right now is on getting `main` all green. One thing that caught my eye on this is why the addition of those `_` arguments on the tests?

tests/unit/test_install.py
m-abulazm: we patch `webbrowser.open` although we dont assert against it. The test method signature did not reflect that

PR #2420, opened 2026-05-08
Add Snowflake profiler source with PAT auth + DuckDB extraction and TCO summaries

jneil17:
### What does this PR do?

Adds a **Snowflake source** to the Lakebridge profiler (`databricks labs lakebridge configure-database-profiler` / `execute-database-profiler`). End-to-end this:

1. Prompts for Snowflake connection details + a Programmatic Access Token (PAT) and saves credentials.
2. Extracts from `SNOWFLAKE.ACCOUNT_USAGE` (warehouse usage, query history, storage, users, account info, plus optional pipe / autoclustering / mat-view refresh credits) into a timestamped DuckDB file (`profiler_extract_YYYYMMDD_HHMMSS.db`).
3. Runs post-extraction DuckDB computations to produce summary tables: `query_mapper`, `warehouse_mapper`, `classified_queries`, `annual_usage_by_category`, `storage_summary`, and `migration_complexity`.
4. Ships a full PAT-setup walkthrough (with screenshots) under `docs/lakebridge/docs/assessment/profiler/snowflake.mdx`.

- **New profiler source registered** in `assessments/_constants.py` (`PROFILER_SOURCE_SYSTEM`, `PLATFORM_TO_SOURCE_TECHNOLOGY_CFG`, `CONNECTOR_REQUIRED`).
- **`SnowflakeConnector`** in `connections/database_manager.py` is now implemented (was `NotImplementedError`). Uses `snowflake-sqlalchemy` + PAT (password authenticator). Eagerly opens a connection so PAT / network-policy errors surface up front with actionable guidance and a docs link.
- **`ConfigureSnowflakeAssessment`** in `assessments/configure_assessment.py` collects connection + profiler config (lookback 1–365 days, default 365 to match the legacy notebook).
- **Pipeline changes** (`assessments/pipeline.py`):
  - DB filename now timestamped (`profiler_extract_<ts>.db`) so reruns don't clobber prior extracts.
  - SQL templates support `{lookback_days}` substitution from saved config.
  - Optional steps that fail no longer abort the run; they log a warning and (for `adjusted_rate`) write a sensible default row.
  - Every extracted table gets an `extract_timestamp` column.
- **Migration complexity analysis** (`assessments/snowflake_computations.py`) classifies each unique query as native/simple/medium/complex by parsing function calls against bundled Snowflake + Photon function lists. Defaults to a 6h wall-clock timeout; on timeout the user is prompted once to retry without a limit.
- **Resource-path resolution** (`_constants.py::_find_product_path_prefix`) now works for both `databricks labs install` layouts *and* editable `pip install -e .` source layouts — earlier the CLI couldn't find SQL resources outside the labs install path.
- **Auth ergonomics**: `cli.py::_warn_on_databricks_auth_ambiguity` warns (non-blocking) when `~/.databrickscfg` has multiple profiles on the same host or when `DATABRICKS_TOKEN` is set alongside cfg profiles — both produced opaque SDK errors during testing.
- **New deps**: `snowflake-connector-python>=3.6.0`, `snowflake-sqlalchemy>=1.7.0`.

- **Post-merge follow-up tracked**: the two `# TODO` comments in `assessments/configure_assessment.py` and `connections/database_manager.py` reference a branch URL for the docs walkthrough (`feature/snowflake-profiler`). After this PR merges and the docs site rebuilds, both must be flipped to the published URL:
  - `configure_assessment.py` → `https://databrickslabs.github.io/lakebridge/docs/assessment/profiler/snowflake#create-a-snowflake-personal-access-token-pat`
  - `database_manager.py` (`docs_url` stem) → `https://databrickslabs.github.io/lakebridge/docs/assessment/profiler/snowflake` (anchors are concatenated at usage sites — `#part-3--bypass-the-network-policy-temporary`, `#part-2--create-a-personal-access-token`)
  - Greppable with: `grep -rn "feature/snowflake-profiler/docs"`. Will be done in a small follow-up PR.
- `tests/unit/test_install.py::test_configure_all_override_installation` switched from picking source `0` to `1` because `Prompts.choice` sorts alphabetically — adding `snowflake` shifted `synapse` from index 0 to 1. Worth a sanity-check that the test is asserting the intent we want.
- `pipeline.py` now silently swallows failures of optional steps with a warning + default row for `adjusted_rate`. Other optional steps just log; if you'd rather keep the explicit error, flag it.
- Migration complexity is intentionally synchronous + slow on big histories (6h default, opt-in unbounded retry). It runs *after* the rest of the extract is on disk, so a Ctrl+C still leaves a usable DuckDB.
- PAT auth uses Snowflake's `password` authenticator (PATs are accepted there). External browser auth is wired up but untested in this PR.
- New Snowflake deps are pulled in at the project level — they're heavy. Acceptable, but worth confirming.

### Functionality

- [x] added relevant user documentation (`docs/lakebridge/docs/assessment/profiler/snowflake.mdx` + index update)
- [ ] added new CLI command
- [x] modified existing command: `databricks labs lakebridge execute-database-profiler`
- [x] new profiler source: Snowflake (via `ACCOUNT_USAGE`)

### Tests

- [x] manually tested end-to-end against a live Snowflake account (PAT auth, full 365-day lookback, all summary tables produced)
- [x] updated `tests/unit/test_install.py::test_configure_all_override_installation` for the new alphabetical source ordering
- [x] **added unit tests** (30 tests, all passing locally):
  - `tests/unit/assessment/test_snowflake_configurator.py` — full credential flow + lookback validation loop.
  - `tests/unit/assessment/test_snowflake_computations.py` — seeds an in-memory DuckDB extract and asserts all summary tables, mapper sizes, credit math, latest-snapshot storage filter, `extract_timestamp` coverage, and the `duckdb.InterruptException` → `MigrationComplexityTimeout` conversion.
  - `tests/unit/connections/test_snowflake_connector.py` — `_raise_friendly_connect_error` branches (network-policy + PAT credential errors point at the right docs anchors), unmapped errors propagate unchanged, `parse_snowflake_account` / `validate_snowflake_account` URL handling.
  - Engine-mocking tests for `_connect` deliberately skipped — would require mocking `snowflake.sqlalchemy` and add little coverage value beyond the friendly-error tests.
- [ ] added integration tests (manual run against live Snowflake covers this; full CI integration would require a managed Snowflake account)


docs/lakebridge/docs/assessment/profiler/index.mdx
```diff
@@ -88,3 +89,26 @@ Each execution will create a timestamped snapshot of your source environment.
...
+The profiler locates its resource files via the installed package, so `pip install -e .` works from any path — no need to copy files into `~/.databricks/labs/lakebridge/lib/`.
+
+### Multiple Databricks profiles or a stale `DATABRICKS_TOKEN`
+
+`configure-database-profiler` prints a warning when `~/.databrickscfg` has multiple profiles pointing at the same host, or when `DATABRICKS_TOKEN` is set alongside a config file. Either can cause the SDK to pick the wrong credentials. If you hit auth errors:
+
+```bash
+unset DATABRICKS_TOKEN            # if you want the config file to win
+databricks auth login --profile <your_profile>
+```
```
m-abulazm: troubleshooting installs is useful but out of scope of this PR. please remove

docs/lakebridge/docs/assessment/profiler/snowflake.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+#### Step 8. Set the bypass window and grant access
+
+Choose a duration (max **24 hours**) and click **Grant access**.
+
+![Bypass window modal](/img/snowflake_pat/08-bypass-window-modal.png)
+
+:::note
+This is a short-term escape hatch for testing. For anything persistent, attach a proper network policy that allow-lists the IP you'll be connecting from.
+:::
```
m-abulazm: that is quite detailed. helpful for users short term. but hard to maintain.

is there official docs to point to instead of maintaining this here?

lakebridge/resources/assessments/snowflake/README.md
```diff
@@ -0,0 +1,120 @@
+# Enhanced Snowflake Assessment Module
```
m-abulazm: this file is not for end-users and communicates internal details. please remove

src/databricks/labs/lakebridge/assessments/_constants.py
```diff
@@ -1,14 +1,41 @@
...
+            candidates.append(pkg_path.parents[3])
+    except (ModuleNotFoundError, TypeError, ValueError):
+        pass
+    for candidate in candidates:
+        if (candidate / _SRC_MARKER).is_dir():
+            return candidate
+    return candidates[0]
+
+
+PRODUCT_PATH_PREFIX = _find_product_path_prefix()
```
m-abulazm: is this used?

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -178,13 +178,105 @@ def _configure_credentials(self) -> str:
...
+        logger.info(
+            "  3. If needed, bypass the network policy:   on the token's ... menu > Bypass requirement for network policy"
+        )
+        # Print the URL outside the logger so the long log prefix doesn't wrap and break click-to-open.
+        # TODO: switch back to https://databrickslabs.github.io/lakebridge/docs/assessment/profiler/snowflake#create-a-snowflake-personal-access-token-pat
+        # once the feature/snowflake-profiler PR merges and the docs site rebuilds.
+        print("\nFull walkthrough with screenshots:")
+        print(
+            "  https://github.com/databrickslabs/lakebridge/blob/feature/snowflake-profiler/docs/lakebridge/docs/assessment/profiler/snowflake.mdx#create-a-snowflake-personal-access-token-pat\n"
+        )
```
m-abulazm: please use `logger`

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -178,13 +178,105 @@ def _configure_credentials(self) -> str:
...
+            "  1. Grab connection details from Snowsight: user menu > Account > View account details > Config File tab"
+        )
+        logger.info(
+            "  2. Generate a PAT:                         user menu > Settings > Authentication > Programmatic access tokens"
+        )
+        logger.info(
+            "  3. If needed, bypass the network policy:   on the token's ... menu > Bypass requirement for network policy"
+        )
+        # Print the URL outside the logger so the long log prefix doesn't wrap and break click-to-open.
+        # TODO: switch back to https://databrickslabs.github.io/lakebridge/docs/assessment/profiler/snowflake#create-a-snowflake-personal-access-token-pat
```
m-abulazm: have the switch already. they will be released together

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -178,13 +178,105 @@ def _configure_credentials(self) -> str:
...
+            "warehouse": self.prompts.question("Enter warehouse name", default="COMPUTE_WH"),
+            "database": self.prompts.question("Enter database name", default="SNOWFLAKE"),
+            "schema": self.prompts.question("Enter schema name", default="ACCOUNT_USAGE"),
+            "role": self.prompts.question("Enter role", default="ACCOUNTADMIN"),
+            "authenticator": "password",
+            "password": self.prompts.password("Enter Programmatic Access Token (PAT)"),
+        }
+
+        # Profiler Settings
+        logger.info("📊 Assessment will profile all objects and metrics comprehensively")
```
m-abulazm: no emojis

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -178,13 +178,105 @@ def _configure_credentials(self) -> str:
...
+
+        logger.info(f"Credential template created for {source}.")
+        return source
+
+    def _prompt_lookback_days(self) -> int:
+        # Snowflake retains 365 days in ACCOUNT_USAGE; clamp anything outside that range.
+        while True:
+            raw = self.prompts.question(
+                "Enter lookback window in days (1-365)",
+                default="365",
```
m-abulazm: how big can this data get? should we default to 30?

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -19,7 +20,10 @@
 
 logger = logging.getLogger(__name__)
 
-DB_NAME = "profiler_extract.db"
+
+def _make_db_name() -> str:
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+    return f"profiler_extract_{timestamp}.db"
```
```suggestion
    return f"lakebridge_snowflake_profiler_extract_{timestamp}.db"
```

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -19,7 +20,10 @@
 
 logger = logging.getLogger(__name__)
 
-DB_NAME = "profiler_extract.db"
+
+def _make_db_name() -> str:
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
```
```suggestion
    timestamp = datetime.now(tz=UTC).strftime("%Y%m%d")
```

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -35,13 +39,23 @@ class StepExecutionResult:
     error_message: str | None = None
 
 
+DEFAULT_LOOKBACK_DAYS = 365
```
m-abulazm: follows other comment

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -178,13 +178,105 @@ def _configure_credentials(self) -> str:
...
+        snowflake_profiler = {
+            "lookback_days": lookback_days,
+            "exclude_system_objects": False,  # Include everything
+            "include_performance_metrics": True,
+            "include_cost_analysis": True,
+        }
+
+        credential = {
+            "secret_vault_type": secret_vault_type,
+            "secret_vault_name": secret_vault_name,
```
m-abulazm: this is always None?

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -35,13 +39,23 @@ class StepExecutionResult:
...
+DEFAULT_ADJUSTED_RATE = 2.00
+
+
 class PipelineClass:
-    def __init__(self, config: PipelineConfig, executor: DatabaseManager | None):
+    def __init__(self, config: PipelineConfig, executor: DatabaseManager | None, query_params: dict | None = None):
         self.config = config
         self.executor = executor
+        self.query_params = query_params or {"lookback_days": DEFAULT_LOOKBACK_DAYS}
```
m-abulazm: why are snowflake specifics making it into this class?

m-abulazm: please explain your intention with the changes to this class?

src/databricks/labs/lakebridge/assessments/profiler.py
```diff
@@ -65,23 +73,81 @@ def _setup_extractor(platform: str) -> DatabaseManager | None:
...
+            logger.info(f"Profile extraction completed for {platform}: {result}")
         except FileNotFoundError as e:
             logger.error(f"Configuration file not found for source {platform}: {e}")
             raise FileNotFoundError(f"Configuration file not found for source {platform}: {e}") from e
         except Exception as e:
             logger.error(f"Error executing pipeline for source {platform}: {e}")
             raise RuntimeError(f"Pipeline execution failed for source {platform} : {e}") from e
 
+        # Snowflake-specific post-extraction computations.
+        if platform == "snowflake":
```
m-abulazm: what is stopping us from running this inside the profiler at the end

m-abulazm (changes requested):
thanks for the work, I went through the production code. generally, the snowflake specific changes to the pipeline class should be avoided. and remove the code of the old profiler.

docs/lakebridge/docs/assessment/profiler/snowflake.mdx
jneil17: Good call out, we can linked to Snowflake official site and just use that, it is not as clean but will be better: https://docs.snowflake.com/en/user-guide/programmatic-access-tokens#generating-a-programmatic-access-token

lakebridge/resources/assessments/snowflake/README.md
jneil17: Will remove

src/databricks/labs/lakebridge/assessments/pipeline.py
jneil17: The change with the look back period is because the Query History tables sometimes can be MASSIVE and we want to give the ability configure how far we go. We can hold off on implementing until we see the need but is a possible concern for some of our larger accounts

docs/lakebridge/docs/assessment/profiler/index.mdx
jneil17: done - also dropped _warn_on_databricks_auth_ambiguity from cli.py since it was the source of the docs section. Happy to spin out into a separate PR if useful

docs/lakebridge/docs/assessment/profiler/snowflake.mdx
jneil17: Done — the page now intros the profiler and links straight to Snowflake's PAT docs. Screenshots deleted. Also condensed the in-CLI walkthrough to a single docs link. ebec7bbb 1b747a31

src/databricks/labs/lakebridge/assessments/_constants.py
jneil17: removed, devs can use databricks labs install lakebridge@<branch> for end-to-end testing. ebec7bbb

src/databricks/labs/lakebridge/assessments/configure_assessment.py
jneil17: done

src/databricks/labs/lakebridge/assessments/pipeline.py
jneil17: Reverted all Snowflake-specific additions from pipeline.py — timestamping, query_params, DEFAULT_LOOKBACK_DAYS, DEFAULT_ADJUSTED_RATE, optional-step fallback, extract_timestamp injection. Snowflake-specifics now live in snowflake_computations.py and the configurator. Lookback hardcoded to 365 for now. 4f290e16

src/databricks/labs/lakebridge/assessments/configure_assessment.py
jneil17: good catch — dropped the field. Was always None for Snowflake. ebec7bbb

src/databricks/labs/lakebridge/assessments/pipeline.py
jneil17: removed, covered by the query_params thread below. 4f290e16

src/databricks/labs/lakebridge/assessments/profiler.py
jneil17: It already runs at the end of Profiler._execute() — run_post_extraction() is called after the pipeline finishes, gated on platform  == 'snowflake'. After this commit the profiler.py diff vs main is just 2 imports and a 3-line conditional; the timeout/retry/prompt logic, the post-extract orchestration, and everything else Snowflake specific lives in snowflake_computations.py.  Snowflake is the only profiler with a post-extraction phase today — synapse does its computation inside pipeline steps and mssql doesn't have derived tables. A registry/hook would be premature abstraction for one entry. Happy to refactor when a second platform shows up.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
jneil17: removed

src/databricks/labs/lakebridge/assessments/configure_assessment.py
jneil17: We can change it to 30, we will need to make sure we call this out in our documentation

jneil17: Looked at this more carefully. The profiler's primary use case is annual migration TCO conversations — 365 days matches Snowflake's billing year, gives us seasonality, and we verified the credit math against WAREHOUSE_METERING_HISTORY exactly at that window. At 30 days the numbers come in ~12× lower than actual spend and annual_usage_by_category becomes misleading. On the data-size concern even a busy test account runs end-to-end in under 2 minutes at 365 days. Happy to add a configurable prompt later if a customer hits a real wall, but think 365 is the right default for now

src/databricks/labs/lakebridge/assessments/profiler.py
m-abulazm: this should be part of TCO app as it is a long running task and will benefit from using a proper job run to manage it

src/databricks/labs/lakebridge/assessments/snowflake_computations.py
```diff
@@ -0,0 +1,684 @@
...
+        "This step classifies every unique query as native/simple/medium/complex "
+        "by parsing function calls against Snowflake + Photon function lists. "
+        "Large query histories can legitimately take many hours. "
+        "Retrying without a timeout lets it run to completion; the rest of the "
+        "extract is already saved so you can cancel (Ctrl+C) anytime."
+    )
+    if not prompts.confirm("Retry migration complexity analysis without a timeout?"):
+        logger.warning("Skipped migration complexity analysis; `migration_complexity` table will be missing.")
+        return
+    run_migration_complexity(db_path, timeout_seconds=None)
```
m-abulazm: I am not sure if this will work that we can capture the user stdin properly. I would not do that here. we have a dedicated configure where any user input can be captured if needed

src/databricks/labs/lakebridge/assessments/snowflake_computations.py
```diff
@@ -0,0 +1,684 @@
+"""
```
m-abulazm: please check my other comment that this should go to TCO app

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -92,7 +99,63 @@ def _create_connector(db_type: str, config: JsonObject) -> DatabaseConnector:
...
+        role = str(connection_config.get("role", "ACCOUNTADMIN"))
+        authenticator = str(connection_config.get("authenticator", "password"))
+
+        # Build Snowflake SQLAlchemy URL based on authentication method
+        if authenticator == "externalbrowser":
+            # For external browser authentication, don't include password in URL
+            snowflake_url = (
+                f"snowflake://{user}@{account}/"
+                f"{database}/{schema}?warehouse={warehouse}&role={role}&authenticator=externalbrowser"
+            )
```
m-abulazm: password is hardcoded in configure step, lets keep changes focused to what we will use

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -0,0 +1,126 @@
...
+    if not account_identifier:
+        return False
+
+    # Snowflake account identifiers typically contain letters, numbers, and hyphens
+    # They cannot be empty and should not contain spaces or special characters
+    pattern = r'^[A-Za-z0-9\-_]+$'
+    return bool(re.match(pattern, account_identifier))
+
+
+def build_snowflake_url(
```
m-abulazm: this is duplicate of `SnowflakeConnector`

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -0,0 +1,126 @@
...
+    ]
+
+    print("Testing Snowflake account parsing:")
+    for test_input in test_cases:
+        parsed = parse_snowflake_account(test_input)
+        valid = validate_snowflake_account(parsed)
+        print(f"  Input: {test_input}")
+        print(f"  Parsed: {parsed}")
+        print(f"  Valid: {valid}")
+        print()
```
m-abulazm: test code to be removed

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -0,0 +1,68 @@
+name: snowflake_assessment
```
m-abulazm, line 1: need to verify if we need a ddl step or not

m-abulazm (changes requested):
a few minor comments and the main one still about post calculations that I think should move to a background job in TCO app

src/databricks/labs/lakebridge/assessments/snowflake_computations.py
jneil17: Not relevant anymore, the file got dropped in 601e03dc per the TCO app thread. Any future timeout/retry UX we can collect at configure time like you suggested.

src/databricks/labs/lakebridge/assessments/snowflake_computations.py
jneil17: Done, dropped in 601e03dc. Will land in the TCO app PR.

src/databricks/labs/lakebridge/connections/database_manager.py
jneil17: Removed the externalbrowser branch and the authenticator key from the saved credentials. PAT password is the only path we actually use. 601e03dc

src/databricks/labs/lakebridge/connections/snowflake_utils.py
jneil17: yea agreed. dropped build_snowflake_url, URL construction only lives in SnowflakeConnector._connect() now. 601e03dc

src/databricks/labs/lakebridge/connections/snowflake_utils.py
jneil17: Removed in 601e03dc.

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
jneil17: Dug into this. No DDL step needed. PipelineClass._save_to_db issues CREATE TABLE step_name AS SELECT * FROM _result_frame at pipeline.py:303, so each SQL step creates its own table from the result schema. Same pattern as the legacy_synapse profiler that landed in main (all SQL, no DDL).

uv.lock
```diff
@@ -37,6 +37,15 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/74/f5/9373290775639cb67a2fce7f629a1c240dce9f12fe927bc32b2736e16dfc/argcomplete-3.6.3-py3-none-any.whl", hash = "sha256:f5007b3a600ccac5d25bbce33089211dfd49eab4a7718da3f10e3082525a92ce", size = 43846, upload-time = "2025-10-20T03:33:33.021Z" },
 ]
 
+[[package]]
```
jneil17, line 40: Adding snowflake-connector-python + snowflake-sqlalchemy to pyproject.toml means we have to update uv.lock, uv lock --check fails otherwise. After merging main the only new entries are those two plus their transitive deps (filelock, pyopenssl, sortedcontainers). asn1crypto isn't a new addition anymore, main already pulls it via databricks-sdk. Also rewrote registry URLs to public PyPI in 70ee0184 so lint-uv passes.

docs/lakebridge/docs/assessment/profiler/index.mdx
```diff
@@ -33,6 +33,7 @@ Key capabilities:
 | Source Platform | Configuration Status |
 |:---------------:|:-------------------:|
 | <a href="./synapse" style={{ fontWeight: 'bold', color: '#1976d2', textDecoration: 'underline' }}>Azure Synapse</a> | &#x2705; |
+| <a href="./snowflake" style={{ fontWeight: 'bold', color: '#1976d2', textDecoration: 'underline' }}>Snowflake</a> | &#x2705; |
```
m-abulazm, line 36: need to update overview.mdx as well

src/databricks/labs/lakebridge/assessments/_constants.py
```diff
@@ -5,16 +5,22 @@
...
 PLATFORM_TO_SOURCE_TECHNOLOGY_CFG = {
     "synapse": "src/databricks/labs/lakebridge/resources/assessments/synapse/pipeline_config.yml",
+    "snowflake": "src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml",
     "mssql": "src/databricks/labs/lakebridge/resources/assessments/mssql/pipeline_config.yml",
     "legacy_synapse": "src/databricks/labs/lakebridge/resources/assessments/legacy_synapse/pipeline_config.yml",
 }
 
 # TODO modify this PLATFORM_TO_SOURCE_TECHNOLOGY.keys() once all platforms are supported
-PROFILER_SOURCE_SYSTEM = ["synapse", "mssql", "legacy_synapse"]
+PROFILER_SOURCE_SYSTEM = ["synapse", "snowflake", "mssql", "legacy_synapse"]
```
m-abulazm: please merge main. oracle is newly there

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -81,7 +88,51 @@ def health_check(self) -> bool:
...
+                f"  - Original error: {msg}"
+            ) from error
+        if "Invalid OAuth access token" in msg or "Incorrect username or password" in msg:
+            raise ConnectionError(
+                "Snowflake rejected the credentials. The PAT may be expired, revoked, mistyped, or for a different user.\n"
+                f"  - Regenerate a token: {docs_url}\n"
+                f"  - Original error: {msg}"
+            ) from error
+        # No friendly mapping for this error — re-raise with cause preserved.
+        raise error
```
m-abulazm: we have `health_check` method on the connector that should be used.
and as a result, this can be done generically for all connectors

tests/unit/connections/test_snowflake_connector.py
```diff
@@ -0,0 +1,93 @@
...
+from databricks.labs.lakebridge.connections.database_manager import SnowflakeConnector
+from databricks.labs.lakebridge.connections.snowflake_utils import (
+    parse_snowflake_account,
+    validate_snowflake_account,
+)
+
+# ---- friendly connect-error mapping -------------------------------------------------
+
+
+def _raise_friendly(error: Exception):
```
m-abulazm: to be clarified given other [comment](https://github.com/databrickslabs/lakebridge/pull/2420/changes#r3288666326)

tests/unit/test_install.py
```diff
@@ -901,7 +901,9 @@ def test_configure_all_override_installation(  # FIXME
             r"Enter .* schema name": "tpch_sf1000",
             r"Enter target Databricks catalog name": "tpch",
             r"Enter target Databricks schema name": "1000gb",
-            # Profiler Configuration Prompts
+            # Profiler Configuration Prompts.
+            # Prompts.choice sorts alphabetically, so index 0 = legacy_synapse when
+            # PROFILER_SOURCE_SYSTEM is ["synapse", "snowflake", "mssql", "legacy_synapse"].
```
m-abulazm:
```suggestion
```

m-abulazm (changes requested):
almost there.

m-abulazm (approved):
Great. thanks

src/databricks/labs/lakebridge/assessments/_constants.py
```diff
@@ -5,17 +5,24 @@
...
     "synapse": "src/databricks/labs/lakebridge/resources/assessments/synapse/pipeline_config.yml",
+    "snowflake": "src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml",
     "oracle": "src/databricks/labs/lakebridge/resources/assessments/oracle/pipeline_config.yml",
     "mssql": "src/databricks/labs/lakebridge/resources/assessments/mssql/pipeline_config.yml",
     "legacy_synapse": "src/databricks/labs/lakebridge/resources/assessments/legacy_synapse/pipeline_config.yml",
 }
 
 # TODO modify this PLATFORM_TO_SOURCE_TECHNOLOGY.keys() once all platforms are supported
-PROFILER_SOURCE_SYSTEM = ["synapse", "mssql", "legacy_synapse", "oracle"]
+PROFILER_SOURCE_SYSTEM = ["synapse", "snowflake", "mssql", "legacy_synapse", "oracle"]
```
dgomez04: `Prompts.choice` sorts options alphabetically before display, so the explicit order here is decorative. `sorted(PLATFORM_TO_SOURCE_TECHNOLOGY_CFG.keys())` would convey the same thing without inviting future re-orderings.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -217,6 +217,51 @@ def _configure_credentials(self) -> str:
...
+        snowflake_connection = {
+            "account": self.prompts.question(
+                "Enter Snowflake account URL (e.g., myorg-myaccount.snowflakecomputing.com)"
+            ),
+            "user": self.prompts.question("Enter username"),
+            "warehouse": self.prompts.question("Enter warehouse name", default="COMPUTE_WH"),
+            "database": self.prompts.question("Enter database name", default="SNOWFLAKE"),
+            "schema": self.prompts.question("Enter schema name", default="ACCOUNT_USAGE"),
+            "role": self.prompts.question("Enter role", default="ACCOUNTADMIN"),
+            "password": self.prompts.password("Enter Programmatic Access Token (PAT)"),
```
dgomez04: The configurator writes the PAT under `password:` because Snowflake's `password` authenticator accepts PAT values, but a user opening the credential files would later see: 

```yaml
snowflake:
  connection:
    user: ME
    password: ABC...<60 chars>
```

Nothing flags this is a rotating token rather than a SQL password. Either:
- Rename the YAML field to `pat:`.
- Keep `password:` but add a preceding comment when the file is written like `# Snowflake PAT`

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -217,6 +217,51 @@ def _configure_credentials(self) -> str:
...
+
+    def _configure_credentials(self) -> str:
+        cred_file = self._credential_file
+        source = self._source_name
+
+        logger.info(
+            "\n(local | env) \nlocal means values are read as plain text \nenv means values are read "
+            "from environment variables fall back to plain text if not variable is not found\n",
+        )
+        secret_vault_type = str(self.prompts.choice("Enter secret vault type (local | env)", ["local", "env"])).lower()
```
dgomez04, line 231: When `env` is selected, the YAML is meant to hold the environment variable names that `EnvGetter` resolves via `os.getenv`.

The prompt order here is: vault type -> connection fields -> `Enter Programmatic Access Token (PAT)`, with no reminder that the user is now supposed to type a variable name, not the token. 

**Possible fix:** when `secret_vault_type == "env"`, change the prompt label to "Enter env var name holding the PAT".

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -81,7 +88,21 @@ def health_check(self) -> bool:
...
+
+        account = parse_snowflake_account(str(connection_config["account"]))
+        user = str(connection_config["user"])
+        warehouse = str(connection_config.get("warehouse", "COMPUTE_WH"))
+        database = str(connection_config.get("database", "SNOWFLAKE"))
+        schema = str(connection_config.get("schema", "ACCOUNT_USAGE"))
+        role = str(connection_config.get("role", "ACCOUNTADMIN"))
+        password = str(connection_config["password"])
+
+        snowflake_url = f"snowflake://{user}:{password}@{account}/{database}/{schema}?warehouse={warehouse}&role={role}"
```
dgomez04: Snowflake uses **JWT-based bearer tokens** for PATs using `Base64URL` encoding, meaning it can include include characters like `/`, `+`, `=`, and `@`. When any of those land in the f-string, SQLAlchemy's URL parser won't have a way to know which characters are PAT and which are the URL structure.
- A PAT containing `@` (e.g. `abc@xyz123...`) splits the `userinfo` segment early — SQLAlchemy treats `xyz123...` as the `host` and `account` as part of the `path`. Connection fails with something like `Could not parse rfc1738 URL` or, worse, succeeds against a wrong host.
- A PAT containing `/` rolls into the `path` component, so `database` and `schema` get pushed past the end and Snowflake sees an empty database. The user sees Database 'XYZ' does not exist or not authorized.
- ...

In every case the user sees an opaque error that does not point at the PAT. They'll regenerate the token, get a new one with different special chars, and the bug will appear intermittent.

Fix is a one-liner — and it brings this connector in line with the rest of the file. `MSSQLConnector._connect` (line 131) and `OracleConnector._connect` (line 145) already build their URLs through `sqlalchemy.engine.URL.create()`, which percent-encodes username/password automatically:

```python
from sqlalchemy.engine import URL
snowflake_url = URL.create(
    drivername="snowflake",
    username=user,
    password=password,
    host=account,
    database=f"{database}/{schema}",
    query={"warehouse": warehouse, "role": role},
)
return create_engine(snowflake_url)
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -81,7 +88,21 @@ def health_check(self) -> bool:
 
 class SnowflakeConnector(_BaseConnector):
     def _connect(self) -> Engine:
-        raise NotImplementedError("Snowflake connector not implemented")
+        # Extract connection details from config. The SDK types JSON values loosely;
+        # narrow to a plain dict so the dict-style accesses below typecheck.
+        raw = self.config.get("connection", self.config)
```
dgomez04: Every other connector reads flat into the fields, `SnowflakeConnector`'s current fallback path has no caller and no tests. If a future refactor accidentaly drops the `connection:` wrapper from the configurator's YAML, the user gets a error far downstream. Additionally, it also invites for schema drift to happen.

**Suggested fix** — pick the nested shape and commit to it:
```
connection_config: dict[str, Any] = self.config["connection"]
```

src/databricks/labs/lakebridge/connections/database_manager.py
dgomez04, line 1: @m-abulazm —  `engine.connect()` was removed in `74eacf2` per your "use health_check generically" suggestion, but I don't see a `health_check` call wired in anywhere in this pull request ( at least not in `SnowflakeConnector`, not at the end of `_configure_credentials`, not in `Profiler._setup_extractor`). PAT or network-policy errors now surface deep inside the pipeline on the first SQL fetch instead of at configure time.

Is the generic rollout landing in a separate pull request, or did you mean for it to land here?

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -0,0 +1,60 @@
...
+    if "/" in account:
+        account = account.split("/")[0]
+
+    # Remove trailing slashes
+    account = account.rstrip("/")
+
+    return account
+
+
+def validate_snowflake_account(account_identifier: str) -> bool:
```
dgomez04: This is dead code. It's tested in `test_snowflake_connector.py` but it's never called. Wiring it adds real value today because a typo like "my org-acct" (with a space) is silently accepted and explotes at SQL-fetch time.

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -0,0 +1,68 @@
...
+version: "1.0"
+extract_folder: "."
```
dgomez04: `extract_folder: "."` makes DuckDB write to the local directory. Compared with other profilers, this is an anti-pattern. Please align with all other profilers. This connects to #2461. 

It also seems like `synapse` prompts for the output directory - is this something we want to standardize on? @m-abulazm

src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history.sql
```diff
@@ -0,0 +1,45 @@
...
+        WHEN TOTAL_ELAPSED_TIME >= 60000 THEN 'MEDIUM'
+        ELSE 'SHORT'
+    END as QUERY_DURATION_CATEGORY,
+    CURRENT_TIMESTAMP() as EXTRACT_TIMESTAMP
+FROM SNOWFLAKE.ACCOUNT_USAGE.QUERY_HISTORY
+-- No NULL filters: legacy ingested SELECT * with no WHERE clause and we need
+-- parity. Cloud-services-only queries (NULL warehouse_name) carry the bulk of
+-- CREDITS_USED_CLOUD_SERVICES; system/replication queries have NULL user_name.
+-- Downstream analytics should LEFT JOIN + COALESCE so NULLs degrade cleanly.
+WHERE START_TIME >= DATEADD('day', -365, CURRENT_TIMESTAMP())
```
dgomez04: **Lookback window is hardcoded to 365 days across 7 SQL files, with no runtime override.** Pattern seen across `warehouse_usage.sql`, `query_history.sql`, `storage_usage.sql`, `user_activity.sql`, `pipe_usage.sql`, `automatic_clustering.sql`, `materialized_view_refresh.sql`.

The configurator doesn't prompt for a lookback window, `pipeline.py` does no `{lookback_days} `substitution, and `pipeline_config.yml` doesn't expose it as a parameter. The only way to narrow the window today is to hand-edit the SQL resources inside the installed package.

**In practice,** 365 days of `SNOWFLAKE.ACCOUNT_USAGE.QUERY_HISTORY` for an active account is easily tens to hundreds of millions of rows.

There's also review history pointing at this in `af7b919`, where the default was lowered to 30 days per @m-abulazm's feedback and reverted in `4ba6b7f`, without the configurability piece coming back.

Either:
- `ConfigureSnowflakeAssessment` collects `lookback_days` (1–365, default 365), persists it in `.credentials.yml`, and `pipeline.py` substitutes `{lookback_days}` at SQL load time.
- Reduce the lookback or document this decision in `snowflake.mdx` along with how to override.

src/databricks/labs/lakebridge/resources/assessments/snowflake/storage_usage.sql
dgomez04, line 1: **Document why this file doesn't filter system databases.** `database_objects.sql` filters out `SNOWFLAKE` and `UTIL_DB`

```sql
WHERE DELETED IS NULL
    AND DATABASE_NAME NOT IN ('SNOWFLAKE', 'UTIL_DB')
```

This is intentional because `SNOWFLAKE` shouldn't appear in `DATABASE_STORAGE_USAGE_HISTORY`, and `UTIL_DB`only appears in legacy accounts where it's a real customer database.

One way to defuse this is a header comment on this file along the lines of:
```sql
-- NOTE: intentionally no NOT IN ('SNOWFLAKE', 'UTIL_DB') filter here, even
-- though database_objects.sql filters those out. DATABASE_STORAGE_USAGE_HISTORY
-- only includes billed databases, so SNOWFLAKE (a shared DB) never appears,
-- and applying the filter zeroes out fresh/sandbox accounts. See 19746e1.
```

src/databricks/labs/lakebridge/resources/assessments/snowflake/function_lists/photon_functions.json
dgomez04, line 1: These functions are not consumed anymore. They should either be **deleted**,  or add a `README.md` on why they're being preserved.

src/databricks/labs/lakebridge/resources/assessments/snowflake/function_lists/snowflake_functions.json
dgomez04, line 1: These functions are not consumed anymore. They should either be deleted, or add a README.md on why they're being preserved.

tests/unit/assessment/test_snowflake_configurator.py
```diff
@@ -0,0 +1,57 @@
...
+        # Prompts.choice sorts options alphabetically; "env" is index 0, "local" is index 1.
+        r"Enter secret vault type \(local \| env\)": str(sorted(["local", "env"]).index("env")),
+        r"Enter Snowflake account URL.*": "myorg-myaccount.snowflakecomputing.com",
+        r"Enter username": "TEST_USER",
+        r"Enter warehouse name": "COMPUTE_WH",
+        r"Enter database name": "SNOWFLAKE",
+        r"Enter schema name": "ACCOUNT_USAGE",
+        r"Enter role": "ACCOUNTADMIN",
+        r"Enter Programmatic Access Token \(PAT\)": "fake-pat-token",
+        r"Do you want to test the connection to snowflake\?": "no",
```
dgomez04: This never tests the connection - I'm not sure how much value this brings.

tests/unit/connections/test_snowflake_connector.py
dgomez04, line 1: This file should be renamed. `SnowflakeConnector` is never imported, it only tests the URL helpers.

No test in this pull request ever invokes or tests the connector. 

At-least, I'd recommend to mock the connector and test the URL logic: 
(a) URL is built correctly for happy path 
(b) URL escapes a PAT containing /, @, +, %, (c) `KeyError` is raised cleanly when account is missing.

tests/unit/test_install.py
```diff
@@ -901,7 +901,6 @@ def test_configure_all_override_installation(  # FIXME
             r"Enter .* schema name": "tpch_sf1000",
             r"Enter target Databricks catalog name": "tpch",
             r"Enter target Databricks schema name": "1000gb",
-            # Profiler Configuration Prompts
             r"Select the source technology": "0",
```
dgomez04: With `PROFILER_SOURCE_SYSTEM = ["synapse", "snowflake", "mssql", "legacy_synapse", "oracle"]` and `Prompts.choice` sorting alphabetically, answer "0" selects `legacy_synapse`.

tests/unit/test_install.py
dgomez04: This points to `synapse_assessment` when the source tech is `legacy_synapse` (see comment above). Worth a sanity check to make sure we're testing what we expect.

docs/lakebridge/docs/assessment/profiler/snowflake.mdx
```diff
@@ -0,0 +1,47 @@
...
+databricks labs lakebridge configure-database-profiler
+```
+
+Pick **snowflake** when prompted, then paste the connection details and PAT.
+
+```bash
+databricks labs lakebridge execute-database-profiler
+```
+
+The extract is written to a DuckDB file (`profiler_extract.db`) in the configured output directory. It contains raw `SNOWFLAKE.ACCOUNT_USAGE` data — warehouse usage, query history, storage, users, and (when applicable) pipe / autoclustering / materialized-view refresh credits. Inspect it with:
```
dgomez04: The extract is written to a DuckDB file (`profiler_extract.db`) in the configured output directory" is overstating things — there's no configured output directory; it's just current working directory. **This description should be tightened.**

pyproject.toml
```diff
@@ -46,6 +46,8 @@ dependencies = [
   "requests>=2.28.1,<3",  # Matches databricks-sdk (and 'types-requests' below), to avoid conflicts.
   "pandas~=2.3.1", # Required for new configure assessment
   "redshift-connector~=2.1",
+  "snowflake-connector-python>=3.6.0",
```
dgomez04: A tighter floor like `>=3.13` (matches the lock) would prevent silent install regressions.

pyproject.toml
```diff
@@ -46,6 +46,8 @@ dependencies = [
...
+  "snowflake-sqlalchemy>=1.7.0",
```
dgomez04: Same for `snowflake-sqlalchemy>=1.7.0` vs locked `1.9.0`.

.gitignore
dgomez04, line 1: There are personal-workflow ignores that shouldn't be included in the shared `.gitignore`.

dgomez04 (changes requested):
Left comments - please reach out if any clarifications are needed.

src/databricks/labs/lakebridge/connections/database_manager.py
m-abulazm: the user is prompted if they want to test the connection. as well as flagging the profiler config `connector required` to `True`

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
m-abulazm: we need a dedicated PR to fix this asap for all sources. we have the issue https://github.com/databrickslabs/lakebridge/issues/2480 and an external contributor already opened a PR to fix it. lets use any sane value here and fix on the dedicated PR

src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history.sql
m-abulazm: nice catch

src/databricks/labs/lakebridge/resources/assessments/snowflake/function_lists/photon_functions.json
m-abulazm: we need to move those to profiler-etl

src/databricks/labs/lakebridge/resources/assessments/snowflake/function_lists/snowflake_functions.json
m-abulazm: same

tests/unit/assessment/test_snowflake_configurator.py
m-abulazm: we need integration tests https://github.com/databrickslabs/lakebridge/issues/2477

m-abulazm: not part of this PR

src/databricks/labs/lakebridge/assessments/_constants.py
jneil17: done in 8cc5ae1a. switched to `sorted(PLATFORM_TO_SOURCE_TECHNOLOGY_CFG.keys())`, which is what the TODO already called for, so the list can't drift from the config.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
jneil17: done in 8cc5ae1a. renamed the field to `pat`. we build the URL via `URL.create` ourselves so the key name is internal only, no need to keep `password`. the connector reads `connection["pat"]` now and the docs note it.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
jneil17: fixed in e5809f5d. when `secret_vault_type == "env"` the prompt now reads "Enter the environment variable name holding the PAT" instead of asking for the token. added a configurator test covering both local and env modes.

src/databricks/labs/lakebridge/connections/database_manager.py
jneil17: good catch, fixed in e5809f5d. swapped the f-string for `URL.create` exactly like the MSSQL/Oracle connectors, so the structural chars (`/`, `=`, `@`) get percent-escaped and SQLAlchemy can't misread the token. added a test asserting the escaping.

src/databricks/labs/lakebridge/connections/database_manager.py
jneil17: dropped in e5809f5d. went with the nested shape you suggested, `connection_config = self.config["connection"]`, and it raises if the block is missing. no silent fallback.

src/databricks/labs/lakebridge/resources/assessments/snowflake/storage_usage.sql
jneil17: documented in 2a670cad, refined in e4e769f8 with your wording: `DATABASE_STORAGE_USAGE_HISTORY` only lists billed databases so SNOWFLAKE never appears, and UTIL_DB is a real customer db in legacy accounts, so filtering would zero out their storage. added that as a header comment.

tests/unit/connections/test_snowflake_connector.py
jneil17: resolved with the rename. `test_snowflake_utils.py` covers the URL helper + connector URL building; the connection check itself runs through the generic health_check path.

tests/unit/test_install.py
jneil17: fixed in e5809f5d. the test selects snowflake by computed index off the sorted list instead of "0", so it tests what it claims.

tests/unit/test_install.py
jneil17: fixed in e5809f5d. now selects snowflake and asserts the snowflake extract path, so source tech and path line up.

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -0,0 +1,68 @@
+name: snowflake_assessment
+version: "1.0"
+extract_folder: "~/.databricks/labs/lakebridge_profilers/snowflake_assessment"
```
dgomez04, line 3: Once #2488 goes through we can remove this @m-abulazm.

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -0,0 +1,68 @@
...
+    flag: active
+
+  - name: user_activity
+    type: sql
+    extract_source: src/databricks/labs/lakebridge/resources/assessments/snowflake/user_activity.sql
+    mode: overwrite
+    frequency: daily
+    flag: active
+
+  # Background-service usage tables. Optional because accounts that don't use
```
dgomez04, line 47: The last three optional steps use `flag: active` and the pipeline will hard-fail the whole run if any active step errors. They will only degrade cleanly if the query returns zero rows, not when it errors. 

We need to either soften the or make these truly optional, but that would require changes to `pipeline.py`.

docs/lakebridge/docs/assessment/profiler/snowflake.mdx
dgomez04, line 1: In `pipeline.py:269:275`, on steps that generate zero rows, `_save_to_db` logs a warning and returns before creating the table. In the current state, this means that downstream consumners must tolerate missing tables. We should at least document this behavior.

tests/unit/test_install.py
```diff
@@ -1000,10 +1005,8 @@ def test_configure_all_override_installation(  # FIXME
     )
 
     expected_profiler_dash_config = ProfilerDashboardConfig(
-        source_tech="legacy_synapse",
-        extract_file_path=str(
-            Path("~/.databricks/labs/lakebridge_profilers/synapse_assessment/profiler_extract.db").expanduser()
-        ),
+        source_tech="snowflake",
+        extract_file_path="~/.databricks/labs/lakebridge_profilers/snowflake_assessment/profiler_extract.db",
```
dgomez04, line 1009: Switches the expected `extract_file_path` from an `.expanduser()`-wrapped value to a literal `~` path.

docs/lakebridge/docs/assessment/profiler/snowflake.mdx
```diff
@@ -0,0 +1,48 @@
...
+databricks labs lakebridge configure-database-profiler
+```
+
+Pick **snowflake** when prompted, then paste the connection details and PAT.
+
+```bash
+databricks labs lakebridge execute-database-profiler
+```
+
+The extract is written to a DuckDB file at `~/.databricks/labs/lakebridge_profilers/snowflake_assessment/profiler_extract.db`. It contains raw `SNOWFLAKE.ACCOUNT_USAGE` data — warehouse usage, query history, storage, users, and (when applicable) pipe / autoclustering / materialized-view refresh credits. By default each extract looks back 90 days. Inspect it with:
```
dgomez04, line 41: The documentation tells users the extract lives at `~/.databricks/labs/lakebridge_profilers/snowflake_assessment/profiler_extract.db`. If a user pastes that into the dashboard prompt, `install.py:450` stores it verbatim and the dashboard validates with:

```python
        # Validate the extract file path
        if not os.path.exists(local_file_path):
            logger.error(f"Local file not found: {local_file_path}")
            return False
```

`os.path.exists("~/…")` does not expand `~` → "Local file not found." This is pre-existing behavior, worth following up on.

dgomez04 (approved):
Please rewrite the pull request description to match the shipped diff. It currently advertises a much larger, different change.

Other than that, left small asks but don't block the pull request.

PR #2423, opened 2026-05-08
Fix: forward aggregate report_type to TriggerReconAggregateService

moomindani:
### What does this PR do?
Forwards `report_type=="aggregate"` from `TriggerReconService.trigger_recon` to `TriggerReconAggregateService.trigger_recon_aggregates` so programmatic callers that pass an aggregate report type get the actual aggregate comparison instead of a silent no-op.

`TriggerReconService.trigger_recon` validates `report_type` against `_RECON_REPORT_TYPES = {"schema","data","row","all","aggregate"}`, so `"aggregate"` is accepted at entry. However, `_do_recon_one` only branches on `{"schema","all"}` and `{"data","row","all"}`. The aggregate case falls through, no comparison runs, and `recon_capture.start()` records the run with `status=true`. Callers see a successful reconciliation even when source and target differ.

The CLI dispatch in `execute.py` already routes `AGG_RECONCILE_OPERATION_NAME` to `TriggerReconAggregateService.trigger_recon_aggregates`, but programmatic users of the public Reconcile API (notebooks, integrations) hit the silent no-op without this change.

This PR adds an early dispatch at the top of `trigger_recon`. The import is deferred because `trigger_recon_aggregate_service` already imports `trigger_recon_service` and a top-level import would be circular.

- Deferred import inside the function — done to avoid the circular import; behavior identical to `execute.py`'s existing dispatch
- All existing report types (`schema`, `data`, `row`, `all`) keep the original path

### Linked issues
None

### Functionality
- [ ] added new CLI command
- [x] modified existing command: programmatic API of `TriggerReconService.trigger_recon`

### Tests
- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

Manual end-to-end on Databricks Serverless with two tables of different row counts:
- Before: `aggregate=None`, `status=true`, no rows in `recon.aggregate_*`
- After: `aggregate=False`, `mismatch=1`, rows present in `recon.aggregate_metrics`

Unit tests added in `tests/unit/reconcile/test_verify_reconcile.py`:
- `test_trigger_recon_dispatches_aggregate_to_aggregate_service`: forwarding works and the data-path setup is skipped
- `test_trigger_recon_does_not_dispatch_non_aggregate`: `schema` / `data` / `row` / `all` keep the original path

---

Reopened from #2401 on an upstream branch to bypass the fork-PR OIDC restriction on JFrog auth (CI cannot run on fork PRs). All review comments and history are preserved on the original PR.


m-abulazm (changes requested):
we do not support this behavior. users can run aggregate reconcile using cli `databricks labs lakebridge aggregates_reconcile` or the service directly if using python

moomindani:
Reopening after addressing the cyclic-dependency concern we discussed.

The previous version used a deferred import inside `trigger_recon` to avoid the module-load cycle but kept the semantic dependency in both directions. Per the design feedback, that is not the right shape.

This revision breaks the cycle by extracting the shared helpers (`create_recon_dependencies`, `verify_successful_reconciliation`, `get_schemas`, plus a new `cleanup_intermediate_persist`) into a new module `reconcile/recon_service_helpers.py`. Both `TriggerReconService` and `TriggerReconAggregateService` now depend only on the helpers module — they no longer import each other. `trigger_recon_service.py` can import `TriggerReconAggregateService` at module load time without any cycle, and the deferred-import / pylint disable is gone.

Dependency graph after the change:
- `recon_service_helpers.py` — no service-level imports.
- `trigger_recon_aggregate_service.py` → `recon_service_helpers.py` only.
- `trigger_recon_service.py` → `recon_service_helpers.py` + `trigger_recon_aggregate_service.py` (one-way).

Behavior is unchanged: `report_type="aggregate"` on the public `TriggerReconService.trigger_recon` dispatches to `TriggerReconAggregateService.trigger_recon_aggregates`. Tests cover both dispatch and the non-aggregate path. Full unit suite passes (1160), pylint 10.00/10.

---

The diff is larger than the actual new logic — most of it is mechanical. Breakdown:

| File | Change | What's in it |
|---|---|---|
| `recon_service_helpers.py` (new) | +150 | 3 helpers moved verbatim from `TriggerReconService`; only `cleanup_intermediate_persist` (5 lines) is new |
| `trigger_recon_service.py` | -158 net | 3 helpers removed (moved out). New logic: 5-line aggregate dispatch + 1-line cleanup call |
| `trigger_recon_aggregate_service.py` | +18 | `TriggerReconService.X(...)` → bare `X(...)` (helper imports); no behavior change |
| `test_verify_reconcile.py` | +87 | 2 new dispatch tests; existing tests just drop the `TriggerReconService.` prefix |

The actual new behavior is: (1) the 5-line aggregate dispatch in `trigger_recon`, (2) the 5-line `cleanup_intermediate_persist` helper that dedupes the dbfs cleanup block (and resolves a pylint similarity warning), and (3) the 2 new dispatch tests. Everything else is moves and call-site renames.

moomindani:
@m-abulazm ping on this one, and a question rather than a nudge.

Your review said this behavior isn't something lakebridge supports, and that users should call `aggregates_reconcile` via the CLI or use the service directly. My follow-up comment addressed the cyclic-dependency shape but not that point, so the PR has been sitting in limbo since May.

If forwarding `report_type="aggregate"` through `TriggerReconService` isn't a direction you want, I'm happy to close this — no objection. If it is worth having, I'll rebase (it conflicts with main now) and re-run CI. Which would you prefer?

m-abulazm:
lets close this as discussed

PR #2390, opened 2026-04-22
Fix CHAR/VARCHAR length overflow when writing reconcile intermediate data

moomindani:
### What does this PR do?

Strip CHAR(n)/VARCHAR(n) length constraints from DataFrames before writing intermediate data to Delta during reconciliation. This prevents `DELTA_EXCEED_CHAR_VARCHAR_LIMIT` errors when source data contains space-padded CHAR values.

### Root cause

Some data sources (e.g., Teradata) return CHAR(n) values with space padding via JDBC, resulting in values that exceed the declared column length (e.g., a CHAR(16) column returning 16 digits + 16 spaces = 32 characters). Delta enforces CHAR/VARCHAR length constraints through column metadata (`__CHAR_VARCHAR_TYPE_STRING`), causing writes to fail for these padded values.

This was observed with Teradata via Lakehouse Federation but not with Lakebase (PostgreSQL) via Lakehouse Federation.

### Fix

Strip all column metadata via `col.alias(name, metadata={})` before writing intermediate DataFrames to Delta. This removes the constraint that Delta uses for length enforcement. The intermediate data is temporary and does not need metadata preservation.

### Linked issues

Fixes #2389

### Tests

- [x] manually tested with Teradata via Lakehouse Federation
- [x] added unit tests
- [ ] added integration tests

## Test plan

- [x] `test_strip_char_varchar_constraints_strips_metadata` — verifies CHAR/VARCHAR metadata is stripped
- [x] `test_strip_char_varchar_constraints_preserves_types` — verifies column types are preserved
- [x] All existing reconcile unit tests pass

PR #2428, opened 2026-05-09
Fix CHAR/VARCHAR length overflow when writing reconcile intermediate data

### LINKED ISSUE #2389: [BUG]: Reconcile fails with DELTA_EXCEED_CHAR_VARCHAR_LIMIT when source has CHAR/VARCHAR padded values

### Category of Bug / Issue

Reconcile bug

### Current Behavior

When reconciling data from external sources via Lakehouse Federation (e.g., Teradata), the reconcile process fails with `DELTA_EXCEED_CHAR_VARCHAR_LIMIT` when writing intermediate DataFrames to Delta volumes.

Teradata `CHAR(n)` columns contain space-padded values. The intermediate Delta table/volume preserves the `VARCHAR(n)` type constraint from the source schema, causing the write to fail when padded values exceed the declared length.

Version: v0.12.2

### Expected Behavior

Reconcile intermediate data should be written successfully regardless of source column padding. The intermediate tables/volumes are temporary and should use `STRING` (unbounded) instead of `VARCHAR(n)`/`CHAR(n)` to avoid length constraint violations.

### Steps To Reproduce

1. Set up Lakehouse Federation connection to Teradata
2. Create a Foreign Catalog pointing to a Teradata database with `CHAR(n)` columns
3. Run Lakebridge Reconcile with `data_source="databricks"` using the Foreign Catalog as source
4. Source table has `CHAR(16)` columns with space-padded values
5. Reconcile fails when writing intermediate DataFrame to Volume

\`\`\`
[DELTA_EXCEED_CHAR_VARCHAR_LIMIT] Value "4297690382031830                " exceeds char/varchar type length limitation.
Failed check: ((src_kyf IS NULL) OR (length(src_kyf) <= 16)).

recon_capture.py, line 92, in _write_df_to_volumes
    df.write.format(self._format).save(path)
\`\`\`

### Operating System

macOS

### Version

latest via Databricks CLI

moomindani:
### What does this PR do?

Strip CHAR(n)/VARCHAR(n) length constraints from DataFrames before writing intermediate data to Delta during reconciliation. This prevents `DELTA_EXCEED_CHAR_VARCHAR_LIMIT` errors when source data contains space-padded CHAR values.

### Root cause

Some data sources (e.g., Teradata) return CHAR(n) values with space padding via JDBC, resulting in values that exceed the declared column length (e.g., a CHAR(16) column returning 16 digits + 16 spaces = 32 characters). Delta enforces CHAR/VARCHAR length constraints through column metadata (`__CHAR_VARCHAR_TYPE_STRING`), causing writes to fail for these padded values.

This was observed with Teradata via Lakehouse Federation but not with Lakebase (PostgreSQL) via Lakehouse Federation.

### Fix

Strip all column metadata via `col.alias(name, metadata={})` before writing intermediate DataFrames to Delta. This removes the constraint that Delta uses for length enforcement. The intermediate data is temporary and does not need metadata preservation.

### Linked issues

Fixes #2389

### Tests

- [x] manually tested with Teradata via Lakehouse Federation
- [x] added unit tests
- [ ] added integration tests

## Test plan

- [x] `test_strip_char_varchar_constraints_strips_metadata` — verifies CHAR/VARCHAR metadata is stripped
- [x] `test_strip_char_varchar_constraints_preserves_types` — verifies column types are preserved
- [x] All existing reconcile unit tests pass

---

Reopened from #2390 on an upstream branch to bypass the fork-PR OIDC restriction on JFrog auth (CI cannot run on fork PRs). All review comments and history are preserved on the original PR.


m-abulazm (changes requested):
this looks like a deeper issue with teradata jdbc driver:
1. https://stackoverflow.com/questions/70596812/how-to-avoid-blank-spaces-while-loading-data-from-teradata-to-databricks
2. https://support.teradata.com/community?id=community_question&sys_id=ca9847a71b97fb00682ca8233a4bcb41
3. https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_5.html#BGBJECGD

anyway the better fix right now would be to trim variable length columns in `databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py:250` for databricks which will apply to teradata foreign catalogs

m-abulazm:
> this looks like a deeper issue with teradata jdbc driver:
> 
> 1. https://stackoverflow.com/questions/70596812/how-to-avoid-blank-spaces-while-loading-data-from-teradata-to-databricks
> 2. https://support.teradata.com/community?id=community_question&sys_id=ca9847a71b97fb00682ca8233a4bcb41
> 3. https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_5.html#BGBJECGD
> 
> anyway the better fix right now would be to trim variable length columns in `databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py:250` for databricks which will apply to teradata foreign catalogs

also please investigate the current query that gets built. it should already have trim as it is specified as the universal transformation

moomindani:
Looked into the existing transformation pipeline. Findings:

1. The universal default at `expression_generator.py:251` does include `trim`, and for `databricks` source the mapping has no CHAR/VARCHAR override (only ARRAY), so variable-length columns do fall through to the universal `[coalesce, trim]`.

2. But the path that hits the DELTA_EXCEED_CHAR_VARCHAR_LIMIT error bypasses `add_transformations` entirely. `SamplingQueryBuilder.build_query_with_alias()` (used in `Reconciliation._get_mismatch_data` to fetch the target sampling model at `reconciliation.py:360`) constructs the SELECT without calling `add_transformations` — unlike its sibling `build_query()`. So the resulting DataFrame keeps the CHAR(n) type and trailing-space-padded values, and the constraint fires when that DF is written to the intermediate volume.

3. Adding `partial(trim)` to `DataType_transform_mapping["databricks"]` alone wouldn't help this path because the transformation pipeline isn't invoked on that query.

Two proper fixes:
   A. Make `build_query_with_alias` go through `add_transformations` so the target sampling query trims variable-length columns at the source — consistent with `build_query()`.
   B. Keep the metadata-stripping in `_write_df_to_volumes` as defense at the write boundary.

I think A is cleaner and aligns with what you described. Happy to also add an explicit databricks CHAR/VARCHAR override per your suggestion for clarity (no-op vs the universal fallback, but documents intent). Want me to swap the PR to that shape?

m-abulazm:
@moomindani @bishwajit-db 
> Looked into the existing transformation pipeline. Findings:
> 
> 1. The universal default at `expression_generator.py:251` does include `trim`, and for `databricks` source the mapping has no CHAR/VARCHAR override (only ARRAY), so variable-length columns do fall through to the universal `[coalesce, trim]`.
> 2. But the path that hits the DELTA_EXCEED_CHAR_VARCHAR_LIMIT error bypasses `add_transformations` entirely. `SamplingQueryBuilder.build_query_with_alias()` (used in `Reconciliation._get_mismatch_data` to fetch the target sampling model at `reconciliation.py:360`) constructs the SELECT without calling `add_transformations` — unlike its sibling `build_query()`. So the resulting DataFrame keeps the CHAR(n) type and trailing-space-padded values, and the constraint fires when that DF is written to the intermediate volume.
> 3. Adding `partial(trim)` to `DataType_transform_mapping["databricks"]` alone wouldn't help this path because the transformation pipeline isn't invoked on that query.
> 
> Two proper fixes: A. Make `build_query_with_alias` go through `add_transformations` so the target sampling query trims variable-length columns at the source — consistent with `build_query()`. B. Keep the metadata-stripping in `_write_df_to_volumes` as defense at the write boundary.
> 
> I think A is cleaner and aligns with what you described. Happy to also add an explicit databricks CHAR/VARCHAR override per your suggestion for clarity (no-op vs the universal fallback, but documents intent). Want me to swap the PR to that shape?

`tgt_sampler.build_query_with_alias()` always runs for target tables which are databricks. how do we get `DELTA_EXCEED_CHAR_VARCHAR_LIMIT` for databricks tables? could it be a recon configuration problem? I also need to see the entire stack trace to have more context

moomindani:
@m-abulazm Thanks for pushing back — and please disregard my earlier `build_query_with_alias` comment, that analysis pointed at the wrong code path. Let me correct it.

**Why a "databricks" read hits the CHAR limit — it's about preserved type metadata, not the access mechanism.** The repro (issue #2389) uses `data_source="databricks"` against a **Lakehouse Federation Foreign Catalog over Teradata**. The distinction that matters here isn't JDBC-vs-not (Federation does talk to Teradata over JDBC under the hood) — it's whether the source DataFrame carries the `CHAR(n)` length as Spark column metadata:

| Path | source engine | column type | CHAR constraint metadata | volume write |
|------|--------------|-------------|--------------------------|--------------|
| Direct JDBC (`data_source="teradata"`) | teradata | typically unbounded `StringType` | dropped | OK |
| Federation (`data_source="databricks"`) | databricks | `CharType(16)` w/ `__CHAR_VARCHAR_TYPE_STRING` | **preserved** | **fails** |

Spark's JDBC reader generally maps `CHAR(16)` to an unbounded `StringType` (the length constraint is dropped), whereas UC Lakehouse Federation preserves the remote `CHAR(16)` as a length-typed Spark column. That's why this only reproduces on the Federation path, and why from the engine's point of view it looks like a "databricks" source — which is exactly the behavior you were questioning.

**Where it actually fails:** not the sampling query, but the source↔target hash-join checkpoint in `reconcile_data` (`compare.py:80`):

```python
df = source.alias("src").join(target.alias("tgt"), ...).selectExpr(...src_*, ...tgt_*)
df = persistence.write_and_read_df_with_volumes(df)   # -> _write_df_to_volumes (recon_capture.py:92)
```

The failing column `src_kyf` is a `src_`-prefixed **source** column carried into that joined DataFrame — not the target sample. The Teradata `CHAR(16)` value (`"4297690382031830"` + trailing spaces = 32 chars) violates the `length <= 16` constraint preserved on that column when the DF is checkpointed to the Delta volume.

**Stack trace.** Unfortunately I can't pull a fuller driver-side trace — the repro is in a customer environment we don't have access to, so what's captured in #2389 is all we have:
```
[DELTA_EXCEED_CHAR_VARCHAR_LIMIT] Value "4297690382031830                " exceeds char/varchar type length limitation.
Failed check: ((src_kyf IS NULL) OR (length(src_kyf) <= 16)).
  File "recon_capture.py", line 92, in _write_df_to_volumes
    df.write.format(self._format).save(path)
```
The trace doesn't show the caller frames, so I'm inferring the `compare.py:80` checkpoint from the `src_` prefix rather than confirming it directly — but every intermediate volume write funnels through the same `_write_df_to_volumes` sink regardless of caller.

**Config problem?** It's a legitimate, reproducible Lakehouse Federation setup rather than a misconfiguration — repro steps are in #2389 (Foreign Catalog as source + `data_source="databricks"`). It only surfaces for federated CHAR/VARCHAR sources, which is why native-to-native recon never hits it.

**Why the fix sits in `_write_df_to_volumes`:** all three intermediate volume writes (`compare.py:80`, `compare.py:453`, `reconciliation.py:376`) funnel through `write_and_read_df_with_volumes` → `_write_df_to_volumes`. Stripping CHAR/VARCHAR metadata at that single sink covers all of them regardless of which engine/source produced the constraint.

I deliberately avoided the trim-based approach (adding `partial(trim)` to the source transform mapping): trimming the actual values would change reconciliation semantics — it would mask genuine trailing-space differences between source and target. The metadata-strip keeps the real values intact and only drops the length constraint on the temporary intermediate, so the comparison stays faithful.

m-abulazm:
@moomindani please check if this is resolved on main branch. as well teradata is supported now as a source. if a teradata source still fails which should not, the real fix to to [apply TRIM](https://databrickslabs.github.io/lakebridge/docs/reconcile/configuration/#transformations) as the [official docs](https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_5.html#BGBJECGD) recommend

PR #2430, opened 2026-05-10
Allowing max_sample_size through specification value

ganeshdogiparthi-db:
### What does this PR do?

  Adds `sampling_options.specifications.value` to the reconcile Table config so users can control how many rows land in each sample bucket (mismatch, missing-in-source, missing-in-target) instead of being stuck at the hardcoded
  `_SAMPLE_ROWS = 50`. Previously, large sample sizes (e.g. 777) blew up the UNION-based sampling query against Spark Connect with `[RETRIES_EXCEEDED]`. Databricks sources now go through a temp-view sampling path that scales cleanly
  up to the new engine-aware ceiling.
  
  - `SamplingSpecifications`: `type` now defaults to `COUNT`, `value` defaults to `50`. Accepts `int | float`; `__post_init__` coerces COUNT values via `int(value)`, rejects bools/non-numeric, and keeps the `FRACTION` type disabled.
  - `SamplingOptions`: defaults wired so `sampling_options: { specifications: { value: N } }` is the minimal user-facing form.
  - `NormalizeReconConfigService` clamps engine-aware:
    - Databricks source → `[1, 50_000]` (silent cap)
    - Non-Databricks source → `[1, 50]` (cap above 50 with a WARNING log)
    - `value <= 0` floors to 50
  - `Table.get_max_sample_size()` returns the configured value (or 50). Plumbed through `compare.reconcile_data` and the threshold-mismatch limit in `reconciliation.py`.
  - `SamplingQueryBuilder` for the Databricks engine uses a UUID-named temp view of recon keys (`recon_keys_<hex>`) and joins source against it, replacing the per-row UNION pattern that didn't scale.
  - Flooring/capping logic removed from `Sampler._validate_sampling_options` since `NormalizeReconConfigService` now owns it.

  - Non-Databricks sources are intentionally capped at 50 — the temp-view path is Databricks-only for now.
  - The Databricks sampling query still relies on `filters.source` push-down for fast execution. If the source filter is `(col1, col2) IN (SELECT … FROM …)` Spark can't push it down to Parquet (observed 2.88B-row scans on
  tpcds_sf1000). Lead the filter with a literal predicate when scale matters.
  - `FRACTION` type stays disabled (raises `ValueError`). Value-range check is still present so re-enabling FRACTION later doesn't silently truncate via `int()`.

  ### Linked issues

  Resolves #..

  ### Functionality

  - [ ] added new CLI command
  - [x] new config field: `sampling_options` on `Table`


src/databricks/labs/lakebridge/reconcile/query_builder/sampling_query.py
```diff
@@ -109,6 +110,17 @@ def _get_join_clause(self, key_cols: list):
...
 
     def _get_with_clause(self, df: DataFrame) -> exp.Select:
+        if get_key_from_dialect(self.engine) == "databricks":
+            return self._with_clause_from_temp_view(df)
+        return self._with_clause_from_union(df)
+
+    @staticmethod
+    def _with_clause_from_temp_view(df: DataFrame) -> exp.Select:
+        view_name = f"recon_keys_{uuid.uuid4().hex}"
+        df.createOrReplaceTempView(view_name)
```
bishwajit-db, line 120: Temp view is session-scoped and never dropped, so it accumulates across many sample queries in a long-running job. Drop it after the query executes (return the view name and call `spark.catalog.dropTempView(name)`).

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
```diff
@@ -38,6 +45,20 @@ def _normalize_sampling(self, table: Table):
                 if normalized_sampling.stratified_columns
                 else None
             )
+            requested = normalized_sampling.specifications.value or _DEFAULT_MAX_SAMPLE_SIZE
+            is_databricks = isinstance(self.source, DatabricksDataSource)
+            if requested <= 0:
+                normalized_sampling.specifications.value = _DEFAULT_MAX_SAMPLE_SIZE
+            elif is_databricks:
+                normalized_sampling.specifications.value = min(requested, _MAX_SAMPLE_SIZE_DATABRICKS)
+            elif requested > _DEFAULT_MAX_SAMPLE_SIZE:
```
bishwajit-db, line 54: Silent regression for non-Databricks sources: cap drops from 400 to 50. `Sampler.sample()` uses engine-agnostic `DataFrame.sample()`/`.limit()`; only the UNION CTE is the bottleneck. Keep the old `[50, 400]` ceiling for non-Databricks, or flag this as a breaking change in release notes.

src/databricks/labs/lakebridge/reconcile/recon_config.py
```diff
@@ -47,23 +47,30 @@ class HashAlgoMapping:
...
 class SamplingSpecifications:
-    type: SamplingSpecificationsType
-    value: float
+    type: SamplingSpecificationsType = SamplingSpecificationsType.COUNT
+    value: int | float = 50
 
     def __post_init__(self):
         if not isinstance(self.type, SamplingSpecificationsType):
             self.type = SamplingSpecificationsType(str(self.type).lower())
         # Disabled
```
bishwajit-db, line 56: Put the `'FRACTION' type is disabled` raise first and unconditional; keep the range check below as inert future-restore code. Also fix the double space in `"greater than  0"`.

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
```diff
@@ -38,6 +45,20 @@ def _normalize_sampling(self, table: Table):
                 if normalized_sampling.stratified_columns
                 else None
             )
+            requested = normalized_sampling.specifications.value or _DEFAULT_MAX_SAMPLE_SIZE
```
bishwajit-db: `value or _DEFAULT_MAX_SAMPLE_SIZE` and the `requested <= 0` branch overlap (`value=0` never reaches the second). Drop one; prefer the explicit `<= 0` check and remove the `or` fallback.

src/databricks/labs/lakebridge/reconcile/query_builder/sampling_query.py
ganeshdogiparthi-db: This can be done. But it may lead to unnecessary breaks because all the comparisons and sampling are lazy evaluations. If we have to drop it, we have to cache the mismatch, missing df. Is this Okay for you ?

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
ganeshdogiparthi-db: Have we tested whether 400 works for a non-Databricks source? When I tested with 750, it failed for a table with two primary columns.

src/databricks/labs/lakebridge/reconcile/recon_config.py
ganeshdogiparthi-db: This was implemented this way for test coverage purposes. If I reverse it, the test coverage starts making noise.

src/databricks/labs/lakebridge/reconcile/query_builder/sampling_query.py
ganeshdogiparthi-db: Also, I don’t think view leaks will be significant. We are using UUIDs for the view names. Additionally, we are not dropping the views for threshold comparison.

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
ganeshdogiparthi-db: Done

src/databricks/labs/lakebridge/reconcile/recon_config.py
ganeshdogiparthi-db: Fixed typo.

PR #2151, opened 2025-11-17
Add SQL Server Profiler

goodwillpunning:
### What does this PR do?

This PR adds a new profiler for `mssql`. 

### Relevant implementation details

This profiler closely follows the implementation of the Azure Synapse profiler. However, this implementation provides a `last_execution_time` param for all `mssql` server queries, allowing for a future scheduler component to hook in an repeat queries. 

### Caveats/things to watch out for when reviewing:

This profiler will not support on-prem `mssql` servers.

### Linked issues

N/A

### Functionality

- [ ] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs lakebridge ...`
- [X] added new profiler component
- [ ] ... +add your own

### Tests

- [X] manually tested
- [ ] added unit tests
- [ ] added integration tests

PR #2431, opened 2026-05-11
[Feat/2093] Collapse DuckDB write helpers and move shared profiler infrastructure to `assessments/common/`

### LINKED ISSUE #2093: [Tech Debt] Collapse the db insert into single method to support both append and overwrite.
Seems that there are 2 similar functions for inserting profiler data, but radically different behavior. Workspace artifacts and monitoring metrics leverage this function, which does **not** support an append. However, serverless SQL pools and dedicated SQL pools leverage a function called `save_resultset_to_db`, which **does** support and append. Why the difference in behavior for these sets of Synapse objects? Perhaps combine the functions into a single function and use the `mode` param to dictate the behavior?

_Originally posted by @goodwillpunning in https://github.com/databrickslabs/lakebridge/pull/2020#discussion_r2387836818_

dgomez04:
# Collapse DuckDB insert helpers and move shared profiler helpers to `assessments/common/`

Closes #2093.

## Summary

`synapse/common/duckdb_helpers.py` carried two parallel write helpers — `save_resultset_to_db` (for `FetchResult` from SQL pools, supported `overwrite`/`append` with an inlined per-table schemas dict) and `insert_df_to_duckdb` (for `pd.DataFrame` from Azure SDK / monitoring, only `overwrite`, no schema). They behaved differently on errors (one swallowed, one re-raised), shared no code, and had drifted enough that adding the SQL Server profiler in #2151 was forced to reach across into `synapse/common/` and inline 13 new MSSQL schemas into the same Synapse-named function.

This pull request replaces both helpers with a single `save_to_duckdb(df, table_name, db_path, mode="overwrite", schema=None)` and lifts the shared profiler helpers out of `synapse/common/` to a new `assessments/common/` package so MSSQL (and future profilers) no longer have to import from a Synapse-named module.

### Folder layout after this PR

```
assessments/
├── common/                          NEW (sibling of synapse/, mssql/)
│   ├── cli.py                       arguments_loader
│   └── duckdb_helpers.py            save_to_duckdb, get_max_column_value_duckdb, _table_exists
├── synapse/
│   └── common/
│       ├── functions.py             only Azure SDK clients now (arguments_loader removed)
│       └── schemas.py               NEW: SYNAPSE_SCHEMAS
└── mssql/
    └── common/
        └── schemas.py               NEW: MSSQL_SCHEMAS
```

`synapse/common/duckdb_helpers.py` is gone. Zero cross-backend imports remain between `synapse/` and `mssql/`.

### `save_to_duckdb` behavior matrix

| mode | schema | table state | action |
|---|---|---|---|
| `overwrite` | provided | any | DROP, recreate with schema, INSERT (skip insert if df empty) |
| `overwrite` | none | exists | TRUNCATE, INSERT (preserves DDL types from prior run) |
| `overwrite` | none | missing | `CREATE TABLE AS SELECT *` (with `LIMIT 0` if df is empty) |
| `append` | provided | exists | INSERT |
| `append` | provided | missing | CREATE with schema, INSERT |
| `append` | none | exists | INSERT (DataFrame schema must match) |
| `append` | none | missing | `CREATE TABLE AS SELECT *` from first batch (logs a warning — brittle for incremental appends) |
| `append` | any | empty df | no-op |

The "schema = explicit string" path is what makes incremental appends robust: pandas dtype inference can drift between batches when a column is null-only in one batch but typed in the next, breaking the INSERT. With an explicit schema DuckDB owns the type and casts on insert.

## Commits (7, each independently buildable and bisectable)

1. **`Introduce shared profiler helpers in assessments/common/`** — adds the new shared modules, the per-backend `synapse/common/schemas.py`, slims `synapse/common/duckdb_helpers.py` to back-compat shims so unmigrated callers keep working, adds 10 unit tests.
2. **`Migrate workspace_extract.py to assessments/common/`** — 14 sites.
3. **`Migrate monitoring_metrics_extract.py to assessments/common/`** — 3 sites.
4. **`Migrate dedicated_sqlpool_extract.py to assessments/common/`** — 7 sites; uses `SYNAPSE_SCHEMAS`.
5. **`Migrate serverless_sqlpool_extract.py to assessments/common/`** — 10 sites; uses `SYNAPSE_SCHEMAS`.
6. **`Migrate mssql extracts to assessments/common/ and pin MSSQL_SCHEMAS`** — adds `mssql/common/schemas.py` (13 entries lifted from #2151), migrates `mssql/info_extract.py` (9 sites) and `mssql/activity_extract.py` (4 sites).
7. **`Remove deprecated profiler helpers from synapse/common/`** — deletes the file and removes the `arguments_loader` re-export.

Total: 47 call sites across 6 files migrated.

## Behavior changes (worth flagging in review)

- **Errors propagate consistently.** The old `save_resultset_to_db` swallowed exceptions, which meant a profiler run could silently drop data while reporting success. `pipeline.py`'s `_run_python_script` already converts re-raised script errors into the `{"status": "error"}` JSON it expects, so this is the correct behavior.
- **`overwrite` without schema now TRUNCATEs an existing table** instead of dropping and recreating with whatever pandas inferred. This preserves the DDL-declared types from a prior run — relevant for the workspace/monitoring extracts where the table was first created via CTAS.
- **Empty DataFrame with no columns** is now skipped with a warning instead of attempting an invalid `CREATE TABLE`.

## Argument-order swap

`insert_df_to_duckdb(df, db_path, table_name)` → `save_to_duckdb(df, table_name, db_path)`. Mechanical, but worth a sanity scan in the workspace and monitoring migrations.

## Why?

#2151 had to reach into `synapse/common/` from MSSQL because there was no shared profiler module to import from. Splitting "collapse the helpers" and "give them a proper home" into two PRs would mean PR #2 depends on PR #1 in non-trivial ways, and the merge conflict that kicked off this rebase would just resurface for the next contributor. Done together, the result is one coherent story and removes the cross-cutting smell entirely.

The `arguments_loader` move is in the same spirit and was free to bundle here because every migrated file was already updating its imports.

## Testing

- 10 new unit tests in `tests/unit/assessment/test_duckdb_helpers.py` cover overwrite/append, explicit-schema dtype pinning, empty DataFrames, and a regression guard for the incremental-append dtype drift case.
- Full `tests/unit/assessment/` suite passes (85 tests).
- `ruff check` and `pylint -E,F` clean across the touched packages.
- **Integration tests haven't been ran as of the current state of the pull request for either Synapse or MSSQL. Reviewers please advice.**

## Follow-ups (out of scope)

- **Cross-run history accumulation for `workspace_*` and `metrics_*` tables.** They're written in `overwrite` mode today, so each profiler run wipes the prior run's data. With `save_to_duckdb` accepting an optional schema, switching them to `append` is now a one-line-per-call change once a watermark column is agreed on. Will open a separate issue.


src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+
+import logging
+
+import duckdb
+import pandas as pd
+
+logger = logging.getLogger(__name__)
+
+
+def save_to_duckdb(
```
m-abulazm, line 22: lets split this method into two one for overwrite and another for append so it is easier to read.

src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+    df: pd.DataFrame,
+    table_name: str,
+    db_path: str,
+    mode: str = "overwrite",
```
m-abulazm: we can use Literal type to tighten the interface here so linter will make sure only valid values will come `typing.Literal["append" | "overwrite"]`

m-abulazm (changes requested):

src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
dgomez04: Set `[Literal](mode: Literal["overwrite", "append"])` and exported a `SaveMode` alias.

src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
dgomez04: I decided to apply the split internally and keep a single, public entry point. 
- Added `_save_overwrite` and `_save_append` private helpers. 
- Kept the single entry point because the call sites use `mode = "overwrite" if idx == 0 else "append"`, doing an external split would've added a lot of boilerplate code.

m-abulazm (approved):

sundarshankar89 (approved):
LGTM

PR #2441, opened 2026-05-13
legacy sql dw profiler

### LINKED ISSUE #2224: [BUG]: Profiler Support for Legacy DWH for Synapse Profiler

### Category of Bug / Issue

Profiler bug

### Current Behavior

Currently, We generate the list of sqlpools to the profiler by listing them using the workspace api

```python
dedicated_sqlpools = workspace.list_sql_pools()
all_dedicated_pools_list = [pool for poolPages in dedicated_sqlpools for pool in poolPages]
```

- Azure SQL Data Warehouse (Gen1/Gen2) was a standalone service
- Azure Synapse Analytics was introduced later as a unified workspace
- Legacy SQL DW instances were not automatically migrated to Synapse Workspaces
- Customers can still run legacy SQL DW without a workspace
- These legacy instances don't expose workspace-level APIs

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

```shell

```

### Logs Confirmation

### Sample Query

```shell

```

### Operating System

macOS

### Version

latest via Databricks CLI

m-abulazm:
### What does this PR do?
profile legacy sql dw

added extraction queries

### Linked issues

Resolves #2224 

### Functionality

- [ ] added new CLI command
- [X] modified existing command: `databricks labs lakebridge configure-database-profiler`

### Tests

- [x] manually tested
- [X] added unit tests
- [ ] added integration tests


docs/lakebridge/docs/assessment/profiler/legacy_sqldw.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+# Legacy SQL DW Profiler Details
+
+- [Prerequisites](#prerequisites)
+- [Profiled Tables and Views](#profiled-tables-and-views)
+- [Configure Connection to Legacy SQL DW](#configure-connection-to-legacy-sql-dw)
+- [Execute the Profiler](#execute-the-profiler)
+
+:::info
+"Legacy SQL DW" refers to **Azure Synapse dedicated SQL pool** —
```
dgomez04: Personally, I think we should rename this to something like `synapse-dedicated-pool` or something around those lines. The current name can be confusing, considering we offer multiple flavors of "SQL Server". 

I'd expect this to be a subset of the `synapse` profiler in the future.

docs/lakebridge/docs/assessment/profiler/legacy_synapse.mdx
```diff
@@ -0,0 +1,213 @@
...
+:::info
+"Legacy SQL DW" refers to **Azure Synapse dedicated SQL pool** —
+MPP-based platforms that expose Parallel Data Warehouse (PDW) Dynamic Management Views (`sys.dm_pdw_*`).
+For serverless SQL pools, Spark pools, and the Synapse workspace control plane, use the
+[Synapse profiler](./synapse.mdx) instead.
+:::
+
+## Prerequisites
+
+### 1. Download
```
dgomez04, line 21: We can probably tighten the requirements so people understand *why* they need to install these: 
- ODBC driver for SQL Server — required for `pyodbc` to talk to the dedicated SQL pool. ([download](https://learn.microsoft.com/en-us/sql/connect/odbc/download-odbc-driver-for-sql-server?view=sql-server-ver18))
- (Windows only) Microsoft Visual C++ Redistributable (x64) — required by `pyodbc` and the ODBC driver, which are native binaries linked against the MSVC runtime. ([download](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170))

docs/lakebridge/docs/assessment/profiler/legacy_sqldw.mdx
```diff
@@ -0,0 +1,213 @@
...
+## Prerequisites
+
+### 1. Download
+- ODBC driver for SQL Server (https://learn.microsoft.com/en-us/sql/connect/odbc/download-odbc-driver-for-sql-server?view=sql-server-ver18)
+- (Windows only) Visual C++ (https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170)
+
+### 2. SQL DW Authentication
+
+:::warning Attention:
+Multi-factor Authentication (MFA) is **not** supported. The Legacy SQL DW profiler connects via ODBC, which does not support MFA.
```
dgomez04: The MS ODBC Driver 17/18 supports several AAD modes that handle MFA: `ActiveDirectoryInteractive`, `ActiveDirectoryDeviceCode`, `ActiveDirectoryServicePrincipal`, plus token injection via `SQL_COPT_SS_ACCESS_TOKEN`. The only AAD mode that genuinely cannot do MFA is ActiveDirectoryPassword.

It's not that ODBC doesn't support MFA - it's the Lakebridge Connector currently only wires up SQL Authentication for this source. 

This could be reworded to something like: 

> The Legacy SQL DW profiler currently authenticates via SQL Authentication only. AAD-based auth modes (interactive, service principal, managed identity) are not yet wired up for this source — please use a SQL login with MFA disabled, and follow <link to follow-up issue> for AAD support.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -72,6 +72,45 @@ def run(self):
         logger.info(f"{source.capitalize()} Assessment Configuration Completed")
 
 
+# TODO: should we just reuse sql server profiler?
```
dgomez04: I believe this is something we should resolve right now or at least raise an issue for our unification proposals. 

This class is a near-duplicate of `ConfigureSqlServerAssessment` with a few values hardcoded and an extra `database` prompt.

src/databricks/labs/lakebridge/resources/assessments/synapse/dedicated_sqlpool_extract.py
```diff
@@ -140,7 +140,7 @@ def execute():
 
             table_name = "dedicated_session_requests"
             prev_max_end_time = get_max_column_value_duckdb("end_time", table_name, db_path)
-            session_request_query = SynapseQueries.list_dedicated_requests(prev_max_end_time)
+            session_request_query = SynapseQueries.list_dedicated_requests(sqlpool_name, prev_max_end_time)
```
dgomez04, line 146: This change looks unrelated to the `legacy-sql-dw` profiler. We should probably split this into a separate pull request?

tests/unit/test_install.py
```diff
@@ -1001,7 +1001,7 @@ def test_configure_all_override_installation(  # FIXME
     )
 
     expected_profiler_dash_config = ProfilerDashboardConfig(
-        source_tech="mssql",
+        source_tech="legacy sql dw",
```
dgomez04: Why does the expected default in this test now flip from `mssql` to `legacy sql dw`?

src/databricks/labs/lakebridge/resources/assessments/legacy_synapse/databases.sql
dgomez04, line 1: The table in `legacy_sqldw.mdx` claims this query "Lists each database with its edition and service objective" and lists `sys.database_service_objectives` as a source - but this query only returns `name` from `sys.databases`. Edition and service-objective columns never make it into extract. 

```sql
SELECT
    db.name AS database_name,
    dso.edition,
    dso.service_objective,
    dso.elastic_pool_name,
    CURRENT_TIMESTAMP AS extract_ts
FROM sys.databases AS db
LEFT JOIN sys.database_service_objectives AS dso
  ON db.database_id = dso.database_id;
```

**Or** just fix the documentation to drop the `sys.database_service_objectives` and the "edition and service objective" promises.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -188,9 +227,10 @@ def create_assessment_configurator(
...
         "mssql": ConfigureSqlServerAssessment,
         "synapse": ConfigureSynapseAssessment,
+        "legacy sql dw": ConfigureLegacySqlDwAssessment,
     }
 
     if source_system not in configurators:
         raise ValueError(f"Unsupported source system: {source_system}")
 
-    return configurators[source_system](product_name, prompts, source_system, credential_file)
+    return configurators[source_system](product_name, prompts, source_system, credential_file)  # type: ignore[abstract]
```
dgomez04: The `# type: ignore[abstract]` here is silencing a real mypy complaint: with three+ entries, mypy joins the value types up to `type[AssessmentConfigurator]`, which is abstract (because of `@abstractmethod _configure_credentials`), so the call site looks like it could be instantiating the `ABC`.

Functionally fine — none of the three registered classes are abstract — but the cleaner fix is to type the `dict` by callable signature instead of by class hierarchy:
```python
from collections.abc import Callable
ConfiguratorFactory = Callable[[str, Prompts, str, Path | str | None], AssessmentConfigurator]
configurators: dict[str, ConfiguratorFactory] = {
    "mssql": ConfigureSqlServerAssessment,
    "synapse": ConfigureSynapseAssessment,
    "legacy sql dw": ConfigureLegacySqlDwAssessment,
}
```
That removes the `# type: ignore` and future-proofs the registry: if someone tries to add `AssessmentConfigurator` itself, mypy will reject it (abstract classes aren't callable in the type system's view).

src/databricks/labs/lakebridge/resources/assessments/legacy_sqldw/pipeline_config.yml
```diff
@@ -0,0 +1,52 @@
+name: legacy_sqldw_assessment
+version: "1.0"
+extract_folder: "/tmp/data/legacy_sqldw_assessment"
```
dgomez04: Three things bundled here: 
- `/tmp/...` isn't a real Windows path - `Path("/tmp/...").expanduser()` in `pipeline.py` will resolve it to `<current drive>:\tmp\data\legacy_sqldw_assessment`, which "works" but isn't where any Windows user expects an extract to land. **The synapse profiler already solves this with `~/.databricks/labs/lakebridge_profilers/synapse_assessment`, which `expanduser()` resolves correctly on every OS.**
-  Two of the three profilers (`mssql`, `legacy sql dw`) write to `/tmp/data/...`, while `synapse` writes to `~/.databricks/labs/lakebridge_profilers/....` A user running multiple profilers ends up needing two different `--extract-file` paths in their `create-profiler-dashboard` command for no good reason.
- Volatility of `/tmp`. On macOS, the OS purges files in `/tmp` periodically (3+ days). A multi-hour profiler run shouldn't live there.

Could we standardize on the synapse-style path here, e.g.
```bash
extract_folder: "~/.databricks/labs/lakebridge_profilers/legacy_sqldw_assessment"
```
If that's done, the dashboard-upload example in legacy_sqldw.mdx ("Publish Profiler Summary Dashboard" section) needs to update to match.

**Note.** The mssql profiler has the same issue today.

dgomez04 (changes requested):
Left some comments. 

Even though we're operating in lift-and-shift mode today, I don't want us to keep deferring the consolidation and proper architecture of the profilers. Functionality is not sufficient - adoption will stall if the tool is painful to use - putting the user first matters. 

In more concrete thoughts: this profiler shares ~80% of its surface with the existing `synapse` profiler - it's worth a serious look whether this should be a mode within the `synapse` profiler rather than a separate one. We should track this issue separetly to not block this pull request.

src/databricks/labs/lakebridge/resources/assessments/synapse/dedicated_sqlpool_extract.py
m-abulazm: for simplicity, just added the fix here

src/databricks/labs/lakebridge/assessments/configure_assessment.py
m-abulazm: good point. fixed

docs/lakebridge/docs/assessment/profiler/legacy_sqldw.mdx
m-abulazm: I like that. fixed

src/databricks/labs/lakebridge/resources/assessments/legacy_synapse/databases.sql
m-abulazm: fixed docs

src/databricks/labs/lakebridge/resources/assessments/legacy_sqldw/pipeline_config.yml
m-abulazm: as noted on oracle PR, lets fix this in a followup

src/databricks/labs/lakebridge/assessments/configure_assessment.py
m-abulazm: good point. fixed

dgomez04 (approved):
LGTM 👍

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -80,6 +80,7 @@ def _create_connector(db_type: str, config: JsonObject) -> DatabaseConnector:
         "mssql": MSSQLConnector,
         "tsql": MSSQLConnector,
```
dgomez04: `"tsql": MSSQLConnector` is dead code — no profiler caller ever passes `"tsql"` as `source_tech`/`platform`/`db_type`, and it isn't registered in `PROFILER_SOURCE_SYSTEM`, `CONNECTOR_REQUIRED`, or `PLATFORM_TO_SOURCE_TECHNOLOGY_CFG`. Feel free to just remove.

src/databricks/labs/lakebridge/assessments/_constants.py
```diff
@@ -6,17 +6,17 @@
 PLATFORM_TO_SOURCE_TECHNOLOGY_CFG = {
     "synapse": "src/databricks/labs/lakebridge/resources/assessments/synapse/pipeline_config.yml",
     "mssql": "src/databricks/labs/lakebridge/resources/assessments/mssql/pipeline_config.yml",
+    "synapse_dedicated_sqlpool": (
```
sundarshankar89: What is the CUJ we are landing here? The user chooses directly legacy WH, or they select Synapse, and then we fork? 

I like them to select Synapse and then fork so we never expose them as `synapse_dedicated_sqlpool` to the customer.

sundarshankar89 (changes requested):
I m fine with the core changes, only concern I have is how this is wired we shouldn't be creating a new source rather it should be fork inside synapse profiler.

src/databricks/labs/lakebridge/assessments/_constants.py
m-abulazm: renamed to legacy_synapse

sundarshankar89 (approved):
LGTM

PR #2454, opened 2026-05-18
Pytest: quieter tests

asnare:
## Changes

This PR quietens down the output when running pytest: logs and output are only shown for tests that fail.

The `pip` install logic included a redirection configuration stdin/stdout/stderr that should have had no effect ("pass through") but somehow interfered with the mechanism that pytest uses to capture logs.


sundarshankar89 (approved):
LGTM

gueniai (approved):
LGTM

PR #2458, opened 2026-05-19
Remove explicit dependency: `cryptography`

asnare:
## Changes

This PR removes the explicit dependency on the `cryptography` package: previously we used this directly in our code for setting up connections to Snowflake, but it hasn't been needed since #2362.

The package is still present as a transient dependency, and therefore present in the lockfile, but we no longer need to declare an explicit dependency or track the supported versions ourselves.

### Linked Issues

This removes the need for #2447.

### Tests

- existing unit tests
- existing integration tests


sundarshankar89 (approved):
LGTM

gueniai (approved):

PR #2305, opened 2026-02-19
Redshift Profiler PR1: Redshift connection and credential support in database_manager

ysmx-github:
## Changes

This PR adds Redshift as a supported profiler assessment platform. 

### What does this PR do?

- Adds Redshift as a supported platform for the profiler assessment (alongside Synapse).
- Implements Redshift credential flows: database password, federated user, secrets manager (ARN), temporary credentials db user, temporary credentials IAM; allows SSL for all; configurator prompts and connection handling in database_manager.py.
- Files changed
  - .gitignore
  - pyproject.toml
  - src/databricks/labs/lakebridge/connections/database_manager.py

### Relevant implementation details

- Credentials: database password, federated user, secrets manager (ARN), temporary credentials db user, temporary credentials IAM. Federated and Secrets Manager use an optional lazy boto3 import.

### Linked issues

Resolves #..

### Functionality

- [ ] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs lakebridge ...`
- [ ] ... +add your own

### Tests

1. Manually Tested all credential flows on all clusters in  AWS Sandbox account aws-sandbox-field-eng (332745928618)
tests/resources/assessments/pipeline_config_main_redshift.yml: pipeline config that runs that script.
- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests

PR #2467, opened 2026-05-21
Fix version specification for `redshift-connector`

asnare:
## Changes

This PR fixes the version specification for the `redshift-connector` dependency, introduced in #2305: as written it allows for higher 2.x versions (e.g. 2.2) as well as downgrades to earlier versions of the connector (e.g. 2.1.7) that were not used during development or tested.

When using `~=` with semantic versioning, the complete minimum version needs to be specified.

### Linked issues

Noticed when reviewing #2466, which downgrades `redshift-connector` to escape a version constraint on `lxml`. Coincidentally includes #2464.


m-abulazm (approved):

gueniai (approved):
LGTM once conflicts are resolved

Issue #2449, opened 2026-05-15
BigQuery profiler CI infra: provision test project + Labs vault entries


take60:
## Context

PR #2445 adds the BigQuery profiler. The integration test (`tests/integration/assessments/test_bigquery_pipeline.py`) is currently `xfail`-gated on `TEST_BQ_PROJECT_ID` and `TEST_BQ_REGION` env vars (per reviewer ask in #2445 comment thread).

For CI to actually exercise the BQ profiler, those env vars need to be provisioned in the Databricks-Labs CI vault, similarly to how the Synapse and MSSQL integration tests get their credentials via the `databrickslabs/sandbox/acceptance` action.

## What's needed

- A BQ test project (dedicated CI project owned by the Labs platform team, OR read-only access to an existing project like `gcp-sandbox-field-eng`).
- Vault entries for:
  - `TEST_BQ_PROJECT_ID`
  - `TEST_BQ_REGION`
- Authentication mechanism for CI runners (likely service-account JSON key or Workload Identity Federation via `gcloud auth application-default login` in a setup step).

## Why now

Without this infra the BQ integration test stays in `xfail` mode in CI, which makes regressions easy to miss. Same gap exists today for external contributors without vault access — that's expected for them, but Lakebridge CI itself should run the test for real.

Filed per #2445 review comment #23.

m-abulazm:
duplicate of #2477

PR #2445, opened 2026-05-14
Add BigQuery profiler

take60:
### What does this PR do?

Adds a new BigQuery profiler so customers can drive BQ → Databricks migration sizing through the same `configure → execute → visualize` CLI flow Synapse and MSSQL support today.

### Relevant implementation details

This profiler closely follows the Synapse and MSSQL implementations. 16 SQL files plus the substitution engine are vendored one-time from the standalone GCP-native BQ profiler (`databricks-field-eng/fieldeng-platform-profilers`); Lakebridge is canonical post-port. The extract step runs serially across `(project, region)` and in parallel across the 16 SQLs within each iteration (`max_parallel_sqls`, default 8); results accumulate in memory then concat-and-overwrite to DuckDB — re-runs are idempotent. `google-cloud-bigquery-storage` is included so `to_dataframe()` streams over the BQ Storage Read API rather than paginating REST (~3× faster wall-clock on smoke). The pricing-analysis step (`bq_pricing_analysis.py`) computes the cost-projection layer in DuckDB. The Lakeview dashboard is cloned from the existing `bq_profiling` workspace dashboard and parameterized for per-customer catalog/schema.

### Caveats/things to watch out for when reviewing

- Schema default for new source-techs changes from `profiler` to `<source_tech>_profiler` (`bigquery_profiler` for BQ); Synapse keeps `profiler` for back-compat.
- The integration test (`test_bigquery_pipeline.py`) is env-gated on `TEST_BQ_PROJECT_ID` / `TEST_BQ_REGION`; it skips cleanly until vault provisioning lands, mirroring the Synapse pattern for external contributors.
- The deployment job currently uses classic compute (`compute.ClusterSpec` in `deployment/job.py`); this is an existing Lakebridge core limitation, not BQ-specific. The BQ profiler works around it the same way Synapse/MSSQL do (assumes classic-compute-allowed workspace).

### Linked issues

N/A

### Functionality

- [X] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command
- [X] added new profiler component

### Tests

- [X] manually tested (smoke against `gcp-sandbox-field-eng`; full end-to-end dashboard verified in a personal FEVM workspace)
- [X] added unit tests
- [X] added integration tests (env-gated)

PR #2472, opened 2026-05-21
Add BigQuery profiler (continuation of #2445)

take60:
Adds a BigQuery source-tech to the Lakebridge profiler.

## What it does

Runs 16 region-qualified `INFORMATION_SCHEMA` queries against the customer's BigQuery project(s) (configurable via `project.region` pairs) and writes 12 analysis tables into a local DuckDB file at `~/.databricks/labs/lakebridge_profilers/bigquery_assessment/profiler_extract.db`.

Pipeline structure mirrors the existing Synapse / MSSQL / legacy_synapse / Oracle source-techs:
- `configure-database-profiler` interactive prompt → writes `credentials.yml`
- `execute-database-profiler --source-tech bigquery` → runs `bq_metadata_extract.py`

## Design highlights

- **Auth**: ADC-only (`gcloud auth application-default login` or `GOOGLE_APPLICATION_CREDENTIALS` for non-interactive). Service-account impersonation supported via ADC.
- **Pair syntax**: `project.region` — matches Google's [fully-qualified resource-path convention](https://cloud.google.com/iam/docs/full-resource-names#bigquery), bulk-entry friendly for multi-project sizing.
- **Region qualifier in SQL**: required by BQ for INFORMATION_SCHEMA views per [docs.cloud.google.com](https://docs.cloud.google.com/bigquery/docs/information-schema-intro#region_qualifier) — the client's `location` parameter doesn't substitute for it.
- **Shared helpers**: uses `assessments.common.{cli, duckdb_helpers}` from #2431 — no duplication.

## Verification

- `make fmt && make lint && make test` clean (1187 unit tests pass, 2 skipped, 3 xfailed).
- End-to-end smoke test produces the expected 12 DuckDB tables.

## Follow-up PRs

- BigQuery profiler dashboard (Lakeview JSON) — separate PR after this merges.
- Integration test + `sandbox_bigquery_*` fixtures — separate PR after this merges.
- Continuation of #2445 (closed; see #2449 for the JFrog/OIDC fork-PR tracking).

---

### Review follow-ups
Addressed the blocking `table_storage` billing-model split in this PR (commit `60e08f13`; verified against a real BQ project).

The non-blocking notes are tracked for a follow-up in #2493:
- streaming/write-API profiling-window filter
- `timeline_analysis` `bytes_processed` diagnostic
- catch `GoogleAPIError` in the per-SQL handler
- CHANGELOG note for the profiler-schema default change


m-abulazm:
@take60 please resolve merge conflicts

take60:
resolved, thanks @m-abulazm

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/table_storage.sql
```diff
@@ -0,0 +1,44 @@
...
+     `%s`.INFORMATION_SCHEMA.TABLE_STORAGE
+   WHERE total_physical_bytes > 0
+   AND table_type IN ('BASE TABLE', 'MATERIALIZED VIEW')
+)
+SELECT *, 
+(active_physical_tb + long_term_physical_tb + time_travel_physical_tb + fail_safe_physical_tb) as total_physical_tb_to_migrate
+FROM TABLE_STORAGE_ANALYSIS;""", 
+metadatalevel)
+USING 
+metadatalevel AS metadatalevel
```
udaysat-db: This is the pre-billing-model version of the query. Upstream it was fixed to detect each dataset's `storage_billing_model` from `INFORMATION_SCHEMA.SCHEMATA.ddl` and roll up per (project, region, billing_model). Without it, storage cost is understated for any dataset on PHYSICAL billing. Suggested replacement (note it now uses two `%s` and passes `metadatalevel` twice — internally consistent with the substituter, which only rewrites the `SET metadatalevel` line):

```suggestion
-- billing model
-- ----------------------------------------------------------------------------
-- Each dataset is configured with `storage_billing_model = LOGICAL` (default)
-- or `PHYSICAL`. Both `active_logical_bytes` and `active_physical_bytes` are
-- populated in INFORMATION_SCHEMA.TABLE_STORAGE regardless — only the bytes
-- matching the dataset's billing model are actually billed. We extract the
-- billing model from each dataset's DDL via REGEXP and roll up per (project,
-- region, billing_model). Datasets that don't specify the option fall through
-- to LOGICAL (BQ default). The post-analysis job applies the per-region rate
-- to the right bytes for each model.

DECLARE metadatalevel STRING DEFAULT 'region-us';

-- SET metadatalevel to the format <project>.<region>
SET metadatalevel = 'my-gcp-project.region-us';

EXECUTE IMMEDIATE
FORMAT("""
WITH dataset_billing_model AS (
  SELECT
    schema_name,
    COALESCE(
      REGEXP_EXTRACT(ddl, r'''storage_billing_model\\s*=\\s*['"]?([^'",\\s)]+)'''),
      'LOGICAL'
    ) AS storage_billing_model
  FROM `%s`.INFORMATION_SCHEMA.SCHEMATA
),
table_storage_with_model AS (
  SELECT ts.*, COALESCE(dbm.storage_billing_model, 'LOGICAL') AS storage_billing_model
  FROM `%s`.INFORMATION_SCHEMA.TABLE_STORAGE ts
  LEFT JOIN dataset_billing_model dbm ON ts.table_schema = dbm.schema_name
  WHERE ts.total_physical_bytes > 0
    AND ts.table_type IN ('BASE TABLE', 'MATERIALIZED VIEW')
)
SELECT
    @metadatalevel as metadatalevel,
    SUM(IF(deleted=false, active_logical_bytes, 0)) / power(1024, 4) AS active_logical_tb,
    SUM(IF(deleted=false, long_term_logical_bytes, 0)) / power(1024, 4) AS long_term_logical_tb,
    SUM(active_physical_bytes) / power(1024, 4) AS active_physical_tb,
    SUM(active_physical_bytes - time_travel_physical_bytes) / power(1024, 4) AS active_no_tt_physical_tb,
    SUM(long_term_physical_bytes) / power(1024, 4) AS long_term_physical_tb,
    SUM(time_travel_physical_bytes) / power(1024, 4) AS time_travel_physical_tb,
    SUM(fail_safe_physical_bytes) / power(1024, 4) AS fail_safe_physical_tb,
    SUM(active_physical_bytes + long_term_physical_bytes
        + time_travel_physical_bytes + fail_safe_physical_bytes) / power(1024, 4)
      AS total_physical_tb_to_migrate,
    storage_billing_model,
    COUNT(DISTINCT table_schema) AS dataset_count
FROM table_storage_with_model
GROUP BY metadatalevel, storage_billing_model
ORDER BY storage_billing_model;""",
metadatalevel, metadatalevel)
USING
metadatalevel AS metadatalevel
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/analysis_types.json
```diff
@@ -0,0 +1,654 @@
...
+                    "type": "double",
+                    "nullable": true,
+                    "metadata": {}
+                },
+                {
+                    "name": "total_physical_tb_to_migrate",
+                    "type": "double",
+                    "nullable": true,
+                    "metadata": {}
+                }
```
udaysat-db, line 74: The `table_storage` schema needs the two new columns appended so the DuckDB write and downstream ingest match the corrected query:

```suggestion
                },
                {
                    "name": "storage_billing_model",
                    "type": "string",
                    "nullable": true,
                    "metadata": {}
                },
                {
                    "name": "dataset_count",
                    "type": "long",
                    "nullable": true,
                    "metadata": {}
                }
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
```diff
@@ -0,0 +1,43 @@
+-- This procedure summarizes streaming inserts, focusing on request counts,
```
udaysat-db: Heads-up: vs. the current upstream version, this file and `write_api_summary.sql` dropped the profiling-window filter (`WHERE start_timestamp >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL @profiling_window_in_days DAY)` and the corresponding `DECLARE`/`SET`/`USING`). As written they scan **all** streaming/write-API history regardless of the configured window — different results and more bytes scanned. Worth restoring for consistency with the other windowed queries.

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/timeline_analysis.sql
```diff
@@ -0,0 +1,116 @@
+-- This procedure runs through timelined historical workloads on BQ
```
udaysat-db, line 1: The upstream version also emits a `bytes_processed` per (hour × workload) diagnostic (the `JOB_BYTES_TERMINAL` / `BYTES_BY_HOUR_WORKLOAD` CTEs, plus a `bytes_processed` field in `analysis_types.json`). Not present here. Fine to defer, just flagging the drift.

src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
```diff
@@ -0,0 +1,297 @@
...
+            executor.submit(
+                _run_sql_for_iteration, sql_filename, sql_substituter, bq_client, project_region
+            ): sql_filename
+            for sql_filename in sql_files
+        }
+        for future in as_completed(future_to_file):
+            sql_filename = future_to_file[future]
+            try:
+                analysis_type, df, elapsed = future.result()
+            except (RuntimeError, ValueError, OSError) as exc:
```
udaysat-db: `except (RuntimeError, ValueError, OSError)` won't catch BigQuery failures — `bq_client.query(...).to_dataframe()` raises `google.api_core.exceptions.GoogleAPIError` subclasses (`BadRequest`, `Forbidden`, `NotFound`), which don't inherit from any of those three. So for the most common failure mode this `except` is skipped and you lose the per-`sql_filename`/`project_region` log attribution (it still surfaces via the top-level handler, just without context). Suggest catching `google.api_core.exceptions.GoogleAPIError` (or broadening to `Exception`, since you re-`raise`).

src/databricks/labs/lakebridge/install.py
```diff
@@ -464,25 +464,30 @@ def _prompt_for_new_profiler_dashboard_installation(self) -> ProfilerDashboardCo
...
         logger.info("Configuring profiler dashboard metadata.")
         catalog = self._configure_catalog()
-        schema = self._configure_schema(
-            catalog,
-            "profiler",
-        )
+        # Schema-per-source convention: each source-tech writes to its own `<source>_profiler`
+        # schema. Customers upgrading from a pre-convention install can copy their existing
+        # `profiler` schema to the new `<source>_profiler` schema manually.
+        schema_default = f"{source_tech}_profiler" if source_tech else "profiler"
```
udaysat-db, line 481: The schema default change from `"profiler"` → `f"{source_tech}_profiler"` affects **all** existing source-techs, not just BigQuery — a Synapse/Oracle user re-running `configure` will now be defaulted to a new schema. The migration note is good; just flagging that this behavior change rides inside an "Add BigQuery" PR and may warrant a changelog mention.

udaysat-db (changes requested):
Thanks for picking this up! The extraction scaffolding and the Lakebridge pipeline integration look clean, and mirroring the Synapse/Oracle source-tech pattern is the right call.

One **blocking** issue: this PR is based on an older snapshot of the GCP-native profiler, and the `table_storage` query has since been corrected upstream to split storage by **LOGICAL vs PHYSICAL billing model**. Without that split, physical-billing datasets are mis-costed in the migration sizing (both `active_logical_bytes` and `active_physical_bytes` are always populated, so a single rollup silently bills the wrong set of bytes). Fix + the matching schema fields are suggested inline.

A few **non-blocking** notes below (streaming window filter, a `bytes_processed` diagnostic, an exception-handling gap, and a cross-source schema-default change) — take or leave for this PR vs. a follow-up.

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/table_storage.sql
take60: Done in `60e08f13` — adopted the per-billing-model rollup.

One necessary tweak vs. the verbatim suggestion: the regex backslashes had to be doubled (`\\s` instead of `\s`). The regex lives inside the outer `FORMAT("""...""")` non-raw string, so BigQuery's outer string parser rejected the single-backslash version with `400 Syntax error: Illegal escape sequence: \s` before the inner raw string ever applied. Doubling them makes the outer string emit `\s` into the dynamic SQL as intended.

Verified end-to-end against a real BQ project (region `us`): `table_storage` now rolls up per `(project, region, storage_billing_model)` — e.g. `(LOGICAL, 35 datasets)` — with both new columns populated.

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/analysis_types.json
take60: Done in `60e08f13` — appended `storage_billing_model` (string) and `dataset_count` (long) to the `table_storage` schema, matching the corrected query output. Verified via a real-BQ extract (12 tables, new columns populated).

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
take60: Good catch — deferring to follow-up #2493. Confirmed: this snapshot dropped the `@profiling_window_in_days` filter so streaming/write-API queries scan all history. Will restore it there.

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/timeline_analysis.sql
take60: Thanks — deferring the `bytes_processed` per (hour × workload) diagnostic to follow-up #2493, as suggested.

src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
take60: Good catch — deferring to follow-up #2493. Agreed: `(RuntimeError, ValueError, OSError)` misses `google.api_core.exceptions.GoogleAPIError` subclasses, so the per-SQL/(project, region) log attribution is lost for the common BQ failure mode (the top-level handler still catches it). Will catch `GoogleAPIError` there.

src/databricks/labs/lakebridge/install.py
take60: Good flag — deferring to follow-up #2493. The `"profiler"` → `f"{source_tech}_profiler"` default does affect all source-techs; will add a CHANGELOG note (or scope it to bigquery) in that PR.

take60:
@udaysat-db thanks for the careful review! Summary of changes:

  **Blocking:** `table_storage` billing-model split 
  - Adopted your suggested query + the two new schema fields (`storage_billing_model`,
  `dataset_count`).
  - One tweak vs. the verbatim suggestion: doubled the regex backslashes (`\\s`, not `\s`) —
  the regex is inside the outer `FORMAT("""…""")` (non-raw) string, so BigQuery rejected the
  single-backslash version with `400 Syntax error: Illegal escape sequence: \s`. Doubling
  makes the outer string emit `\s` as intended.
  - Verified end-to-end on a real BigQuery project: rolls up per `(project, region, storage_billing_model)`, both new columns populated (e.g. `LOGICAL`, 35 datasets).

  **Non-blocking — deferred to #2493** (replied on each inline thread):
  - restore the `@profiling_window_in_days` filter on `streaming_summary` /
  `write_api_summary`
  - add the `bytes_processed` per (hour × workload) diagnostic to `timeline_analysis`
  - catch `google.api_core.exceptions.GoogleAPIError` in the per-SQL handler (current
  `(RuntimeError, ValueError, OSError)` misses it → per-SQL log attribution lost)
  - CHANGELOG note for the `"profiler"` → `f"{source_tech}_profiler"` schema-default change

  **Housekeeping:**
  - Rebased onto latest `main` + regenerated `uv.lock` (branch had fallen behind → CI merge
  ref hit a lockfile-parse error). Green now.

 Please take a look! thanks!

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/consumption_beyond_commitments_standard.sql
```diff
@@ -0,0 +1,430 @@
...
+billing report. Change the start_time and end_time values to match the desired
+window. */
+
+/* The following three variables (start_time, end_time, and edition_to_check)
+are the only variables that you need to set in the script.
+
+During daylight savings time, the start_time and end_time variables should
+follow this format: 2024-02-20 00:00:00-08. */
+
+SET start_time = current_timestamp() - INTERVAL profiling_window_in_days DAY;
```
udaysat-db, line 37:
```suggestion
SET profiling_window_in_days = 180;
SET start_time = current_timestamp() - INTERVAL profiling_window_in_days DAY;
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/consumption_beyond_commitments_enterprise.sql
```diff
@@ -0,0 +1,430 @@
...
+billing report. Change the start_time and end_time values to match the desired
+window. */
+
+/* The following three variables (start_time, end_time, and edition_to_check)
+are the only variables that you need to set in the script.
+
+During daylight savings time, the start_time and end_time variables should
+follow this format: 2024-02-20 00:00:00-08. */
+
+SET start_time = current_timestamp() - INTERVAL profiling_window_in_days DAY;
```
udaysat-db, line 37:
```suggestion
SET profiling_window_in_days = 180;
SET start_time = current_timestamp() - INTERVAL profiling_window_in_days DAY;
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/consumption_beyond_commitments_enterprise_plus.sql
```diff
@@ -0,0 +1,430 @@
...
+billing report. Change the start_time and end_time values to match the desired
+window. */
+
+/* The following three variables (start_time, end_time, and edition_to_check)
+are the only variables that you need to set in the script.
+
+During daylight savings time, the start_time and end_time variables should
+follow this format: 2024-02-20 00:00:00-08. */
+
+SET start_time = current_timestamp() - INTERVAL profiling_window_in_days DAY;
```
udaysat-db, line 37:
```suggestion
SET profiling_window_in_days = 180;
SET start_time = current_timestamp() - INTERVAL profiling_window_in_days DAY;
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
```diff
@@ -0,0 +1,43 @@
+-- This procedure summarizes streaming inserts, focusing on request counts, 
+-- row counts, input bytes, and specific error conditions.
+
+DECLARE metadatalevel STRING DEFAULT 'region-us';
+    
+-- SET metadatalevel to the format <project>.<region>
+SET metadatalevel = 'my-gcp-project.region-us';
```
udaysat-db:
```suggestion
DECLARE metadatalevel STRING DEFAULT 'region-us';
DECLARE profiling_window_in_days INT64 DEFAULT 180;

-- SET metadatalevel to the format <project>.<region>
SET metadatalevel = 'my-gcp-project.region-us';
SET profiling_window_in_days = 180;
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
```diff
@@ -0,0 +1,43 @@
...
+            total_requests,
+            0)) AS user_error,
+        SUM(
+            IF(
+            error_code IN ('CONNECTION_ERROR','INTERNAL_ERROR'),
+            total_requests,
+            0)) AS server_error,
+        SUM(IF(error_code IS NULL, 0, total_requests)) AS total_error
+    FROM
+    `%s`.INFORMATION_SCHEMA.STREAMING_TIMELINE_BY_PROJECT
```
udaysat-db:
```suggestion
    `%s`.INFORMATION_SCHEMA.STREAMING_TIMELINE_BY_PROJECT
    WHERE start_timestamp >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL @profiling_window_in_days DAY)
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
```diff
@@ -0,0 +1,43 @@
...
+    GROUP BY
+    start_timestamp
+    ORDER BY
+    start_timestamp DESC;
+""", metadatalevel)
+USING 
+metadatalevel AS metadatalevel;
```
udaysat-db:
```suggestion
USING
metadatalevel AS metadatalevel,
profiling_window_in_days AS profiling_window_in_days;
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/write_api_summary.sql
```diff
@@ -0,0 +1,41 @@
+-- This procedure calculates a per minute breakdown of successful and failed 
+-- append requests, split into error code categories
+
+DECLARE metadatalevel STRING DEFAULT 'region-us';
+    
+-- SET metadatalevel to the format <project>.<region>
+SET metadatalevel = 'my-gcp-project.region-us';
```
udaysat-db:
```suggestion
DECLARE metadatalevel STRING DEFAULT 'region-us';
DECLARE profiling_window_in_days INT64 DEFAULT 180;

-- SET metadatalevel to the format <project>.<region>
SET metadatalevel = 'my-gcp-project.region-us';
SET profiling_window_in_days = 180;
```

udaysat-db (commented):
Thanks for turning the `table_storage` change around — this looks right. You adopted the billing-model split query and added both `storage_billing_model` and `dataset_count`; the `GROUP BY (metadatalevel, storage_billing_model)` rollup and the schema line up cleanly, and the `\\s` escaping inside the non-raw `FORMAT("""…""")` is handled correctly. Appreciate you verifying it end-to-end on a live project before pushing. ✅

One thing I'd like to pull **into this PR** rather than defer: the configured `profiling_window_in_days` is silently ignored by **5 of the 13 windowed queries**. Compiling with a window of `30`, I traced the output:

- **`streaming_summary.sql`, `write_api_summary.sql`** — no time filter at all (the `WHERE start_timestamp >= …` predicate and its `USING` binding aren't present).
- **`consumption_beyond_commitments_{standard,enterprise,enterprise_plus}.sql`** — the substitution rule anchors on `SET profiling_window_in_days = 180;`, but these three only `DECLARE … DEFAULT 180;` with no `SET` line, so the rule matches nothing and the default stands. (The `consumption_through_*` variants *have* the `SET` line — that's why they work.)

The profiler runs every query per `(project, region)` and fans out across many pairs, so the window is the primary lever for bounding a large run. With these five ignoring it, a narrowed window still pays full 180-day cost and the outputs span inconsistent time horizons. Inline suggestions restore parity.

On the other deferred items: the **`bytes_processed` diagnostic** and **CHANGELOG note** are fine in #2493. The **`GoogleAPIError` handler** (`except (RuntimeError, ValueError, OSError)` at `bq_metadata_extract.py:155`) is also OK to defer, but worth flagging it's effectively dead for the primary failure mode — BQ raises `google.api_core.exceptions.*`, none of those three, so per-SQL attribution is lost. One-line fix if you want to grab it here.

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/write_api_summary.sql
```diff
@@ -0,0 +1,41 @@
...
+        SUM(
+            IF(
+            error_code IN (
+                'DEADLINE_EXCEEDED','ABORTED', 'INTERNAL', 'UNAVAILABLE',
+                'DATA_LOSS', 'UNKNOWN'),
+            total_requests,
+            0)) AS server_error,
+        SUM(IF(error_code = 'OK', 0, total_requests)) AS total_error,
+        FROM
+        `%s`.INFORMATION_SCHEMA.WRITE_API_TIMELINE
```
udaysat-db:
```suggestion
        `%s`.INFORMATION_SCHEMA.WRITE_API_TIMELINE
    WHERE start_timestamp >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL @profiling_window_in_days DAY)
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/write_api_summary.sql
```diff
@@ -0,0 +1,41 @@
...
+    GROUP BY
+    start_timestamp
+    ORDER BY
+    start_timestamp DESC;
+""", metadatalevel)
+USING 
+metadatalevel AS metadatalevel;
```
udaysat-db:
```suggestion
USING
metadatalevel AS metadatalevel,
profiling_window_in_days AS profiling_window_in_days;
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/substitutions.json
```diff
@@ -0,0 +1,340 @@
...
+    },
+    {
+        "file_path": "sql-client-run/streaming_summary.sql",
+        "substitutions": [
+            {
+                "search_text": "SET metadatalevel = 'my-gcp-project.region-us';",
+                "find_text": "my-gcp-project.region-us",
+                "replace_with_var": "project_region",
+                "is_rule_active": true
+            }
```
udaysat-db:
```suggestion
            {
                "search_text": "SET metadatalevel = 'my-gcp-project.region-us';",
                "find_text": "my-gcp-project.region-us",
                "replace_with_var": "project_region",
                "is_rule_active": true
            },
            {
                "search_text": "SET profiling_window_in_days = 180;",
                "find_text": "180",
                "replace_with_var": "profiling_window_in_days",
                "is_rule_active": false
            }
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/substitutions.json
```diff
@@ -0,0 +1,340 @@
...
+    },
+    {
+        "file_path": "sql-client-run/write_api_summary.sql",
+        "substitutions": [
+            {
+                "search_text": "SET metadatalevel = 'my-gcp-project.region-us';",
+                "find_text": "my-gcp-project.region-us",
+                "replace_with_var": "project_region",
+                "is_rule_active": true
+            }
```
udaysat-db:
```suggestion
            {
                "search_text": "SET metadatalevel = 'my-gcp-project.region-us';",
                "find_text": "my-gcp-project.region-us",
                "replace_with_var": "project_region",
                "is_rule_active": true
            },
            {
                "search_text": "SET profiling_window_in_days = 180;",
                "find_text": "180",
                "replace_with_var": "profiling_window_in_days",
                "is_rule_active": false
            }
```

src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
take60: Grabbed it here in `47e14bfb` — broadened to `except Exception` (it re-raises, so nothing's swallowed; now logs the per-SQL / `(project, region)` context for `GoogleAPIError` too, matching the top-level handler). Removing it from #2493.

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/consumption_beyond_commitments_standard.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/consumption_beyond_commitments_enterprise.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/resources/consumption_beyond_commitments_enterprise_plus.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/streaming_summary.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/write_api_summary.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/write_api_summary.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/sql-client-run/write_api_summary.sql
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/substitutions.json
take60: Applied in `47e14bfb` ✅

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/substitutions.json
take60: Applied in `47e14bfb` ✅

take60:
@udaysat-db pulled both items into this PR (`47e14bfb`):

**Window filter** — the 5 queries that ignored `profiling_window_in_days` now honor it:
- `consumption_beyond_commitments_{standard,enterprise,enterprise_plus}.sql` — added the `SET profiling_window_in_days = 180;` line so the existing substitution rule anchors.
- `streaming_summary.sql` / `write_api_summary.sql` — added the DECLARE/SET, the `WHERE start_timestamp >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL @profiling_window_in_days DAY)` filter, and the `USING` binding, plus their `substitutions.json` rules.

Verified on a live project with a 7-day window: `streaming_summary` 127→1 rows, `write_api_summary` 79→0 — filter is applied.

**GoogleAPIError handler** — broadened `bq_metadata_extract.py` per-SQL `except` to `Exception` (re-raises, so nothing's swallowed; now keeps per-SQL/(project,region) log attribution for BQ errors).

Remaining defers (`bytes_processed` diagnostic, CHANGELOG note) tracked in #2493. PTAL !

udaysat-db (approved):
Thanks @take60 — all addressed, LGTM. ✅

- **Window filter**: all 5 previously-broken queries now honor `profiling_window_in_days` (verified 13/13 windowed queries substitute the configured value locally). Nice live confirmation with the 7-day run (`streaming_summary` 127→1, `write_api_summary` 79→0).
- **GoogleAPIError**: broadening the per-SQL handler to `except Exception` (re-raise + per-`(project,region)` attribution) is a clean fix.
- `table_storage` billing-model split looks good.

Good to defer `bytes_processed` and the CHANGELOG note to #2493. 🚢

take60:
Thanks, @udaysat-db !

Rebased onto latest `main` to clear the conflict. The only overlap was in `install.py`: #2488 reworked the profiler extract path into a `--output-folder`-driven prompt, which collided with this PR's per-source `extract_file_path` default. Resolved by taking #2488's new prompt and dropping the now-redundant default (the schema-per-source change is unaffected). No code changes beyond that — re-approve when you get a chance

src/databricks/labs/lakebridge/resources/assessments/bigquery/pipeline_config.yml
```diff
@@ -0,0 +1,20 @@
+name: bigquery_assessment
+version: "1.0"
+extract_folder: "~/.databricks/labs/lakebridge_profilers/bigquery_assessment"
```
udaysat-db: Leftover from the path refactor: every other source-tech's `pipeline_config.yml` dropped `extract_folder`, but this one still has it. It's now dead config (`load_config_from_yaml` reads only `name`/`version`/`steps`), so removing it keeps things consistent:
```suggestion
```

src/databricks/labs/lakebridge/install.py
```diff
@@ -463,21 +463,24 @@ def _prompt_for_new_profiler_dashboard_installation(self) -> ProfilerDashboardCo
         source_tech = self._prompts.choice("Select the source technology", PROFILER_SOURCE_SYSTEM)
         extract_file_path = self._prompts.question("Enter the path to the profiler output file").strip()
```
udaysat-db: This dropped the default that pointed at the generated extract file. The profiler now writes a date+version-stamped name (`profiler_extract_{platform}_{version}_{YYYYMMDD}.db`) the user can't easily predict, so the dashboard step has no way to default to it — they'd have to read the profiler log and hand-type the path. Could this default to the resolved file under `default_output_folder(source_tech)` (e.g. the latest matching extract) so the profiler→dashboard handoff stays automatic? Also: the profiler prompt asks for an output *folder*, this asks for an output *file* — worth aligning.

udaysat-db (commented):
Took another look after the latest push (`1af259f9`) — the extract-path refactor (centralizing the DB path and the new `profiler_extract_{platform}_{version}_{date}.db` filename) reads cleanly, and the per-platform folder base is preserved. A few things to tidy before merge — two inline, two here:

- **`assessments/pipeline.py` `make_profiler_db_filename`** — `datetime.now()` means each run on a new day (or version) writes a fresh `.db` with no cleanup, and nothing downstream points at "the latest". If the timestamped accumulation is intentional (traceability) 👍 — just flagging it's a behavior change from the fixed `profiler_extract.db`, and it's what makes a sensible dashboard-path default (see the `install.py` note) non-trivial.
- **`assessments/profiler.py` `Profiler.create`** — `PLATFORM_TO_SOURCE_TECHNOLOGY_CFG[platform]` now indexes directly instead of `.get(platform, None)`, so an unrecognized platform raises `KeyError` rather than the prior graceful `None`. cli validates `source_tech` upstream so it's likely unreachable, but a clearer error (or keeping the guarded lookup) would be friendlier.

take60:
@udaysat-db thanks — all four tidied in `0299b67e`:

**1. `extract_folder` (BigQuery config):** removed, so it matches the post-#2488 shape. (FWIW the `redshift/*` configs still carry it — flagging, but left out of this PR's scope.)

**2. Dashboard extract-path default (`install.py`):** added `_default_extract_file_path(source_tech)` — defaults the prompt to the newest `profiler_extract_{source}_*.db` under `default_output_folder(source_tech)`, so the profiler→dashboard handoff stays automatic despite the timestamped filename (required prompt when none exists yet). Kept it asking for a file since the dashboard needs the specific `.db`; the default means it's rarely hand-typed now.

**3. Timestamped filename accumulation:** treating it as intentional traceability per your 👍 — #2's default covers the "nothing points at latest" downside. A retention/cleanup policy feels like a separate #2488 follow-up; happy to file one.

**4. `Profiler.create` lookup:** now `.get(platform)` with a clear `ValueError` listing supported platforms, instead of the bare `KeyError`.

PTAL !

src/databricks/labs/lakebridge/resources/assessments/bigquery/pipeline_config.yml
take60: Done in `0299b67e` ✅

src/databricks/labs/lakebridge/install.py
take60: Done in `0299b67e` ✅

take60: Superseded by #2494 — which just landed in `main` and addresses this exact concern with a prompt hint (`Look for "Profiler extract written to" in the execute logs`). Rebased onto it and dropped my custom default so the fix lives in one place. (`c06ed87e`)

take60:
Rebased onto `main` again (`c06ed87e`) — #2494 touched the same profiler-dashboard prompt line. Resolved by adopting #2494's hinted prompt and dropping my earlier smart default (and its helper/import), so the predictability fix stays in one place. The `Profiler.create` `.get` guard and the schema-per-source default are unchanged. No other overlap — verified #2494's `arguments_loader`/`pipeline`/`profiler` changes don't affect the BigQuery path.

take60:
@udaysat-db @m-abulazm this is green again after rebasing past the recent profiler changes on `main` (#2488  and #2494), and all your review points are resolved. Since it keeps re-conflicting with the active profiler work, I've turned on auto-merge (squash) , so a re-approve when you have a sec is the only thing left and it'll merge itself. Thanks for the reviews!

udaysat-db (approved):
Both follow-ups from my last pass are addressed at `c06ed87e` — LGTM. ✅

- **`pipeline_config.yml`** — dead `extract_folder` removed; the file is now `name`/`version`/`steps`, consistent with the other source-techs.
- **Dashboard extract path** (`install.py`) — rebased onto #2494, custom default dropped, and the prompt now carries the inline hint (`Look for "Profiler extract written to" in the execute logs`). This resolves both the unpredictable-path concern and the earlier file-vs-folder mismatch, with the fix living in one place.

Thanks for the thorough back-and-forth and the live verifications throughout. 🚢

take60:
Thanks @udaysat-db for the review!!

@gueniai @akashjaiswaldbr I would really appreciate if if you could review this when you have time.

docs/lakebridge/docs/assessment/profiler/bigquery.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+:::warning Attention:
+Each source-tech writes to its own default schema (`bigquery_profiler` for BigQuery, `mssql_profiler` for MSSQL, `synapse_profiler` for Synapse). Reusing one schema across source-techs is not supported and may overwrite tables.
+:::
+
+This deploys:
+
+- The Lakeview dashboard `lakebridge_bigquery_profiler_summary` to your workspace.
+- An ingestion job that uploads the DuckDB extract to UC and writes one Delta table per DuckDB table under `<catalog>.<schema>`.
+- A `validation_report` Delta table summarizing extract health (WARN-everywhere — non-blocking).
```
m-abulazm: is this part of this PR? please remove, those visuals will be part of the desktop app

akashjaiswaldbr (commented):
A few additional notes on top of the earlier round (these don't overlap with the billing-model / profiling-window / except-type / dashboard-default threads). All three are about config/runtime robustness rather than the SQL itself.

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/vars.json
```diff
@@ -0,0 +1,4 @@
+{
+    "profiling_window_in_days": "365",
```
akashjaiswaldbr: `vars.json` looks like dead config — it's referenced nowhere in the Python or tests. Runtime substitution values come exclusively from the `SqlSubstituter` constructor kwargs in `_run_iteration` (`project_region`, `profiling_window_in_days`). Worse, the values here disagree with what actually runs: `profiling_window_in_days: "365"` contradicts the real default of **180** (the configure prompt, the `execute()` fallback, and every SQL's baked `SET`), and `time_format: "%Y-%m-%d"` contradicts `timeline_analysis.sql`, which hard-sets `TIME_FORMAT` to hourly `'%Y-%m-%dT%H'`. A maintainer reading this file will draw the wrong conclusions. Suggest either wiring it up as the single source of truth or deleting it.

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/sql_substituter.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        rules = self._rules_by_filename.get(sql_filename, [])
+        if not rules:
+            logger.debug(f"No substitution rules for {sql_filename}; using SQL as-is.")
+            return raw_sql
+
+        lines = raw_sql.splitlines(keepends=True)
+        for sub in rules:
+            var_name = sub.get("replace_with_var")
+            if var_name not in self._variables:
```
akashjaiswaldbr: `substitute()` applies a rule whenever its `replace_with_var` is present in the kwargs and never inspects `is_rule_active`. Since `_run_iteration` always passes `profiling_window_in_days`, all 13 `profiling_window_in_days` rules in `substitutions.json` fire despite being marked `"is_rule_active": false`. (The `time_format` rules don't fire — but only because `time_format` is never passed, not because the flag is `false`.) The docstring documents this divergence, but a per-rule boolean that has zero runtime effect and reads as the *opposite* of what happens is a maintenance trap — note an accepted suggestion in this PR added rules with `is_rule_active: false` presumably expecting them to be off. Suggest either dropping the field from the vendored schema or honoring it.

src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
```diff
@@ -0,0 +1,297 @@
...
+        if not sql_files:
+            raise RuntimeError("All SQL files excluded by config; nothing to extract.")
+
+        substitutions = json.loads(_load_resource_text("common", "substitutions.json"))
+        analysis_types = json.loads(_load_resource_text("common", "analysis_types.json"))
+
+        accumulators: dict[str, list[pd.DataFrame]] = {at: [] for at in set(SQL_FILE_TO_ANALYSIS_TYPE.values())}
+        accumulator_lock = threading.Lock()
+
+        for pair in pairs:
```
akashjaiswaldbr, line 215: Resilience for multi-project sizing: this per-pair loop has no isolation, and `_run_iteration` re-raises on the first SQL failure, so a single project with (say) a missing reservations-admin grant or one unreadable region aborts extraction for *all* other projects — and since the write happens once at the end with overwrite semantics, nothing is salvaged. For a tool whose headline use case is bulk multi-project sizing, consider soft-failing a single `(project, region)` (log + continue) and surfacing a per-pair status in the final JSON payload, so one permission gap doesn't nuke the whole run. (Distinct from the earlier note on the exception *type* at line 155 — this is about the continue-vs-abort policy.)

src/databricks/labs/lakebridge/resources/assessments/bigquery/pipeline_config.yml
```diff
@@ -0,0 +1,19 @@
+name: bigquery_assessment
+version: "1.0"
+steps:
+  - name: bq_metadata_extract
+    type: python
+    extract_source: src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
+    mode: overwrite
+    frequency: once
+    flag: active
+    dependencies:
```
m-abulazm: please pull in main. this functionality is gone and dependencies are specified in project dependencies. you only have to add the new ones "google ones I guess" and run `make lock-dependencies`

src/databricks/labs/lakebridge/resources/assessments/bigquery/pipeline_config.yml
```diff
@@ -0,0 +1,19 @@
+name: bigquery_assessment
+version: "1.0"
+steps:
+  - name: bq_metadata_extract
+    type: python
+    extract_source: src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
```
m-abulazm: this is gone as well and it is now a cli argument configured by end users

docs/lakebridge/docs/assessment/profiler/bigquery.mdx
take60: Removed — dropped the entire `## Generate the dashboard` section (the dashboard JSON isn't part of this PR anyway; it's a post-merge follow-up). Done in `8559207b`.

src/databricks/labs/lakebridge/resources/assessments/bigquery/pipeline_config.yml
take60: Done. Rebased onto `main` and picked up #2485 (profiler Python steps now run on the parent interpreter), so the per-step `dependencies:` block is dead — removed it. Moved the BigQuery deps (`google-cloud-bigquery`, `google-cloud-bigquery-storage`, `google-auth`, `db-dtypes`) into `[project].dependencies` next to the existing azure ones and ran `make lock-dependencies`. `8559207b`.

src/databricks/labs/lakebridge/resources/assessments/bigquery/pipeline_config.yml
take60: The user-facing extract path (`extract_folder`) was already dropped in `0299b67e` — post-#2488 it's the `--output-folder` CLI argument. The file now matches the other source-techs' schema exactly (`name`/`version`/`steps`, each step with `name`/`type`/`extract_source`/`mode`/`frequency`/`flag`); no path/folder config field remains. I kept `extract_source` since that's the in-repo script path `pipeline.py` reads to locate the step (all siblings carry it) — let me know if you meant something else here. `8559207b`.

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/vars.json
take60: Deleted. It was dead config — referenced nowhere in the code or tests — and its values had drifted from what actually runs (`profiling_window_in_days: 365` vs the real default of 180; `time_format: %Y-%m-%d` vs the hourly `%Y-%m-%dT%H` baked into `timeline_analysis.sql`). Runtime substitution values come solely from the `SqlSubstituter` constructor kwargs, so there was no single-source-of-truth to preserve. `8559207b`.

src/databricks/labs/lakebridge/resources/assessments/bigquery/common/sql_substituter.py
take60: Dropped the `is_rule_active` field entirely from `substitutions.json` (all rules) and removed the docstring note. Since it had zero runtime effect, removing it is behavior-preserving — the windowed substitutions verified earlier (e.g. `streaming_summary` honoring the profiling window) still fire. The substituter contract is now simply: a rule applies iff its `replace_with_var` is supplied in the constructor kwargs (absent → SQL keeps its baked-in default). `8559207b`.

src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
take60: Implemented. The per-`(project, region)` loop now soft-fails: a failing pair is logged and skipped instead of aborting the run, the pairs that succeed are still written to DuckDB, and each pair's outcome is reported in the final JSON payload under a new `pairs` array. If *every* pair fails the step raises so the top-level handler reports a structured error (rather than "success" over an empty/stub-only extract).

Verified end-to-end against real BigQuery with a 2-pair config (a nonexistent project + `gcp-sandbox-field-eng.us`):

```json
"pairs": [
  {"project": "lakebridge-nonexistent-zzz999", "region": "us", "status": "error",
   "message": "404 ... Not found: Project lakebridge-nonexistent-zzz999"},
  {"project": "gcp-sandbox-field-eng", "region": "us", "status": "success"}
]
```

→ overall `status: success`, exit 0, and the healthy pair's 12 tables were written (e.g. `jobs_timeline_by_reservations` 1534 rows). An all-pairs-fail config returns `status: error` with a non-zero exit. Added unit tests for both the soft-fail and all-fail paths. `8559207b`.

take60:
@m-abulazm @akashjaiswaldbr thanks for the latest round of review. All comments are addressed and the branch is green again (rebased onto latest `main`, past #2485 — profiler Python steps now run on the parent interpreter).

**@m-abulazm**
- `pipeline_config.yml`: removed the per-step `dependencies:` block (dead post-#2485) and moved the BigQuery deps into `[project].dependencies` alongside the existing azure ones, then re-ran `make lock-dependencies`.
- The extract path is no longer a config field — it's been the `--output-folder` CLI argument since #2488. The file now matches the other source-techs' schema (`name`/`version`/`steps`).
- `bigquery.mdx`: removed the dashboard section (those visuals belong to the desktop app / a follow-up).

**@akashjaiswaldbr**
- Deleted `common/vars.json` (dead config whose values had drifted from what actually runs).
- Dropped the no-op `is_rule_active` field from `substitutions.json` — behavior-preserving, the windowed substitutions still apply.
- Added per-`(project, region)` resilience: a failing project/region is now logged and skipped instead of aborting the whole run, the pairs that succeed are still written, and each pair's outcome is reported in the result payload under a new `pairs` array. If every pair fails the step reports a top-level error. Verified end-to-end against real BigQuery with a bad project plus a good one — the bad pair surfaced its 404 and was skipped, the good pair's 12 tables were written, exit 0. Added unit tests for both the soft-fail and all-fail paths.

`make fmt && make lint && make test` clean (pylint 10.0/10) and CI is green. PTAL when you have a moment.

src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
m-abulazm: why are we doing this? is at a required feature? because multiple projects and multiple regions does make it more complex than just a single project and region

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -274,6 +302,60 @@ def _configure_credentials(self) -> str:
...
+        )
+        max_parallel_sqls = int(
+            self.prompts.question(
+                "Enter max parallel SQLs per (project, region) iteration", default="8", valid_number=True
+            )
+        )
+
+        logger.info("Please configure profiler settings:")
+        bigquery_profiler = {
+            "redact_query_text": self.prompts.confirm("Redact query text in extracted data?"),
```
m-abulazm: this needs to be always redacted

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -274,6 +302,60 @@ def _configure_credentials(self) -> str:
...
+        logger.info("Please provide BigQuery connection settings:")
+        logger.info(
+            "\nProject/region pairs are entered as `project.region`, comma-separated. "
+            "Example: my-proj-a.us, my-proj-b.eu-west-1\n"
+        )
+        pairs_raw = self.prompts.question("Enter BigQuery project/region pairs (project.region, comma-separated)")
+        pairs = _parse_project_region_pairs(pairs_raw)
+
+        profiling_window_days = int(
+            self.prompts.question("Enter profiling window in days", default="180", valid_number=True)
```
```suggestion
            self.prompts.question("Enter lookback window in days to profile", default="180", valid_number=True)
```

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -274,6 +302,60 @@ def _configure_credentials(self) -> str:
...
+        )
+        secret_vault_type = str(self.prompts.choice("Enter secret vault type (local | env)", ["local", "env"])).lower()
+        secret_vault_name = None
+
+        logger.info("Please provide BigQuery connection settings:")
+        logger.info(
+            "\nProject/region pairs are entered as `project.region`, comma-separated. "
+            "Example: my-proj-a.us, my-proj-b.eu-west-1\n"
+        )
+        pairs_raw = self.prompts.question("Enter BigQuery project/region pairs (project.region, comma-separated)")
```
```suggestion
        pairs_raw = self.prompts.question("Enter BigQuery project and region pairs (Format: comma-separated project.region. Example: my-proj-a.us, my-proj-b.eu-west-1)")
```

src/databricks/labs/lakebridge/assessments/profiler.py
```diff
@@ -31,7 +31,12 @@ def __init__(self, platform: str, pipeline_configs: PipelineConfig | None = None
 
     @classmethod
     def create(cls, platform: str) -> "Profiler":
-        pipeline_config_path = PLATFORM_TO_SOURCE_TECHNOLOGY_CFG[platform]
+        pipeline_config_path = PLATFORM_TO_SOURCE_TECHNOLOGY_CFG.get(platform)
+        if pipeline_config_path is None:
+            raise ValueError(
+                f"Unsupported profiler source platform: {platform!r}. "
+                f"Supported: {sorted(PLATFORM_TO_SOURCE_TECHNOLOGY_CFG)}"
+            )
```
m-abulazm: the cli does this check before calling this method

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -274,6 +302,60 @@ def _configure_credentials(self) -> str:
...
+            "exclude_streaming_metrics": self.prompts.confirm("Exclude streaming and write API summary?"),
+        }
+
+        credential = {
+            "secret_vault_type": secret_vault_type,
+            "secret_vault_name": secret_vault_name,
+            source: {
+                "pairs": pairs,
+                "profiling_window_days": profiling_window_days,
+                "max_parallel_sqls": max_parallel_sqls,
```
m-abulazm: those should be under profiler

m-abulazm (changes requested):

src/databricks/labs/lakebridge/resources/assessments/bigquery/bq_metadata_extract.py
take60: @m-abulazm big enterprises profile many projects at once, and since `INFORMATION_SCHEMA` is region-scoped, often multiple regions too.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
take60: Done in `1690c3d2` — query text is now always redacted: removed the prompt and hardcoded `redact_query_text: true`.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
take60: Applied in `1690c3d2`.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
take60: Applied in `1690c3d2` — also dropped the now-redundant hint line above it, since the prompt carries the format + example.

src/databricks/labs/lakebridge/assessments/profiler.py
take60: Done in `1690c3d2` — removed the guard; `Profiler.create` indexes `PLATFORM_TO_SOURCE_TECHNOLOGY_CFG[platform]` directly, since the CLI validates `source_tech` against `PROFILER_SOURCE_SYSTEM` before calling it.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
take60: Done in `1690c3d2` — moved `profiling_window_days` and `max_parallel_sqls` under `profiler` in the credentials; updated the extract reads, the tests, and the doc example.

m-abulazm (approved):
LGTM

src/databricks/labs/lakebridge/install.py
m-abulazm: this does not affect many customers

dgomez04 (approved):
Approved as discussed with @m-abulazm  👍

PR #2485, opened 2026-05-26
Run profiler Python steps with the parent interpreter instead of per-step venvs

### LINKED ISSUE #2484: Synapse and MSSQL profilers fail with `ModuleNotFoundError: redshift_connector` in per-step venv
## Summary

After `bafbefe8` (Redshift Profiler PR1, #2305), `src/databricks/labs/lakebridge/connections/database_manager.py` does an unconditional top-level `import redshift_connector`. The Synapse and MSSQL profiler pipelines create a fresh venv per step and only install what's listed in their `pipeline_config.yml` `dependencies:` lists, neither of which include `redshift-connector`. As a result, every Synapse/MSSQL profiler run fails the moment a step's script transitively imports `database_manager`.

## Affected steps

- `synapse.dedicated_sqlpool_extract` (imports `synapse_connection_helpers` → `database_manager`)
- `synapse.serverless_sqlpool_extract` (same chain)
- `mssql.activity_extract` (imports `mssql/common/connector.py` → `database_manager`)
- `mssql.info_extract` (same chain)

Oracle and legacy_synapse are not affected — their steps are SQL files executed in-process by the parent (which has `redshift-connector` from `pyproject.toml`).

## Repro

1. `make dev`
2. `databricks labs lakebridge configure-database-profiler` (pick `synapse` or `mssql`)
3. `databricks labs lakebridge execute-database-profiler --source-tech synapse`

A failing step's per-step venv produces:

```
File ".../synapse/dedicated_sqlpool_extract.py", line 11, in <module>
    from databricks.labs.lakebridge.connections.synapse_connection_helpers import create_synapse_connection
File ".../connections/synapse_connection_helpers.py", line 2, in <module>
    from databricks.labs.lakebridge.connections.database_manager import DatabaseManager
File ".../connections/database_manager.py", line 16, in <module>
    import redshift_connector
ModuleNotFoundError: No module named 'redshift_connector'
```

dgomez04:
## The problem

The `synapse` and `sql-server` profilers create a fresh virtualenv per step and
install only the dependencies each step declares in its `pipeline_config.yml`.
Since #2305, `connections/database_manager.py` imports `redshift_connector` at
module scope, and the extract steps import `database_manager` transitively.

Because no step declared `redshift-connector`, every clean run failed with
`ModuleNotFoundError: No module named 'redshift_connector'` (#2484).
This is structural, not specific to `redshift_connector`: any runtime dependency
a step imports transitively but does not re-declare breaks the same way. The
per-step venv is also slow (venv creation + a `pip install` per step) and forces
every step author to hand-maintain a dependency list.

## What this does

Drops the per-step virtualenv entirely. The profiler already runs in-process
under the labs-managed venv (`~/.databricks/labs/lakebridge/state/venv`), which
has lakebridge and all runtime dependencies installed. Each step script now runs
with that interpreter (`sys.executable`) — no venv creation, no `pip install` at
step time, no network.

The `azure-*` packages the Synapse/MSSQL steps need are now declared as regular
lakebridge runtime dependencies, so they are present in the labs-managed venv
like every other dependency.
- `assessments/pipeline.py`: `_execute_python_step` runs via `sys.executable`;
  removed `_create_venv`, `_install_dependencies`, `_link_parent_site_packages`,
  and the now-unused imports.
- `assessments/profiler_config.py`: removed the now-unused `Step.dependencies`
  field.
- `pyproject.toml`: added `azure-identity`, `azure-mgmt-sql`,
  `azure-monitor-query`, `azure-synapse-artifacts` to the core dependencies
  (previously dev-only in the `lint` group).
- `labs.yml`: no extras needed.
- `synapse`/`mssql` `pipeline_config.yml`: removed the now-redundant per-step
  `dependencies`.
- tests: removed the obsolete per-step venv / dependency-failure tests.

Closes #2484.


m-abulazm:
@dgomez04 please streamline with https://github.com/databrickslabs/lakebridge/pull/2467

dgomez04:
> @dgomez04 please streamline with #2467

@m-abulazm done - should be good to go.

src/databricks/labs/lakebridge/resources/assessments/mssql/pipeline_config.yml
```diff
@@ -19,6 +19,7 @@ steps:
       - azure-identity
       - azure-monitor-query==1.3.0b1
       - pyodbc
+      - redshift-connector~=2.1.13 # See #2484
```
m-abulazm:
```suggestion
      - redshift-connector~=2.1.13
```

m-abulazm (commented):
please remove the comments

dgomez04:
Reworked per our discussion. Instead of linking the parent `site-packages` into a per-step venv, this now drops the per-step venv entirely: each Python step runs with `sys.executable` — the labs-managed interpreter the profiler already runs under, which has lakebridge and all runtime deps installed. Nothing is installed at step time.

The only non-runtime packages (the `azure-*` set the Synapse/MSSQL steps need) move into a `profiler` extra in `pyproject.toml`, installed into the labs venv via the new `extras: profiler` entry in `labs.yml`. So adding a dependency means editing the extra, not per-step config lists.

Net effect addresses the points raised:
- No per-step venv creation and no runtime `pip install` (faster, and no more hand-maintained per-step dependency lists).
- No risk of leaking step-pinned library versions — everything resolves from the single installed environment.
- The original `redshift_connector` failure (and the whole class of transitive-import failures in #2484) is gone, since steps share the parent environment.

One requirement worth flagging: the `labs.yml` `extras` field needs a Databricks CLI new enough to include databricks/cli#2288 (Feb 2025); on older CLIs the extra won't install. Noted in the description.

tests/unit/assessment/test_pipeline_venv.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+    run_python_script = getattr(PipelineClass, "_run_python_script")
+    # Should not raise: the script imports a parent package and reports success, proving the step
+    # resolves dependencies from sys.executable without creating or populating a per-step venv.
+    run_python_script(sys.executable, script, str(tmp_path / "out.db"), "unused-credentials.yml")
+
+
+@pytest.mark.parametrize("platform", ["synapse", "mssql"])
+def test_shipped_python_steps_declare_no_dependencies(platform: str) -> None:
```
m-abulazm: please get rid of `Step#dependencies` instead of this test

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -145,81 +142,14 @@ def _execute_ddl_step(self, step: Step):
...
-        except CalledProcessError as e:
-            # Log detailed output at debug level for troubleshooting
-            logging.debug(
-                f"Failed to install dependencies (exit code {e.returncode})\n" f"stdout: {e.stdout}\nstderr: {e.stderr}"
-            )
-            logging.error(f"Failed to install dependencies: {e.stderr}")
-            raise RuntimeError(f"Failed to install dependencies: {e.stderr}") from e
+        # Run the step script with the interpreter the profiler already runs under (the
+        # labs-managed venv). It has lakebridge and all runtime dependencies installed, plus the
+        # source-specific `profiler` extra packages, so there is nothing to install at step time.
```
```suggestion
        # Run the step script with the labs-managed venv
```

tests/resources/assessments/db_extract_dep.py
```diff
@@ -39,6 +39,9 @@ def execute():
         packages = ['databricks_labs_dqx', 'databricks_labs_ucx']
         df = check_specific_packages(packages)
         logger.info(f'DataFrame columns: {df.columns}')
+        missing = df.loc[df["Status"] == "Not Installed", "Package"].tolist()
+        if missing:
+            raise RuntimeError(f"Required packages are not installed: {', '.join(missing)}")
```
m-abulazm: what is the usefulness of this? we have a linter

tests/unit/assessment/test_pipeline_venv.py
```diff
@@ -0,0 +1,40 @@
...
+import pytest
+import yaml
+
+from databricks.labs.lakebridge.assessments._constants import PLATFORM_TO_SOURCE_TECHNOLOGY_CFG
+from databricks.labs.lakebridge.assessments.pipeline import PipelineClass
+
+_REPO_ROOT = Path(__file__).parents[3]
+
+
+def test_python_step_runs_under_parent_interpreter(tmp_path: Path) -> None:
```
m-abulazm: I would not test this. we use python apis like `sys.executable` and `Popen`

pyproject.toml
```diff
@@ -55,6 +55,17 @@ Documentation = "https://databrickslabs.github.io/lakebridge"
 Issues = "https://github.com/databrickslabs/lakebridge/issues"
 Source = "https://github.com/databrickslabs/lakebridge"
 
+[project.optional-dependencies]
+# Source-specific packages required by the Synapse profiler steps. These are not lakebridge
+# runtime dependencies; they are installed into the labs-managed venv via the `extras` entry
+# in labs.yml so the profiler step scripts can resolve them at runtime.
```
m-abulazm:
```suggestion
```

pyproject.toml
```diff
@@ -55,6 +55,17 @@ Documentation = "https://databrickslabs.github.io/lakebridge"
...
+profiler = [
```
m-abulazm: we need this as runtime dependencies to develop the app. I am interested what is the result of running on a fresh directory i.e. after running `make clean`

m-abulazm (changes requested):
Please update the PR description as well

pyproject.toml
dgomez04: Good call, moved the four `azure-*` into the core `[project] depedencies` rather than keeping them as optional `profiler` extra, and drop the `extras: profiler` from `labs.yaml`. 

On a fresh `make clean && make dev`: `uv sync` installs `azure-identity`, ``azure-mgmt-sql`,` azure-monitor-query`, and `azure-synapse-artifacts` into the labs-managed venv as part of the core dependency set, so the Synapse/MSSQL profiler step scripts resolve them at runtime with nothing installed per-step. 

The tradeoff is that every install now carries the `azure-*` set even for users who never run those profilers.

tests/unit/assessment/test_pipeline_venv.py
dgomez04: Removed

tests/resources/assessments/db_extract_dep.py
dgomez04: Removed this and the obsolete scaffolding entirely (`db_extract_dep.py`, `pipeline_config_failure_dependency.yml`, and `test_run_python_dep_failure_pipeline`).

tests/unit/assessment/test_pipeline_venv.py
dgomez04: Done

m-abulazm (approved):

gueniai (approved):

PR #2488, opened 2026-06-01
Make profiler extract location user-configurable via --output-folder

### LINKED ISSUE #2461: Make profiler extract location user-configurable via `--output-folder`
Profiler `extract_folder` values differ across sources today:

  | Profiler | `extract_folder` | Cross-platform? |
  |---|---|---|
  | `synapse` | `~/.databricks/labs/lakebridge_profilers/synapse_assessment` | ✅ |
  | `mssql` | `/tmp/data/mssql_assessment` | ❌ |
  | `legacy_synapse` (PR #2441) | `/tmp/data/legacy_synapse_assessment` | ❌ |
  | `oracle` (PR #2187) | `/tmp/data/oracle_assessment` | ❌ |

  The drift causes three real problems:

  1. **Windows portability.** `/tmp/...` isn't a valid Windows path. `Path("/tmp/...").expanduser()` resolves to `<drive>:\tmp\data\...`, which "works" but isn't where any Windows user expects an extract to
  land.
  2. **macOS volatility.** `/tmp` is purged periodically (3+ days). A multi-hour profiler run shouldn't live there.
  3. **Inter-profiler scripting.** A user running multiple profilers has to guess different output paths for downstream tooling (e.g. `create-profiler-dashboard --extract-file`) for no good reason.

  ### Acceptance criteria

  - CLI flag `--output-folder` works on `execute-database-profiler`, behaves like `transpile`'s flag.
  - When omitted, prompt surfaces the default path and accepts user input.
  - `extract_folder` removed from all four `pipeline_config.yml`s.
  - `PipelineConfig.extract_folder` defaults to `None`; default is injected at execute time.
  - Unit tests cover: flag-provided override, flag-omitted prompt-default, invalid path (parent missing) raises.
  - Docs (`mssql.mdx`, `legacy_synapse.mdx`, `oracle.mdx`, `synapse.mdx`) document the new flag in the `execute-database-profiler` section.

  ### Out of scope

  - `create-profiler-dashboard --extract-file` behavior
  - `configure-database-profiler` does NOT get the flag

  ### References

  - PR #2441 review thread on `legacy_synapse/pipeline_config.yml` ("as noted on oracle PR, lets fix this in a followup").
  - PR #2187 review thread on `oracle/pipeline_config.yml:3` (same issue).
  - Branch `bug/fix_profiler_path_and_queries` — does the YAML normalization. We rebase on top once it lands.
  - Transpile's `--output-folder` for the implementation pattern: `cli.py:111` (command), `cli.py:332-351` (`_TranspileConfigChecker.use_output_folder` / `_prompt_output_folder`).

m-abulazm:
### What does this PR do?
* Surfaces the profiler's DuckDB extract path as a CLI choice instead of a hardcoded value in per-source pipeline_config.yml files with the default `~/.databricks/labs/lakebridge_profilers/<source>_assessment`. Mirrors the shape of the transpile command's `--output-folder` flag.
*  the DuckDB output file is now named `profiler_extract_<platform>_<version>_<YYYYMMDD>.db` to identify runs against different sources or with different lakebridge versions no longer overwrite each other; the file is identifiable when found on its own.

### Linked issues

Resolves #2461

### Functionality

- [ ] added new CLI command
- [X] modified existing command: `databricks labs lakebridge execute-database-profiler --output-folder`
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/cli.py
```diff
@@ -1024,10 +1028,25 @@ def execute_database_profiler(w: WorkspaceClient, source_tech: str | None = None
...
+        ).strip()
+    _validate_output_folder(Path(output_folder))
+
+    profiler = Profiler.create(source_tech)
     # TODO: Add extractor logic to ApplicationContext instead of creating inside the Profiler class
-    profiler.profile()
+    profiler.profile(output_folder=Path(output_folder))
+
+
+def _validate_output_folder(output_folder: Path) -> None:
```
dgomez04: _During the *first-run*, this rejects the default the prompt just suggested._

The default output folder is `~/.databricks/labs/lakebridge_profilers/<platform>_assessment`, so `parent` here is `~/.databricks/labs/lakebridge_profilers`. That directory is only created at profiler execution time - `os.makedirs(venv_path_prefix, …)` `in _execute_python_step` (`pipeline.py:153`) or `_create_dir` (`pipeline.py:325`), never at install/configure time.

A fresh user who installs: `configure-database-profiler`, `execute-database-profiler`, presses `Enter` to accept the default will hit `Invalid path for '--output-folder', parent does not exist.` We'd be rejecting our own suggested value.

It's also inconsistent with `PipelineClass._create_dir`, which already does `mkdir(parents=True, exist_ok=True)` - the pipeline is happy to create the whole tree, only this pre-flight check insists the parent pre-exists.

We can either: 
- Drop the parent check and rely on `mkdir(parents=True)`, wrapping it to re-raise `OSError`/`PermissionError` as a friendly `ValueError`.
- Ensure `~/.databricks/labs/lakebridge_profilers` exists before validating so the default always passes.

dgomez04: There are no unit tests for `_validate_output_folder` or the new prompt/flag path.

src/databricks/labs/lakebridge/assessments/profiler.py
```diff
@@ -49,13 +47,15 @@ def profile(
...
         pipeline_config: PipelineConfig | None = None,
+        output_folder: Path | None = None,
     ) -> None:
         platform = self._platform.lower()
         if not pipeline_config:
             if not self._pipeline_config:
                 raise ValueError(f"Cannot Proceed without a valid pipeline configuration for {platform}")
             pipeline_config = self._pipeline_config
-        self._execute(platform, pipeline_config, extractor)
+        resolved_output_folder = output_folder or default_output_folder(platform)
```
dgomez04, line 57: *Non-blocking*. `default_output_folder(platform)`** is computed once at `cli.py:1035` as the prompt default, and again here. 

Is this fallback really needed? Only ever fires for non-CLI / programmatic callers.

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1024,10 +1028,25 @@ def execute_database_profiler(w: WorkspaceClient, source_tech: str | None = None
...
+    if not output_folder.parts:
+        raise ValueError("Output folder cannot be empty")
+    parent = output_folder.expanduser().parent
```
dgomez04: `_validate_output_folder` expands `~` but never resolves relative paths to absolute. A user running `execute-database-profiler --output-folder data` gets a path relative to whatever directory the command was invoked from, and the DuckDB extract silently lands in current working directory rather than anywhere predictable.

We can either resolve to absolute paths here or keep relative paths but document the relative-to-CWD behavior in the flag help text.

labs.yml
```diff
@@ -115,6 +115,8 @@ commands:
     flags:
       - name: source-tech
         description: (Optional) The technology/platform of the sources to Profile
+      - name: output-folder
+        description: (Optional) Folder where the extracted DuckDB file is written; if omitted, prompts with a smart default
```
dgomez04, line 119: If we keep relative paths, I'd add something like `Relative paths resolve against the current working directory; if omitted, prompts with a smart default.`

dgomez04 (changes requested):
I think we need to fix the first-run breakage before shipping this issue.

src/databricks/labs/lakebridge/cli.py
m-abulazm: dropping the check

src/databricks/labs/lakebridge/assessments/profiler.py
m-abulazm: since all the function args are optional, I followed the same hence we need to handle empty case.

src/databricks/labs/lakebridge/cli.py
m-abulazm: relative paths have to be supported

labs.yml
m-abulazm: I am not sure how helpful is that, this is the definition of a relative path

src/databricks/labs/lakebridge/cli.py
m-abulazm: dropped

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -13,13 +14,17 @@
...
 from databricks.labs.lakebridge.assessments.profiler_config import PipelineConfig, Step
 from databricks.labs.lakebridge.connections.credential_manager import cred_file
 from databricks.labs.lakebridge.connections.database_manager import DatabaseManager, FetchResult
 
 logger = logging.getLogger(__name__)
 
-DB_NAME = "profiler_extract.db"
+
+def make_profiler_db_filename(platform: str) -> str:
+    # Lazy import to avoid circular import at module load.
```
dgomez04: Isn't this comment inaccurate? We can just delete.

src/databricks/labs/lakebridge/install.py
```diff
@@ -461,12 +461,7 @@ def _configure_new_profiler_dashboard_installation(self) -> ProfilerDashboardCon
     def _prompt_for_new_profiler_dashboard_installation(self) -> ProfilerDashboardConfig:
         logger.info("Please answer a few questions to configure the Lakebridge profiler dashboard.")
         source_tech = self._prompts.choice("Select the source technology", PROFILER_SOURCE_SYSTEM)
-        extract_file_path = self._prompts.question(
-            "Enter the path to the profiler extract file:",
-            default=str(
-                Path("~/.databricks/labs/lakebridge_profilers/synapse_assessment/profiler_extract.db").expanduser()
-            ),
-        )
+        extract_file_path = self._prompts.question("Enter the path to the profiler output file").strip()
```
dgomez04, line 464: With the extract filename now `profiler_extract_<platform>_<version>_<YYYYMMDD>.db`, this prompt having no default means the user must hand-type a non-guessable, changing path to wire up the dashboard. 

Consider defaulting to the output folder (or auto-selecting the newest `profiler_extract_*.db` in it), or persisting the path execute-database-profiler wrote. This doesn't really matter since we'll be removing it anyways.

dgomez04 (approved):
Left small notes, LGTM.

src/databricks/labs/lakebridge/install.py
sundarshankar89: are we going to have this, I belive we wanted to remove this and move to desktop app directly, but we should ahve guideance note on where to find the files.

sundarshankar89 (approved):
LGTM small nit: if we are removing the dashboard install, we can remove it subsequent pr but we are going to keep that in some sort of guideance will help.

src/databricks/labs/lakebridge/install.py
m-abulazm: https://github.com/databrickslabs/lakebridge/pull/2494/changes/6f9866740e4dbe8e1d78ce863447a7ce3ef0c4eb

PR #2453, opened 2026-05-18
create dedicated test cluster for recon e2e tests

m-abulazm:
### What does this PR do?
creates a test cluster to run recon e2e tests
### Relevant implementation details
use pytester and pin spark version

### Tests

- [X] manually tested
- [ ] added unit tests
- [ ] added integration tests

PR #2339, opened 2026-03-18
Add Redshift connector to Recon

bishwajit-db:
Add Redshift connector to Recon

PR #2490, opened 2026-06-02
Add fingerprint pre-check to Recon (Redshift)

ameersalman33:
### What does this PR do?

Adds an **opt-in fingerprint pre-check** to Recon. When `fingerprint_precheck=True` and the source has a registered query builder, Recon runs a sketch-based detection pass (MD5-sub-bucketed aggregates over both sides) **before** the row-hash compare pipeline.

- **MATCH** -> Recon short-circuits in seconds; no full table scan, no JOIN.
- **MISMATCH** -> an algebraic solver returns the differing row hashes; a surgical Stage-2 fetch pulls just those rows and feeds them into the existing `compare.reconcile_data` flow. If the mismatch is systemic (>15% of sub-buckets), the precheck defers to the existing row-hash pipeline.
- **Ineligible** -> falls through to the existing row-hash pipeline silently.

The flag defaults to `False`; existing behaviour is unchanged.

The algorithm is byte-identical to the dataprint sketch-based reconciliation library (Databricks-internal); this MR is the first dataprint-into-lakebridge integration. Redshift is the first source dialect — adding Snowflake / Oracle / TSQL is one `FingerprintQueryBuilder` subclass plus one registry entry, with no orchestrator changes.

- `trigger_recon_service._run_fingerprint_or_reconcile_data` is the single decision point. Static eligibility (flag off, unsupported dialect, report type not data, no join columns, filters / transformations / thresholds configured) is centralised in `classify_ineligibility` (`fingerprint/orchestrator.py`) and runs before any source-side compute. One config-time reason — `unmapped_target_column_mapping` — is detected later, inside the precheck (because it requires the target schema), and surfaces via the typed `UnmappedTargetColumnMappingError` raised by `align_columns`; the trigger catches it before the runtime-failure branch and routes it through `FingerprintRunMetadata.ineligible(...)`. Every reason is one of the values on the `IneligibilityReason` enum and is recorded on `recon_metrics.fingerprint_metrics.ineligibility_reason`, so adoption queries can count each cause without needing to grep cluster logs.
- Source-side reads use upstream's `RemoteQueryReader` / `remote_query()` TVF unmodified — **no new connector code**. Stage-1 aggregation pushdown was verified empirically against a 1 M-row Redshift fixture on a DBR 17.3 cluster: both direct JDBC (4.28 s median) and the upstream `remote_query()` TVF path (5.44 s median) push the GROUP BY to Redshift and return identical aggregated row counts.
- Stage-1 detection is parallelised across source / target via `ThreadPoolExecutor(max_workers=2, thread_name_prefix="fp-stage2")`. Failure semantics match the serial version; the trigger-layer catch wraps the whole block.
- New config fields on `ReconcileConfig`:
  - `fingerprint_precheck: bool = False` (the main flag)
  - `fingerprint_treat_empty_as_null: bool = False` (collapse `''` to NULL in the per-column hash payload, end-to-end on both sides)
  - `fingerprint_row_count_override: int | None = None` (pin the adaptive sub-bucket / bucket tier when `DESCRIBE DETAIL` cannot read `numRecords`)
- `ReconcileConfig.__version__` bumps from `2 -> 3`. The new `v2_migrate` introduces `fingerprint_precheck` and folds two legacy spellings (`redshift_fingerprint_precheck`, `use_fingerprint_precheck`) into it. Combined with upstream's `v1_migrate`, the chain `v1 -> v2 -> v3` runs automatically via `Installation.load(ReconcileConfig)`. Existing deployments upgrade with zero user action.

### Pre-existing fixes that ride along (upstream PR #2339)

While integrating dataprint end-to-end against Redshift on a real cluster, two pre-existing correctness bugs in the upstream Redshift connector MR (PR #2339) surfaced. Neither is dataprint-specific — both crash the existing row-hash recon path on real customer schemas — but they sat directly in the integration path, so they are fixed inline. Both fixes live in `reconcile/query_builder/expression_generator.py` and are pinned by regression tests.

| # | Fix | What broke before |
|---|---|---|
| 1 | **Add `TIMESTAMP` / `TIMESTAMPTZ` transform handlers to the Databricks dialect** (target side) | The Redshift dialect already emits source-side `COALESCE(TO_CHAR(ts, 'YYYY-MM-DD HH24:MI:SS.US'), '_null_recon_')` (always 6 fractional digits, fixed-width). The Databricks dialect block had **no** override for `TIMESTAMP` / `TIMESTAMPTZ`, so the target side fell through to the universal default `TRIM(COALESCE(col, '_null_recon_'))`, where Spark's implicit `cast(timestamp AS string)` emits a variable-length fractional component (omitted entirely for zero-microsecond timestamps; `'2023-10-02 18:08:43'` vs source `'2023-10-02 18:08:43.000000'`). The byte-width drift made the per-row SHA2 disagree for **every** logically-identical `TIMESTAMP` / `TIMESTAMPTZ` row in any Redshift -> Databricks reconcile. Fix: add `COALESCE(DATE_FORMAT(_, 'yyyy-MM-dd HH:mm:ss.SSSSSS'), '_null_recon_')` for both types so the two sides are byte-identical. |
| 2 | **Add `BOOLEAN` transform handler to the Redshift dialect** (source side) | The Redshift block defined overrides only for `SUPER` / `DATE` / `TIMESTAMP` / `TIMESTAMPTZ` and had no dialect-level `default`. Any `BOOLEAN` column fell through to the universal default `TRIM(COALESCE(col, '_null_recon_'))`, which Redshift rejects during *output schema resolution* (before any rows are read) with `function pg_catalog.btrim(boolean) does not exist`. Result: any customer schema containing a single `BOOLEAN` column crashes row-hash recon end-to-end on the Redshift connector shipped in PR #2339. Fix: explicit `COALESCE(CASE WHEN col THEN 'true' WHEN NOT col THEN 'false' ELSE NULL END, '_null_recon_')` so the rendered string matches Spark's `cast(boolean AS string)` byte-for-byte. The existing unit suite did not catch this because the test fixtures hand-build query strings and never exercise the dialect transform-mapping path. |

### Hardening from internal review

Before pushing upstream this PR went through one round of internal review on the contributor's fork. Three substantive code changes landed: pinning timezone-aware target columns to UTC in `fingerprint/spark_target.py` so a non-UTC `spark.sql.session.timeZone` does not produce Stage-1 false-mismatches; falling through to the full pipeline on a `build_mismatch_output` exception in `trigger_recon_service.py` to match the fail-open contract every other branch already follows; and casting Redshift strings to `VARCHAR(65535)` instead of bare `VARCHAR` (which truncates at 256). The remaining changes are cleanups (unused `ColumnAlignment.exclude_columns` removed, no-op reorder in `connectors/source_adapter.py` reverted, flaky wall-clock assertion in `test_fetch_parallel.py` replaced with a deterministic distinct-thread-id check, and `test_expression_generator.py` pins exact rendered SQL on each dialect).

- **DBR 17.3+ requirement**: source-side reads via `remote_query()` require classic clusters on DBR 17.3+ or SQL warehouses (Pro / Serverless) on version 2025.35+. This is inherited from upstream's `RemoteQueryReader` adoption, not a fingerprint-specific limitation; the runtime requirement is documented in `docs/lakebridge/docs/reconcile/index.mdx`.
- **MISMATCH-state cost**: at 1 M scale, fingerprint pays ~20 s for Stage-1 detection unconditionally and Stage-2 currently feeds into the same `compare.reconcile_data` JOIN rather than replacing it, so on MISMATCH it costs an extra 16-94 s vs. row-hash-only mode. The MATCH path is the headline win (38.7% wall-clock improvement on 1 M rows: 55.5 s vs 90.6 s); the production motivation is billion-row scale, where the row-hash pipeline's full-table JOIN is what becomes intractable. Stage-1 hash persistence as Stage-2 input is filed as a follow-up.
- **Test fixture file**: the only non-fingerprint, non-`expression_generator` test file modified is `tests/unit/test_install.py` (six `"version": 2 -> 3` fixture bumps, a direct consequence of the migration step we own). Not scope creep.
- **`success_count` formula in `verify_successful_reconciliation`** is mathematically wrong (yields counts greater than `total`). Pre-existing in upstream PR #2259 (commit `e56c79c3d`); sits right next to fingerprint code in `trigger_recon_service.py`. **Not fixed in this MR** to keep scope contained; filed as a separate issue.

### Linked issues

Resolves #..

### Functionality

- [x] added relevant user documentation (`docs/lakebridge/docs/reconcile/configuration.mdx` *Fingerprint Pre-check (Experimental)* section + `docs/lakebridge/docs/reconcile/index.mdx` runtime requirements)
- [ ] added new CLI command
- [x] new opt-in feature behind `ReconcileConfig.fingerprint_precheck` (default `False`)

### Tests

- [x] added unit tests (+1,400 LOC across 14 new test modules covering: query builders, algebraic solver, eligibility classifier, fetch parallelisation, `v1 -> v2 -> v3` config migration, recon-capture typed schema, fingerprint dispatch, Stage-1 / Stage-2 contract symmetry)
- [x] **All 1383 unit tests pass** (2 skipped, 3 xfailed); `tests/unit/reconcile/` alone runs **307 reconcile tests in 1.13 s**. The 6 failing `test_cli_analyze.py` cases (Informatica analyzer binary) are pre-existing on `main` (verified on stash-clean HEAD) and unrelated.
- [x] regression tests for the two pre-existing fixes: `test_expression_generator.py` pins handler presence, format string, and source / target byte alignment for `TIMESTAMP` / `TIMESTAMPTZ`, plus the `BOOLEAN` `CASE WHEN` rendering on Redshift.
- [x] correctness validated end-to-end on a 1 M-row Redshift / Delta fixture across the 20-scenario dual-mode parity matrix: **39 / 40** cells PASS, **1** scenario shows a known fingerprint-solver fallback edge with verdict + missing-rows agreement on both sides — only the cap-bounded `mismatch` count differs (fingerprint reports the true 10000, normal reports the cap-50 sample).
- [x] linter clean: pylint **10.00/10** on touched src; ruff, black, mypy green across the full `reconcile/` surface plus `config.py`.
- [ ] added integration tests (the recon e2e cluster gate landed in #2453; integration coverage will be added in a follow-up MR alongside the cluster fixture)

### DIFF
--- a/src/databricks/labs/lakebridge/reconcile/fingerprint/fingerprint_hash_columns.py
+++ b/src/databricks/labs/lakebridge/reconcile/fingerprint/fingerprint_hash_columns.py
@@ -1,0 +1,55 @@
+"""Resolve which columns participate in fingerprint detection.
+
+Mirrors HashQueryBuilder.build_query column resolution exactly so the fingerprint
+hashes the same set, in the same order, the row-hash path would. The two
+implementations are pinned together by regression tests, but the duplication is
+a known maintenance hazard — extracting a shared ``compute_hash_columns`` helper
+into ``query_builder/`` is deferred until the second source dialect lands and
+informs the right abstraction boundary (forcing the shape on Redshift alone
+risks over-fitting).
+"""
+
+from __future__ import annotations
+
+from databricks.labs.lakebridge.reconcile.connectors.data_source import DataSource
+from databricks.labs.lakebridge.reconcile.connectors.dialect_utils import DialectUtils
+from databricks.labs.lakebridge.reconcile.recon_config import Table, Schema
+
+
+def _strip(name: str) -> str:
+    """Return ``name`` lowercased and without ANSI/source delimiters.
+
+    Bridges the two naming conventions: ``Table.join_columns`` are bare,
+    ``Schema.column_name`` (and ``get_select_columns`` output) are ANSI-delimited.
+    Set operations on the two would otherwise produce duplicate entries.
+    """
+    return DialectUtils.unnormalize_identifier(name).lower()
+
+
+def hash_columns_ordered_for_reconcile(
+    table_conf: Table,
+    schema: list[Schema],
+    layer: str,
+    data_source: DataSource,
+) -> list[str]:
+    """Mirror HashQueryBuilder hash column set + sort order (case-insensitive sort_key)."""
+    join_keys = {_strip(c): c for c in (table_conf.join_columns or [])}
+    select_keys: dict[str, str] = {}
+    for col_name in table_conf.get_select_columns(schema, layer):
+        select_keys.setdefault(_strip(col_name), col_name)
+    threshold_keys = {_strip(c) for c in table_conf.get_threshold_columns(layer)}
+    drop_keys = {_strip(c) for c in table_conf.get_drop_columns(layer)}
+
+    merged: dict[str, str] = {}
+    for stripped, original in {**select_keys, **join_keys}.items():
+        if stripped in threshold_keys or stripped in drop_keys:
+            continue
+        merged[stripped] = original
+
+    hash_cols_with_sort = []
+    for original in merged.values():
+        sort_key = DialectUtils.unnormalize_identifier(
+            data_source.normalize_identifier(original).ansi_normalized
+        ).lower()
+        hash_cols_with_sort.append((sort_key, original))
+    return [c for _, c in sorted(hash_cols_with_sort, key=lambda x: x[0])]
--- a/src/databricks/labs/lakebridge/reconcile/fingerprint/orchestrator.py
+++ b/src/databricks/labs/lakebridge/reconcile/fingerprint/orchestrator.py
@@ -1,0 +1,729 @@
+import logging
+import time
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+
+from pyspark.sql import DataFrame, SparkSession
+from pyspark.sql.functions import expr
+from sqlglot import Dialect
+
+from databricks.labs.lakebridge.config import DatabaseConfig
+from databricks.labs.lakebridge.reconcile.compare import (
+    _HASH_COLUMN_NAME,
+    capture_mismatch_data_and_columns,
+    reconcile_data as compare_reconcile_data,
+)
+from databricks.labs.lakebridge.reconcile.connectors.data_source import DataSource
+from databricks.labs.lakebridge.reconcile.connectors.dialect_utils import DialectUtils
+from databricks.labs.lakebridge.reconcile.fingerprint.constants import pick_sub_bucket_count
+from databricks.labs.lakebridge.reconcile.fingerprint.engine import (
+    DetectionResult,
+    DetectionVerdict,
+    detect_and_solve,
+)
+from databricks.labs.lakebridge.reconcile.fingerprint.exceptions import (
+    UnmappedTargetColumnMappingError,
+    UnsupportedDataSourceError,
+)
+from databricks.labs.lakebridge.reconcile.fingerprint.metadata import (
+    FETCH_PATH_V1_SANDWICH,
+    INELIGIBLE_COLUMN_THRESHOLDS_CONFIGURED,
+    INELIGIBLE_FILTERS_CONFIGURED,
+    INELIGIBLE_FLAG_DISABLED,
+    INELIGIBLE_NO_JOIN_COLUMNS,
+    INELIGIBLE_REPORT_TYPE_NOT_DATA,
+    INELIGIBLE_TABLE_THRESHOLDS_CONFIGURED,
+    INELIGIBLE_TRANSFORMS_CONFIGURED,
+    INELIGIBLE_UNSUPPORTED_DIALECT,
+)
+from databricks.labs.lakebridge.reconcile.fingerprint.query_builders.base import FingerprintQueryBuilder
+from databricks.labs.lakebridge.reconcile.fingerprint.query_builders.redshift import RedshiftFingerprintQueryBuilder
+from databricks.labs.lakebridge.reconcile.fingerprint.fingerprint_hash_columns import (
+    hash_columns_ordered_for_reconcile,
+)
+from databricks.labs.lakebridge.reconcile.fingerprint.row_count import fetch_target_row_count
+from databricks.labs.lakebridge.reconcile.fingerprint.spark_target import (
+    build_target_filter_subquery,
+    compute_target_fingerprint,
+)
+from databricks.labs.lakebridge.reconcile.query_builder.hash_query import HashQueryBuilder
+from databricks.labs.lakebridge.reconcile.recon_capture import AbstractReconIntermediatePersist
+from databricks.labs.lakebridge.reconcile.recon_config import Table, Schema
+from databricks.labs.lakebridge.reconcile.recon_output_config import (
+    DataReconcileOutput,
+    MismatchOutput,
+)
+
+logger = logging.getLogger(__name__)
+
+# Fingerprint defaults to keeping '' distinct from NULL so the detection-side serialization
+# matches the row-hash path (TRIM in expression_generator does not collapse '' to NULL).
+# Flipping this to True silently disagrees with row-hash on every NULL <-> '' flip.
+_DEFAULT_TREAT_EMPTY_AS_NULL = False
+
+
+@dataclass(frozen=True)
+class ColumnAlignment:
+    column_mapping: dict[str, str] | None
+
+
+@dataclass(frozen=True)
+class FingerprintResult:
+    verdict: DetectionVerdict
+    source_rows: DataFrame | None = None
+    target_rows: DataFrame | None = None
+    solved_count: int = 0
+    unsolved_sb_count: int = 0
+    total_mismatched_sbs: int = 0
+    detection_elapsed_ms: int = 0
+    sub_bucket_count: int = 0
+    bucket_count: int = 0
+    target_row_count: int | None = None
+    row_count_source: str | None = None
+    fetch_path: str | None = None
+
+
+@dataclass(frozen=True)
+class _TierSelection:
+    """Adaptive (sub_bucket_count, bucket_count) for one run.
+
+    Source and target must use identical values or sub-bucket IDs won't align across
+    the GROUP BY join.
+    """
+
+    sub_bucket_count: int
+    bucket_count: int
+    target_row_count: int | None
+    row_count_source: str
+
+
+def classify_ineligibility(
+    *,
+    flag_enabled: bool,
+    data_source: str,
+    report_type: str,
+    table_conf: Table,
+) -> str | None:
+    """Return the ineligibility reason for the pre-check, or None when eligible.
+
+    First-match-wins; flag/source-level reasons are surfaced before per-table config so
+    adoption queries can distinguish "feature off" from "feature on but table ineligible".
+    """
+    if not flag_enabled:
+        return INELIGIBLE_FLAG_DISABLED
+    if data_source not in fingerprint_supported_sources():
+        return INELIGIBLE_UNSUPPORTED_DIALECT
+    if report_type not in {"data", "row", "all"}:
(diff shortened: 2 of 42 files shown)


src/databricks/labs/lakebridge/reconcile/fingerprint/fingerprint_hash_columns.py
```diff
@@ -0,0 +1,55 @@
...
+    """Return ``name`` lowercased and without ANSI/source delimiters.
+
+    Bridges the two naming conventions: ``Table.join_columns`` are bare,
+    ``Schema.column_name`` (and ``get_select_columns`` output) are ANSI-delimited.
+    Set operations on the two would otherwise produce duplicate entries.
+    """
+    return DialectUtils.unnormalize_identifier(name).lower()
+
+
+def hash_columns_ordered_for_reconcile(
```
bishwajit-db, line 29: This re-derives the hash-column set and ordering that `HashQueryBuilder.build_query` already computes (`hash_query.py:55-76`): the same `sorted((join | select_columns) - threshold_columns - drop_columns)`, followed by the same "sort by unnormalized identifier, case-insensitive" rule. The docstring acknowledges the duplication, but it can be removed now rather than deferred: lift `hash_query.py:55-76` into a method on `HashQueryBuilder` (e.g. `ordered_hash_columns()`) and call it here.

The `select_columns` / `threshold_columns` / `drop_columns` properties already wrap the same `Table.get_*` methods this function calls, so this is a pure extraction with no behavior change, leaving a single definition of "which columns participate, in what order" instead of two pinned together by tests.

bishwajit-db (commented):
A few review comments focused on reusing existing reconcile machinery, plus one correctness note on the terminal MATCH path. Details inline.

src/databricks/labs/lakebridge/reconcile/fingerprint/query_builders/redshift.py
```diff
@@ -0,0 +1,133 @@
...
+    embedded quotes for the same reason — keep the two boundaries consistent.
+    """
+    escaped = bare.replace(_REDSHIFT_IDENTIFIER_QUOTE, _REDSHIFT_IDENTIFIER_QUOTE * 2)
+    return f"{_REDSHIFT_IDENTIFIER_QUOTE}{escaped}{_REDSHIFT_IDENTIFIER_QUOTE}"
+
+
+class RedshiftFingerprintQueryBuilder(FingerprintQueryBuilder):
+    """Generate Redshift SQL for fingerprint detection and surgical row fetch."""
+
+    def serialize_column(self, col_name: str, col_type: str) -> str:
```
bishwajit-db, line 31: This serializer, together with `spark_target._serialize_column_spark` / `_serialize_column_spark_sql`, hand-rebuilds the per-column type handling the row-hash path already does generically via `QueryBuilder._default_transformer` + `DataType_transform_mapping` + `transform_expression` (`base.py:125-159`, `expression_generator.py`). That mapping is dialect-keyed and already carries the Redshift TIMESTAMP/TIMESTAMPTZ/DATE/BOOLEAN handlers and the Databricks ones, including the two added in this PR.

Rendering each column through the transform map (Redshift dialect for the source concat operands, Databricks dialect for the Spark SQL string, then `F.expr(...)` for the Stage-1 `Column`) would keep only the MD5 -> bucket-modulus arithmetic as fingerprint-specific code, which is the legitimate new part. The type serialization is not, and maintaining three copies that must stay byte-identical is the maintenance hazard the module docstrings describe. (See the separate comment on timezone handling for the correctness consequence of these copies drifting.)

src/databricks/labs/lakebridge/reconcile/fingerprint/query_builders/redshift.py
```diff
@@ -0,0 +1,133 @@
...
+            # produces 'true'/'false' to match Spark's cast(bool AS string).
+            cast_expr = f"CASE WHEN {quoted} THEN 'true' WHEN NOT {quoted} THEN 'false' ELSE NULL END"
+        elif col_type_lower in {"timestamptz", "timestamp with time zone"}:
+            # Parity with the row-hash compare path: ``TO_CHAR`` with a fixed
+            # ``YYYY-MM-DD HH24:MI:SS.US`` format pins microsecond width so the
+            # byte stream matches the Spark target's ``DATE_FORMAT(_,
+            # 'yyyy-MM-dd HH:mm:ss.SSSSSS')``. Default ``CAST(_ AS VARCHAR)``
+            # would emit variable-width fractional seconds and silently disagree
+            # with row-hash on the same row, dropping it from Stage-2.
+            cast_expr = f"TO_CHAR({quoted} AT TIME ZONE 'UTC', 'YYYY-MM-DD HH24:MI:SS.US')"
```
bishwajit-db, line 48: This branch diverges from the row-hash path, which emits `COALESCE(TO_CHAR({}, 'YYYY-MM-DD HH24:MI:SS.US'), ...)` with no `AT TIME ZONE` (`expression_generator.py`, redshift TIMESTAMPTZ). The target side mirrors the divergence: `spark_target._serialize_column_spark` uses `TO_UTC_TIMESTAMP(col, CURRENT_TIMEZONE())` (line 214) while the row-hash Databricks handler does not normalize.

Consequence: on a non-UTC `spark.sql.session.timeZone`, Stage-1 can reach a MATCH verdict that disagrees with the full row-hash pipeline on timezone-aware timestamps. The MATCH path is terminal: `fingerprint_match_output()` returns a zeroed output and does not fall open, so a divergent MATCH is a silently missed mismatch rather than a fall-through. Either reconcile the timestamp handling on both sides with the row-hash path, or adopt the shared transform map (see the serializer-reuse comment), which removes the divergence by construction. Please also add a test that exercises detection with a non-UTC session timezone over a timestamptz column.

src/databricks/labs/lakebridge/reconcile/fingerprint/orchestrator.py
```diff
@@ -0,0 +1,729 @@
...
+        mismatch_count=0,
+        missing_in_src_count=0,
+        missing_in_tgt_count=0,
+        mismatch=MismatchOutput(),
+        missing_in_src=None,
+        missing_in_tgt=None,
+    )
+
+
+def build_mismatch_output(
```
bishwajit-db, line 203: This already calls `compare.capture_mismatch_data_and_columns`, but then re-implements the column-diff derivation owned by `compare._get_mismatch_df` / `_get_mismatch_columns`: the null-safe `<=>` recompute of each `<col>_match`, the per-row `NOT <col>_match` filter, and the `concat_ws` of differing columns. In the normal path this is produced by `Reconciliation._get_sample_data` (`reconciliation.py:306`), which already wraps capture and handles `mismatch_columns`.

Routing the prefetched MISMATCH frames back through `_get_sample_data` (or factoring its tail into a shared helper both paths call) avoids a second copy of the null-safe-equality semantics. As written, any change to the canonical mismatch-column logic must be mirrored here, and the inline comment block already documents how easy that logic is to get wrong.

src/databricks/labs/lakebridge/reconcile/fingerprint/orchestrator.py
```diff
@@ -0,0 +1,729 @@
...
+    # explicitly so the four call sites cannot diverge silently.
+    treat_empty_as_null: bool
+
+
+# sqlglot renders the table placeholder per dialect: Spark/Databricks emits ``:tbl``,
+# Postgres-family dialects emit ``%(tbl)s`` (pyformat). Substitute both forms.
+_TBL_PLACEHOLDERS = (":tbl", "%(tbl)s")
+
+
+def _substitute_tbl_placeholder(query: str, replacement: str) -> str:
```
bishwajit-db, line 438: This duplicates the placeholder convention `RedshiftDataSource.read_data` already applies (`redshift.py:59`: `.replace("%(tbl)s", ...).replace(":tbl", ...)`). The orchestrator now encodes a connector-internal detail, and a second source dialect with a different placeholder would break silently here.

There is no existing seam to reuse directly, since the connector hardcodes `schema.table` as the replacement and offers no hook to inject a `(SELECT ... WHERE ...)` subquery. The substitution should live behind an abstraction rather than being reproduced in the orchestrator: either have the connector expose a way to substitute a sub-select for the table reference, or have `FingerprintQueryBuilder` own table-reference wrapping so the placeholder contract stays in one place.

PR #2494, opened 2026-06-04
Add optional cli argument `cred_file_path` to `execute-database-profiler` cli command

### LINKED ISSUE #2480: Add optional cli argument `creds_file_path` to `execute-database-profiler` and `test-profiler-connection` cli commands
Extend cli commands to support additional optional argument `creds_file_path`. This is necessary to support multiple profilers within the desktop app. 

The cli credentials management stays as is (it does not support multiple credentials). if the optional argument is there, the cli uses it otherwise it loads the .credentials file from user home per status quo

- [ ] Support in `execute-database-profiler` cli command
- [ ] Support in `test-profiler-connection` cli command

m-abulazm:
### What does this PR do?
Lets users use a non-default credentials file using cli argument **`--cred-file-path <path>`** . Falls back to `cred_file(PRODUCT_NAME)` (the file written by `configure-database-profiler`). 

 * `databricks labs lakebridge test-profiler-connection` already supported `--cred-file-path`.
 * Uses **singular** `cred-file-path` **not** `creds-file-path`

### Linked issues

Resolves #2480 

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/cli.py
dgomez04, line 1: On `cli.py:30-34` the production extract entry point still rejects any credential file whose name doesn't end in `credentials.yml`, but this pull request's whole purpose is to let users pass an abritrary `--cred-file-path`. For platforms likes `synapse` where `CONNECTOR_REQUIRED["synapse"] = False`, the credentials file is consumed only by the extract subprocess that calls this loader - `_setup_extractor` is skipped entirely. 

So an arbitrary path passes the `.exists()` check in `execute_database_profiler`, gets forwarded as `--credential-config-path`, and then dies right here with `ValueError: Credential config file must have 'credentials.yml' extension`. The user sees a generic `RuntimeError: Pipeline execution failed for source synapse` with the real reason buried in the subprocess output.

The new integration tests don't catch this because they run the test resource rather than the production loader.

Two options: 
1. If arbitrary filenames are intended, drop this check here (the production path), not just in the test resource.
2. If the `credentials.yml` suffix is a real requirement, keep this check but validate it early in `execute_database_profiler` so the user gets an immediate, clear error instead of an opaque pipeline failure — and document the constraint in `labs.yml`. In that case the test resource should keep the check too.

**Note this is necessary but not sufficient for `synapse`: even with a valid filename, the synapse extract scripts ignore the path and read the default file.**

dgomez04: Even with a valid filename, the **`synapse`** extract scripts ignore the forwarded path and rebuild the credential manager from the default file:

- `workspace_extract.py#L26`
- `serverless_sqlpool_extract.py#L58`
- `monitoring_metrics_extract.py#L29`
- `dedicated_sqlpool_extract.py#L30`

Each unpacks `creds_file` from `arguments_loader` but calls `create_credential_manager(PRODUCT_NAME, EnvGetter())` without passing it, so the `--cred-file-path` value is discarded and the default `~/.databricks/labs/lakebridge/.credentials.yml` is read instead.

They should forward it: `create_credential_manager(PRODUCT_NAME, EnvGetter(), creds_path=Path(creds_file))`.

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1022,8 +1023,8 @@ def execute_database_profiler(
     user = ctx.current_user
     logger.debug(f"User: {user}")
     # check if cred_file is present which has the connection details before running the profiler
-    file = cred_file(PRODUCT_NAME)
-    if not file.exists():
+    creds_path = Path(cred_file_path) if cred_file_path else cred_file(PRODUCT_NAME)
+    if not creds_path.exists():
         raise_validation_exception(
```
dgomez04: When the user passes an explicit `--cred-file-path` that's missing, telling them to run `configure-database-profiler` is misleading, that writes the default file, not their path. Consider a message like `Credential file not found at {creds_path}` for the explicit case.

src/databricks/labs/lakebridge/assessments/profiler.py
```diff
@@ -45,33 +46,31 @@ def path_modifier(*, config_file: str | Path, path_prefix: Path = PRODUCT_PATH_P
     def profile(
         self,
         *,
-        extractor: DatabaseManager | None = None,
```
dgomez04, line 48: `profile()` drops the `extractor` parameter that used to allow injecting a pre-built `DatabaseManager`. No in-repo caller uses it, but the desktop app is the stated consumer of this feature. Can we confirm it doesn't inject an `extractor`? Either way, this is a public-signature change that should be noted in the pull request description.

tests/resources/assessments/db_extract.py
```diff
@@ -37,13 +37,6 @@ def execute():
         '--credential-config-path', type=str, required=True, help='Path string containing credential configuration'
     )
     args = parser.parse_args()
-    credential_file = args.credential_config_path
```
dgomez04, line 40: Removing the extension validation here makes the new integration tests pass with `test_creds.yml`, but it diverges from production (`common/cli.py`) and from the sibling resource `db_extract_dep.py`, which both still enforce it. Whatever the decision on the constraint, keep the test resources consistent with production so the integration tests actually exercise the path users hit.

dgomez04 (approved):
Approved with changes. Please fix the `common/cli.py` issue before merging.

sundarshankar89 (approved):
LGTM

src/databricks/labs/lakebridge/assessments/profiler.py
m-abulazm: this was an unused parameter so worth a cleanup

src/databricks/labs/lakebridge/cli.py
m-abulazm: nice catch

tests/resources/assessments/db_extract.py
m-abulazm: removed from common/cli.py

src/databricks/labs/lakebridge/cli.py
m-abulazm: done

PR #2496, opened 2026-06-08
Add Reconcile RCA dashboard

m-abulazm:
A new Lakeview dashboard (rca_metrics) for root-cause analysis of reconcile runs. Replaces the legacy reconciliation_metrics dashboard.

Two sections:
- Recon Jobs: Jobs Summary + Run Details tables
- Table Health: KPI counters, Latest Run per Table (with failing columns joined from column_metrics), Runs Over Time and Failed Rows Over Time grouped bars, Failures by Table + Successes by Table pies

Data model enhancement: rework the nested reconciliation data model for
better usability and queryability.

Remove the LSQL library dependency from the dashboard deployment path.


BesikiML:
1. Migration path lacks automated test 
Can we add at least one integration test that:

- Creates a legacy details table with data ARRAY<MAP<...>>
- Runs the upgrade
- Asserts row counts and a sample mismatch row in the new schema

2. *_backup tables are never dropped
Migration renames old tables to details_backup / aggregate_details_backup but never cleans them up. Over time this doubles storage for every upgraded workspace.

Is that intentional for rollback? If so, worth documenting in the PR body. If not, consider a follow-up or optional cleanup step.

3. Removed Lakeview retry logic
The old DashboardDeployment had @retried(on=[InternalError, DeadlineExceeded], timeout=3min) around dashboard create/update. The new lakeview.create / update / publish path has no retries.

Was this intentional? Transient Lakeview failures were a known issue (per old comments). Worth keeping retries on the new SDK calls if flakiness still happens in CI/deploy.

4. Naive catalog/schema substitution (low)
raw = raw.replace("remorph.reconcile.", f"{config.catalog}.{config.schema}.")
Works for the placeholder convention, but a blind string replace could corrupt unrelated occurrences inside the 142KB JSON. Safer long-term: structured substitution or a more constrained token (e.g. {{CATALOG}}.{{SCHEMA}}.).

5. Minor test inconsistency
Most recon capture tests use the ws fixture; test_store_run_context still uses mock_workspace_client. Not blocking, but inconsistent with the direction from #2521.

bishwajit-db:
Re point 2 (`*_backup` tables): keeping them is intentional. The migration is a one-way, destructive transformation of user data (rename original, recreate, backfill), so the original tables are preserved as `*_backup` to serve as a rollback path in case the transformed data turns out wrong for some edge case. We did not want an upgrade step to silently delete user data, and the upgrade log explicitly points at the backup table so admins know where the raw rows live. Dropping the backups is left to the workspace admin once the migrated tables are verified; an optional cleanup step can be added as a follow-up if there is demand. Will make sure this is called out in the PR description as well.

Points 1, 3 and 5 will be addressed in upcoming commits on this PR.

BesikiML (approved):
LGTM

ganeshdogiparthi-db:
The changes look good.

One question: If an upgrade fails midway, do we recover automatically on the next run, or is manual cleanup required? If manual intervention is needed, could we either make the process resumable by detecting existing `*_backup` objects and continuing, or at least document the recovery steps?

bishwajit-db:
The upgrade isn't auto-resumable today; if it fails midway the raw rows are preserved in the `*_backup` tables and recovery is manual (rename the backup back and rerun). I've documented the recovery steps in the installation docs.

ganeshdogiparthi-db (approved):
LGTM

PR #2497, opened 2026-06-09
Support max_sample_size via sampling_options with engine-aware bounds

### LINKED ISSUE #2559: [FEATURE]: Increase the number of records displayed in the Reconcile output dashboard

  Reconcile

  ### Problem statement

  Increase the number of sample rows captured per result set (mismatch / missing-in-source / missing-in-target) in the Reconcile output. The current fixed default is 50.

  ### Proposed Solution
  
  Let the user set the count via `sampling_options.specifications.value`, with engine-aware upper bounds:

  - Databricks sources (including foreign catalogs): up to **50,000** — the Databricks path registers the sample keys as a temp view, so the generated SQL stays small regardless of
  sample size.
  - Non-Databricks sources: up to **400** — higher values would exceed the JDBC / inline-UNION query-length limits.
  - `value <= 0` falls back to 50; values above the source's cap are clamped with a WARNING log.

  ### Additional Context

  Applies to `data` / `all` report types (row-level reconcile). Aggregates reconciliation is out of scope and still samples a fixed 50 rows.

ganeshdogiparthi-db:
## Changes
  
  ### What does this PR do?

  Adds a `sampling_options` block to the reconcile table config so users can control how many sample rows are captured per result set (mismatch, missing-in-source,
  missing-in-target) instead of the fixed 50. Previously large sample sizes on Databricks blew up the UNION-based sampling query against Spark Connect; the Databricks path now uses
  a temp view so it scales.
  
  - `SamplingSpecifications` now defaults to `count` / `50` and validates the value (coerces to int, rejects bools/non-numeric, keeps `fraction` disabled).
  `Table.get_max_sample_size()` exposes the configured value.
  - Sample-size bounds are applied in `NormalizeReconConfigService` (engine-aware), not in the samplers:
    - Databricks source: clamped to `[1, 50_000]`.
    - Non-Databricks source: honored up to `400`, larger values capped at 400 with a warning.
    - `value <= 0` falls back to 50.
  - `SamplingQueryBuilder` registers the recon keys as a temp view for the Databricks engine (`SELECT * FROM recon_keys_<uuid>`) so the generated SQL stays small regardless of
  sample size; other dialects keep the inline-UNION derived table.
  - The old fixed `[50, 400]` clamp was removed from the samplers.

  - Non-Databricks sources are intentionally capped at 400 (matches prior behaviour); only Databricks honors larger sizes up to 50,000.
  - `fraction` sampling type stays disabled and raises a config error.

  ### Linked issues
  

  Resolves #2559 
  
  ### Functionality

  - [ ] added new CLI command
  - [x] new config field: `sampling_options` on the reconcile Table config

  ### Tests
  

  - [x] manually tested — end-to-end on Databricks source (default → 50, 100000 → capped to 50000) and Redshift source (default → 50, 500 → capped to 400 with warning), covering
  column mapping, drop columns and schema validation
  - [x] added unit tests
  - [x] added integration tests


src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
```diff
@@ -38,6 +46,20 @@ def _normalize_sampling(self, table: Table):
                 if normalized_sampling.stratified_columns
                 else None
             )
+            requested = normalized_sampling.specifications.value
+            is_databricks = isinstance(self.source, DatabricksDataSource)
+            if requested <= 0:
+                normalized_sampling.specifications.value = _DEFAULT_MAX_SAMPLE_SIZE
```
bishwajit-db: `dataclasses.replace(table.sampling_options)` produces a shallow copy, so `normalized_sampling.specifications` is still the same `SamplingSpecifications` instance as the input table's. The clamping below then mutates `specifications.value` in place, which rewrites the un-normalized input `Table` as a side effect.

`normalize_recon_table_config` starts with `dataclasses.replace(table)` specifically so the input is left untouched, and this breaks that contract: re-normalizing the same config, or inspecting/logging the raw config after a run, will show already-clamped values.

Could we copy the specifications before mutating? e.g.

```python
normalized_sampling = dataclasses.replace(
    table.sampling_options,
    specifications=dataclasses.replace(table.sampling_options.specifications),
)
```

A small unit test asserting the input table's `sampling_options` is unchanged after normalization would lock this in.

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
```diff
@@ -38,6 +46,20 @@ def _normalize_sampling(self, table: Table):
                 if normalized_sampling.stratified_columns
                 else None
             )
+            requested = normalized_sampling.specifications.value
+            is_databricks = isinstance(self.source, DatabricksDataSource)
```
bishwajit-db: This makes the normalize service depend on a concrete connector class, and the rule it encodes ("which engines can safely handle large sample sizes") really belongs to the connector, not to the normalizer. It also means any future Spark-backed connector would silently inherit the 400 cap.

Suggestion: express this as a capability on `DataSource`, e.g. a `max_sample_size` (or `supports_large_sampling`) property that `DatabricksDataSource` overrides. The normalizer then just clamps to `self.source.max_sample_size` with no knowledge of specific connectors, and the import of `DatabricksDataSource` here goes away.

Not blocking, but it would keep the layering clean and make the next connector a one-line change.

src/databricks/labs/lakebridge/reconcile/recon_config.py
```diff
@@ -200,6 +207,11 @@ def __post_init__(self):
         self.drop_columns = to_lower_case(self.drop_columns) if self.drop_columns else None
         self.join_columns = to_lower_case(self.join_columns) if self.join_columns else None
 
+    def get_max_sample_size(self) -> int:
```
bishwajit-db, line 229: With the clamps removed from the samplers and `Table.get_max_sample_size()` returning the raw configured value, all bounds enforcement now lives solely in `NormalizeReconConfigService`. Both trigger services normalize, so the CLI flows are covered, but the invariant "this Table has been normalized" is now implicit:

- Anyone driving `Reconciliation` / `SamplerFactory` programmatically with an un-normalized `Table` gets no guardrails at all. A negative value flows into `df.limit(-2)`, which raises in Spark; a very large value on a non-Databricks source re-creates the oversized inline UNION this PR is fixing.
- The updated sampler tests had to route through `NormalizeReconConfigService` to pass, which is a hint that the samplers alone no longer guarantee safe behavior.

Centralizing the policy is a good call. Could we make the contract explicit, either by documenting on `get_max_sample_size()` (and the samplers) that values are assumed pre-normalized, or by keeping a cheap floor (`max(value, 1)`) in `get_max_sample_size()` as a last line of defense?

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
```diff
@@ -9,6 +11,12 @@
     ColumnThresholds,
 )
 
+logger = logging.getLogger(__name__)
+
+_DEFAULT_MAX_SAMPLE_SIZE = 50
```
bishwajit-db: The default sample size of 50 is now defined independently in five places:

- `_SAMPLE_ROWS = 50` in `compare.py`
- `_SAMPLE_ROWS = 50` in `reconciliation.py`
- `_DEFAULT_MAX_SAMPLE_SIZE = 50` here
- the `value: int | float = 50` default on `SamplingSpecifications`
- the hardcoded `return 50` fallback in `Table.get_max_sample_size()`

(`_MIN_SAMPLE_COUNT = 50` in `sampler.py` also encodes the same number for its default-fallback paths.)

If the default ever changes, these will drift independently and the effective behavior will depend on which code path a run takes. Suggest defining a single `DEFAULT_SAMPLE_ROWS` in one shared location (e.g. `recon_config.py` or `constants.py`) and importing it everywhere else.

bishwajit-db (commented):
Thanks for the PR. The approach looks good overall: centralizing the bounds at normalization and switching the Databricks path to a temp view both make sense, and the test coverage is solid. A few comments inline, one of which is a real (if low-severity) mutation bug in the normalize service.

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
ganeshdogiparthi-db: Fixed.

src/databricks/labs/lakebridge/reconcile/recon_config.py
ganeshdogiparthi-db: 1. Lower-bound/value validation (non-positive → default, type coercion) is moved into SamplingSpecifications.__post_init__ (the dataclass), so it always holds regardless of caller.
  2. Engine-aware upper-bound capping stays in NormalizeReconConfigService, since it depends on the source dialect's max_sample_size.

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
ganeshdogiparthi-db: done

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
ganeshdogiparthi-db: done

bishwajit-db (approved):
LGTM

src/databricks/labs/lakebridge/reconcile/connectors/data_source.py
```diff
@@ -12,6 +12,9 @@
 
 
 class DataSource(ABC):
+    # Max sample rows this engine can materialize via the inline-UNION sampling query.
+    # Databricks overrides this (temp-view sampling scales much higher).
+    max_sample_size: int = 400
```
m-abulazm, line 17: I would not have this here. this property is not that dynamic and only databricks overrides it

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
```diff
@@ -32,12 +35,23 @@ def normalize_recon_table_config(self, table: Table) -> Table:
...
                 else None
             )
+            requested = normalized_sampling.specifications.value
+            max_allowed = self.source.max_sample_size
+            if requested > max_allowed:
+                logger.warning(
+                    f"sample size={requested} exceeds the limit of {max_allowed} "
+                    f"for {type(self.source).__name__}; capping to {max_allowed}."
+                )
+                normalized_sampling.specifications.value = max_allowed
```
m-abulazm, line 54: to be removed. add verification to the class itself

src/databricks/labs/lakebridge/reconcile/recon_config.py
```diff
@@ -57,23 +61,32 @@ class HashAlgoMapping:
...
 
 
 @dataclass
 class SamplingOptions:
-    method: SamplingOptionMethod
-    specifications: SamplingSpecifications
+    method: SamplingOptionMethod = SamplingOptionMethod.RANDOM
+    specifications: SamplingSpecifications = field(default_factory=SamplingSpecifications)
     stratified_columns: list[str] | None = None
     stratified_buckets: int | None = None
```
m-abulazm, line 94: this is a chance to simplify these class. Fraction is not supported and complicates the interface. stratified_columns and buckets are only need with `STRATIFIED` sampling. this has to improve

src/databricks/labs/lakebridge/reconcile/recon_config.py
```diff
@@ -210,6 +223,12 @@ def __post_init__(self):
         self.drop_columns = to_lower_case(self.drop_columns) if self.drop_columns else None
         self.join_columns = to_lower_case(self.join_columns) if self.join_columns else None
 
+    def get_max_sample_size(self) -> int:
+        """Sample-row cap. The engine-aware upper bound is applied by NormalizeReconConfigService."""
+        if self.sampling_options is None:
+            return DEFAULT_SAMPLE_ROWS
+        return int(self.sampling_options.specifications.value)
```
m-abulazm, line 233: I would add the maximum here if needed

src/databricks/labs/lakebridge/reconcile/recon_config.py
m-abulazm: thinking more about it, do we even want the user to configure this?

m-abulazm (dismissed):

src/databricks/labs/lakebridge/reconcile/connectors/data_source.py
ganeshdogiparthi-db: This was added to the base class based on Bishwajit's earlier review. The feedback was to define the cap at the source level rather than as a module-level constant in the normalize service.

Keeping a default value of 400 in the base class and overriding it to 50,000 for Databricks ensures every connector has a safe default bound without requiring each connector to redeclare it.

src/databricks/labs/lakebridge/reconcile/normalize_recon_config_service.py
ganeshdogiparthi-db: The lower-bound validation and type coercion already live in SamplingSpecifications.__post_init__. The upper cap cannot be enforced there because the specification is constructed before the source is known, and the cap itself is source-specific (source.max_sample_size).

I can move the logic to the DataSource class as a clamp_sample_size() method so that the validation lives alongside the source-specific limit it depends on. Would that address the concern?

src/databricks/labs/lakebridge/reconcile/recon_config.py
ganeshdogiparthi-db: We can clean thisup in next PR's. This is not related to my changes.

src/databricks/labs/lakebridge/reconcile/recon_config.py
ganeshdogiparthi-db: Engine cap can't live on Table — it has no source reference;

BesikiML:
Thanks for the update. Overall this looks solid, and docs/tests coverage is strong.

I have one concern:

### Potential issue: Databricks temp-view cleanup in sampling path

In `SamplingQueryBuilder` (Databricks path), each sampling query now creates a unique temp view (`recon_keys_<uuid>`) via `createOrReplaceTempView(...)`.

I checked the reconcile flow and don’t see a corresponding drop/cleanup for those views after query execution.

Because the view name is unique per call, this can accumulate temp views during longer runs (many tables and multiple sampling passes), which may increase Spark session metadata/memory overhead.

**Suggestion:** track the created view name and explicitly drop it after the read step (or wrap create/query/drop in a scoped helper).

m-abulazm (commented):
Unblocking although the UX still needs improvement

BesikiML:
1. Aggregate reconcile ignores sampling_options
reconcile_data threads max_sample_size=table_conf.get_max_sample_size(), but reconcile_agg_data_per_rule still hardcodes DEFAULT_SAMPLE_ROWS:

```
missing_in_src=missing_in_src.limit(DEFAULT_SAMPLE_ROWS),
missing_in_tgt=missing_in_tgt.limit(DEFAULT_SAMPLE_ROWS),
```
A user setting sampling_options: { specifications: { value: 500 } } on an aggregates table gets 500 for row-level reconcile but 50 for aggregates-reconcile. Either wire get_max_sample_size() here or document aggregates as a known limitation.

2. Temp-view cleanup   **resolved**
My earlier concern about recon_keys_<uuid> views accumulating is addressed by _drop_sample_temp_views() in TriggerReconService, called per table in a finally block. Scoped to our prefix via SHOW VIEWS, looks good.

3. Silent lower-bound fallback
value <= 0 silently resets to 50 in SamplingSpecifications.__post_init__, while upper-bound capping logs a WARNING. Consider a warning on the lower bound too for consistency (acknowledged m-abulazm's UX note).

4. PR hygiene (nit)
Resolves #.. in the PR body should link a real issue or be removed.
SUPPORTED_OPERATIONS / discover-tables constants in recon_config.py look like merge artifacts, unrelated to sampling.
5. Test coverage gap (nit)
Codecov shows _drop_sample_temp_views (~12 lines in trigger_recon_service.py) largely untested. A small unit test mocking spark.sql("SHOW VIEWS") would close that gap.

Happy to approve once aggregate sampling is either wired or explicitly documented.

ganeshdogiparthi-db:
- A user setting sampling_options: { specifications: { value: 500 } } on an aggregates table gets 500 for row-level reconcile but 50 for aggregates-reconcile. Either wire get_max_sample_size() here or document aggregates as a known limitation.
  - Actually, the sampling logic is never used for aggregate reconciliation, so I documented it as a known limitation.
- Silent lower-bound fallback
value <= 0 silently resets to 50 in SamplingSpecifications.post_init, while upper-bound capping logs a WARNING. Consider a warning on the lower bound too for consistency (acknowledged m-abulazm's UX note).
  - Done. Also added this in test case.
 
- PR hygiene (nit)
Resolves #.. in the PR body should link a real issue or be removed.
SUPPORTED_OPERATIONS / discover-tables constants in recon_config.py look like merge artifacts, unrelated to sampling.
  - Done.
- Test coverage gap (nit)
Codecov shows _drop_sample_temp_views (~12 lines in trigger_recon_service.py) largely untested. A small unit test mocking spark.sql("SHOW VIEWS") would close that gap.
  - There was an issue with our process. The `drop_views` function I added was private. If I wrote a test for the private function, the linter complained. If I didn't write a test, the code coverage check failed. To avoid both issues, I moved the function out of the class, made it public, and added a test for it.

BesikiML:
Well, if @m-abulazm  does not have any concerns about it, I can approve it.

ganeshdogiparthi-db:
> Well, if @m-abulazm does not have any concerns about it, I can approve it.

@m-abulazm has already reviewed and commented on this PR. Isn't that sufficient, or are you looking for another round of approval?

BesikiML (approved):

PR #2499, opened 2026-06-10
Add Snowflake rate_sheet extract for TCO rate/tier

jneil17:
## What
Adds a `rate_sheet` extract to the Snowflake profiler that pulls the effective per-credit/unit rate from `SNOWFLAKE.ORGANIZATION_USAGE.RATE_SHEET_DAILY` — a 90-day average, scoped to the current account and split by rating type. The same rows also carry the account tier (`SERVICE_LEVEL`) and the billing currency.

Also logs the absolute path of the DuckDB extract on successful completion, so users can find the output file (this benefits all profilers, not just Snowflake).

## Why
The downstream TCO / value model needs the customer's effective `$/credit` rate and Snowflake tier to convert consumed credits into dollars and to derive their contract discount. The profiler emitted credits but no rate, so the dollar conversion had to be sourced manually.

## Changes
- New `resources/assessments/snowflake/rate_sheet.sql`
- Register the `rate_sheet` step in the Snowflake `pipeline_config.yml`
- Log the extract output path on successful completion in `pipeline.py`

## Access requirement
`ORGANIZATION_USAGE` lives in the shared `SNOWFLAKE` database. The existing default role `ACCOUNTADMIN` can read it from an ORGADMIN-enabled account (or grant a custom role the `ORGANIZATION_USAGE_VIEWER` database role). When access isn't available the step simply returns 0 rows and is skipped — the run still succeeds.

## Testing
- Ran the Snowflake profiler end-to-end against an ORGADMIN-enabled account; `rate_sheet` populated with tier (Enterprise), region, currency (USD), and effective rate ($3.00/credit).
- `tests/unit/assessment/` unit tests pass (87 passed).

Classifying usage into the value model (BI/ETL split, discount derivation) is downstream work outside this PR.

### DIFF
--- a/src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
+++ b/src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
@@ -65,3 +65,13 @@ steps:
     mode: overwrite
     frequency: daily
     flag: active
+
+  # Effective rate per credit/unit and the account tier (SERVICE_LEVEL). Requires
+  # SNOWFLAKE.ORGANIZATION_USAGE access (ACCOUNTADMIN in an ORGADMIN-enabled account).
+  # Feeds the downstream TCO model.
+  - name: rate_sheet
+    type: sql
+    extract_source: src/databricks/labs/lakebridge/resources/assessments/snowflake/rate_sheet.sql
+    mode: overwrite
+    frequency: once
+    flag: active
--- a/docs/lakebridge/docs/assessment/profiler/snowflake.mdx
+++ b/docs/lakebridge/docs/assessment/profiler/snowflake.mdx
@@ -9,6 +9,7 @@ import Admonition from '@theme/Admonition';
 ## Prerequisites
 
 - A Snowflake user with permission to create Personal Access Tokens (PATs) and to query the `SNOWFLAKE.ACCOUNT_USAGE` schema (e.g., the `ACCOUNTADMIN` role or a role granted `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`).
+- The `rate_sheet` step also reads `SNOWFLAKE.ORGANIZATION_USAGE.RATE_SHEET_DAILY` for the effective credit rate and account tier. This is available to `ACCOUNTADMIN` in an ORGADMIN-enabled account (the first account in an organization is ORGADMIN-enabled by default, so single-account orgs are covered), or to a custom role granted the `ORGANIZATION_BILLING_VIEWER` database role. If the account isn't ORGADMIN-enabled — or the organization is on a reseller contract — the view returns no rows, so `rate_sheet` is skipped and the rest of the profile still completes.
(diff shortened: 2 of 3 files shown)


src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -75,6 +75,7 @@ def execute(self) -> list[StepExecutionResult]:
             logger.error(error_msg)
             raise RuntimeError(error_msg)
 
+        logger.info(f"Profiler extract written to: {self._db_path}")
```
m-abulazm: this log is duplicate of log in `databricks.labs.lakebridge.assessments.profiler.Profiler._execute` or do you find it better here?

jneil17: I just wanted it to print out the location of the path when it finishes it, when I did the commit I don't think I had the other logger. Let me remove this and use the one that already exists. It is just helpful to have at the end of all the runs to print out the location so it is easier for the user to find it in their files

jneil17: Removed in 18acf503 — relying on the existing `Profiler._execute` log, which already prints the resolved `db_path` at the end of the run, so this was a duplicate. Thanks for catching it.

jneil17: removed in 18acf503 — relying on the existing Profiler._execute log, which already prints the resolved db_path at the end of the run, so this was a duplicate. Thanks for catching it.

dgomez04 (approved):

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -65,3 +65,13 @@ steps:
     mode: overwrite
     frequency: daily
     flag: active
+
+  # Effective rate per credit/unit and the account tier (SERVICE_LEVEL). Requires
+  # SNOWFLAKE.ORGANIZATION_USAGE access (ACCOUNTADMIN in an ORGADMIN-enabled account).
```
m-abulazm, line 70: do we need to update docs to mention this.

jneil17: Good call — added a prerequisites note in 5b728eed (docs/.../profiler/snowflake.mdx).

On the access question: confirmed against Snowflake's docs/KB. RATE_SHEET_DAILY is readable by ACCOUNTADMIN in an ORGADMIN-enabled account — and the first account in an org is ORGADMIN-enabled by default, so single-account orgs (the common case) are covered out of the box. For a non-ORGADMIN-enabled account, or a reseller contract, the view returns 0 rows (it does not error), so the step is simply skipped and the rest of the profile completes — no hard failure. A custom role can also be granted ORGANIZATION_BILLING_VIEWER. The docs note captures this.

m-abulazm (approved):

PR #2502, opened 2026-06-11
Simplify BigQuery SQL resources

m-abulazm:
## Changes

  ### What does this PR do?
-   Refactors how the BigQuery profiler parameterizes and loads its SQL templates
  - Collapses the BigQuery scripting boilerplate (`DECLARE`/`SET`/`EXECUTE IMMEDIATE FORMAT(...)`) in the templates down to plain queries
  - Flattens the resource layout to a single `bigquery/resources/` dir and simplifies resource loading.

  - `substitute(raw_sql, params)` is variable-agnostic and stateless, so other source-techs can reuse it.
  - The 16 templates now use `{{project_region}}` / `{{profiling_window_in_days}}`.  `time_format` is inlined as
  the constant `'%Y-%m-%dT%H'` (it was never a substituted variable).
  - Deletes `substitutions.json` and the old line-anchored substituter.
  - Resource layout flattened: all SQL + `analysis_types.json` moved under `bigquery/resources/`

  - `substitute()` injects values **verbatim, unescaped** — `project_region` flows into a table-path identifier (which can't be parameterized), so input validation must stay upstream (it's built from credential config today).

  ### Functionality
  - [x] internal refactor of the BigQuery profiler resources 

  ### Tests
  - [x] added unit tests


take60 (approved):
Reviewed and smoke-tested against real BigQuery — LGTM.

Nice simplification. Replacing the line-anchored substituter + `substitutions.json` with the stateless `{{placeholder}}` `substitute()` is a big readability win, and collapsing the `DECLARE`/`SET`/`EXECUTE IMMEDIATE FORMAT(...)` boilerplate down to plain queries makes the templates much easier to follow. The single-pass substitution (a substituted value containing `{{...}}` is never re-scanned) and the raise-on-unfilled-placeholder behavior are good safety properties, and the verbatim-injection caveat is clearly documented — `project_region` validation staying upstream in the credential config is the right call.

Smoke test (PR head, against an internal BigQuery project, reservations + streaming included):
- All 16 SQL templates executed against live `INFORMATION_SCHEMA` with **no query errors** — confirms the boilerplate collapse + `{{...}}` substitution end-to-end.
- 12 tables produced, names unchanged; row counts in line with before (`jobs_timeline_by_reservations` 1453, `fulfillment_analysis` 445, `timeline_analysis` 542, `streaming_summary` 40, `workload_types` 2, `table_storage` 1).
- `table_storage` retains the full schema including `metadata_level`, `storage_billing_model`, and `dataset_count`.
- `status: success`, exit 0.

Also confirmed it preserves the per-`(project, region)` soft-fail resilience and the `profiler`-nested credential config. Thanks for cleaning this up.

PR #2503, opened 2026-06-11
[DEV] Refactor DatabaseConfig usages in reconcile runtime and tests

### LINKED ISSUE #2417: [DEV] Refactor `DatabaseConfig` usages
`DatabaseConfig` is explicitly marked as a backwards-compat shim. Replace usages with the new classes `SourceConnectionConfig` and `TargetConnectionConfig`

BesikiML:
## Summary
- Refactored reconcile runtime internals to use `SourceConnectionConfig` and `TargetConnectionConfig` instead of `DatabaseConfig`.
- Updated `Reconciliation`, `TriggerReconService`, and `TriggerReconAggregateService` to pass source/target connection config objects explicitly.
- Updated integration tests in reconcile query-builder and aggregate/capture paths to align with the new config model.
- Removed remaining `DatabaseConfig` usage from `tests/integration/reconcile/test_recon_capture.py`.
## Backward Compatibility
- Preserved `ReconCapture` constructor compatibility for legacy `DatabaseConfig`-style callers during the migration period.
## Validation
- `uv run python -m pytest tests/integration/reconcile/test_recon_capture.py -q`  
  - `23 skipped`
- `uv run python -m pytest tests/integration/reconcile/query_builder/test_execute.py -q`  
  - `1 passed, 15 skipped`
- `uv run python -m pytest tests/integration/reconcile/test_aggregates_reconcile.py -q`  
  - `4 passed, 2 skipped`
- `uv run python -m pytest tests/integration/reconcile/test_aggregates_recon_capture.py -q`  
  - `1 skipped`
## Notes
- `uv.lock` changes from local environment sync were intentionally excluded from this PR.

### Linked issues
Fixed issue : https://github.com/databrickslabs/lakebridge/issues/2417

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


uv.lock
```diff
@@ -10,7 +10,7 @@ resolution-markers = [
 ]
 
 [options]
-exclude-newer = "2026-06-04T14:11:46.713298Z"
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
```
m-abulazm: this needs to be reverted. please use the Makefile commands e.g. `make dev` to install dependencies otherwise you will get different dependencies

m-abulazm (changes requested):
Please use the Makefile commands e.g. make dev to install dependencies otherwise you will get different dependencies

uv.lock
BesikiML: done

m-abulazm (changes requested):
please remove all `DatabaseConfig` usage from the code

src/databricks/labs/lakebridge/reconcile/recon_capture.py
```diff
@@ -273,13 +280,11 @@ def _generate_recon_main_id(
...
     ) -> int:
         full_source_table = (
-            f"{self.database_config.source_schema}.{table_conf.source_name}"
-            if self.database_config.source_catalog is None
-            else f"{self.database_config.source_catalog}.{self.database_config.source_schema}.{table_conf.source_name}"
-        )
-        full_target_table = (
-            f"{self.database_config.target_catalog}.{self.database_config.target_schema}.{table_conf.target_name}"
+            f"{self.source_connection.schema}.{table_conf.source_name}"
+            if self.source_connection.catalog is None
```
m-abulazm: catalog is not optional. the goal of this refactoring is to simplify the code with the new type classes. please reflect that

m-abulazm (changes requested):

src/databricks/labs/lakebridge/reconcile/recon_capture.py
BesikiML: done

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -293,6 +297,9 @@ def _get_reconcile_aggregate_output(
                 if data_source_exception:
                     rule_reconcile_output = DataReconcileOutput(exception=str(data_source_exception))
                 else:
+                    assert joined_df is not None
+                    assert src_data is not None
+                    assert tgt_data is not None
```
m-abulazm: please do not use `assert` in production code. should revert or what is the goal here?

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -348,13 +355,13 @@ def _get_sample_data(
...
-                    self._database_config.source_catalog,
-                    self._database_config.source_schema,
+                    self._source_connection.catalog,
+                    self._source_connection.schema,
                     table_conf.source_name,
                 )
 
         return DataReconcileOutput(
-            mismatch=mismatch,
+            mismatch=cast(Any, mismatch),
```
m-abulazm: ?

m-abulazm (changes requested):
remove the class itself as well since it is not used anymore

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: Agreed, those `asserts` were a mypy workaround, not intentional runtime checks. mypy couldn’t narrow `joined_df` / `src_data` / `tgt_data` after the `try`/`except` when they were initialized as `None`.

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: Same story, that was a mypy hack. `mismatch` starts as `None` but `DataReconcileOutput.mismatch` expects `MismatchOutput`.

BesikiML:
@m-abulazm 
Good catch, agreed. `DatabaseConfig` and the `database_config` property was kept temporarily as part of the v2 migration but production code no longer uses them

src/databricks/labs/lakebridge/reconcile/recon_output_config.py
```diff
@@ -22,7 +22,7 @@ class DataReconcileOutput:
     mismatch_count: int = 0
     missing_in_src_count: int = 0
     missing_in_tgt_count: int = 0
-    mismatch: MismatchOutput = field(default_factory=MismatchOutput)
+    mismatch: MismatchOutput | None = None
```
m-abulazm: why are you changing this class

src/databricks/labs/lakebridge/reconcile/reconciliation.py
```diff
@@ -259,44 +262,40 @@ def _get_reconcile_aggregate_output(
...
                 )
-            except DataSourceRuntimeException as e:
-                data_source_exception = e
-
-            # For each Aggregated Query, reconcile the data based on the rule
-            for rule in src_query_with_rules.rules:
-                if data_source_exception:
-                    rule_reconcile_output = DataReconcileOutput(exception=str(data_source_exception))
-                else:
+                for rule in src_query_with_rules.rules:
```
m-abulazm, line 286: this diff looks off

m-abulazm (changes requested):

src/databricks/labs/lakebridge/reconcile/recon_output_config.py
BesikiML: This was not part of the original `DatabaseConfig` refactor.
In `507f692fa` I removed `cast(Any, mismatch)` from `_get_sample_data` and temporarily defaulted to `MismatchOutput()` when sampling was skipped. That broke integration tests that expect `actual.mismatch is None` when there are no mismatches (e.g. `test_reconcile_data_without_mismatches_and_missing`).
On `main`, `_get_sample_data` already returns `mismatch=None` in those cases — the `default_factory=MismatchOutput` on the dataclass field wasn't reflected at runtime. `2a8da1fab` restores that behavior and makes the type match:
- `mismatch: MismatchOutput | None = None`
- `None` guard before accessing `.mismatch.*` in `recon_capture` and tests.

src/databricks/labs/lakebridge/reconcile/reconciliation.py
BesikiML: This restructure is from your earlier feedback to remove production `assert`s in aggregate reconciliation.

**Before:**
- `try/except` captured `data_source_exception`
- A single `for rule` loop used `if data_source_exception: ... else: assert joined_df/src_data/tgt_data is not None; reconcile...`.

**After:**
- Success path: `read → join → for rule: reconcile` inside `try`
- Failure path: `for rule: DataReconcileOutput(exception=...)` in `except`
No functional change ,  still one output per rule, with exceptions propagated the same way. The diff is noisy because the loop moved inside the `try`/`except` blocks instead of sitting between them.

m-abulazm (approved):
LGTM

bishwajit-db (approved):
LGTM

PR #2504, opened 2026-06-12
log source system in recon commands

m-abulazm:
### What does this PR do?
add source system and report type to user agent extras


gueniai (approved):
LGTM

PR #2511, opened 2026-06-17
capture variant in execute-profiler

m-abulazm:
### What does this PR do?
Introduce the concept of a variant of a profiler. the user is prompted to select a variant when executing profiler if supported

- **Variant model** — add `SOURCE_SYSTEM_VARIANTS` ; `PROFILER_SOURCE_SYSTEM` is now the flat list of real source systems. Remove `REDSHIFT_VARIANTS`, and `source_system_family()`.
- **Pipeline resolution** — Remove `SOURCE_SYSTEM_TO_PIPELINE_CFG`, replace the path map with convention-based `get_pipeline(source_system, variant)` → `resources/assessments/<source>/<variant>/pipeline_config.yml`.
- **CLI** — `execute-database-profiler` accepts `--variant`. `parse_profiler_variant` prompts for a variant when the source has variants and none is given.

- `configure-profiler` and `test-profiler-connection` currently do not accept variant.
- integration with desktop app needs to be defined

### Functionality

- [ ] added new CLI command

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests


dgomez04 (changes requested):
The addition looks good to me, however: 

On `src/databricks/labs/lakebridge/deployment/profiler_dashboard.py:70`:

The pull request flattened `PROFILER_SOURCE_SYSTEM` so Redshift is now selected as `redshift`, and that value flows into `ProfilerDashboardConfig.source_tech` via `install.py:463`. This guard still matches the old prefix: `"redshift".startswith("redshift_")` is `False`, so the short circuit doesn't fire.

`install` will try to deploy a dashboard/ingestion job, which isn't the expected behavior. 

On `tests/unit/deployment/test_profiler_dashboard_redshift.py:30`:

The real flow produces `source_tech="redshift"`, not `redshift_{variant}`. This test still asserts against the old spelling, which is why it stays green despite the `_is_dashboard_unsupported` regression. We should switch `source_tech=f"redshift_{variant}"` to `source_tech="redshift"` so the test exercises what the product actually emits — it should fail until the guard is fixed.

dey-abhishek:
+1 on the variant model — this is the clean generalization we need for the **Teradata profiler** (`core` / `pdcr`), which today rides the same flattened `teradata_{variant}` +
  `source_system_family()` scheme this PR removes.

  the TD branch layout already matches the new convention:
  
  ```
  resources/assessments/teradata/core/pipeline_config.yml
  resources/assessments/teradata/pdcr/pipeline_config.yml
  ```

  So on my side this should drop in as just:

  ```python
  SOURCE_SYSTEM_VARIANTS = {
      "redshift": ("serverless", "provisioned", "provisioned_multi_az"),
      "teradata": ("core", "pdcr"),
  }
  ```

  **One alignment note on the dashboard guard:** Teradata also ships no profiler dashboard, so `_is_dashboard_unsupported` needs to cover it too. With the guard currently being:

  ```python
  return source_tech.lower().startswith("redshift_")
  ```

  Teradata wouldn't be caught regardless of the flattening. Suggest fixing this **model-driven** — e.g. a source-keyed "no dashboard" flag alongside `CONNECTOR_REQUIRED` / `SOURCE_SYSTEM_VARIANTS` —
  rather than per-source string prefixes, so that:

  - both Redshift and Teradata are covered, and
  - future variant-based sources don't re-trip the same bug.

  No changes needed for this PR to land first — just flagging so the dashboard fix accounts for Teradata too.

dgomez04 (approved):
We will fix the dashboard in a follow-up, as discussed with Mohammed. Approved.

dey-abhishek (commented):
Really like the direction of this PR — the variant model cleans things up nicely. 

  One thing I wanted to flag before it merges, since it affects both Redshift and our upcoming Teradata work:

  Once Redshift is flattened to `"redshift"`, the dashboard guard doesn't catch it anymore:

  ```python
  # profiler_dashboard.py
  return source_tech.lower().startswith("redshift_")   # "redshift" no longer matches
  ```

  When that check misses, `install` goes ahead and tries to deploy a dashboard + ingestion job for Redshift — but Redshift doesn't have a dashboard, so that's not what we want.

  While digging in, I noticed the bigger picture: the dashboard deploy looks for a template at `assessments/dashboards/{source_tech}`, and **only `synapse` and `mssql` actually have one**. Snowflake, Oracle, legacy_synapse, BigQuery, Redshift, and Teradata all have none. So rather than listing out the sources *without* a dashboard (which we'd have to keep updating), it'd be simpler and more 
  future-proof to flip it around and just list the two that *do* have one.

  Something like this in `assessments/_constants.py`:

  ```python
  # Source systems that ship a profiler dashboard template under
  # resources/assessments/dashboards/<source>. Everything else just gets the
  # DuckDB extract delivered to the UC Volume (no dashboard/ingestion job).
  DASHBOARD_SUPPORTED = {"synapse", "mssql"}
  ```

  and in `profiler_dashboard.py`:
  
  ```python
  from databricks.labs.lakebridge.assessments import DASHBOARD_SUPPORTED

  @staticmethod
  def _is_dashboard_unsupported(source_tech: str) -> bool:
      return source_tech.lower() not in DASHBOARD_SUPPORTED
  ```

  One small related thing: `tests/unit/deployment/test_profiler_dashboard_redshift.py` still uses `source_tech=f"redshift_{variant}"`, which is why it stays green today. Switching it to `source_tech="redshift"` (what the product now emits) would let it actually catch this.

With that change, Redshift, Teradata, and any other dashboard-less source all skip the dashboard step cleanly while still getting their extract to the UC Volume — and adding a dashboard for a new source later is just one line. Thanks again for putting this together!

dey-abhishek (approved):
Approved as discussed for follow-up actions

dey-abhishek (approved):
LGTM

PR #2512, opened 2026-06-17
implement missing bigquery test connection and cleanup profiler config

m-abulazm:
### What does this PR do?
Fix connection testing implementation that relied on magic if conditions and duplicate config that was not modeled correctly and could be broken easily

Introduce `AssessmentConfigurator.test_connection()` to fetch the saved creds and delegates to `_check_connection()`:
    - base → `DatabaseManager` health check (snowflake, oracle, mssql, legacy_synapse, redshift)
    - `ConfigureSynapseAssessment` → `validate_synapse_pools` (per SQL pool)
    - `ConfigureBigQueryAssessment` → `validate_bigquery_pairs` (per project/region)

This removes the `if source_tech == "synapse"` special-case  in `test-profiler-connection`  and the hand-maintained `CONNECTOR_REQUIRED` source-system flag which can be derived automatically

### Functionality

- [X] modified existing command: `databricks labs lakebridge test-profiler-connection`
- [X] modified existing command: `databricks labs lakebridge configure-database-profiler`

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/connections/bigquery_connection_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+
+    failures: dict[str, str] = {}
+    for pair in pairs:
+        label = f"{pair['project']}.{pair['region']}"
+        logger.info(f"Testing connection to BigQuery {label}...")
+        try:
+            client = bigquery.Client(project=pair["project"], location=pair["region"])
+            client.query("SELECT 1").result()
+            logger.info(f"BigQuery {label} connection successful")
+        except GoogleAPICallError as e:
```
dgomez04: This `except` is too narrow and is what's turning `test-python` red across all versions. `test_validate_bigquery_pairs_aggregates_failures` raises a generic exception, which escapes uncaught `RuntimeError: boom`.

More importantly, the real failure `google.auth.exceptions.DefaultCredentialsError` is not a `GoogleAPICallError `, so it would propagate a raw traceback.

In the `configure-database-profiler` flow that's unhandled, since `AssessmentConfigurator.run()` only catches `ConnectionError`.

What about mirroring synapse and adding `except Exception: # pylint: disable=broad-exception-caught`?

dgomez04 (approved):
The changes are fine, fix the pending comment so that CI is green before merging.

dey-abhishek (changes requested):
## Code Review

The PR correctly removes the brittle `CONNECTOR_REQUIRED` dict and `if source_tech == "synapse"` special-case, replacing them with a clean polymorphic `_check_connection()` override per source. The design is sound. The following issues should be addressed before merge.

---

### Findings

| Severity | Location | Finding |
|----------|----------|---------|
| **Medium** | `connections/bigquery_connection_helpers.py:868` | `bigquery.Client` is never closed — resource leak |
| **Medium** | `cli.py:682-687` (`_table_recon_exists`) | Only catches `NotFound`; corrupted JSON files raise `ValueError`/`SerdeError` and crash the command |
| **Medium** | `cli.py` | `visualize_profiler_results` command was silently removed — breaking change not mentioned in the PR description |
| **Low** | `connections/database_manager.py:886-889` | `TrustServerCertificate` logic `str(x) == 'False'` is fragile — breaks if credential file stores value as string `"no"` or `"false"` |
| **Low** | `connections/bigquery_connection_helpers.py:865` | Field `"region"` in the credential file is passed as `location=` to the BigQuery client — naming inconsistency with GCP terminology |

---

### Details

**1. `bigquery.Client` resource leak** (`connections/bigquery_connection_helpers.py:868`)

`bigquery.Client` holds an underlying HTTP session and must be closed after use.

```python
# before
client = bigquery.Client(project=pair["project"], location=pair["region"])
client.query("SELECT 1").result()

# after
with bigquery.Client(project=pair["project"], location=pair["region"]) as client:
    client.query("SELECT 1").result()
```

---

**2. `_table_recon_exists` missing error cases** (`cli.py:682-687`)

If the workspace file exists but contains malformed JSON, `Installation.load` raises `ValueError` (or `SerdeError`), which propagates uncaught and crashes `_run_auto_configure_recon_tables`. The pattern in `install.py` already catches both.

```python
# before
except NotFound:
    return False

# after
except (NotFound, ValueError):
    return False
```

---

**3. Silent removal of `visualize_profiler_results` command** (`cli.py`)

The command is deleted with no deprecation notice, no mention in the PR description, and no migration path. Anyone calling `databricks labs lakebridge visualize-profiler-results` will get a hard failure. Please document this in the PR description and changelog/release notes.

---

**4. `TrustServerCertificate` logic is fragile** (`connections/database_manager.py:886-889`)

`str("no") == 'False'` evaluates to `False`, so a manually set `trust_server_certificate: "no"` in the credential file would silently flip to `"yes"`.

```python
# before
"TrustServerCertificate": (
    "no" if str(self.config.get('trust_server_certificate', 'False')) == 'False' else "yes"
),

# after
"TrustServerCertificate": "yes" if self.config.get('trust_server_certificate', False) else "no",
```

---

**5. `"region"` vs `"location"` naming** (`connections/bigquery_connection_helpers.py`)

The credential file stores the key as `"region"` and the user prompt uses `project.region`, but it is passed to `bigquery.Client(location=...)`. BigQuery's API and docs use *location* (e.g., `US`, `EU`, `us-central1`). Consider renaming the credential key to `"location"` for clarity, or at minimum document the mapping in the prompt text.

dey-abhishek (approved):
Finding #1 fixed ✅. Remaining findings (#2–#5) acknowledged and tracked for a follow-up PR. LGTM.

PR #2516, opened 2026-06-19
Update expected SQL assertion

### LINKED ISSUE #2515: Transpile integration tests fail on stale exact-match of Databricks `[UNRESOLVED_ROUTINE]` search path
## Summary

The `transpile` acceptance tests started failing around 2026-06-18.

**Root cause:** `test_transpile_teradata_sql` (Bladebridge) and the Morpheus equivalent assert the **verbatim** Databricks `[UNRESOLVED_ROUTINE]` error for `select cole(hello) world from table`, including the full function search path. Databricks changed that search path server-side — it now includes `system.ai` and reorders `session`/`builtin`:

- expected (hard-coded in tests): `` [`system`.`builtin`, `system`.`session`, `catalog`.`schema`] ``
- actual (Databricks now returns): `` [`system`.`session`, `system`.`builtin`, `system`.`ai`, `catalog`.`schema`] ``

This is a platform change in the workspace the suite runs against. This is *not* a validation regression

`select cole(hello) ...` was **always** expected to fail validation — the test asserts exactly one validation error for it (`validation_error_count: 1`) plus an exception block in the output. Validation behaviour is unchanged and correct; only the **captured error string** differs. The assertion that actually fails is `assert_sql_match` on the `dummy_function.sql` output block (`tests/integration/transpile/common_utils.py:35`), which diffs the full failure block including the search path.

## Affected assertions

- `tests/integration/transpile/test_bladebridge.py`
- `tests/integration/transpile/test_morpheus.py`

The unit test (`tests/unit/helpers/test_validation.py`) only checks the `[UNRESOLVED_ROUTINE]` substring, so it is unaffected.

## Proposed fix

Make the assertions resilient to the server-controlled search-path text (normalize/ignore the bracket contents) rather than re-pinning the literal string — Databricks will likely keep adding `system.*` schemas, and re-pinning would just break again on the next change.

m-abulazm:
### What does this PR do?
Fix integration test assertion after databricks workspace behavior change

 - before: `` [`system`.`builtin`, `system`.`session`, `catalog`.`schema`] ``
 - after:  `` [`system`.`session`, `system`.`builtin`, `system`.`ai`, `catalog`.`schema`] ``

 A more resilient fix (wildcarding the path inside the assertion) was considered and deliberately skipped in favour of the minimal change.

### Linked issues

Resolves #2515 

### Functionality

- [X] update test code

### Tests

- [ ] manually tested
- [ ] added unit tests
- [X] added integration tests

### DIFF
--- a/tests/integration/transpile/test_bladebridge.py
+++ b/tests/integration/transpile/test_bladebridge.py
@@ -318,7 +318,7 @@ PRIMARY KEY (col1,col3) )
 TBLPROPERTIES('delta.feature.allowColumnDefaults' = 'supported');"""
     expected_validation_failure_sql = """-------------- Exception Start-------------------
 /*
-[UNRESOLVED_ROUTINE] Cannot resolve routine `cole` on search path [`system`.`builtin`, `system`.`session`, `catalog`.`schema`].
+[UNRESOLVED_ROUTINE] Cannot resolve routine `cole` on search path [`system`.`session`, `system`.`builtin`, `system`.`ai`, `catalog`.`schema`].
 */
 select cole(hello) world from table;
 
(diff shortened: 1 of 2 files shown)


asnare (approved):
Thanks for picking this up so quickly.

bishwajit-db (approved):
LGTM

PR #2527, opened 2026-06-24
Add BigQuery as a reconcile source

take60:
## What
Adds **BigQuery** as a reconcile source, at parity with the other sources (`schema` / `row` / `data` / `all` / `aggregate`). BigQuery was already supported by the transpiler and profiler; this closes the gap for reconcile.

## How it works
BigQuery uses the same **Lakehouse Federation `remote_query`** path as Snowflake/Oracle/etc. — no new dependencies. `BigQueryDataSource` uses backtick-quoted 3-part `project.dataset.table` names and reads metadata from `INFORMATION_SCHEMA.COLUMNS`. sqlglot already ships a BigQuery dialect, so query generation and schema comparison are reused unchanged.

## Changes
- **Connector** `reconcile/connectors/bigquery.py` (`BigQueryDataSource`); registered in `ReconSourceType` and `source_adapter.create_adapter`; install prompt + `recon_capture` display name.
- **Schema type handling**: the connector's `INFORMATION_SCHEMA` query canonicalizes the few BigQuery types sqlglot can't bridge to Databricks (`BIGNUMERIC`→`string`, bare `NUMERIC`→`decimal(38,9)`, `TIME`→`string`, `JSON`→`variant`, `RANGE<T>`→`struct<…>`); everything else is left to sqlglot.
- **Row hashing**: adds a BigQuery hash algorithm (`TO_HEX(SHA256())`, matching Databricks `sha2(...,256)`) and a **scale-aware decimal transform** (`FORMAT('%.<scale>f', col)`) so BigQuery's trailing-zero-stripped numeric strings match Spark's scale-padded `DECIMAL` strings.
- **Docs + tests**: supported-sources row, a BigQuery config tab (with a compute note), connector tests, a type-coverage guardrail test, and an adapter test.

## Testing
- `make fmt` / `make lint` (pylint 10.0/10.0) and `make test` — green (1319 passed).
- **End-to-end on a real workspace**: a BigQuery source (UC Federation connection) reconciled against an identical Databricks copy via the deployed reconcile **job** — `schema`, `row`, and `data` all matched (0 mismatches / 0 missing).

## Notes / limitations
- **Compute**: BigQuery reads use `remote_query`, which requires **Databricks Runtime 17.3+** or **serverless** compute (the reconcile job's default cluster may run an older runtime — point it at a DBR 17.3+ cluster via `job_overrides.existing_cluster_id`, or run serverless). This applies to all Lakehouse Federation reconcile sources. Documented in the config tab.
- `INTERVAL` maps to two Databricks columns, which the 1:1 schema comparison can't represent — surfaces as a visible mismatch (documented).


m-abulazm (changes requested):
added UC connection with name `bigquery_sandbox` for the e2e test

take60:
@m-abulazm thanks for setting up the connection.

  All 4 BigQuery acceptance tests now fail on the same single cause — a connection grant, not
  code:

  PERMISSION_DENIED: User does not have USE CONNECTION on Connection 'bigquery_sandbox'
  (the deployed-job path shows it masked as bigquery_TEST_CATALOG, but it's the same
  connection.)

  Two things from your side:

  1. Grant USE CONNECTION on the BigQuery connection to the acceptance test principal
  2. Heads-up for the next step: once the grant is in, remote_query will try to materialize
  results into the read dataset (public) — so please also make sure that dataset is writable by
  the connection's service account, otherwise we'll hit that right after. (Not failing on this
  yet — execution stops at the grant first.)

  Thanks

bishwajit-db:
**Materialization dataset: undocumented write requirement + no config knob**

In production the BigQuery materialization target always defaults to the read dataset (`_mat_dataset` falls back to `schema`; `create_adapter` never sets `materialization_dataset`). Two gaps follow from that:

1. **Docs:** the source dataset must be writable by the connection's service account (remote_query materializes results there), but only the DBR 17.3+ requirement is documented. Suggest a line in the BigQuery config notes, e.g.: "ensure the source dataset, or a dedicated materialization dataset, is writable by the connection's service account."

2. **Config:** `materialization_dataset` is a constructor-only arg with no path from config (`SourceConnectionConfig` has no field, `create_adapter` never passes it), so it's only settable in test code. Adding it to `SourceConnectionConfig` and plumbing it through `create_adapter` would let users keep the source dataset read-only. It also fixes `list_schemas`, which passes `_mat_dataset("")` and resolves to an empty `materializationDataset` when none is configured.

The doc line unblocks the PR; the config change can be a follow-up.

take60:
Thanks @bishwajit-db for the review!
     
(1) Docs — done here: added a "Writable dataset required" note (source or a dedicated materialization dataset must be writable by the connection's SA). Also fixed the naming text for two-part — project comes from the UC connection, schema is the dataset, catalog mirrors it.

(2) Config — agreed, taking as a follow-up: add materialization_dataset to SourceConnectionConfig and thread it through create_adapter (also fixes list_schemas' empty materializationDataset). Separate PR so this can land on the doc fix. #2529

@m-abulazm thanks for setting up UC connection and grant the permission. added testing and verified.

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -0,0 +1,150 @@
...
+        self,
+        engine: Dialect,
+        reader: RemoteQueryReader,
+        materialization_dataset: str | None = None,
+    ):
+        self._engine = engine
+        self._reader = reader
+        # BigQuery's remote_query rejects `database`; a `query` push requires `materializationDataset`
+        # (a writable BigQuery dataset where results are materialized). Defaults to the dataset being
+        # read; set explicitly when the source dataset is read-only for the connection's service account.
```
m-abulazm:
```suggestion
```

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -0,0 +1,150 @@
...
+from databricks.labs.lakebridge.reconcile.connectors.data_source import DataSource
+from databricks.labs.lakebridge.reconcile.connectors.models import NormalizedIdentifier
+from databricks.labs.lakebridge.reconcile.connectors.remote_query_reader import RemoteQueryReader
+from databricks.labs.lakebridge.reconcile.connectors.dialect_utils import DialectUtils
+from databricks.labs.lakebridge.reconcile.recon_config import JdbcReaderOptions, Schema
+
+logger = logging.getLogger(__name__)
+
+
+class BigQueryDataSource(DataSource):
```
m-abulazm: lets use catalog as `materializationDataset` and document that clearly plus logs when using it

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -0,0 +1,150 @@
...
+      * BIGNUMERIC -> string  (precision 76 exceeds Databricks DECIMAL max 38; STRING preserves it)
+      * NUMERIC    -> decimal(38,9)  (bare NUMERIC is fixed 38/9; sqlglot would emit DECIMAL(10,0))
+      * TIME       -> string  (Databricks has no TIME type)
+      * JSON       -> variant
+      * RANGE<T>   -> struct<start <T>, end <T>>
+    All other types (INT64, FLOAT64, BOOL, STRING, BYTES, DATE, DATETIME, TIMESTAMP, NUMERIC(p,s),
+    GEOGRAPHY, ARRAY, STRUCT) are left raw because sqlglot translates them correctly.
+    INTERVAL is intentionally not mapped: it migrates to two Databricks columns, which the one-to-one
+    schema comparison cannot represent, so such columns surface as a visible mismatch.
+    """
```
m-abulazm:
```suggestion
```

m-abulazm: please avoid comments. Readable well-written code does not need comments

src/databricks/labs/lakebridge/reconcile/query_builder/base.py
```diff
@@ -144,6 +145,10 @@ def _get_transform(datatype: str):
             except sqlglot.errors.ParseError:
                 logger.warning(f"Could not parse datatype {datatype} for source {source_dialect}")
 
+            # BigQuery decimals need scale-aware padding to match Spark's CAST(DECIMAL AS STRING).
+            if source_dialect == "bigquery" and parsed == exp.DataType.Type.DECIMAL.value:
```
m-abulazm: we support this through `DataType_transform_mapping` I dont want to introduce another magical place that does things conditionally. 

we should improve `DataType_transform_mapping` to better support this. please revert this and add the new refactoring on a new dedicated PR

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
```diff
@@ -253,6 +279,12 @@ def _get_is_string(column_types_dict: dict[str, DataType], column_name: str) ->
 
 DataType_transform_mapping: dict[str, dict[str, list[partial[exp.Expression]]]] = {  # pylint: disable=invalid-name
     "universal": {"default": [partial(coalesce, default='_null_recon_', is_string=True), partial(trim)]},
+    "bigquery": {
+        # BigQuery CONCAT/|| and TRIM/COALESCE require STRING operands, so cast every column to STRING
+        # before concatenation (mirrors the tsql default). DATE/INT64/NUMERIC/STRING cast deterministically;
+        # TIMESTAMP/FLOAT64 string formatting parity with Databricks is a follow-up (explicit FORMAT needed).
```
m-abulazm:
```suggestion
        # TODO: add timestamps and numbers handling
```

src/databricks/labs/lakebridge/install.py
```diff
@@ -363,15 +363,27 @@ def _prompt_for_source_connection_config(self, dialect: str) -> SourceConnection
             catalog = self._prompts.question("Enter Oracle service name")
         elif dialect == ReconSourceType.DATABRICKS.value:
             catalog = self._prompts.question("Enter source Databricks catalog name", default="hive_metastore")
+        elif dialect == ReconSourceType.BIGQUERY.value:
+            # The BigQuery project is abstracted by the UC connection (its default project scopes
+            # unqualified names), so no project is prompted; the connector uses two-part dataset.table.
+            # catalog is set to the dataset below (it must be non-empty to round-trip through serde).
```
m-abulazm:
```suggestion
```

src/databricks/labs/lakebridge/install.py
```diff
@@ -363,15 +363,27 @@ def _prompt_for_source_connection_config(self, dialect: str) -> SourceConnection
...
         if dialect == ReconSourceType.ORACLE.value:
             schema_prompt = "Enter Oracle database name"
+        elif dialect == ReconSourceType.BIGQUERY.value:
+            schema_prompt = "Enter source BigQuery dataset name"
 
         schema = self._prompts.question(schema_prompt)
 
+        if dialect == ReconSourceType.BIGQUERY.value:
+            # BigQuery has no separate catalog; mirror the dataset so the value is non-empty (the
+            # connector ignores it). The project is abstracted by the UC connection.
```
m-abulazm:
```suggestion
```

m-abulazm (changes requested):
Nice work, a few things before this is good to merge:

**1. Use `catalog` as `materializationDataset`.** Rather than threading a separate `materialization_dataset` arg through the connector and defaulting it to the source dataset, let's just use `catalog` for this. We already carry `catalog` around for every source, it's already non-empty in the config and reusing it avoids inventing a parallel concept. 

**2. Drop the comments.** There's a lot of explanatory commentary here — the module docstring, the inline notes on the `_SCHEMA_QUERY` CASE, the `read_data` placeholder substitution, `list_schemas`, etc. Readable, well-written code doesn't need this. Please lift out anything that's genuinely surprising into the names/structure of the code itself and delete the rest.

**3. Revert the type canonicalization; do it through `DataType_transform_mapping`.** We already support this through `DataType_transform_mapping` and I don't want another magical spot that conditionally rewrites types (The special-cased `bigquery_decimal_transform` branch in `_get_transform`). Let's improve `DataType_transform_mapping` to handle these cases properly instead. Please revert this part here and bring the refactoring on a dedicated follow-up PR so it can be reviewed on its own.

Happy to pair on the `DataType_transform_mapping` extension if useful.

take60:
Thanks @m-abulazm — addressed all three (CI green):

1. **`catalog` as `materializationDataset`** — the connector now uses `catalog` for it (with a log line); dropped the `materialization_dataset` constructor arg. That made the separate config knob unnecessary, so I closed #2529.
2. **Comments** — removed the docstring prose and inline comments in the connector / install / fixtures.
3. **Decimal transform** — reverted `bigquery_decimal_transform` and the `_get_transform` branch; left a `# TODO`. Doing it properly via `DataType_transform_mapping` is tracked in #2537 (happy to pair).

PTAL.

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
take60: Done — `catalog` is now used as the `materializationDataset`, documented in the class docstring and logged on each read.

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
take60: Removed.

src/databricks/labs/lakebridge/reconcile/query_builder/base.py
take60: Reverted the conditional branch. Tracking the `DataType_transform_mapping` refactor in #2537.

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
take60: Applied — left the `# TODO: add timestamps and numbers handling` marker; full handling in #2537.

src/databricks/labs/lakebridge/install.py
take60: Removed.

src/databricks/labs/lakebridge/install.py
take60: Removed.

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
bishwajit-db: @take60 @m-abulazm  Repurposing catalog for the materialization dataset in BigQuery doesn’t feel natural to me. In the data ecosystem, catalog has a well-established meaning, so reusing it simply because the field is currently unused could be confusing and may not convey the intended semantics.

If certain connectors require connector-specific settings, wouldn’t it be clearer to introduce those settings explicitly instead of overloading an existing field? Of course, if the team believe catalog is semantically appropriate in this context, then that would be a different discussion.

take60: @bishwajit-db @m-abulazm — let me lay out the trade-off so we can land on one approach together.

The root tension: BigQuery's hierarchy is `project.dataset.table`, but the project is abstracted by the UC connection — we reference tables two-part (`dataset.table`), so the project never appears in the query. That leaves `catalog` (which semantically maps to *project*) with nothing to hold for BigQuery, which is how it ended up set to the dataset and reused as the `materializationDataset`.

@bishwajit-db's point stands: the materialization dataset is really a third concept — a writable scratch dataset — that's neither the project (`catalog`) nor the read dataset (`schema`), so overloading `catalog` for it is semantically off.

The fully-clean model would be:
- `catalog` → project (optional/empty here, since the connection supplies it)
- `schema` → dataset (read)
- a dedicated `materialization_dataset` field (writable scratch)

The cost is that making `catalog` optional touches `SourceConnectionConfig` and threads the new field through `create_adapter` (≈ what #2529 proposed).

Both options are reasonable and I'm happy to do the implementation either way — could we align on the direction? My slight lean is the explicit field for clarity, but I'll go with wherever we land.

src/databricks/labs/lakebridge/install.py
```diff
@@ -363,15 +363,22 @@ def _prompt_for_source_connection_config(self, dialect: str) -> SourceConnection
             catalog = self._prompts.question("Enter Oracle service name")
         elif dialect == ReconSourceType.DATABRICKS.value:
             catalog = self._prompts.question("Enter source Databricks catalog name", default="hive_metastore")
+        elif dialect == ReconSourceType.BIGQUERY.value:
```
m-abulazm, line 366: FIXME

tests/integration/reconcile/conftest.py
```diff
@@ -456,6 +460,55 @@ def teradata_recon_config(recon_cluster: str, recon_schema: SchemaInfo, make_vol
...
+
+    deployment_overrides = ReconcileJobConfig(
+        existing_cluster_id=recon_cluster,
+        tags={"lakebridge": "reconcile_test"},
+    )
+    logger.info(f"Using recon job overrides: {deployment_overrides}")
+
+    assert recon_schema.catalog_name
+    assert recon_schema.name
+    # "schema" report: the BigQuery source and the generic DIAMONDS target don't share columns, so
```
m-abulazm: FIXME

docs/lakebridge/docs/reconcile/configuration.mdx
```diff
@@ -105,6 +105,33 @@ Place the file in `.lakebridge/` in your Databricks workspace home folder.
...
+
+        :::note BigQuery type mapping
+        A few BigQuery types have no exact Databricks equivalent and are reconciled against their standard migration
+        target: `BIGNUMERIC`→`STRING`, `TIME`→`STRING`, `JSON`→`VARIANT`, `RANGE<T>`→`STRUCT<start, end>`,
+        `DATETIME`→`TIMESTAMP_NTZ`. If your migration used a different target, the `schema` report
+        flags those columns for review. `INTERVAL` migrates to two Databricks columns and is not
+        supported by `schema` reconciliation.
+        :::
+
+        :::warning Writable dataset required
```
m-abulazm, line 121: FIXME

tests/unit/reconcile/connectors/test_bigquery_schema_types.py
```diff
@@ -0,0 +1,82 @@
...
+    ("STRING", "STRING", "STRING", False),
+    ("BYTES", "BYTES", "BINARY", False),
+    ("DATE", "DATE", "DATE", False),
+    ("DATETIME", "DATETIME", "TIMESTAMP_NTZ", False),
+    ("TIMESTAMP", "TIMESTAMP", "TIMESTAMP", False),
+    ("NUMERIC(10, 2)", "NUMERIC(10, 2)", "DECIMAL(10, 2)", False),
+    ("GEOGRAPHY", "GEOGRAPHY", "STRING", False),
+    ("ARRAY<INT64>", "ARRAY<INT64>", "ARRAY<BIGINT>", False),
+    ("STRUCT<a INT64>", "STRUCT<a INT64>", "STRUCT<a: BIGINT>", False),
+    # NO-NATIVE / sqlglot-mishandled types — canonicalized in Stage 1
```
m-abulazm: FIXME

tests/unit/helpers/test_recon_config_utils.py
```diff
@@ -7,7 +7,9 @@
 from databricks.sdk.errors.platform import ResourceDoesNotExist
 from databricks.sdk.service.workspace import SecretScope
 
-SOURCE_DICT = {"databricks": "0", "mssql": "1", "oracle": "2", "redshift": "3", "snowflake": "4", "synapse": "5"}
+# Indices follow the alphabetically-sorted ReconSourceType values (Prompts.choice sorts):
+# bigquery, databricks, mssql, oracle, redshift, snowflake, synapse, teradata
+SOURCE_DICT = {"databricks": "1", "mssql": "2", "oracle": "3", "redshift": "4", "snowflake": "5", "synapse": "6"}
```
m-abulazm: "bigquery": 0

tests/unit/reconcile/connectors/test_bigquery.py
```diff
@@ -0,0 +1,151 @@
...
+    assert "when data_type = 'TIME' then 'string'" in schema_query
+    assert "when data_type = 'JSON' then 'variant'" in schema_query
+    assert "when data_type = 'RANGE<DATE>' then 'struct<start date, end date>'" in schema_query
+
+
+def test_list_schemas_and_tables():
+    engine, reader = initial_setup()
+    dfds = BigQueryDataSource(engine, reader)
+
+    # SCHEMATA is project-level and unqualified (the connection's default project scopes it).
```
m-abulazm: FIXME

m-abulazm (commented):
I added some comments to review closely in the IDE

take60:
Thanks @m-abulazm — addressed all the FIXMEs (CI green: `lint-uv` fixed, acceptance 173/173):

1. **`install.py`** — now prompts "Enter BigQuery project ID" and sets `catalog` = the project (dropped the old `catalog = schema` fallback).
2. **`test_recon_config_utils.py`** — added the missing `"bigquery": "0"` to `SOURCE_DICT`.
3. **`conftest.py`** — removed the stale "schema report" comment (config is `report_type="all"`).
4. **`configuration.mdx`** — reworded: `lakebridge_reconcile` is required (create it, writable, same region as the source), not a "default."
5. **`bigquery.py`** — made `get_schema`/`list_tables`/`list_schemas` reference `INFORMATION_SCHEMA` three-part with `catalog` (project), consistent with `read_data`; updated the read tests to pass the real project.
6. **`test_bigquery_schema_types.py`** — I read this one as "drop the comment," so I removed it and kept the `_SCHEMA_QUERY` type CASE + guardrail test (same per-connector pattern used for types sqlglot can not bridge). If you meant something deeper there, happy to follow up.

PTAL.

m-abulazm (approved):

bishwajit-db (approved):
LGTM

Issue #2483, opened 2026-05-26
MSSQL profiler: a single profiler run can only profile one user database


dgomez04:
## Background

Once #2482 lands (or in any future MSSQL profiler arrangement that uses `type: sql` / `type: ddl` instead of imperative `type: python`), all 13 extracts in `mssql/pipeline_config.yml` run through a single `MSSQLConnector` pinned to the `database` field in `.credentials.yml`. 7 of those 13 extracts read from per-database catalogs and therefore only return rows for one database per profiler run:

| Extract | Source | Scope |
|---|---|---|
| `tables` | `INFORMATION_SCHEMA.TABLES` | per-database |
| `views` | `INFORMATION_SCHEMA.VIEWS` | per-database |
| `columns` | `INFORMATION_SCHEMA.COLUMNS` | per-database |
| `routines` | `INFORMATION_SCHEMA.ROUTINES` | per-database |
| `indexed_views` | `sys.views` JOIN `sys.indexes` | per-database |
| `db_sizes` | `sys.database_files` | per-database |
| `table_sizes` | `sys.dm_db_partition_stats` | per-database |

The other 6 extracts (`sys_info`, `databases`, `query_stats`, `proc_stats`, `sessions`, `cpu_utilization`) read from server-scoped DMVs and return the same data regardless of which DB the connection lives in.

This is not a regression *introduced* by #2482 — the deleted `info_extract.py` / `activity_extract.py` hardcoded `db_name="master"`, so the per-DB queries returned `master`'s system catalog (i.e. nothing useful). #2482 makes those queries actually work for the configured database, which exposes the underlying single-DB-per-run limitation.

## What the user sees

For a server with N user databases, getting comprehensive schema + sizing data today requires:

1. Re-run `configure-database-profiler` to change the `database` field
2. Re-run `execute-database-profiler --source-tech mssql`
3. Manually move `profiler_extract.db` out of the way between runs (each run `TRUNCATE`s the DuckDB tables because every step has `mode: overwrite`)
4. Stitch the per-DB DuckDB files together by hand

`tables` / `views` / `columns` rows happen to carry `TABLE_CATALOG` (the DB name), so they can be disambiguated post-hoc — but `indexed_views`, `db_sizes`, `table_sizes`, and `routines` extracts have no DB identifier projected, so even manual stitching loses provenance.

## Precedent in this repo

[...]

PR #2536, opened 2026-06-26
Add multi-database profiling support for SQL Server

m-abulazm:
### What does this PR do?

The mssql profiler previously assessed only a single database per run. This PR adds automatic detection between single-database and multi-database profiling based on the SQL Server edition and the configured database, and introduces a reusable variant-resolution mechanism to support it.

**mssql auto-detection**
- A configured database scopes profiling to that one database (`single_db`).
- A blank database lets the edition decide: Azure SQL Database (`EngineEdition = 5`) falls back to `single_db`; on-prem SQL Server and Azure SQL Managed Instance profile every accessible database (`multi_db`).
- Credential configuration now accepts a blank database name (blank = all databases) for mssql.

**Multi-database extracts** (`mssql/multi_db/`, new)
- New pipeline config and SQL extracts that fan out across every online, accessible user database (system DBs excluded), tagging each row with a `database_name` column.
- UNION-based dynamic SQL for catalog views (tables, views, columns, routines, indexed views); cursor + `USE` for database-scoped DMVs (`db_sizes`, `table_sizes`) that can't use three-part naming.
- All extracts and DDL gain a `DATABASE_NAME` column. The original config moves to `mssql/single_db/`.

### Linked issues

Progresses #2483 

### Functionality

- [ ] added new CLI command
- [X] modified existing command: `databricks labs lakebridge execute-database-profiler --source-tech mssql`
- [ ] ... +add your own

### Tests

- [X] manually tested: Azure SQL and On-prem SQL Server on Docker
- [X] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -130,7 +130,13 @@ def _configure_credentials(self) -> None:
                 "login_timeout": self.prompts.question("Enter login timeout (seconds)", default="30"),
                 "server": self.prompts.question("Enter the fully-qualified server name"),
                 "port": int(self.prompts.question("Enter the port details", valid_number=True)),
-                "database": self.prompts.question("Enter the database name"),
+                # mssql: blank profiles every accessible database (on-prem / Managed Instance); a name scopes
+                # to that one database. legacy_synapse (shares this configurator) needs the dedicated-pool name.
+                "database": (
```
dgomez04, line 137: This skips a common middle case: customers want to scope the migration to a _subset_ of databases on a shared instance. A comma-separated allowlist here would be a natural fit.

We can either address the changes in this pull request or open an issue and follow-up.

src/databricks/labs/lakebridge/assessments/variants.py
dgomez04, line 1: Stepping back on the architecture before we land anything. **I don't think `single_db` vs `multi_db` is the right axis.**

Comparing the two pipeline configurations, they are identical except for 7 of the ~13 extracts. And those that do only differ in database scope. 

What `EngineEdition` actually tells us is the **engine variant** - each variant differs in system-table surface as we had discussed. That's the thing that should select a query pool / pipeline configuration. 

The curent `== 5 ? single_db : multi_db` check both ignores this axis and mis-fires on it. 

Synapse (6), serverless (9), Fabric (12), and SQL Edge (11) all fall into the else and would run the on-prem-style sys.databases + USE + three-part-INFORMATION_SCHEMA SQL, which isn't valid there.

The shape I'd propose: 
- ` **EngineEdition` maps to engine variant, which selects a query pool.** Pick the dialect-appropriate set of extracts, with a safe default for unknown editions instead of assuming on-prem multi database.
- **Database scope is a user input** (`all` or comma-separated list), independent of the enigne. 
- **Engine capability** decides how a multi-database scope executes: cross-database dynamic SQL on on-prem/MI; per-connection looping (or "scope the connected DB only") on Azure SQL Database, which can't do cross-DB from one connection.

**Note.** Scoping a single query pool by `all`/subset needs a way to inject `WHERE [name] IN (...)` into the SQL, which the `type: sql` pipeline can't do today, so this likely comes with adding parameter substitution to the pipeline.

src/databricks/labs/lakebridge/assessments/variants.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    resolvers = resolvers if resolvers else VARIANT_RESOLVERS
+    spec = SOURCE_SYSTEM_VARIANTS.get(source_system)
+    if spec is None:
+        if variant:
+            logger.warning(f"Ignoring variant '{variant}': source system '{source_system}' has no variants.")
+        return None
+
+    if AUTO in spec:
+        if variant != AUTO:
```
dgomez04: Every normal `mssql` run reaches here with `variant=None`, so `None != AUTO` is true and we emit the warning on each run. The user never passed a variant, so "ignoring" one is confusing. 

Suggest guarding the warning: `if variant and variant != AUTO:` so it only fires when an explicit choice is discarded.

src/databricks/labs/lakebridge/assessments/variants.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    cred_manager = create_credential_manager("mssql", EnvGetter(), creds_path=cred_file_path)
+    connect_config = cred_manager.get_credentials("mssql")
+    if connect_config.get("database"):
+        # The user picked a specific database -> profile only that one, regardless of edition.
+        return "single_db"
+    with DatabaseManager("mssql", connect_config) as db_manager:
+        # SERVERPROPERTY returns sql_variant, which pyodbc cannot fetch (ODBC type -16); CAST to int.
+        result = db_manager.fetch("SELECT CAST(SERVERPROPERTY('EngineEdition') AS INT) AS engine_edition")
+    engine_edition = int(result.rows[0][0])
```
dgomez04, line 43: **Optional Improvement:**  `int(result.rows[0][0])` will `IndexError` if the probe ever returns no rows. Low risk for `SERVERPROPERTY`, but a guarded failure ("could not determine SQL Server edition; defaulting to ...") reads better than a raw `IndexError`.

src/databricks/labs/lakebridge/assessments/variants.py
```diff
@@ -0,0 +1,88 @@
...
+def resolve_mssql_variant(cred_file_path: Path | None) -> str:
+    """Pick the SQL Server profiler variant from the configured database and the server edition.
+
+    A configured database scopes profiling to just that database (``single_db``) on any edition. When the
+    database is left blank the edition decides: Azure SQL Database falls back to the connected database
+    (``single_db``); on-prem SQL Server and Azure SQL Managed Instance profile every database (``multi_db``).
+    """
+    cred_manager = create_credential_manager("mssql", EnvGetter(), creds_path=cred_file_path)
+    connect_config = cred_manager.get_credentials("mssql")
+    if connect_config.get("database"):
```
dgomez04: Blank database on Azure SQL Database (`EngineEdition == 5`) returns `single_db`, after which `MSSQLConnector` connects with `database=None` (the login's default, often master). Is profiling the default/master database the intended result, or should blank+Azure-SQL-DB require a database name?

dgomez04 (changes requested):
Let's align on the engine-variant vs. database-scope model (See `variants.py`).

src/databricks/labs/lakebridge/assessments/variants.py
m-abulazm: that is a misconfiguration by the user

src/databricks/labs/lakebridge/assessments/variants.py
m-abulazm: Synapse, serverless, Fabric are not covered by our mssql profiler

dey-abhishek:
Follow-up on the existing Azure SQL DB + blank/all-databases thread after the latest changes:

I verified the current path still effectively routes `database=\"*\"` to `master`:

- `configure_assessment.py`: MSSQL database prompt defaults to `*`
- `assessments/variants.py`: `*` triggers edition probe; Azure SQL DB (`EngineEdition=5`) resolves to `single_db`
- `connections/database_manager.py`: `database=\"*\"` is coerced to `master`
- `assessments/profiler.py`: extraction uses the same credential config

So for Azure SQL DB, accepting defaults can still profile `master` (or fail for contained users without `master` access), instead of the intended user database.

Suggested fix: for Azure SQL DB + `single_db`, require a concrete database name (or rewrite to a concrete DB before extraction), and avoid relying on the `* -> master` fallback in that path.

dey-abhishek:
The new multi-db metadata extracts build dynamic `UNION ALL` across multiple databases:

- `mssql/multi_db/tables.sql`
- `mssql/multi_db/views.sql`
- `mssql/multi_db/columns.sql`
- `mssql/multi_db/routines.sql`
- `mssql/multi_db/indexed_views.sql`

On instances with mixed database collations, unioning text columns directly can fail with collation conflicts (for example, schema/table/object name fields from different DB collations).

Suggested fix: normalize all projected string expressions in every UNION branch to one chosen collation consistently (e.g., `CAST(... AS NVARCHAR(...)) COLLATE <chosen_collation>` for each text column in each branch).

dey-abhishek:
This PR adds leading `DATABASE_NAME` to existing MSSQL extract tables (`tables`, `views`, `columns`, `routines`, `indexed_views`, `table_sizes`).

Pipeline behavior today appears to be:

- DDL step skips when table already exists
- overwrite path does `TRUNCATE <table>` then `INSERT INTO <table> SELECT * FROM _result_frame`

So rerunning profiler into an existing same-day/same-version DuckDB file created before this schema change can fail due to column count/order mismatch (old table shape vs new result shape).

Suggested fix: add schema migration/recreate logic when expected DDL has changed, or use explicit column-list inserts after schema migration.

dey-abhishek (changes requested):
Added PR review comments above @m-abulazm

dgomez04 (approved):
LGTM - adding a comment to the original issue for comma-separated lists.

m-abulazm:
> The new multi-db metadata extracts build dynamic `UNION ALL` across multiple databases:
> 
> * `mssql/multi_db/tables.sql`
> * `mssql/multi_db/views.sql`
> * `mssql/multi_db/columns.sql`
> * `mssql/multi_db/routines.sql`
> * `mssql/multi_db/indexed_views.sql`
> 
> On instances with mixed database collations, unioning text columns directly can fail with collation conflicts (for example, schema/table/object name fields from different DB collations).
> 
> Suggested fix: normalize all projected string expressions in every UNION branch to one chosen collation consistently (e.g., `CAST(... AS NVARCHAR(...)) COLLATE <chosen_collation>` for each text column in each branch).

lets solve this in a followup. it does not fit on this PR

m-abulazm:
> So rerunning profiler into an existing same-day/same-version DuckDB file created before this schema change can fail due to column count/order mismatch (old table shape vs new result shape).

this is not supported by our profilers and not in scope of this PR to handle

dey-abhishek (approved):
LGTM

PR #2539, opened 2026-06-29
Add bigquery e2e recon test

m-abulazm:
### What does this PR do?
* use dataset as materialization dataset
### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [X] added e2e tests

### DIFF
--- a/src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
+++ b/src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
@@ -49,9 +49,8 @@ class BigQueryDataSource(DataSource):
         self._engine = engine
         self._reader = reader
 
-    def _read(self, query: str, catalog: str, options: JdbcReaderOptions | None = None) -> DataFrame:
-        logger.info(f"Using catalog '{catalog}' as the BigQuery materializationDataset")
-        return self._reader.read_data(query, catalog, "materializationDataset", "query", options)
+    def _read(self, query: str, materialization_dataset: str) -> DataFrame:
+        return self._reader.read_data_direct(query, "query", {"materializationDataset": materialization_dataset})
 
     def read_data(
         self,
@@ -61,11 +60,11 @@ class BigQueryDataSource(DataSource):
         query: str,
         options: JdbcReaderOptions | None,
     ) -> DataFrame:
-        table_ref = f"`{schema}`.`{table}`"
+        table_ref = f"`{catalog}.{schema}.{table}`"
         table_query = query.replace(":tbl", table_ref).replace("@tbl", table_ref)
         try:
             logger.info(f"Fetching data using query: \n`{table_query}`")
-            df = self._read(table_query, catalog, options)
+            df = self._read(table_query, schema)
             return df.select([col(column).alias(column.lower()) for column in df.columns])
         except (RuntimeError, PySparkException) as e:
             return self.log_and_throw_exception(e, "data", table_query)
@@ -85,7 +84,7 @@ class BigQueryDataSource(DataSource):
         try:
             logger.debug(f"Fetching schema using query: \n`{schema_query}`")
             logger.info(f"Fetching Schema: Started at: {datetime.now()}")
-            df = self._read(schema_query, catalog)
+            df = self._read(schema_query, schema)
             schema_metadata = df.select([col(c).alias(c.lower()) for c in df.columns]).collect()
             logger.info(f"Schema fetched successfully. Completed at: {datetime.now()}")
             return [self._map_meta_column(field, normalize) for field in schema_metadata]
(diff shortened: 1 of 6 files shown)


src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -95,15 +94,15 @@ def get_schema(
     def list_schemas(self, catalog: str) -> list[str]:
         query = BigQueryDataSource._LIST_SCHEMAS_QUERY
         try:
-            df = self._read(query, catalog)
+            df = self._read(query, catalog)  # User has to create a dataset with the value
```
take60: Small flag on `list_schemas`: it uses `catalog` (the project) as the `materializationDataset`, but that needs to be a writable *dataset*, not a project — so in production (where `catalog` is the project) it would point at a dataset that does not exist. The test passes since it calls `list_schemas("lakebridge")` with a real dataset.

Likely fine as a follow-up (an explicit `materialization_dataset` field would cover it) — just flagging so discovery does not trip up later.

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -49,9 +49,8 @@ def __init__(self, engine: Dialect, reader: RemoteQueryReader):
         self._engine = engine
         self._reader = reader
 
-    def _read(self, query: str, catalog: str, options: JdbcReaderOptions | None = None) -> DataFrame:
-        logger.info(f"Using catalog '{catalog}' as the BigQuery materializationDataset")
-        return self._reader.read_data(query, catalog, "materializationDataset", "query", options)
+    def _read(self, query: str, materialization_dataset: str) -> DataFrame:
+        return self._reader.read_data(query, materialization_dataset, "materializationDataset")
```
take60:
```suggestion
    _MATERIALIZATION_DATASET = "lakebridge_reconcile"

    def _read(self, query: str) -> DataFrame:
        return self._reader.read_data(query, self._MATERIALIZATION_DATASET, "materializationDataset")
```

Hardcode a dedicated materialization dataset so every read materializes into `lakebridge_reconcile` (not the source dataset / project). This fixes `list_schemas` and lets the source dataset stay read-only.

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -61,11 +60,11 @@ def read_data(
         query: str,
         options: JdbcReaderOptions | None,
     ) -> DataFrame:
-        table_ref = f"`{schema}`.`{table}`"
+        table_ref = f"`{catalog}.{schema}.{table}`"
         table_query = query.replace(":tbl", table_ref).replace("@tbl", table_ref)
         try:
             logger.info(f"Fetching data using query: \n`{table_query}`")
-            df = self._read(table_query, catalog, options)
+            df = self._read(table_query, schema)
```
take60:
```suggestion
            df = self._read(table_query)
```

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -85,7 +84,7 @@ def get_schema(
         try:
             logger.debug(f"Fetching schema using query: \n`{schema_query}`")
             logger.info(f"Fetching Schema: Started at: {datetime.now()}")
-            df = self._read(schema_query, catalog)
+            df = self._read(schema_query, schema)
```
take60:
```suggestion
            df = self._read(schema_query)
```

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
take60:
```suggestion
            df = self._read(query)
```

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -95,15 +94,15 @@ def get_schema(
...
+            df = self._read(query, catalog)  # User has to create a dataset with the value
             return [row.schema_name for row in df.select(col("schema_name").alias("schema_name")).collect()]
         except (RuntimeError, PySparkException) as e:
             return self.log_and_throw_exception(e, "schemas", query)
 
     def list_tables(self, catalog: str, schema: str) -> list[str]:
         query = BigQueryDataSource._LIST_TABLES_QUERY.format(schema=schema)
         try:
-            df = self._read(query, catalog)
+            df = self._read(query, schema)
```
take60:
```suggestion
            df = self._read(query)
```

take60 (commented):
Suggesting we go with the hardcoded `lakebridge_reconcile` materialization dataset you mentioned. It makes all reads materialize into a dedicated dataset (fixes the `list_schemas` project-as-dataset case and lets the source dataset stay read-only), while `catalog` stays = project.

One prerequisite to document: `lakebridge_reconcile` must be created in the **same region** as the source data and be writable by the connection's service account. (The class docstring on line 23 also needs a small update since `catalog` is no longer the materializationDataset.)

PR #2543, opened 2026-06-30
Sample Snowflake query history to bound extract size (#2532)

### LINKED ISSUE #2532: [ISSUE]: Snowflake profiler extract is too large on high-volume accounts
## Context

With anonymized numbers from one (of several) Snowflake instances:

- `query_history` over the default **90-day** window: **~28.07M rows**
- Daily volume is **steady, not spiky**. ~312k queries/day average, peaking at ~427k/day.
- `QUERY_TEXT` alone  averages `~500 chars`, **max `~1.0M chars`**, **~13 GB** of text.
- Resulting extract size is around **~20 GB** for a single instance.

This customer has multiple instances, each expected to produce similarly sized files. 

## Root Cause

Three compounding factors:

1. **Full `QUERY_TEXT` over a hardcoded 90-day window, no row cap.** `src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history.sql` selects `QUERY_TEXT` with `WHERE START_TIME >= DATEADD('day', -90, CURRENT_TIMESTAMP())` and `ORDER BY START_TIME DESC`. `QUERY_TEXT` dominates the output size.

2. **Whole result materialized in memory, then written once.** `assessments/pipeline.py::_execute_sql_step` calls `executor.fetch(query)` then `_save_to_db(...)`. `connections/database_manager.py::_BaseConnector.fetch` does `result.fetchall()` (entire result set in memory), and `_save_to_db` then calls `result.to_df()` (a second full copy as a pandas DataFrame) before a single DuckDB write. Nothing is written incrementally. For 28M rows this also makes the run appear hung for a long time and risks out-of-memory issues on smaller machines.

3. **No compression / export / column-pruning options.** The `.db` is shared as-is; there's no built-in way to produce a smaller, transport-friendly artifact.

## Related Issues

- #1469, #2486, #1417 

## Environment
- Lakebridge version: `0.14.0`

jneil17:
## Summary

`query_history` produced a multi-GB extract driven by `QUERY_TEXT` (avg ~500 chars, up to ~1M). This pull request replaces it with a random sample that always leads to a flat, predictable size. 

## Changes

- **`query_history`** — now a **random sample** (`ORDER BY RANDOM() LIMIT 10000`, 90-day window) with `QUERY_TEXT`, for the downstream skill to classify the workload mix.
- **`query_history_top`** *(new)* — top 20 queries by credits (with `QUERY_TEXT`) so outliers the sample misses are still captured.
- **`query_history_stats`** *(new)* — full-window aggregates (no query text) by `QUERY_TYPE × WAREHOUSE_SIZE`: counts + total/avg/median/p90 credits & elapsed. Carries the true totals the sampled mix is applied against.
- **`user_activity`** — per-user/client summary of `LOGIN_HISTORY` instead of a per-event dump (secondary bloat noted in the issue).

## Notes

- Window (90d) and sample size (10k) are currently hardcoded in SQL; making them CLI-configurable is a possible follow-up.
- `query_history_stats` credits are attributed/estimated and won't perfectly reconcile with `warehouse_usage` (`WAREHOUSE_METERING_HISTORY`), which remains the billing source of truth.

Closes #2532


src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history.sql
```diff
@@ -1,45 +1,62 @@
--- Query History and Performance Analysis
--- Extract query execution patterns, performance metrics, and user activity
+-- Query History — bounded RANDOM sample for workload classification
+--
+-- Issue #2532: the previous step pulled the full 90-day QUERY_HISTORY including
```
dgomez04, line 3: We don't need to mention the issue within the comment, we should make it flatter and straight to the point.

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -22,23 +22,37 @@ steps:
     frequency: daily
     flag: active
 
+  # Bounded RANDOM sample of query history (incl. QUERY_TEXT) for workload
```
dgomez04, line 25: Remove this comment.

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -22,23 +22,37 @@ steps:
...
+  # multi-GB extracts on high-volume accounts (issue #2532).
   - name: query_history
     type: sql
     extract_source: src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history.sql
     mode: overwrite
     frequency: daily
     flag: active
 
-  - name: storage_usage
+  # Top-N most expensive queries (with QUERY_TEXT) so the random sample never
```
dgomez04, line 35: Remove this comment.

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
```diff
@@ -22,23 +22,37 @@ steps:
...
+  - name: query_history_top
     type: sql
-    extract_source: src/databricks/labs/lakebridge/resources/assessments/snowflake/storage_usage.sql
+    extract_source: src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history_top.sql
     mode: overwrite
     frequency: daily
     flag: active
 
-  - name: user_activity
+  # Full-window aggregate (no query text): true totals + per-group distribution
```
dgomez04, line 44: Remove this comment.

src/databricks/labs/lakebridge/resources/assessments/snowflake/user_activity.sql
dgomez04, line 1: The pull request description mentions that `user_activity` still exists but here it's completely removed. Can we confirm whether this is the intended behavior?

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipeline_config.yml
dgomez04, line 1: Each of the new pipeline steps scans the full 90-day `QUERY_HISTORY` window independently. Export size is fixed, but the total execution time can grow by ~3x compared to the old, single-step run. 

I believe this is something we should look into further.

src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history.sql
```diff
@@ -1,45 +1,62 @@
...
+            WHEN '6X-LARGE' THEN 512
+            ELSE 0
+        END
+    ) AS ESTIMATED_CREDITS,
+    (a.CREDITS_ATTRIBUTED_COMPUTE IS NULL) AS IS_CREDITS_ESTIMATED,
+    CURRENT_TIMESTAMP() AS EXTRACT_TIMESTAMP
+FROM SNOWFLAKE.ACCOUNT_USAGE.QUERY_HISTORY qh
+LEFT JOIN attributed a ON qh.QUERY_ID = a.QUERY_ID
+WHERE qh.START_TIME >= DATEADD('day', -90, CURRENT_TIMESTAMP())
+ORDER BY RANDOM()
```
dgomez04, line 61: `ORDER BY RANDOM() LIMIT 10000` may still require Snowflake to touch every row before applying the limit. 

Has execution time been benchmarked? 

Did we consider other alternatives such as `SAMPLE (10000 ROWS)` or hash-based sampling like `(MOD(ABS(HASH(QUERY_ID)), …))`? They could be much cheaper while still giving a representative mix. 

Additionally, is there any reason for us to get 90 days of history rather than 30?

dgomez04 (changes requested):
The core issue is fixed but grows execution time significantly. Let's look into this deeper before approving.

dgomez04:
Closing off as I'll be reworking based on some new insights.

PR #2548, opened 2026-07-02
Improve handling of non-native Bigquery types

m-abulazm:
### What does this PR do?
* `_SCHEMA_QUERY` maps only the BigQuery types sqlglot can't bridge on its own, and only within the same type family:  `NUMERIC` → `decimal(38,9)`, `JSON` → `variant`, `GEOGRAPHY` → `geography`. Everything else passes through to sqlglot. 
* Types with no same-family Databricks equivalent (`BIGNUMERIC`,`TIME`, `RANGE<T>`, or any of these nested in `ARRAY`/`STRUCT`) are left as their original BigQuery type and user gets a warning log naming those columns so the user knows to verify them manually.

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [X] added unit tests
- [X] added integration tests


src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -17,28 +17,32 @@
...
+
     _LIST_SCHEMAS_QUERY = "select schema_name from `{catalog}`.INFORMATION_SCHEMA.SCHEMATA order by schema_name"
     _LIST_TABLES_QUERY = "select table_name from `{catalog}.{schema}`.INFORMATION_SCHEMA.TABLES order by table_name"
     _SCHEMA_QUERY = """select column_name,
                                   case
-                                        when data_type like 'BIGNUMERIC%' then 'string'
-                                        when data_type = 'NUMERIC' then 'decimal(38,9)'
-                                        when data_type = 'TIME' then 'string'
+                                        when data_type = 'NUMERIC' then 'decimal(38, 9)'
+                                        when data_type like 'BIGNUMERIC%' then 'decimal(38, 9)'
```
take60: Question on `BIGNUMERIC` → `decimal(38, 9)`: BIGNUMERIC has precision up to 76 digits, well beyond Databricks' DECIMAL max of 38, so this mapping is lossy/truncating for any value that doesn't fit. 

If a customer migrated BIGNUMERIC→STRING to keep precision , this change would make schema reconcile report a false mismatch on every such column, which seems like the opposite of what Stage-1 canonicalization is for.

Could we either (a) keep the same-family-only philosophy but treat BIGNUMERIC like TIME/RANGE (pass through + warn) instead of asserting decimal(38,9), or (b) make the truncating target configurable/documented so users who chose STRING aren't penalized?

src/databricks/labs/lakebridge/install.py
```diff
@@ -364,15 +364,15 @@ def _prompt_for_source_connection_config(self, dialect: str) -> SourceConnection
         elif dialect == ReconSourceType.DATABRICKS.value:
             catalog = self._prompts.question("Enter source Databricks catalog name", default="hive_metastore")
         elif dialect == ReconSourceType.BIGQUERY.value:
-            catalog = self._prompts.question("Enter BigQuery project ID")
+            catalog = self._prompts.question("Enter BigQuery project name")
```
take60: `catalog` is substituted raw into identifiers like ```{catalog}.{schema}```.INFORMATION_SCHEMA.COLUMNS. GCP's **project ID** (e.g. `my-project-123456`) is the immutable slug valid there; **project name** is just a mutable display label that can contain spaces and isn't a valid identifier — using it would break every query. Could we revert to "project ID"?

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -17,28 +17,32 @@
...
+    * ``JSON`` -> ``variant``.
+
+    Columns with no same-family equivalent (``TIME``, ``RANGE<T>``, ``INTERVAL`` or these nested in
+    ``ARRAY``/``STRUCT``) can be reported as schema mismatches even when the migration is correct — an
+    accepted false negative, logged in ``get_schema`` so the user knows to verify them manually.
     """
 
     _IDENTIFIER_DELIMITER = "`"
 
+    _APPROXIMATE_TYPES = ("time", "bignumeric", "range", "interval")
```
take60: `_SCHEMA_QUERY`'s CASE matches exact strings (`data_type = 'JSON'`), so `STRUCT<a JSON>` falls through unconverted — same issue as nested TIME/RANGE. But `_APPROXIMATE_TYPES` doesn't include `json`/`numeric`, so `_warn_on_approximate_types` never flags it. This PR's own test proves it: `col_struct_json` is one of the 6 `is_valid=false` columns in `test_bigquery_schema_compare`, with no warning logged. Could `json`/`numeric` be added to `_APPROXIMATE_TYPES`?

tests/unit/reconcile/connectors/test_bigquery_schema_types.py
```diff
@@ -1,81 +0,0 @@
-"""Guardrail for BigQuery schema-reconcile type handling.
```
take60, line 1: `test_bigquery_schema_types.py` pinned down which specific type validated against which target — useful if sqlglot ever silently breaks one mapping. The new test only checks aggregate counts (21/15/6), so two mappings breaking/fixing at once could cancel out and still pass. Worth keeping a lighter per-type check alongside it?

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
m-abulazm: good call

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
m-abulazm: added json, do we still need numeric?

m-abulazm: numeric will be mapped to decimal reliably

tests/unit/reconcile/connectors/test_bigquery_schema_types.py
m-abulazm: I added tests in schema compare like all the others, no suprises for developers. and I added `test_sqlglot_has_no_native_databricks_equivalent_for_approximate_types` to pin sqlglot behavior in case it changes

PR #2549, opened 2026-07-02
Add `WORKSPACE_NAME` to Synapse activity extract tables

dgomez04:
## Feedback

Concurrency numbers are currently unreliable when the same pool name exists across multiple workspaces.

Because the Synapse session/request extract tables only track `POOL_NAME` + `EXTRACT_TS`, data from different workspaces collides once it lands in the extract, making it impossible to differentiate them.

## Fix

Added a `WORKSPACE_NAME` column to uniquely identify extracts. This value is pulled from the workspace config's name property, which is already available in both extracts, meaning no extra API calls were required.

Downstream consumers can now accurately group metrics by `(WORKSPACE_NAME, POOL_NAME)` when required.


m-abulazm (approved):

sundarshankar89 (approved):
LGTM

PR #2553, opened 2026-07-03
Add cross-profiler timestamp fidelity regression tests

nnigam19:
## What

Adds unit regression tests verifying that every profiler preserves timestamp
**wall-clock fidelity** when writing its extract into DuckDB. No production code
changes.

Both shared ingestion paths are covered:

- **SQL profilers** (Oracle, Redshift, MSSQL, Snowflake, Teradata) -
  `DatabaseManager.fetch` → `FetchResult.to_df()` → `PipelineClass._save_to_db`,
  driven through the public `PipelineClass.execute()` with a fake executor so the
  real ingestion wiring is exercised end to end.
- **Python profilers** (Synapse, BigQuery) — `save_to_duckdb()`.

## What the tests assert

- A timezone-naive source timestamp stays a naive DuckDB `TIMESTAMP`, with its
  wall-clock reading intact.
- A tz-aware timestamp keeps its instant (stored as `TIMESTAMP WITH TIME ZONE`).
- Naive timestamps do **not** shift when the extract is read under a non-UTC
  session time zone (`America/Loles`, `Asia/Kolkata`, `Pacific/Kiritimati`).

25 parametrized cases across all seven profilers.

## Testing

- `pytest`: 25 passed

### DIFF
--- a/tests/unit/assessment/test_profiler_timestamp_fidelity.py
+++ b/tests/unit/assessment/test_profiler_timestamp_fidelity.py
@@ -1,0 +1,200 @@
+"""Timestamp wall-clock fidelity across *all* Lakebridge profiler ingestion paths.
+
+Every profiler writes its extracted source data into a local DuckDB database
+through one of two *shared* code paths:
+
+* **SQL profilers** (Oracle, Redshift, MSSQL, Snowflake, Teradata)::
+
+      DatabaseManager.fetch -> FetchResult.to_df() -> PipelineClass._save_to_db
+
+* **Python profilers** (Synapse, BigQuery)::
+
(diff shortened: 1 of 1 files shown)


m-abulazm:
Thanks for putting this together, timestamp bugs are nasty and I get the goal. But I don't think we should take these tests.

`save_to_duckdb` and `_save_to_db` don't touch types at all, they just pass the DataFrame through. So these tests aren't checking our code, they're checking that DuckDB and pandas behave the way they behave today.

The per-profiler parametrization also doesn't add coverage; all five profilers hit the same code path with the same fake data, so it's one test running five times.

Issue #2550, opened 2026-07-02
Consolidate the Teradata profiler into a single pipeline and remove `variant` selection.


dgomez04:
## Why

Unlike Redshift's three variants — which are mutually exclusive alternates for the *same* metric — Teradata's two variants aren't mutually exclusive at all:

| Step | `core` | `pdcr` | Disposition |
|---|---|---|---|
| `td_sys_info`, `td_sys_nodes_info`, `td_sys_usage_agg`, `td_sys_disk_utilization`, `td_db_object_types`, `td_user_databases`, `td_dwh_udf` | shared | shared | already universal (identical steps in both pipeline configs) |
| `td_dbql_core_info_extract` | `DBC.DBQLogTbl`, last 7 days | ❌ none | core-only, always available — DBQL is standard Teradata, not optional |
| `td_pdcr_info_agg_extract` | ❌ none | `PDCRINFO.DBQLogTbl_Hst`, last 180 days | pdcr-only, **requires the optional PDCR feature to be installed** |
| `td_pdcr_sp_exe_info_agg_extract` | ❌ none | `PDCRINFO.DBQLogTbl_Hst` (stored-procedure execution stats) | pdcr-only, no `core` equivalent exists |

`core` and `pdcr` don't disagree on how to compute the same metric — `pdcr` is strictly **additive**: longer history plus a stored-procedure metric `core` cannot produce at all. The only reason they're split today is that `PDCRINFO` doesn't exist on Teradata systems without PDCR installed, and a missing-table error today aborts the whole run (#2541).

This yields two groups:
- **Seven universal queries** (already shared verbatim by both pipelines)
- **Three deployment/feature-bound queries**: `td_dbql_core_info_extract` (always run), `td_pdcr_info_agg_extract` + `td_pdcr_sp_exe_info_agg_extract` (optional, PDCR-only)

### SQL Consolidation:
- [ ] Merge `core/` and `pdcr/` into a single Teradata query directory (the seven shared steps already live outside both subfolders and need no change).
- [ ] Always run `td_dbql_core_info_extract` — it has no availability precondition beyond standard DBQL logging.
- [ ] Mark `td_pdcr_info_agg_extract` and `td_pdcr_sp_exe_info_agg_extract` as `optional`/expected-missing (per #2541's contract): they fail with an expected "object does not exist" error on any system without PDCR installed, and that must not abort the run.
[...]

PR #2555, opened 2026-07-08
Add profiler pipeline partial-run contract for optional steps 

### LINKED ISSUE #2541: Decide how the profiler `pipeline` should handle partial / best-effort runs.
### Problem
`assessments/pipeline.py` treats any step failure as fatal to the whole run: `ddl` / `source_ddl` errors abort mid run, and any `sql` error re-raises at the end, even though "succesful steps" tables have already been written to the extract.

The deeper problem is that the engine has no notion of **why** a step failed, so it cannot respond differently to fundamentally different situations. In a profiling/assessment context, failures fall into two categories:

- 1. **Expected absence** - the source legitimately lacks a view/feature due to deployment edition, version, or permissions. The run should still deliver every other metric.
- 2. **Real failure** - we cannot reach the source (authentication, connectivity) or our own query is malformed. This must surface loudly. 

Today, both collapse into `abort` which has two costs: 
- 1. **It forces the variant workaround.** The only way to avoid category (1) today is to pre-segregate queries per deployment so a run never encounters a missing view. #2540 removes for Redshift, and a pattern that recurs across sources.
- 2. **There is no delierate definition of run success.** Success on today's structure means every step succeded, which is likely wrong for a diagnostic tool. 

We need to agree on the engine's failure-handling contract so that we can design the right fix. 

### Solutions
- **Best-effort completeness** (deliver partial results) **vs. trustworthy failure signal** (never let a broken run look successful is the primary risk).
-  Telling **benign absence** from **real error** likely requires understanding error semantics, not just did it throw.
-  **CI/test signal must stay meaningful**, "success" can't become unfalsifiable.
- **Cross-source generality:** the contract applies to every source.

dgomez04:
### Summary
Adds support for **optional pipeline steps** so the profiler can tolerate expected-missing metrics (deployment-specific views/tables that simply aren't present) instead of aborting the whole assessment, while keeping real database and configuration failures fatal.

Also collapses Teradata into a single unified pipeline (PDCR-only steps marked `optional: true`).

### Changes

**Partial-run contract**
- Added an opt-in `optional: true` flag on pipeline steps (`profiler_config.Step`), validated as a boolean.
- Added `ABSENT` to `StepExecutionStatus` (alongside `COMPLETE` / `ERROR` / `SKIPPED`).
- `_process_step` catches `(RuntimeError, ConnectionError)` and maps the outcome by step kind:
  - **optional** → `ABSENT`: the run is not failed; the driver cause is retained on `StepExecutionResult.error_message` and logged once at `WARNING`.
  - **required** → `ERROR`: fails the pipeline.
- `ddl` (local DuckDB) and `source_ddl` (source-side) steps remain strictly fatal on `ERROR`.
- `DatabaseManager.fetch` / `check_connection` normalize driver failures (`SQLAlchemyError`, `ConnectionError`, `redshift_connector` errors) into a single `ConnectionError` carrying the concise first line of the driver message; the full stack trace stays at `DEBUG`. This gives the pipeline one exception type to reason about instead of per-driver handling.

**Teradata unified pipeline (#2550)**
- Merged the `core` / `pdcr` variants into a single `teradata/pipeline_config.yml`; the two PDCR-only SQL steps are marked `optional: true` so a missing PDCR deployment yields `ABSENT` rather than breaking the run.
- Removed Teradata from `SOURCE_SYSTEM_VARIANTS`; legacy `--variant` inputs are ignored with a warning.

#### Unit
- [x] Pipeline policy — optional→`ABSENT`, required→`ERROR`, tolerate-any-failure, `source_ddl` tolerance, `ddl`-fatal (`tests/unit/assessment/test_pipeline_policy.py`)
- [x] `optional` flag validation and defaults (`tests/unit/assessment/test_profiler_config.py`)
- [x] `DatabaseManager` error normalization (`tests/unit/connections/test_database_manager.py`)
- [x] Teradata unified pipeline + variant elimination (`tests/unit/assessment/test_teradata_profiler.py`, `test_profiler_variants.py`, `test_variant_resolution.py`)

#### Integration
- [x] Optional-absence tolerance against a sandbox SQL Server (`tests/integration/assessments/test_pipeline.py`)

Closes #2541


src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -58,15 +59,13 @@ def execute(self) -> list[StepExecutionResult]:
             execution_results.append(result)
             self._log_step_result(result)
 
-            # Fail immediately if a DDL (local DuckDB) or source_ddl (source-side) step failed
-            if step.type in {"ddl", "source_ddl"} and result.status == StepExecutionStatus.ERROR:
+            if step.type == "ddl" and result.status == StepExecutionStatus.ERROR:
```
m-abulazm: should we change this? source_ddl can be marked optional so when required it will be detected

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -150,11 +148,7 @@ def _execute_source_ddl_step(self, step: Step):
             return
 
         logging.info(f"Executing source_ddl step '{step.name}' on source")
-        try:
-            self.executor.fetch(content)
-        except Exception as e:
-            logging.error(f"source_ddl step failed: {str(e)}")
-            raise RuntimeError(f"source_ddl step failed: {str(e)}") from e
+        self.executor.fetch(content)
```
m-abulazm, line 151: was the previous try statement redundant?

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -26,6 +26,7 @@ class StepExecutionStatus(str, Enum):
     COMPLETE = "COMPLETE"
     ERROR = "ERROR"
     SKIPPED = "SKIPPED"
+    ABSENT = "ABSENT"
```
m-abulazm, line 29: I would reuse skipped instead of adding a new class. skipped is also self explanatory when users read it

src/databricks/labs/lakebridge/assessments/pipeline.py
dgomez04: Made the change. 

Making optional `source_ddl` would also require for its dependent `sql` steps to be skipped, that needs an explicit dependency like `requires: [...]` - we can discuss and handle it on a separate pull request.

src/databricks/labs/lakebridge/assessments/pipeline.py
dgomez04: I’d prefer to keep `ABSENT` separate from `SKIPPED`. They’re distinct outcomes:

- `SKIPPED` = the step was never executed (`flag != "active"`).
- `ABSENT` = the step was executed, failed, and was tolerated because it’s optional — and we keep the driver `error_message`, logged at warning level.

Folding it into a single class would lose the signal we're trying to add.

src/databricks/labs/lakebridge/assessments/pipeline.py
dgomez04: `DatabaseManager.fetch` now normalizes driver failures to `ConnectionError`, and `_process_step` already catches `(RuntimeError, ConnectionError)` to map `optional` → `ABSENT` / `required` → `ERROR`. The local `try`/`except` only re-wrapped that as `RuntimeError` and logged; logging now lives in `_log_step_result`.

m-abulazm (approved):
LGTM

dey-abhishek:
## Review

**Gates:** ✅ 118 unit tests pass; branch is mergeable and up to date with `main`.


### The error taxonomy described in the summary isn't in the code

Grepping the tree: there is **no** `errors.py`, no `ErrorCategory` / `SourceFailure` / `SourceQueryError`, no `classify_standard_sqlstate`, and no `parse_source_error` anywhere. Connectors only implement `_connect/fetch/close/health_check` — none classify. `DatabaseManager.fetch()` just widens the caught type and normalizes everything to a builtin `ConnectionError` with the first line of the message, and `_process_step` is already pure tolerate-any-failure:

```python
except (RuntimeError, ConnectionError) as e:
    status = StepExecutionStatus.ABSENT if step.optional else StepExecutionStatus.ERROR
```

No SQLSTATE inspection, no `UNKNOWN` category, no success floor, and `execute()`'s return type is unchanged. So the branch **already implements the minimal design** — there's no classification layer to remove, and my earlier finding was measuring the code against the description's promise rather than the code itself.

I don't think the classifier guarantee is worth building anyway: it can't reliably separate "our bug" from "expected absence" (e.g. a renamed column and a genuinely-missing PDCR database can surface the same Teradata code), and any message the driver regexes fail to parse would fail the very steps `optional` exists to protect. The guardrail would leak in both directions.

### Asks, in priority order

1. **Make the description match the code.** The summary describes an `ErrorCategory` / `SourceFailure` / `parse_source_error` taxonomy that isn't in the diff. Please drop that language so reviewers aren't evaluating against a guarantee the code doesn't provide.
2. **Add a completion summary (the one real gap).** Today the tolerated error is retained on `StepExecutionResult.error_message` and logged at WARNING, but `profiler.py` only logs `repr(list[...])` at INFO and declares success — there's no structured, customer-facing summary of which optional steps degraded and why. That summary is what lets a reader judge "not deployed" vs "failed to read," and it's far cheaper than a classifier.
3. **Keep `ABSENT` distinct from `SKIPPED`** — `SKIPPED` already means `flag != "active"` (disabled on purpose); `ABSENT` means "optional, attempted, failed, tolerated." Worth keeping, not collapsing.

### Non-blocking

- The core/pdcr YAML consolidation (deleting `teradata/core/pipeline_config.yml`, merging into one config, dropping `teradata` from `SOURCE_SYSTEM_VARIANTS` with `--variant` now ignored+warned) is a larger behavior change riding along with the optional flag. It's clean and tested; flagging only in case you'd prefer to land it separately.
- If "not deployed vs failed to read" turns out to be a hard reporting requirement from #2541/#2550, the proportionate move is to whitelist the two Teradata codes (3802/3807) on optional steps only — not a general per-driver classifier.

_(The optional flag, its validation tests, and the variant cleanup all look good.)_

dey-abhishek (approved):
LGTM

PR #2556, opened 2026-07-08
Remove variant selection from the Teradata profiler

### LINKED ISSUE #2550: Consolidate the Teradata profiler into a single pipeline and remove `variant` selection.

dgomez04:
### Summary
Collapses the Teradata `core`/`pdcr` variant split into a single, unified pipeline configuration. This leverages the new optional-step contract to gracefully handle missing `PDCRINFO` deployments, ensuring that foundational metrics execute succesfully while missing optional metrics yield a non-fatal status.

### Changes
- Unified Teradata configuration into a single `teradata/pipeline_config.yml`. Marked the two PDCR-only SQL steps as `optional: true` so a missing deployment yields an `ABSENT` execution status instead of breaking the pipeline. Shared metrics and core extract steps remain required.
- Removed the `teradata` entry from `SOURCE_SYSTEM_VARIANTS` and updated `cli.py` to route Teradata through the generic, variant-less path. Legacy `--variant` inputs are now safely ignored with a warning. Deleted the redundant `core/` and `pdcr` configuration files.

#### Unit
- [x] Variant-less resolution, `core`/`pdcr`metric integration, and step optionality validation (`tests/unit/assessment/test_teradata_profiler.py`)
- [x] Validation of variant-bypassing and legacy flag warning behavior (`test/unit/assessment/test_cli_other.py`)
- [x] Regresiion testing for variant elimination (`tests/unit/assessment/test_profiler_variants.py`)

#### Integration
- [ ] Execution run against a live Teradata instance (requires access to a live environment; no local sandbox available)

> **Stacked on top of #2555. Retarget to `main` once it merges.**

Closes #2550.


dey-abhishek (commented):
@dgomez04 

### Review + live testing against a Teradata trial box

Reviewed the diff and ran the unified pipeline end-to-end against a live Teradata instance (`…env.trial.teradata.com`) that has **no `PDCRINFO` database** — the exact scenario this PR is meant to handle gracefully.

**The PR as-is does not achieve its goal on a box without `PDCRINFO`.** The two `optional: true` PDCR SQL steps hard-failed the pipeline instead of degrading to `ABSENT`:

```
ERROR  Step td_pdcr_info_agg_extract failed with error: [Error 3802] [SQLState 42S02] Database 'pdcrinfo' does not exist.
ERROR  Pipeline execution failed due to errors in steps: td_pdcr_info_agg_extract, td_pdcr_sp_exe_info_agg_extract
```

**Root cause** (in the partial-run contract from #2555, which this is stacked on):
- `teradatasql` raises a `sqlalchemy.exc.OperationalError` whose SQLSTATE is embedded only in the *message text* (`[SQLState 42S02]`), not in any attribute — so the shared `extract_sqlstate()` returns `None`.
- The message fallback `_classify_teradata_message()` only recognizes error `3807` (object missing), but a missing *database* raises error **`3802`**.
- Net → `UNKNOWN` → `status_for_source_error` returns `ERROR` (not `ABSENT`) → pipeline aborts. `42S02` is already in `_ABSENCE_SQLSTATES`, so the classification was correct; only the extraction was failing.

**Fix** (kept isolated to `TeradataConnector`, so no other connector's error handling changes): `TeradataConnector.fetch` now parses `[SQLState …]` out of the driver message and raises a classified `SourceQueryError` directly. `DatabaseManager.fetch` already re-raises `SourceQueryError` untouched. Required (`optional=False`) steps still hard-fail as before — only the two optional PDCR SQL steps degrade.

**After the fix, same live box:**
```
WARNING  Step td_pdcr_info_agg_extract was absent for this deployment: … Database 'pdcrinfo' does not exist.
WARNING  Step td_pdcr_sp_exe_info_agg_extract was absent for this deployment: … Database 'PDCRINFO' does not exist.
INFO     Pipeline execution summary: complete=18 absent=2 error=0 skipped=0
```
The DuckDB extract stays valid — PDCR target tables are created by their DDL steps and left empty (0 rows), no corruption.

Verified via `make test` (full unit suite green besides the known-flaky `test_cli_analyze` timeout), `make fmt` (pylint 10.00/10), and the profiler CLI (`execute-database-profiler --source-tech teradata`). Added unit coverage for the 3802/42S02 → `ABSENCE` path in `tests/unit/connections/test_database_manager.py`.

Pushing the fix onto this branch. Note it touches the classification layer owned by #2555 — flagging in case you'd prefer it land there instead.

dey-abhishek:
@dgomez04 

### Re: `lint-uv` failure (uv.lock mixed with source)

For the record — this `uv.lock` change is **not** from my fix commit (`1a71efed`, which touches only `database_manager.py` + its test). The lock delta comes from **`7fe9c524`** ("Implemented changes to adhere to…") in the #2555 base stack, which regenerated `uv.lock` alongside `profiler.py`.

Looking at `git diff origin/main...HEAD -- uv.lock` (~3470 changed lines):
- The vast majority are **internal-registry URL rewrites** — `https://pypi.org/simple` and `files.pythonhosted.org` → `https://pypi-proxy.dev.databricks.com/...`. That's the internal proxy leaking into the lock file (the same class of issue as the recent "scrub the registry used in CDN URLs" change on `main`).
- The remaining ~94 lines are resolver churn (e.g. a `databricks-sdk` dependency `marker` change), not new dependencies.
- `pyproject.toml` and `.build-constraints.txt` are **unchanged vs `main`** — so there is no intentional dependency change here.

Since `lint-uv` evaluates the aggregate PR diff against `main`, a separate lock-only commit won't clear it; the branch's `uv.lock` needs to match `main` (regenerate against public PyPI via the repo's `lock-dependencies` flow). That fix belongs on the #2555 base branch where `7fe9c524` lives, rather than here. Flagging so it's not attributed to the Teradata changes.

dey-abhishek (approved):
LGTM

PR #2568, opened 2026-07-14
Enable DuckDB native compression for profiler extracts

dgomez04:
### Summary
Profiler extracts are now written with DuckDB's native compression so the single `.db` stays smaller and directly queryable.

- Connect through a shared helper that sets `force_compression=auto`, `zstd_min_string_length=8`, and `storage_compatibility_version=v1.2.0`.
- Keep the existing extract filename/path and downstream queryability unchanged.

| Extract | Baseline | auto + min8 |
|---|---|---|
| Synapse | 4.8 MB | 3.5 MB |
| Spark pool | 56 MB | 11.3 MB |
| Snowflake | 18.7 GB | 2.4 GB (~23s) |

### Test plan
- [x] Unit tests for DuckDB helpers
- [x] Profiler integration tests that assert the `.db` extract is written
- [ ] CI green on the PR


asnare (changes requested):
DuckDB has native (internal) support for compression; we should use that instead of doing it this way. It needs to be specified when modifying the database, and is transparent to readers.

asnare:
> DuckDB has native (internal) support for compression; we should use that instead of doing it this way. It needs to be specified when modifying the database, and is transparent to readers.

I should elaborate a bit here. From a usability point of view I think it's preferable to stick to a single file, and having it directly queryable is a massive win: enabling DuckDB native (Zstd) compression lets us keep this.

That said, I just ran some experiments and it turns out that the compression ratio we get when using DuckDB's Zstd support out-of-the-box is quite poor. Some knobs that improve it dramatically: setting `zstd_min_string_length` to something small like 8 helps it a lot, as does sorting (`ORDER BY ALL`) the table content. It's not as good as just compressing the entire database file, but it's much smaller than the uncompressed size and remains queryable without affecting our usability/handling.

src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
```diff
@@ -18,6 +19,17 @@
 
 SaveMode = Literal["overwrite", "append"]
 
+_PROFILER_DUCKDB_CONFIG: dict[str, bool | float | int | list[str] | str] = {
+    "storage_compatibility_version": "v1.2.0",
```
asnare: I dug into the versions, and I think we should use `v1.4.0`: it's the LTS release, and includes some other (non-zstd) encoding improvements that my experiments show are going to be wins for us.

This means we'll need to bump the `pyproject.yaml` dependency: it turns out we're already pinning to a new version, so this should be fine.

src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
```diff
@@ -18,6 +19,17 @@
...
+    "force_compression": "auto",
+    "zstd_min_string_length": "8",
```
asnare: Nit:
```suggestion
    "zstd_min_string_length": 8,
```

asnare (commented):
This looks good…

Can we make the DuckDB dependency change I've mentioned as a comment?

Something we haven't addressed so far… the format version is really part of our ingestion contract. With this in mind, we have some integration tests that can be updated to verify the format of the DuckDB files we're creating. This can be checked from python with:
```python
def _storage_version(db_path: Path) -> str | None:
    """Fetch the on-disk storage format of a DuckDB file."""
    with duckdb.connect(str(db_path), read_only=True) as conn:
        (tags,) = conn.execute(
            "SELECT tags FROM duckdb_databases() WHERE database_name = current_database()"
        ).fetchone()
    return tags.get("storage_version")
```

Integration tests to verify the version are probably the simplest way to demonstrate this is all working as intended.

Finally, we have some tests that are also creating DuckDB (`tests/resources/` and we should probably update them as well.

src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
dgomez04: Switched to `storage_compatibility_version: "v1.4.0"` and tightened the `duckdb` lower bound to `>=1.4.0,<1.5.0` in `pyproject.toml`. The resolved version was already `1.4.5`.

src/databricks/labs/lakebridge/resources/assessments/common/duckdb_helpers.py
dgomez04: Done :)

asnare (approved):

m-abulazm (approved):
LGTM

PR #2569, opened 2026-07-14
Add profiler correctness tests for Redshift and Synapse

nnigam19:
## What

Adds the first profiler-correctness tests, following review feedback that the
suite should validate profiler *correctness*, not just that the pipeline engine
runs. Test-only, no production changes.

- **`test_profiler_correctness_redshift.py`** (integration) - runs the real
  Redshift provisioned profiler queries against the live Redshift sandbox and
  asserts the DuckDB extract has no stray tables, the always-produced tables are
  present, and each produced table has exactly the expected column set. Skips
  when `REDSHIFT_*` creds are absent (runs in CI).
- **`test_profiler_correctness_synapse.py`** (integration) - runs the real
  Synapse metadata extract queries (`INFORMATION_SCHEMA`: tables/columns/views/
  routines) against the SQL Server sandbox stand-in and asserts each result
  ingests into its pinned `SYNAPSE_SCHEMAS` shape. Skips when `TEST_TSQL_*` creds
  are absent (runs in CI).
- **`test_synapse_schema_contract.py`** - validates every `SYNAPSE_SCHEMAS`
  declaration is valid DuckDB DDL producing exactly the declared columns and
  types. Needs no creds.
- **`profiler_extract_helpers.py`** - shared, non-test helpers (schema parsing,
  DuckDB inspection, env gating) used by the above.

## Testing

- `test_synapse_schema_contract.py`: 41 cases, all passing locally.
- The two integration tests were verified to lint clean and skip gracefully
  without creds, they exercise their live sandboxes in CI, where their exact
  column/schema expectations get their first live validation.

## Known gaps (follow-ups)

- Synapse coverage is the metadata subset only, the Azure-API steps (workspace
  info, monitoring metrics, SQL-pool enumeration) and `SYS.DM_PDW_*` activity
  extracts require a real Synapse workspace and are not covered.
- No Oracle integration test yet - blocked on provisioning an Oracle sandbox and
  credentials.
- Tests assert schema/column correctness, not value-level correctness.

### DIFF
--- a/tests/integration/assessments/profiler_extract_helpers.py
+++ b/tests/integration/assessments/profiler_extract_helpers.py
@@ -1,0 +1,61 @@
+"""Shared helpers for the profiler extract correctness tests.
+
+Kept in a non-``test_`` module so pytest does not collect it and so the schema
+parsing / DuckDB inspection helpers live in one place rather than being
+duplicated across the per-profiler test modules.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import duckdb
+
+from tests.integration.debug_envgetter import TestEnvGetter
+
+# Maps the type tokens used in the pinned profiler schemas (e.g. SYNAPSE_SCHEMAS)
+# to the type name DuckDB reports back via information_schema -- a declared STRING
+# column is a DuckDB VARCHAR, etc.
+DECLARED_TO_DUCKDB_TYPE = {
+    "STRING": "VARCHAR",
+    "BIGINT": "BIGINT",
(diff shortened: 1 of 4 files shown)


m-abulazm (changes requested):
I would prefer e2e tests that are more future proof. I would test the entry point `Profiler` instead of testing internal profiler components. at the beginning we can keep the assertions minimal. dont assert columns, it can be that it ran and terminated successfully, start with redshift and sqlserver as synapse secrets are not there yet.

PR #2570, opened 2026-07-14
Add fingerprint pre-check to Recon (Redshift)

ameersalman-db:
### What does this PR do?

Adds an **opt-in fingerprint pre-check** to Recon, exposed as the user-facing flag `reconcile_optimizer` (default `False`; existing behaviour is unchanged when off). When enabled and the source has a registered fingerprint query builder, Recon runs a sketch-based detection pass (MD5 sub-bucketed aggregates over both sides) **before** the row-hash compare pipeline:

- **MATCH** → short-circuits in seconds; no full scan, no JOIN.
- **MISMATCH** → an algebraic solver recovers the differing row hashes; a surgical Stage-2 fetch pulls just those rows into the existing `compare.reconcile_data` flow. If the mismatch is systemic (`> 15%` of sub-buckets, or `> 50,000` mismatched sub-buckets), it defers to the row-hash pipeline.
- **Ineligible / any error** → falls through to the row-hash pipeline (fail-open).

Redshift is the first source dialect; adding Snowflake / Oracle / TSQL is one `FingerprintQueryBuilder` subclass plus a registry entry, with no orchestrator changes.

> **Stacked on** #<PR-A> ("Fix Redshift↔Databricks row-hash serialization + success-count;
> centralize the type-transform lookup"). Both the shared `get_transform_for_type` lookup and
> `serialize_column_for_hash` live in `query_builder/column_transformer.py` — the
> `ColumnTransformer` service main introduced in #2601, which PR A extends with the
> Redshift/Databricks type handlers. This pre-check reuses that same lookup, so the detection
> byte stream is identical to the row-hash path by construction; PR A must merge first. Review this PR's diff against  
> that branch.

- **Single decision point:** `trigger_recon_service.run_fingerprint_or_reconcile_data`. Static eligibility (flag off, unsupported dialect, non-data report type, no join columns, filters / transforms / thresholds configured) is centralised in `classify_ineligibility` (`fingerprint/orchestrator.py`) and runs before any source compute. One reason — `unmapped_target_column_mapping` — needs the target schema, so it's detected inside the precheck via a typed `UnmappedTargetColumnMappingError` and routed through `FingerprintRunMetadata.ineligible(...)`. Every reason is an `IneligibilityReason` enum value recorded on `recon_metrics.fingerprint_metrics.ineligibility_reason`, so adoption queries need no log-grepping.
- **No new connector code:** source reads use upstream's `RemoteQueryReader` / `remote_query()` TVF unmodified.
- **Serialization reuses the shared transform map:** the fingerprint source (Redshift) and target (Databricks) serializers call `serialize_column_for_hash` (in `query_builder/column_transformer.py`), a thin adapter over the shared `get_transform_for_type` lookup that the row-hash path (`RuleBasedColumnTransformer`) also uses. The fingerprint byte stream is therefore identical to the row-hash pipeline by construction; only the MD5 → sub-bucket/bucket arithmetic is fingerprint-specific.
- **Parallelised Stage-1/Stage-2** across source / target via `ThreadPoolExecutor(max_workers=2)`; failure semantics match the serial version (the executor joins both futures and the first exception propagates).
- **Single new `ReconcileConfig` field:** `reconcile_optimizer: bool = False`. No config version bump and no migration — the field has a default, so existing configs load unchanged.
- **Column-level mismatch detail reuses the existing diff:** the MISMATCH path routes the Stage-2 prefetched frames through the existing `compare.capture_mismatch_data_and_columns` (same function the normal sampled path uses). Its per-column `<col>_match` flag is now computed null-safely (`<=>` rather than `==`) so a NULL-vs-value difference is attributed and NULL-vs-NULL is not — a fix to the **existing** diff, not a new one. Because Stage-2 fetches whole sub-buckets (Stage-1 only proves a sub-bucket holds *some* mismatch), a small `compare.filter_to_row_mismatches` row filter drops the pairs that turn out to match column-by-column.
- **Stage-2 query construction:** the surgical fetch reuses `HashQueryBuilder` (so its hash serialization is identical to the row-hash path) and injects the bucket-filter subquery via `HashQueryBuilder.build_query(..., from_expression=...)`. The subquery is spliced in as a raw string — the **same `query.replace(":tbl", ...)` mechanism every connector's `read_data` already uses** — rather than re-rendered through sqlglot, which would rewrite the hand-built Redshift bucket math (e.g. `MOD(x, n)` → `x % n`, `SUBSTRING(m, 1, 8)` → `SUBSTRING(m FROM 1 FOR 8)`) and mutate the MD5 arithmetic the filter depends on. The splice is anchored to a single `FROM <placeholder>` occurrence and raises if none is present.
- **Fail-open on Stage-2 build failure:** a `build_mismatch_output` exception falls through to the full pipeline instead of failing the table.
- **Typed NULLs in the persisted `fingerprint_metrics` struct** so the vectorized Parquet reader can read the metrics table back and schema equality against `recon_metrics` holds.

### Code-review hardening

Review-driven fixes on this branch; none changes the happy-path behaviour:

- **Case-insensitive `column_mapping` resolution.** `align_columns` validates mapping *targets* case-insensitively, but the target-side lookup in `spark_target._target_col_name` was a case-sensitive `dict.get`. A config whose `source_name` case differed from the source schema (e.g. `custid` vs `CustID`) passed validation and then missed the lookup, hashing the source-named column on the target (wrong verdict) or raising. The map is now keyed by the bare, lower-cased source name and looked up the same way, consistent with the rest of the pipeline.
- **Stage-2 frame caching for `report_type='all'`.** `build_mismatch_output` consumes the prefetched src/tgt frames twice — the `compare_reconcile_data` join and the `capture_mismatch_data_and_columns` join. They are now cached once, serverless-aware (`.cache()` on classic clusters, `write_and_read_df_with_volumes` on serverless) and released in a `finally`, mirroring the core `Reconciliation._get_mismatch_data` pattern, so the Stage-2 fetch (Redshift JDBC pull + Delta scan) and its shuffle are not re-run for the second join.
- **Genuine `<col>_match` flag detection.** `filter_to_row_mismatches` now identifies the match-flag columns by the presence of their `_base`/`_compare` siblings, so a *join key* literally named `*_match` (projected unsuffixed) is not misread as a boolean flag (which would raise on `~col(<non-boolean>)` and, via fail-open, silently drop the Stage-2 output).
- **Table-placeholder splice hardened.** The `from_expression` splice now replaces a single, anchored `FROM <placeholder>` occurrence (not a global `str.replace`) and raises if no placeholder is present — so a stray `:tbl` / `%(tbl)s` in a filter literal or identifier elsewhere in the SQL cannot be corrupted.
- **Hash-column set de-duplicated.** `ordered_hash_columns` and `build_query` share a single `_hash_column_set()` helper — one definition of the `(join ∪ select) − thresholds − drops` algebra and a single `sorted()` (a redundant double-sort was removed).
- **Dead solver guards removed.** Unreachable `< 0` checks on the `abs()`-derived roots in `solve_d1` and `solve_d2_swap` (both roots are already clamped to `[0, MAX_RH_VALUE]`).

- **Depends on the stacked base PR** (#<PR-A>) for the shared serialization map and the Redshift `TIMESTAMP`/`TIMESTAMPTZ`/`BOOLEAN`/`DOUBLE`/`FLOAT` handlers. On its own this branch would not hash those column types identically across engines.
- **DBR 17.3+:** source reads via `remote_query()` need classic clusters on DBR 17.3+ or SQL warehouses (Pro / Serverless) 2025.35+ — inherited from upstream's `RemoteQueryReader`, documented in `docs/lakebridge/docs/reconcile/index.mdx`.
- **MISMATCH cost:** Stage-1 is paid unconditionally and Stage-2 still feeds the same JOIN, so MISMATCH is slower than row-hash-only at small scale. The MATCH path is the win; the motivation is billion-row scale, where the full-table JOIN is intractable. (For `report_type='all'` the Stage-2 frames are cached so the compare and capture joins share one fetch — see Code-review hardening.)
- **`success_count` bug** in `verify_successful_reconciliation` (yields counts > total) is pre-existing (#2259) and is **not** fixed here — filed separately.

### Solver sufficiency — why d≤2 is enough

The solver only resolves a sub-bucket holding ≤2 differing rows (`d=1`, `d=2`-swap `d=2`-extras). Rows map to sub-buckets uniformly by content, so it's the classic occupancy problem. Two shipped mechanisms bound `λ = D / N` (differing rows per sub-bucket): adaptive tiering scales `N` with table size, and the systemic guard bails once `> min(50,000, 15%·N)` sub-buckets mismatch — so the solver only runs at low `λ`, which floors the per-tier solve rate at **≥98.8%** (rising to 99.99%+ at the larger tiers).

**Correctness does not depend on this coverage.** The `<2%` not solved algebraically (≥3-ball buckets, plus rare 32-bit `rh1` collisions the second `rh2` channel flags) are **brute-force fetched, never dropped** — `d<3` is an *efficiency* threshold, not a *correctness* one. Reproducible under `experiments/redshift_fingerprint_validation/proof/solver_sufficiency/`; the `≥98.8%` floor and a no-false-solve property are CI-guarded in `tests/unit/reconcile/fingerprint/test_solver_sufficiency.py`.

### End-to-end validation (real Redshift → Delta)

One self-contained notebook generates identical deterministic data in both engines, reconciles the same tables twice (`reconcile_optimizer` OFF = built-in row-hash, ON = fingerprint), and diffs the audit tables. Both modes produced **identical counts and identical exact records** (mismatch / missing-in-target / missing-in-source) on both a clean-match and a known-mutation scenario, with the pre-check engaged. Re-derivable from the `main` / `metrics` / `details` audit tables.

### Functionality

- [x] user docs (`configuration.mdx` *Fingerprint Pre-check (Experimental)* + `index.mdx`
  runtime requirements)
- [ ] added new CLI command
- [ ] modified existing command
- [x] new opt-in feature behind `ReconcileConfig.reconcile_optimizer` (default `False`)

### Tests

- [x] unit tests across the fingerprint modules (query builders, solver, eligibility,
  fetch parallelisation, recon-capture schema, dispatch, Stage-1 / Stage-2 symmetry) +
  parity tests asserting the fingerprint serializers are byte-identical to the shared
  transform map
- [x] `tests/unit/reconcile/` passes (423 tests, 2 skipped)
- [x] null-safety + genuine-mismatch-filter coverage for the reused compare-layer diff
  (`tests/integration/reconcile/test_compare.py`), plus unit coverage for the code-review
  hardening: case-insensitive `column_mapping` resolution, `report_type='all'` Stage-2
  frame caching, and the `<col>_match`-sibling flag detection
- [x] linter clean: black / ruff / mypy green across `reconcile/` + `config.py`; pylint
  10.00/10 on touched src
- [ ] integration tests (recon e2e cluster gate; coverage follows with the cluster fixture)

### DIFF
--- a/docs/lakebridge/docs/reconcile/configuration.mdx
+++ b/docs/lakebridge/docs/reconcile/configuration.mdx
@@ -138,6 +138,79 @@ Place the file in `.lakebridge/` in your Databricks workspace home folder.
     </TabItem>
 </Tabs>
 
+---
+
+## Fingerprint Pre-check (Experimental)
+
+The fingerprint pre-check is an opt-in optimisation that, when source and target are
+already in sync, replaces the per-row hash-and-join pipeline with a sub-bucket-level
+aggregate comparison. On a 1 M-row Redshift fixture in MATCH state, the precheck
+short-circuits the recon at roughly **30–45 % of the v3 baseline wall-clock**; on
+MISMATCH it surgically fetches only the differing rows instead of streaming the full
+column set across JDBC. When neither the MATCH nor the surgical-fetch path applies the
+recon falls open to the full pipeline so correctness is never compromised.
+
+### When to enable
+
+- The source / target tables are expected to be **mostly identical** (post-migration,
+  ongoing CDC) — the precheck pays off most when MATCH is the common outcome.
+- The runtime targets **DBR 17.3 or later**. The Stage-2 source-side fetch uses
+  Databricks' `remote_query()` table-valued function which requires DBR 17.3+. On
+  earlier runtimes, leave the flag off — the eligibility gate doesn't yet check DBR
+  version, so the JDBC call would fail mid-fetch and trigger the fail-open path.
+- The source dialect has a registered fingerprint query builder. Today only
+  `redshift` is wired. Other dialects fall through to the full pipeline silently.
+
+### Configuration
+
+Add to `recon_config_*.json` at the top level:
+
+```yaml
+reconcile_optimizer: true
+# Optional. When set, overrides the target Delta DESCRIBE DETAIL numRecords
+# lookup used to pick the sub-bucket tier. Use this when the target is non-Delta
+# (DESCRIBE DETAIL returns no numRecords) or when Delta stats are stale and the
+# heuristic lands on a tier that is too coarse / too fine for your workload.
+# Values <= 0 are treated as unset.
+fingerprint_row_count_override: 250000000
+```
+
+### Eligibility rules
+
+The pre-check declines (and the recon proceeds via the full pipeline) when **any** of
+the conditions below hold. These reasons are recorded in
+`recon_metrics.fingerprint_metrics.ineligibility_reason` so adoption queries can
+distinguish "feature off" from "feature on but table ineligible".
+
+| Reason value | Meaning |
+| --- | --- |
+| `flag_disabled` | `reconcile_optimizer` is false (default). |
+| `unsupported_dialect` | The source dialect has no registered fingerprint query builder (today: anything other than `redshift`). |
+| `report_type_not_data` | `report_type` is `schema` (precheck operates on data-level reconciles). |
+| `no_join_columns` | `join_columns` is empty. The precheck needs primary-key columns to disambiguate culprit rows during Stage-2. |
+| `filters_configured` | `filters.source` or `filters.target` is set. The precheck does not project filter predicates into Stage-1 aggregates yet. |
+| `transforms_configured` | `transformations` is set. Custom transformations are not supported on the fingerprint hash path. |
+| `column_thresholds_configured` | `column_thresholds` is set. Threshold semantics conflict with exact-hash comparison. |
+| `table_thresholds_configured` | `table_thresholds` is set. Same rationale as column thresholds. |
+
+A separate runtime gate validates that every `column_mapping.target_name` resolves to
+a real target column before issuing the source-side scan — a typo in the mapping is
+caught at eligibility time, not after Stage-1 has already pulled across JDBC.
+
+### Tuning and observability
+
+- The pre-check selects a sub-bucket / bucket count adaptively from the target's
+  Delta `numRecords` (DESCRIBE DETAIL). On a non-Delta target, or when the metric
+  is missing, it falls back to a static `(1 048 576, 32 768)` pair. Override
+  explicitly via `ReconcileConfig.fingerprint_row_count_override` (an
+  approximate target row count) when the heuristic picks a tier that you can
+  show is wrong for your workload.
+- Every recon writes a `fingerprint_metrics` named-struct to
+  `recon_metrics.fingerprint_metrics` regardless of eligibility, so adoption,
+  fall-open rate, and verdict distribution can be tracked from one query.
+
+---
+
 ## TABLE Config Schema
 
 <Tabs>
--- a/docs/lakebridge/docs/reconcile/index.mdx
+++ b/docs/lakebridge/docs/reconcile/index.mdx
@@ -88,6 +88,18 @@ The User configuring reconcile must have permission to:
 - `USE CATALOG` and `CREATE SCHEMA` on the target catalog
 - `CREATE VOLUME` if using a pre-existing schema on a serverless cluster
 
+### Runtime requirements
+
+- The job / interactive cluster running reconcile must be on **DBR 15.4 LTS or later**
+  for the standard data-comparison path.
+- If `reconcile_optimizer` is enabled (see
+  [Configuration Reference → Fingerprint Pre-check](/docs/reconcile/configuration#fingerprint-pre-check-experimental)),
+  the cluster must be on **DBR 17.3 or later**. The Stage-2 source-side fetch uses
+  Databricks' `remote_query()` table-valued function, which became available on DBR
+  17.3. On earlier runtimes, leave the flag off; otherwise the JDBC call fails
+  mid-fetch and the recon falls open to the full pipeline (correct, but the precheck
+  buys you nothing while paying for itself in cluster time).
+
 ### Serverless cluster support
 
 Reconcile automatically detects the cluster type and optimizes intermediate data persistence accordingly:
--- a/src/databricks/labs/lakebridge/config.py
+++ b/src/databricks/labs/lakebridge/config.py
@@ -291,7 +291,7 @@ class ReconcileJobConfig:
 @dataclass
 class ReconcileConfig:
     __file__ = "reconcile.yml"
-    __version__ = 2
+    __version__ = 3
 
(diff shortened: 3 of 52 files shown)


m-abulazm:
@ameersalman-db some initial thoughts given the PR description

> Config __version__ 2 → 3: new v2_migrate adds reconcile_optimizer and folds the legacy spellings (fingerprint_precheck and older names) into it; the v1 → v2 → v3 chain runs automatically via Installation.load. Zero user action.

This is hallucinated by the AI. there was never a legacy spelling except for local development which affects no users

> New ReconcileConfig fields: reconcile_optimizer: bool = False; fingerprint_row_count_override: int | None = None (pins the adaptive tier when DESCRIBE DETAIL can't read numRecords).

The config changes should be reconsidered

> Pre-existing fixes that ride along (upstream PR https://github.com/databrickslabs/lakebridge/pull/2339)
Integrating against real Redshift surfaced three pre-existing correctness bugs in the Redshift connector MR (https://github.com/databrickslabs/lakebridge/pull/2339). None is dataprint-specific — all corrupt the existing row-hash path on real schemas — but they sit in the integration path, so they're fixed inline in expression_generator.py and pinned by regression tests.

This should go on its own PR

> Session-level UTC pin for cross-engine timestamp determinism. The Databricks target renders timestamps via DATE_FORMAT, which is spark.sql.session.timeZone-dependent, so pin_utc_session sets the session timezone to UTC for the recon. It's gated on the source dialect (redshift) — a row-hash correctness concern shared by the plain compare and fingerprint paths, not gated on reconcile_optimizer — and the original value is restored when the recon completes, so a shared/interactive cluster sees no lasting change.

revert please. leave this up to the user

> Serialization consolidated onto the shared transform map. Earlier revisions had a hand-written per-column serializer on each of three paths, including a per-column UTC pin and a CAST(_ AS VARCHAR(65535)). Both removed: routing through DataType_transform_mapping makes fingerprint serialization identical to the row-hash path by construction (the default TRIM(col) doesn't truncate, so no width cast is needed).

I dont get this, but should be a new PR

> Table-placeholder substitution moved behind HashQueryBuilder.substitute_table(...), which resolves every dialect form (:tbl, %(tbl)s); the orchestrator no longer hard-codes placeholder syntax.

this needs to be reconsidered. we can just build the query completely with table name in the first place

> Null-safe column diff moved to compare.annotate_mismatch_columns(...) rather than hand-rolled in the orchestrator. (Routing through Reconciliation._get_sample_data would re-issue sampling queries and defeat the surgical fetch, so it lives in compare.py.)

introducing a third diff is no-go. we should fix the existing diff

> success_count bug in verify_successful_reconciliation (yields counts > total) is pre-existing (https://github.com/databrickslabs/lakebridge/pull/2259, e56c79c3d), sits next to fingerprint code, and is not fixed here to keep scope contained — filed separately.

new PR. we can group some of the fixes together on one PR

PR #2571, opened 2026-07-15
use correct exit code on profiler error

m-abulazm:
### What does this PR do?

`Profiler._execute` (profiler.py:93-95) and `PipelineClass.execute()` (pipeline.py:67,76) both correctly raise RuntimeError on step failures. The exception gets swallowed one level up, in blueprint's `App._route`:
  ```
  # databricks/labs/blueprint/cli.py:119-126
  try:
      cmd.fn(**kwargs)
  except Exception as err:
      logger.error(f"{err.__class__.__name__}: {err}")
```
  The error is logged and the process falls off the end of main — so the Python entrypoint exits 0 on any profiler failure, and the databricks CLI in turn reports success. This affects every lakebridge command that signals failure with a plain exception, not just the profiler.

  There is an escape hatch already used: SystemExit derives from BaseException, so it sails through blueprint's except Exception and terminates the process with exit code 1. That's exactly why `test-profiler-connection` raises SystemExit at cli.py:1206-1217.

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/cli.py
+++ b/src/databricks/labs/lakebridge/cli.py
@@ -1140,7 +1140,7 @@ def execute_database_profiler(
     try:
         profiler = Profiler.create(source_tech, variant, creds_path)
         profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
-    except (RuntimeError, FileNotFoundError, ValueError) as e:
+    except Exception as e:  # noqa: BLE001
         raise SystemExit(f"Profiler execution failed: {e}") from e
 
 
--- a/src/databricks/labs/lakebridge/cli.py
+++ b/src/databricks/labs/lakebridge/cli.py
@@ -1137,8 +1137,8 @@ def execute_database_profiler(
             default=str(default_output_folder(source_tech)),
         ).strip()
 
-    profiler = Profiler.create(source_tech, variant, creds_path)
     try:
+        profiler = Profiler.create(source_tech, variant, creds_path)
         profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
     except (RuntimeError, FileNotFoundError, ValueError) as e:
         raise SystemExit(f"Profiler execution failed: {e}") from e
--- a/src/databricks/labs/lakebridge/cli.py
+++ b/src/databricks/labs/lakebridge/cli.py
@@ -1138,8 +1138,10 @@ def execute_database_profiler(
         ).strip()
 
     profiler = Profiler.create(source_tech, variant, creds_path)
-    # TODO: Add extractor logic to ApplicationContext instead of creating inside the Profiler class
-    profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
+    try:
+        profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
+    except (RuntimeError, FileNotFoundError, ValueError) as e:
+        raise SystemExit(f"Profiler execution failed: {e}") from e
 
 
 def parse_profiler_variant(prompts: Prompts, source_tech: str, variant: str | None) -> str | None:

### CODE REVIEW

File: src/databricks/labs/lakebridge/cli.py [hunk 1]
```diff
@@ -1138,8 +1138,10 @@ def execute_database_profiler(
         ).strip()
 
     profiler = Profiler.create(source_tech, variant, creds_path)
```
dgomez04: `Profiler.create` can raise `FileNotFoundError` and `ValueError` before we enter the try. Those paths still exit `0` through `blueprint`. We could probably `create` + `profile` on the same `try` block?

src/databricks/labs/lakebridge/cli.py
```diff
@@ -1138,8 +1138,10 @@ def execute_database_profiler(
...
-    # TODO: Add extractor logic to ApplicationContext instead of creating inside the Profiler class
-    profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
+    try:
+        profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
+    except (RuntimeError, FileNotFoundError, ValueError) as e:
+        raise SystemExit(f"Profiler execution failed: {e}") from e
```
dgomez04, line 1144: No unit test covers the new `SystemExit` path.

dgomez04 (approved):
LGTM

dey-abhishek (changes requested):
## Review: correct diagnosis, but the `except` tuple leaves the mssql path uncovered

Verified the root cause holds: blueprint's `App._route` does `except Exception: logger.error(...)`
(`databricks/labs/blueprint/cli.py:119-126`) and falls off the end, so any plain exception → exit 0.
Raising `SystemExit` (a `BaseException`) correctly escapes it. Moving `Profiler.create` inside the
`try` is right too, since `create()` can raise on its own.

### The gap

```python
    try:
        profiler = Profiler.create(source_tech, variant, creds_path)
        profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
    except (RuntimeError, FileNotFoundError, ValueError) as e:   # ← this tuple
        raise SystemExit(f"Profiler execution failed: {e}") from e

There are two regions where a connection can fail:

- Inside profile() → _execute() (profiler.py:80-95): re-wraps everything as RuntimeError.
Covered for all sources. ✅
- Inside Profiler.create() → resolve_variant() (profiler.py:42): runs before that wrapper.
For mssql, resolve_mssql_variant opens a live connection and runs an edition probe
(variants.py:40-42, db_manager.fetch(...)). On failure it raises ConnectionError — from
DatabaseManager.fetch wrapping OperationalError (database_manager.py:309), or from the connector
directly on bad auth (database_manager.py:98,155).

ConnectionError → OSError → Exception, so it is not in (RuntimeError, FileNotFoundError, ValueError):

>>> issubclass(ConnectionError, (RuntimeError, FileNotFoundError, ValueError))
False

So a wrong password / unreachable server during mssql variant detection escapes this handler, hits
blueprint's except Exception, and still exits 0 — the exact bug this PR targets.

Why only mssql

mssql is the only source with an AUTO variant resolver (_constants.py:23-27) that connects during
create(). Every other source (synapse, legacy_synapse, snowflake, oracle, teradata, redshift, bigquery)
first connects inside _execute, so its failure is already re-wrapped as RuntimeError and caught.
(SQLAlchemy create_engine is lazy; redshift's eager connect() still runs inside _setup_extractor,
i.e. inside _execute.) mssql is precisely the source that grew a create-time probe in the multi-db work (#2536).

Likely Fix to be tested

Catch Exception instead of the narrow tuple — closes the mssql gap and is future-proof against the next source that adds a create-time probe. This also matches the sibling test-profiler-connection handler
(cli.py:1215):

    try:
        profiler = Profiler.create(source_tech, variant, creds_path)
        profiler.profile(output_folder=Path(output_folder), cred_file_path=creds_path)
    except Exception as e:  # noqa: BLE001
        raise SystemExit(f"Profiler execution failed: {e}") from e

The narrow tuple buys nothing here since the handler collapses every type into an opaque SystemExit string regardless of the exception class.

dey-abhishek (approved):
LGTM

PR #2289, opened 2026-02-12
Bug resolve synpase profiler errors

eri-adepoju:
### What does this PR do?

1. Pipeline and trigger run extraction – Updates handling of list_pipeline_runs and list_trigger_runs so they correctly process batched yields (lists of dicts) instead of treating each yield as a single run.
2. Serverless SQL pool routines – Adds list_serverless_routines() using sys.objects because information_schema.routines is not available in serverless pools.
3. Server-level DMVs – Reconnects to the master database before querying server-level DMVs (e.g., data_processed), since these views must be queried from master.
4. Whitespace in credentials – Strips leading/trailing whitespace from credentials and config values (user, password, server, database, driver, auth_type, tz_info, development_endpoint) to avoid connection failures from copy-paste or config issues.
5. DataFrame concatenation – Replaces deprecated DataFrame.union() with pd.concat() in monitoring_metrics_extract.py.
6. Documentation – Clarifies Azure auth (DefaultAzureCredential order), local vs CI setup, DMV permissions (VIEW DATABASE STATE, VIEW SERVER STATE, VIEW DEFINITION), and serverless pool catalog views.

### Relevant implementation details

- serverless_sqlpool_extract.py: Uses get_sqlpool_reader(config, 'master', ...) before querying server-level DMVs; routines query switched from list_routines to list_serverless_routines.
- database_manager.py: All credential/config string fields use .strip() before connection
- monitoring_metrics_extract.py: step_name for spark pool metrics moved outside the loop, so it is defined even when the loop is empty.

### Linked issues

Resolves #2287 

### Functionality

- [x] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs lakebridge ...`
- [x] fixed existing functionality

### Tests
- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

PR #2572, opened 2026-07-15
Fix Synapse profiler failures for empty SQL pools and serverless routines schema

dgomez04:
## Summary

Fixes two customer-reported failures when running the Synapse profiler on a serverless-only workspace.

### 1. `workspace_info` fails when there are no dedicated SQL pools

**Root cause:** When `list_sql_pools()` returns an empty list, `pd.json_normalize([])` produces a DataFrame with zero rows and zero columns. `save_to_duckdb` called `conn.register()` before checking for column metadata, so DuckDB raised `Need a DataFrame with at least one column` and aborted the entire `workspace_info` step — even though an empty pool list is a valid outcome.

**Fix:** Reorder `_save_overwrite` in `duckdb_helpers` to check for zero-column DataFrames before registering with DuckDB. When no schema is provided and the table already exists, truncate it (clearing stale data on re-run); otherwise warn and skip table creation. This is a systemic fix that also protects other workspace resources that may return empty lists.

### 2. `serverless_sqlpool_extract` fails on `serverless_routines`

**Root cause:** PR #2289 switched serverless routines extraction to query `sys.objects` via `list_serverless_routines()` (7 columns) because `information_schema.routines` is not supported on serverless pools. When pinned schemas were added in #2431, `serverless_routines` was copied from `dedicated_routines` (23 columns, matching `information_schema.routines`). The resulting column count mismatch caused `Binder Error: table serverless_routines has 23 columns but 7 values were supplied`.

**Fix:** Update `SYNAPSE_SCHEMAS["serverless_routines"]` to match the 7 columns returned by `list_serverless_routines()`: `ROUTINE_SCHEMA`, `ROUTINE_NAME`, `ROUTINE_TYPE`, `CREATED`, `LAST_ALTERED`, `ROUTINE_DEFINITION`, `POOL_NAME`.

## Test plan

- [x] `uv run pytest tests/unit/assessment/test_duckdb_helpers.py tests/unit/assessment/test_synapse_schemas.py`
- [x] New tests cover zero-column overwrite (skip + truncate), empty schema-backed tables, and serverless routines schema alignment with insert


tests/unit/assessment/test_synapse_schemas.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import annotations
```
m-abulazm: please remove

m-abulazm (approved):

gueniai (approved):

PR #2574, opened 2026-07-16
Add ClickHouse profiler

dey-abhishek:
### What does this PR do?

Adds **ClickHouse** as a first-class citizen profiler source on Lakebridge, covering both **ClickHouse Cloud** and **self-managed / OSS**. It conforms to the existing Lakebridge profiler contract (registration, credential configurator, connector, pipeline config, DuckDB output), so it behaves consistently with the other sources (mssql, bigquery, redshift, synapse, oracle, teradata, snowflake).

The **analytics** portion of the standalone tool is out of scope; only the read-only `system.*` metadata profiling is ported.

### Architecture: SQL/DDL steps (not a Python extractor)

The profiler is built from **paired `ddl` + `sql` profiler steps**, matching the mssql/oracle/teradata profilers — there is no monolithic Python extractor or collector framework. Each of the ~64 result sets is a typed DuckDB `CREATE TABLE` (the `ddl` step) plus a read-only `system.*` query (the `sql` step). SQL is easier to maintain and debug than a custom Python script, and it keeps ClickHouse consistent with every other SQL-based source.

- **Source registration** — `clickhouse` added to `PROFILER_SOURCE_SYSTEM` with an `AUTO` variant. The `oss` vs `cloud` variant is auto-detected at run time (no user selection), mirroring the mssql pattern: a `*.clickhouse.cloud` host or the `cloud_mode` server setting decides.
- **Shared vs split SQL** — replicated-metadata queries live once in the `clickhouse/` root and are shared by both variants. The per-node append-only log tables (`query_log`, `session_log`, `query_views_log`, `asynchronous_insert_log`) are split into `oss/` (read directly) and `cloud/` (wrapped in `clusterAllReplicas('default', ...)` so all replicas are covered) — the same shared-root + variant-dir layout as mssql `single_db`/`multi_db`. Log-table steps are `optional: true`, so a build without a given table degrades to `ABSENT` instead of failing the run.
- **Connector** — `ClickHouseConnector` over the `clickhouse-connect` HTTP client (native-client style, like `RedshiftConnector`, not SQLAlchemy). `secure`/port defaults are host-derived (TLS + 8443 for Cloud hosts, plaintext + 8123 otherwise) so it is never insecure-by-default for Cloud. Driver errors are wrapped in `ConnectionError` so the pipeline can classify a missing/disabled `system.*` table as `ABSENT` (mirrors the other connectors).
- **Credential configurator** — `ConfigureClickHouseAssessment` prompts host/port/user/password/secure and an optional Cloud API block (`key_id`/`key_secret`, optional org/service id, optional tier override). There are no `days_back`/`redact` knobs: the query-history window is a fixed 30 days and redaction is always on (see below).
- **Redaction is always-on, in SQL** — sensitive fields (SQL text, auth params, host IPs/allow-lists, row-policy filters) are emitted as a `'[REDACTED]'` literal directly in the `sql` step, so the real value never leaves the source. Array/tuple columns are JSON-encoded with `toJSONString()`. Aggregate metrics, counts, and grant structure are always preserved.
- **DDL-first, so every table always exists** — because the `ddl` step creates the typed table before the `sql` step populates it, the output schema is stable across runs and deployments even when a result set is empty (consistent with the mssql `*_ddl.sql` approach).
- **Cost** — **actual billed cost only**, pulled from the ClickHouse Cloud `usageCost` API. This is the one remaining Python step (`cloud/cost_enrich.py`), run **optionally in the Cloud pipeline only**, since the billing figures live behind the Cloud REST API and SQL cannot reach them. It writes provider/region, real compute sizing, plan tier (recorded as metadata, no pricing math), and the billed cost into `costs_pricing_config`. OSS emits a static `costs_pricing_config` row via SQL (no dollar figures) and always reports a resource footprint.

- Adds a runtime dependency: `clickhouse-connect~=1.4.2` (+ transitive `lz4`, `zstandard`).
- The Cloud-only `cost_enrich.py` is the sole Python step; if `cloud_api` credentials are absent or the API is unreachable it writes a "no enrichment" row and the step still succeeds (it is `optional: true`).
- The query-history window is hardcoded to `INTERVAL 30 DAY` in the log-table SQL.
- Verified end-to-end against a live ClickHouse Cloud service and a local OSS instance: variant auto-detection, all steps `COMPLETE` (with `session_log` correctly `ABSENT` on OSS), 77 tables per run, redaction, Cloud actual-cost + tier detection, OSS footprint-only.

### Linked issues

Resolves #..

### Functionality

- [x] added relevant user documentation (`docs/.../assessment/profiler/clickhouse.mdx` + profiler index entry)
- [ ] added new CLI command
- [x] modified existing command: `databricks labs lakebridge configure-database-profiler` / `execute-database-profiler` (new `clickhouse` source)

### Tests

- [x] manually tested (live ClickHouse Cloud + local OSS, both variants, output verified)
- [x] added unit tests (variants, connector, configurator, Cloud cost enrichment, CLI)
- [ ] added integration tests

---


m-abulazm:
@dey-abhishek why are we using a python script instead of SQL?

dey-abhishek:
@m-abulazm 

Good question. The short answer is that the ClickHouse collectors aren't a fixed set of queries — they make decisions **while they run**, so a plain `.sql` file wouldn't cover it.

A few concrete reasons:

- **The same query has to run differently on Cloud vs OSS.** On Cloud the per-node log tables (`query_log`, `session_log`, etc.) only show the local replica, so we wrap them in `clusterAllReplicas(...)`; on OSS we read them directly. We only know which one we're hitting after probing `is_cloud` at runtime, so a static file would be wrong for one of the two.
- **We want to degrade gracefully on OSS.** Some `system.*` tables/columns just don't exist on older OSS builds. `safe_query()` looks at the ClickHouse error code and treats a *missing object* (16/47/60/81) as an empty result + warning, while still surfacing real failures like permissions or bad SQL. You can't branch on an error code from a SQL script.
- **Part of it isn't SQL at all.** For cost we call the ClickHouse Cloud REST API to get the real service config (provider, region, replica memory/count) and merge that with the query output, so the numbers reflect the actual service instead of a guess from the hostname.
- **There's runtime work around the results too** — redacting sensitive fields, coercing `days_back` to an `int` before it goes into `INTERVAL ... DAY`, and JSON-encoding ClickHouse types like `Decimal`/`UUID`/`bytes`.

It also matches what we already do — **Synapse** and **BigQuery** are Python extract scripts for the same reasons. The `.sql`-file pattern (mssql, oracle, teradata, snowflake) is used where a static query set is enough; anything that needs branching or an API call is Python.

dey-abhishek:
Hi @m-abulazm - The profiler's collector feature has been now rewritten into SQL/DDL steps, as requested

docs/lakebridge/docs/assessment/profiler/clickhouse.mdx
```diff
@@ -0,0 +1,265 @@
...
+- [Cost Attribution](#cost-attribution)
+- [Data Privacy and Redaction](#data-privacy-and-redaction)
+- [Configure Connection to ClickHouse](#configure-connection-to-clickhouse)
+- [Execute the Profiler](#execute-the-profiler)
+
+## Prerequisites
+
+### 1. Download
+- No ODBC driver is required. The ClickHouse profiler connects over the HTTP interface using the
+  [`clickhouse-connect`](https://clickhouse.com/docs/integrations/python) client, which is installed with Lakebridge.
```
m-abulazm: please remove this part. why even mention something if it is not needed nor in-scope

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -503,6 +503,78 @@ def _configure_credentials(self) -> None:
...
+class ConfigureClickHouseAssessment(AssessmentConfigurator):
+    """ClickHouse assessment configuration (ClickHouse Cloud and self-managed / OSS).
+
+    Connection keys (``host``/``port``/``user``/``password``/``secure``) are stored flat so the
+    ``ClickHouseConnector`` and the OSS-vs-Cloud variant probe can read them directly. The profiler
+    itself is SQL/DDL: the query-history window is a fixed 30 days and sensitive fields (SQL text, auth
+    params, host IPs, row-policy filters) are always redacted to ``[REDACTED]`` in the SQL, so there
+    are no ``days_back`` / ``redact`` knobs. An optional ``cloud_api`` block enables pulling the actual
+    billed cost + real sizing + plan tier from the ClickHouse Cloud API (Cloud only).
+    """
```
```suggestion
    """
```

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -503,6 +503,78 @@ def _configure_credentials(self) -> None:
...
+            source_creds["cloud_api"] = cloud_api
+
+        credential = {
+            "secret_vault_type": secret_vault_type,
+            "secret_vault_name": secret_vault_name,
+            source: source_creds,
+        }
+        _save_to_disk(credential, cred_file)
+
+        logger.info(f"Credential template created for {source}.")
```
```suggestion
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -288,6 +293,55 @@ def health_check(self) -> bool:
         return result.rows[0][0] == 101
 
 
+class ClickHouseConnector(DatabaseConnector):
+    """ClickHouse connector over the clickhouse-connect HTTP client.
+
+    Not SQLAlchemy-based (like Redshift), so it subclasses the ABC directly. Works against both
+    ClickHouse Cloud (``secure=True``, port 8443) and self-managed / OSS (port 8123). Only read-only
+    ``system.*`` metadata queries are ever run through it.
```
m-abulazm:
```suggestion

```

src/databricks/labs/lakebridge/resources/assessments/clickhouse/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    if isinstance(value, bool):
+        return value
+    if value is None:
+        return default
+    if isinstance(value, str):
+        return value.strip().lower() in _TRUTHY
+    if isinstance(value, (int, float)):
+        return bool(value)
+    return default
```
m-abulazm: our confirm prompts always return True or False literals. this extra parsing is unnecessary

src/databricks/labs/lakebridge/resources/assessments/clickhouse/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def normalize_secure_and_port(config: dict) -> tuple[bool, int]:
```
m-abulazm: can we embed this inside the connector

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -288,6 +293,55 @@ def health_check(self) -> bool:
...
+        self.config = config
+        self._client: ClickHouseClient = self._connect()
+
+    def _connect(self) -> ClickHouseClient:
+        host = str(self.config["host"])
+        # The configurator always writes `secure` explicitly; this fallback is for a minimal
+        # hand-written credentials file. normalize_secure_and_port forces TLS/8443 for managed
+        # ClickHouse Cloud hosts (which only accept TLS) and defaults to plaintext/8123 for
+        # self-managed / OSS — never insecure-by-default for Cloud. Shared with the extraction
+        # ClickHouseConnection so both paths behave identically.
```
m-abulazm:
```suggestion

```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -288,6 +293,55 @@ def health_check(self) -> bool:
...
+            port=port,
+            username=str(self.config.get("user", "default")),
+            password=str(self.config.get("password", "")),
+            secure=secure,
+        )
+
+    def fetch(self, query: str) -> FetchResult:
+        # Wrap driver errors as ConnectionError so the pipeline can classify them: an optional step
+        # that hits a missing/disabled system table (e.g. system.session_log on an OSS build) degrades
+        # to ABSENT instead of aborting the whole run. Mirrors the other connectors (e.g. Redshift).
```
m-abulazm:
```suggestion
```

src/databricks/labs/lakebridge/resources/assessments/clickhouse/cloud/cost_enrich.py
```diff
@@ -0,0 +1,212 @@
...
+        "tier": tier,
+        "tier_source": tier_source,
+        "is_cloud": True,
+        "note": note,
+        "cloud_service": json.dumps(cloud_service) if cloud_service else None,
+        "actual_billed_cost": json.dumps(actual_billed_cost) if actual_billed_cost else None,
+    }
+
+
+def execute(credential_manager: CredentialManager, db_path: str) -> dict[str, Any]:
```
m-abulazm:
```suggestion
def execute(credential_manager: CredentialManager, db_path: str, client: ClickHouseCloudAPI) -> dict[str, Any]:
```

please pass client to substitute the dependency in the tests

m-abulazm (changes requested):

src/databricks/labs/lakebridge/resources/assessments/clickhouse/cloud/cost_enrich.py
dey-abhishek: Done

src/databricks/labs/lakebridge/connections/database_manager.py
dey-abhishek: Removed

src/databricks/labs/lakebridge/connections/database_manager.py
dey-abhishek: Removed

src/databricks/labs/lakebridge/resources/assessments/clickhouse/__init__.py
dey-abhishek: Done

src/databricks/labs/lakebridge/resources/assessments/clickhouse/__init__.py
dey-abhishek: Done

src/databricks/labs/lakebridge/connections/database_manager.py
dey-abhishek: Removed

src/databricks/labs/lakebridge/assessments/configure_assessment.py
dey-abhishek: Removed

src/databricks/labs/lakebridge/assessments/configure_assessment.py
dey-abhishek: Removed

docs/lakebridge/docs/assessment/profiler/clickhouse.mdx
dey-abhishek: Done

dey-abhishek:
Added feature as requested by @AlanChoi https://github.com/databrickslabs/lakebridge/pull/2574/commits/0e82ff46cb6cc9744f8e3c31e91981ebe1796071

dey-abhishek:
Added UBIGINT changes as per code review comment by @Rubjit

PR #2576, opened 2026-07-20
Collect CPU/DWU utilization for legacy_synapse via Azure Monitor

### LINKED ISSUE #2557: Collect CPU / DWU utilization for `legacy_synapse` via Azure Monitor
### Issue
The `legacy_synapse` profiler does not currently collect CPU or DWU utilization. 

Its pipeline only extracts schema metadata and PDW activity DMVs: `sys.dm_pdw_exec_requests`, `sys.dm_pdw_exec_sessions`, and `sys.dm_pdw_nodes_db_partition_stats`.

By contrast, the workspace-backed `synapse` profiler already pulls dedicated-pool Azure Monitor metrics (including `CPUPercent`, `DWUUsedPercent`, `MemoryUsedPercent`) via `SynapseMetrics.get_dedicated_sql_pool_metrics`.

This gap was explicitly noted as part of #2441 and never implemented.

As per the [Azure Monitor Reference](https://learn.microsoft.com/en-us/azure/azure-monitor/reference/supported-metrics/microsoft-sql-servers-databases-metrics) this data exists as part of `Microsoft.Sql/servers/databases`.

### Relevant Metrics

Metric (REST API name) | Notes
-- | --
cpu_percent | CPU percentage.
dwu_consumption_percent | DWU percentage. Applies only to data warehouses.
dwu_limit / dwu_used | DWU limit / used. Applies only to data warehouses.
memory_usage_percent | Memory percentage. Applies only to data warehouses.
physical_data_read_percent | Data IO percentage (feeds DWU%).
local_tempdb_usage_percent | Local tempdb %. Applies only to data warehouses.
active_queries / queued_queries | Applies only to data warehouses.

m-abulazm:
### What does this PR do?
Closes the gap noted in #2441 and tracked in #2557: the legacy_synapse profiler extracted only schema and PDW activity DMVs, never the dedicated-pool utilization metrics that the workspace synapse profiler already collects.

- SynapseMetrics.get_sql_dw_metrics: queries the standalone-pool (Microsoft.Sql/ servers/databases) REST metric names — cpu_percent, dwu_consumption_percent, dwu_limit/used, memory_usage_percent, physical_data_read_percent, local_tempdb_usage_percent, active_queries, queued_queries.
- New monitoring_metrics_extract.py pipeline step builds the pool's ARM resource id from a new azure block (subscription_id + resource_group; server = FQDN first label, database = pool name) and writes metrics_dedicated_pool_metrics. Fails the step if the azure block is absent — downstream processing requires the metrics.
- Configurator prompts legacy_synapse (only) for subscription id + resource group.

### Linked issues

Resolves #2557

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [X] added unit tests
- [X] added integration tests


tests/unit/assessment/test_assessment.py
dgomez04, line 1: Nothing across the tests added covers `monitoring_metrics_extract.execute()`.

tests/unit/assessment/test_assessment.py
```diff
@@ -130,6 +131,68 @@ def test_configure_sqlserver_credentials_ad_default(tmp_path):
...
+    assessment.run()
+
+    with open(file, 'r', encoding='utf-8') as fstream:
+        credentials = yaml.safe_load(fstream)
+
+    azure = credentials["legacy_synapse"]["azure"]
+    assert azure == {"subscription_id": "sub-123", "resource_group": "rg-analytics"}
+
+
+def test_configure_mssql_has_no_azure_block(tmp_path):
```
dgomez04: `test_configure_sqlserver_credentials` already does `assert credentials == expected_credentials`, and `expected_credentials` has no `azure` key. If someone were to add it, that tests would fail. 

We should be able to remove `test_configure_mssql_has_no_azure_block`

tests/unit/assessment/test_legacy_synapse_metrics.py
```diff
@@ -0,0 +1,55 @@
...
+        "dwu_used",
+        "memory_usage_percent",
+        "physical_data_read_percent",
+        "local_tempdb_usage_percent",
+        "active_queries",
+        "queued_queries",
+    ]
+
+
+def test_get_sql_dw_metrics_builds_dataframe_from_timeseries():
```
dgomez04: I think this test is redundant and the logic it exercises is already used by `get_dedicated_sql_pool_metrics` and others. Remove if you see fit.

dgomez04 (approved):
I think we can expand/consolidate testing but the core implementation LGTM. Approved.

dey-abhishek (commented):
@m-abulazm 

- The monitoring_metrics step isn't optional: true, so a metrics failure (old creds without azure block, or no Monitoring Reader role) fails the entire profiler run — losing schema/activity output too. Consider optional: true, or document it as a breaking config change.                                                      
- Error path does logger.error + sys.exit(1) but omits the structured print(json.dumps({"status":"error"...})) its Synapse sibling emits. Works via fallback, but add it for consistency.                                                                                                                                        
- All 5 aggregations requested for every metric (incl. active_queries/queued_queries) — Azure may reject unsupported aggregations and fail the whole query. Have we verified against a live standalone pool? ; recommend an end-to-end check before merge.

m-abulazm:
> @m-abulazm
> 
> * The monitoring_metrics step isn't optional: true, so a metrics failure (old creds without azure block, or no Monitoring Reader role) fails the entire profiler run — losing schema/activity output too. Consider optional: true, or document it as a breaking config change.

It is not optional and it can be considered a bug that this step was missing. end users with old config is detected and will be prompted to run re-configure. the desktop app already supports those fields

> * Error path does logger.error + sys.exit(1) but omits the structured print(json.dumps({"status":"error"...})) its Synapse sibling emits. Works via fallback, but add it for consistency.

restored 

> * All 5 aggregations requested for every metric (incl. active_queries/queued_queries) — Azure may reject unsupported aggregations and fail the whole query. Have we verified against a live standalone pool? ; recommend an end-to-end check before merge.

tested and works
Screenshot 2026-08-03 at 12 56 25

dey-abhishek (approved):
LGTM

PR #2578, opened 2026-07-20
Fix profiler DuckDB `DECIMAL` overflow across all source systems.

dgomez04:
### Issue

Snowflake profiler runs fail on `warehouse_usage` when writing to DuckDB:

```
Failed to cast value: Casting value "100.813581994" to type DECIMAL(11,9) failed: value is out of range!
```

Source databases return `NUMBER`/`DECIMAL` columns as Python `Decimal` objects. DuckDB samples the first 1,000 object-column rows to infer types and can pick something too narrow (e.g. `DECIMAL(11,9)`, max `99.99`). Large warehouses with `CREDITS_USED >= 100` in a single metering interval hit this during extract.

The same pattern can affect other SQL-based profilers (Redshift, legacy Synapse) when queries return fractional `NUMERIC` values without an explicit floating-point cast.

### Fix

Cast fractional usage/credit/metric columns to floating-point types in profiler SQL so results arrive as `float64` before DuckDB write — no Python-side type coercion.

MSSQL, Oracle, and Teradata are already protected by paired DuckDB DDL steps. Synapse SQL pool extracts already use explicit `SYNAPSE_SCHEMAS`.

### Changes

- **Snowflake:** `::DOUBLE` on credit/usage/rate/metric columns in `warehouse_usage`, `pipe_usage`, `automatic_clustering`, `materialized_view_refresh`, `query_history`, `query_samples`, `storage_usage`, `rate_sheet`
- **Redshift:** `::double precision` on `round`/`avg`/sum outputs across provisioned, serverless, and multi-AZ variants (including `3_rs_nodes`)
- **legacy_synapse:** float literals in `storage_info.sql`
- **BigQuery:** `FLOAT64` instead of `NUMERIC` in `workload_types.sql`

### Tests

- [x] `uv run pytest tests/unit/assessment/test_duckdb_helpers.py -q` (13 passed)
- [x] `make fmt` (passes)
- [ ] Re-run Snowflake profiler against account with `CREDITS_USED >= 100` and confirm `warehouse_usage` completes
- [ ] Spot-check Redshift profiler extract on an account with Spectrum / managed storage metrics


m-abulazm (approved):

gueniai (commented):
Can we add some tests that would prevent a regression if any of the casts are removed/modified?

gueniai (approved):
lgtm

PR #2581, opened 2026-07-21
Fix profiler overwrite table semantics

dgomez04:
### Issue

Profiler `overwrite` mode replaced rows with `TRUNCATE` + `INSERT` while retaining the existing DuckDB table definition. For tables created through inferred schemas, a first run could persist a narrow type such as `DECIMAL(10,8)`. A same-day rerun with larger values then failed during insertion because the stale schema was reused.

### Solution

- Replace inferred-schema tables with `DROP` + `CREATE TABLE AS SELECT` on overwrite so both the definition and data reflect the current run.
- Recreate DDL-owned tables before loading them so explicitly declared types remain authoritative.
- Align the Python profiler DuckDB helper with the same replace-table semantics.
- Update the existing overwrite tests to cover schema replacement and narrow-to-wide Decimal values.

Tests: `.venv/bin/python -m pytest tests/unit/assessment/test_duckdb_helpers.py -q --tb=short` (13 passed)

### Follow-up

We need to consolidate profiler writes behind a single API. `_save_overwrite`, `save_to_duckdb`, and `_save_to_db` currently provide overlapping write paths and can drift in behavior. A separate issue will be created to track that consolidation.

### DIFF
--- a/src/databricks/labs/lakebridge/assessments/pipeline.py
+++ b/src/databricks/labs/lakebridge/assessments/pipeline.py
@@ -48,8 +48,6 @@ class PipelineClass:
         self._db_path = db_path.expanduser()
         self._create_dir(self._db_path.parent)
         self._cred_file_path = cred_file_path
-        # Tables recreated by a ddl step in this run; SQL overwrite inserts into them
-        # rather than DROP+CTAS so declared types are preserved.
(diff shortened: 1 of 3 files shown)


m-abulazm:
Closing in favor of #2597 and carrying over relevant fixes from here

PR #2588, opened 2026-07-22
Add desktop app installation flow

gueniai:
Restructure the installation page to lead with two installation paths — the desktop app and the CLI — then document each as a self-contained flow. Add the Desktop App section (requirements, download, install, first run) with an install walkthrough gif, and consolidate the shared network access, proxy, and mirror guidance into one section both flows link to.

Update the FAQ link for restricted-network endpoints to point at the new Network access section.


dgomez04 (approved):
LGTM

PR #2591, opened 2026-07-27
Fix BigQuery reconcile data type fidelity

take60:
## What

Reconciling a BigQuery source returned wrong answers in both directions. This fixes
three defects, none of which the existing tests caught:

1. **`row`/`data`/`all` reported near-total mismatch on realistic tables** — even when
   the data had migrated correctly. Critical: it is silent and hits the default report
   type.
2. **`schema` accepted lossy migrations** — `DATETIME`→`timestamp`, `INT64`→`int`,
   `FLOAT64`→`float` all passed.
3. **`schema` rejected correct ones** — every `STRUCT` with two or more fields, plus
   `NUMERIC` nested in `ARRAY`/`STRUCT`.

## How it works

**Row hash.** `DataType_transform_mapping["bigquery"]` only cast every column to
`STRING`, and BigQuery does not render those strings the way Spark does:

|---|---|---|
| `1.5` as `NUMERIC` / `DECIMAL(38,9)` | `1.5` | `1.500000000` |
| `0` / `-2.25` | `0` / `-2.25` | `0.000000000` / `-2.250000000` |
| `1.0` as `FLOAT64` / `DOUBLE` | `1` | `1.0` |
| `TIMESTAMP 2023-12-31 19:00:00` | `2023-12-31 19:00:00+00` | `2023-12-31 19:00:00` |
| `TIMESTAMP …00.100000` | `…00.100+00` | `…00.1` |

Each family is now formatted to match Spark: `FORMAT('%.<scale>f', col)` for decimals,
`FORMAT_TIMESTAMP('%F %H:%M:%E*S', col, 'UTC')` / `FORMAT_DATETIME(...)` for the two
timestamp types. `%E*S` emits only the fractional-second digits present, as Spark does,
and the explicit `'UTC'` drops the `+00` suffix. The decimal scale is read from the
column's own declared type, so `decimal(10, 2)` pads to 2 rather than a fixed default.

Only the BigQuery source side changes — the Databricks target transform is shared by all
seven sources and is left untouched.

**Schema compare, false accepts.** sqlglot maps `INT64`/`INT`/`SMALLINT` and
`FLOAT64`/`FLOAT` to one category each, and both `DATETIME` and `TIMESTAMP` to instants,
so the wrong target compared as equal. `DATETIME`→`timestamp` is the worst of these: it
silently shifts every wall-clock value by the session offset. The schema query now
reports the Databricks type an equivalent migration produces (`bigint`, `double`,
`timestamp_ntz`), so a narrower target is rejected. Verified these still render back to
valid BigQuery SQL, which the sampling path depends on.

**Schema compare, false mismatches.** `_validate_parsed_query` normalized `", "` → `","`
on the two *converted* queries but not the two *originals*. BigQuery's
`INFORMATION_SCHEMA` reports `STRUCT<a INT64, b STRING>` with that separator, so both
equality checks failed for every 2+-field `STRUCT`, whatever its element types.
Normalizing both sides fixes it — and lets any source declare a spaced type parameter
such as `decimal(38, 0)`.

Rewrites are deliberately **top-level only**. Inside `ARRAY`/`STRUCT` a match is reached
by round-tripping the Databricks type back to its BigQuery spelling, so an element
rewritten to the Databricks name would never match. Bare nested `NUMERIC` is instead
given its implied `NUMERIC(38, 9)`, which is what `decimal(38, 9)` round-trips to.

## Changes

- **`query_builder/expression_generator.py`** — per-type BigQuery transforms replacing
  the `# TODO`, plus `bigquery_decimal_transform` for scale-aware decimals.
- **`query_builder/base.py`** — pass the declared type (not just the parsed category)
  through `_get_transform`, since the category discards precision/scale.
- **`connectors/bigquery.py`** — `_SCHEMA_QUERY` reports Databricks type names for
  `INT64`/`FLOAT64`/`DATETIME`; nested `NUMERIC` canonicalized. Docstring updated.
- **`schema_compare.py`** — normalize both original queries via a shared `_normalize`.
- **`docs/reconcile/index.mdx`** — new `FLOAT64` footnote (see below).
- **Tests** — 6 new unit tests: a verdict matrix asserting equivalent migrations validate
  and lossy ones do not, a spacing test across all six non-Databricks sources, and
  row-hash assertions covering two decimal scales. The integration fixture gains
  multi-field, nested and nested-`NUMERIC` structs.

## Testing

- `make fmt` / `make lint` — pylint **10.00/10**, mypy clean, ruff clean.
- `make test` — **1490 passed**, 2 skipped, 3 xfailed. No regressions.
- **Live, both engines.** Ran the actual generated hash SQL against BigQuery and a
  Databricks warehouse over identical rows containing `NUMERIC`, `DECIMAL(10,2)`,
  `FLOAT64`, `TIMESTAMP` and `DATETIME`. All three row hashes now match byte-for-byte:

  | id | BigQuery | Databricks |
  |---|---|---|
  | 1 | `397129a9…5b85d5b` | `397129a9…5b85d5b` |
  | 2 | `0042742f…1ddf28c3` | `0042742f…1ddf28c3` |
  | 3 | `1b5a4d50…a366b3ea` | `1b5a4d50…a366b3ea` |

- The integration schema-compare counts were computed by driving the real
  `_validate_parsed_query` rather than by hand, and the test asserts *which* six columns
  mismatch, so a wrong count cannot pass silently. Integration tests are env-gated and
  need a CI run — hence draft.

## Notes / limitations

- **`FLOAT64` is only partially fixed.** Databricks renders doubles with Java's
  `Double.toString` (`1.0E20`, `1.23456789E8`) and BigQuery has no format that
  reproduces it — `FORMAT('%t')` wins the integral case (`1` → `1.0`) but not scientific
  notation. Values outside roughly `1e-3`…`1e7` may still be reported as row mismatches.
  A full fix needs `DataType_transform_mapping["databricks"]`, which is shared by all
  seven sources and is not source-aware, so it is out of scope here. Documented, with
  `column_thresholds` as the workaround; happy to file a follow-up issue.
- **`schema_compare.py` is shared by all seven sources.** The whitespace change was
  regression-checked against every dialect fixture — snowflake 32/36, oracle 31/33,
  tsql 23/28, redshift 16/16, teradata 16/16, bigquery 18/24 — with zero columns
  changing verdict, and a new parametrized test covers all six.
- **Testing table had only `int64`/`string` columns —
  exactly the subset that coerces identically on both engines. The existing integration
  fixture had the same blind spot (single-field `struct<a int64>` only), which is why the
  STRUCT bug went unnoticed. Both are addressed.
- **Not in this PR:** moving `TIME`→`string`, `BIGNUMERIC`→`string` and
  `RANGE<T>`→`struct<start,end>` into the schema query. #2527's description claimed these
  and they were never there, but adding them changes currently-documented behavior and
  deserves its own PR. They remain covered by `_APPROXIMATE_TYPES` and the docs footnote.
- The fixture's `col_bignumeric` target was corrected from `decimal(38,9)` to `string`:
  BigQuery's precision of up to 76 digits exceeds Databricks' `DECIMAL` max of 38.

### DIFF
--- a/docs/lakebridge/docs/reconcile/index.mdx
+++ b/docs/lakebridge/docs/reconcile/index.mdx
@@ -27,12 +27,14 @@ Lakebridge Reconcile validates data fidelity after migration by comparing your s
 | SQL Server | Yes | Yes | Yes | Yes |
 | Redshift | Yes | Yes | Yes | Yes |
 | Teradata | Yes | Yes [^teradata-hash] | Yes [^teradata-hash] | Yes [^teradata-hash] |
-| BigQuery | Yes [^bigquery-types] | Yes | Yes | Yes |
+| BigQuery | Yes [^bigquery-types] | Yes [^bigquery-float] | Yes [^bigquery-float] | Yes [^bigquery-float] |
 | Databricks | Yes | Yes | Yes | Yes |
 
 [^teradata-hash]: Teradata has no portable cryptographic hash in pure SQL, so row-hash report types (`row`, `data`, `all`) require a user-installed hash UDF on the source and an explicit `hash_expression_overrides.source` entry on the recon config. See [Hash Expression](/docs/reconcile/configuration#hash-expression) for wiring.
 
 [^bigquery-types]: BigQuery types with no equivalent in Databricks — `BIGNUMERIC`, `TIME`, `RANGE<T>`, nested `JSON` such as `ARRAY<JSON>`, or these nested inside `ARRAY`/`STRUCT` — are left as their original BigQuery type and may be reported as `schema` mismatches even when the data migrated correctly.
+
+[^bigquery-float]: `FLOAT64` columns whose magnitude falls outside roughly `1e-3` to `1e7` may be reported as row mismatches even when the data migrated correctly: Databricks renders such a value in scientific notation (`1.0E20`) and BigQuery has no format that reproduces it, so the two row hashes differ. Use `column_thresholds` for the column to compare it numerically, or supply your own `transformations` entry. Values inside that range, and all `NUMERIC`/`DECIMAL` columns, hash identically on both sides.
 
 ---
 
--- a/src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
+++ b/src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
@@ -25,7 +25,15 @@ class BigQueryDataSource(DataSource):
 
     Schema type handling
     ---------------------
-    * ``NUMERIC`` -> ``decimal(38, 9)`` (BigQuery's NUMERIC default precision/scale; same-family, exact).
+    The schema query reports each column as the Databricks type an equivalent migration produces, so
+    that schema compare flags a target that is merely in the same sqlglot category:
+
+    * ``INT64`` -> ``bigint``, ``FLOAT64`` -> ``double`` (otherwise ``int``/``float`` compare as equal
+      even though they overflow BigQuery's range).
+    * ``DATETIME`` -> ``timestamp_ntz`` (otherwise ``timestamp`` compares as equal, silently shifting
+      every wall-clock value by the session offset).
+    * ``NUMERIC`` -> ``decimal(38, 9)``, its implied precision/scale; nested ``NUMERIC`` becomes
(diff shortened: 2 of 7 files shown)


take60:
Superseded by #2592, which runs from a branch on this repo so CI can obtain the JFrog OIDC token (every job here failed at "Setup for JFrog" because fork PRs cannot). Same commits.

Issue #1202, opened 2024-11-13
[BUG]: Number without scale in Oracle vs bigint in Databricks - Reconcilation is treating all records as "mismatch"


Tim-Chowdary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Category of Bug / Issue

ReconcileError

### Current Behavior

Problem Statement: When a column with a NUMBER data type (without scale) is involved in a join, mismatched records (both data and record count) are not displayed on the dashboard.
Root Cause: In Oracle, a NUMBER without scale is treated as a decimal, which causes the join criteria to fail, resulting in empty outputs.

Note: By commenting the transformation rule to handle nulls in joins, we were able to display the total number of mismatch records. However, mismatch data is not shown in the detail window. 

This is really to not generate false negatives during hash comparisons

```
 select a ,b, a=b, hash(a), hash(b), hash(a)=hash(b) 
    from (select cast(21.00 AS decimal(38,0))as a, cast(21 AS int) as b);

m-abulazm:
cannot reproduce. according to oracle official docs, numbers without scale are treated as integers

m-abulazm:
reopened after clarifying this behavior is caused by spark handling of oracle number data type

PR #2592, opened 2026-07-27
Fix BigQuery reconcile data type fidelity

take60:
### What does this PR do?

Fixes three data type defects in the BigQuery reconcile source. None were caught by the existing tests: #2527's end-to-end table was `int64`/`string` only, which are exactly the types that render identically on both engines.

| # | Defect | Effect |
|---|---|---|
| 1 | Row-hash strings differ from Spark's | `row`/`data`/`all` report near-total mismatch on any table with a decimal, float or timestamp column — **even when correctly migrated** |
| 2 | `schema` accepts lossy targets | `DATETIME`→`timestamp`, `INT64`→`int`, `FLOAT64`→`float` all pass |
| 3 | `schema` rejects correct ones | Every 2+-field `STRUCT`; `NUMERIC` nested in `ARRAY`/`STRUCT` |

- Files changed
  - src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
  - src/databricks/labs/lakebridge/reconcile/query_builder/base.py
  - src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
  - src/databricks/labs/lakebridge/reconcile/schema_compare.py
  - docs/lakebridge/docs/reconcile/index.mdx
  - tests/unit/reconcile/connectors/test_bigquery.py
  - tests/integration/reconcile/test_schema_compare.py

**Row hash.** `DataType_transform_mapping["bigquery"]` only cast every column to `STRING`, and BigQuery does not render those strings the way Spark does:

|---|---|---|
| `1.5` as `NUMERIC` / `DECIMAL(38,9)` | `1.5` | `1.500000000` |
| `1.0` as `FLOAT64` / `DOUBLE` | `1` | `1.0` |
| `TIMESTAMP …19:00:00` | `…19:00:00+00` | `…19:00:00` |

Each family now gets a matching format: `FORMAT('%.<scale>f', col)` for decimals, `FORMAT_TIMESTAMP('%F %H:%M:%E*S', col, 'UTC')` and `FORMAT_DATETIME('%F %H:%M:%E*S', col)` for the two timestamp types. `%E*S` emits only the fractional-second digits present, as Spark does, and the explicit `'UTC'` drops the `+00` suffix. The decimal scale is read from the column's declared type, so `decimal(10, 2)` pads to 2 — `_get_transform` already receives the full type string, only the parsed category discards it.

**False accepts.** sqlglot collapses every integer width to `INT64` and both float widths to `FLOAT64` when converting a Databricks type back to BigQuery, so a narrower target compares equal to the source. `_SCHEMA_QUERY` now reports the Databricks type an equivalent migration produces (`bigint`, `double`, `timestamp_ntz`), which separates them. Those spellings still convert back to valid BigQuery SQL, which the sampling path relies on.

**False mismatches.** `_validate_parsed_query` normalized `", "` → `","` on the two *converted* queries but not the two *originals*. BigQuery's `INFORMATION_SCHEMA` reports `STRUCT<a INT64, b STRING>` with that separator, so both equality checks failed for any multi-field `STRUCT`, whatever its element types. Both sides are now normalized. Bare nested `NUMERIC` was never canonicalized either and now becomes `NUMERIC(38, 9)`.

Only the BigQuery source side changes; the Databricks target transform is shared by all seven sources and is untouched.

- **`FLOAT64` is only partially fixed.** Spark renders doubles with Java's `Double.toString` (`1.0E20`, `1.23456789E8`) and BigQuery has no format that reproduces it. `FORMAT('%t')` wins the integral case (`1` → `1.0`) but not scientific notation, so magnitudes outside roughly `1e-3`…`1e7` may still be reported as row mismatches. A full fix needs `DataType_transform_mapping["databricks"]`, which every source shares and which is not source-aware — out of scope here. Documented, with `column_thresholds` as the workaround; happy to file a follow-up issue.
- **Type rewriting is deliberately top-level only.** Inside `ARRAY`/`STRUCT`, a match is only reached by round-tripping the Databricks type back to its BigQuery spelling, so an element rewritten to the Databricks name would never match. This is why `INT64` stays `INT64` inside a `STRUCT` but becomes `bigint` at the top level.
- **`schema_compare.py` is shared by all seven sources.** The whitespace change was regression-checked against every dialect fixture — snowflake 32/36, oracle 31/33, tsql 23/28, redshift 16/16, teradata 16/16, bigquery 18/24 — with zero columns changing verdict, and the integration suite passes on all five Python versions.
- **Not in this PR:** moving `TIME`→`string`, `BIGNUMERIC`→`string` and `RANGE<T>`→`struct<start,end>` into the schema query. #2527's description claimed these and they were never added, but doing so changes currently-documented behavior and deserves its own PR. They stay covered by `_APPROXIMATE_TYPES` and the docs footnote.
- The integration fixture's `col_bignumeric` target was corrected from `decimal(38,9)` to `string`: BigQuery's precision of up to 76 digits exceeds Databricks' `DECIMAL` max of 38.

### Linked issues

Progresses #2527

### Functionality

- [ ] added new CLI command
- [x] fixed data type handling for the `bigquery` reconcile source (`schema` / `row` / `data` / `all`)

### Tests

Verified on live BigQuery and Databricks: ran the hash SQL the query builders actually generate over identical rows containing `NUMERIC`, `DECIMAL(10,2)`, `FLOAT64`, `TIMESTAMP` and `DATETIME` columns. All three row hashes now match byte-for-byte (`397129a9…`, `0042742f…`, `1b5a4d50…`); before this change every row differed.

`make test` — 1472 passed, 2 skipped, 3 xfailed. The integration schema-compare counts were computed by driving the real `SchemaCompare` rather than by hand, and the test asserts *which* six columns mismatch, so a wrong count cannot pass silently.

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/lakebridge/reconcile/query_builder/base.py
+++ b/src/databricks/labs/lakebridge/reconcile/query_builder/base.py
@@ -12,6 +12,7 @@ from databricks.labs.lakebridge.reconcile.connectors.snowflake import SnowflakeD
 from databricks.labs.lakebridge.reconcile.exception import InvalidInputException
 from databricks.labs.lakebridge.reconcile.query_builder.expression_generator import (
     DataType_transform_mapping,
+    bigquery_decimal_transform,
     transform_expression,
     build_column,
 )
@@ -140,7 +141,13 @@ class QueryBuilder(ABC):
 
             parsed = datatype
             try:
-                parsed = exp.DataType.build(datatype, source).this.value
+                built = exp.DataType.build(datatype, source)
+                parsed = built.this.value
+                if source_dialect == "bigquery":
+                    # DECIMAL needs the declared scale, which `parsed` discards.
+                    scaled_transform = bigquery_decimal_transform(built)
+                    if scaled_transform is not None:
+                        return scaled_transform
             except sqlglot.errors.ParseError:
                 logger.warning(f"Could not parse datatype {datatype} for source {source_dialect}")
 
--- a/src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
+++ b/src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
@@ -254,8 +254,27 @@ def _get_is_string(column_types_dict: dict[str, DataType], column_name: str) -> 
 DataType_transform_mapping: dict[str, dict[str, list[partial[exp.Expression]]]] = {  # pylint: disable=invalid-name
     "universal": {"default": [partial(coalesce, default='_null_recon_', is_string=True), partial(trim)]},
     "bigquery": {
-        # TODO: add timestamps and numbers handling
+        # CAST(... AS STRING) does not render the way Spark does: BigQuery strips a decimal's trailing
+        # zeros (1.5 vs Spark's 1.500000000), drops the fractional part of a whole FLOAT64 (1 vs 1.0)
+        # and appends a UTC offset to a TIMESTAMP (...+00). Those differences change the row hash, so
+        # each family below is formatted to match Spark exactly. %E*S emits only the fractional-second
+        # digits that are present, which is also what Spark does.
         "default": [partial(anonymous, func="COALESCE(TRIM(CAST({} AS STRING)), '_null_recon_')")],
+        exp.DataType.Type.DOUBLE.value: [partial(anonymous, func="COALESCE(TRIM(FORMAT('%t', {})), '_null_recon_')")],
+        exp.DataType.Type.TIMESTAMPTZ.value: [
+            partial(
+                anonymous,
+                func="COALESCE(FORMAT_TIMESTAMP('%F %H:%M:%E*S', {}, 'UTC'), '_null_recon_')",
+            )
+        ],
+        # A BigQuery DATETIME parses as TIMESTAMP; the schema query reports it as timestamp_ntz, which
+        # parses as TIMESTAMPNTZ. Both spellings reach here, so both are mapped.
+        exp.DataType.Type.TIMESTAMP.value: [
+            partial(anonymous, func="COALESCE(FORMAT_DATETIME('%F %H:%M:%E*S', {}), '_null_recon_')")
+        ],
+        exp.DataType.Type.TIMESTAMPNTZ.value: [
+            partial(anonymous, func="COALESCE(FORMAT_DATETIME('%F %H:%M:%E*S', {}), '_null_recon_')")
+        ],
     },
     "snowflake": {exp.DataType.Type.ARRAY.value: [partial(array_to_string), partial(array_sort)]},
     "oracle": {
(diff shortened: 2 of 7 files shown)


src/databricks/labs/lakebridge/reconcile/query_builder/base.py
```diff
@@ -140,7 +141,13 @@ def _get_transform(datatype: str):
 
             parsed = datatype
             try:
-                parsed = exp.DataType.build(datatype, source).this.value
+                built = exp.DataType.build(datatype, source)
```
m-abulazm, line 144: @ganeshdogiparthi-db bigquery also needs scale and precision for transformations

take60: BigQuery's `INFORMATION_SCHEMA.COLUMNS` has no `numeric_precision` / `numeric_scale` columns — unlike Oracle, Snowflake, Redshift, T-SQL and Teradata, whose schema queries build `decimal(p,s)` from those fields. What BigQuery does expose is the precision and scale inside `data_type` itself (`NUMERIC(10, 2)`, `BIGNUMERIC(50, 20)`, `ARRAY<NUMERIC(10, 2)>`), so this reads them from the parsed type via `.expressions` instead. Only the scale is used, because the row hash is built from the string rendering and precision does not affect it.

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
```diff
@@ -254,8 +254,27 @@ def _get_is_string(column_types_dict: dict[str, DataType], column_name: str) ->
 DataType_transform_mapping: dict[str, dict[str, list[partial[exp.Expression]]]] = {  # pylint: disable=invalid-name
     "universal": {"default": [partial(coalesce, default='_null_recon_', is_string=True), partial(trim)]},
     "bigquery": {
-        # TODO: add timestamps and numbers handling
+        # CAST(... AS STRING) does not render the way Spark does: BigQuery strips a decimal's trailing
```
m-abulazm, line 257: can we use https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#format_numeric_type_as_string

take60:
Split as discussed: the timestamp half is now #2594, cut fresh from `main` and ready to review.

Keeping this one open in draft for the numeric half. It still carries the `FORMAT('%.<scale>f', …)` decimal transform, the `DOUBLE` transform and the `base.py` hook you commented on — none of which are in #2594. Once your numeric investigation lands I will rebase this onto whatever generic mechanism comes out of #1202, or close it if that work supersedes it.

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
take60: Tested it on live BigQuery — the `CAST(... AS STRING FORMAT ...)` form space-pads to the width of the format model, so it does not match Spark as-is:

```
CAST(NUMERIC '1.5' AS STRING FORMAT '999999999999999999999999999.999999999')
  -> "           1.500000000"
FORMAT('%.9f', NUMERIC '1.5')
  -> "1.500000000"     -- matches Spark's cast of decimal(38,9)
```

One related finding for your investigation: BigQuery's `INFORMATION_SCHEMA.COLUMNS` has no `numeric_precision` / `numeric_scale` columns, unlike the other sources, but `data_type` itself carries them — `NUMERIC(10, 2)`, `BIGNUMERIC(50, 20)`, `ARRAY<NUMERIC(10, 2)>`. Bare `NUMERIC` comes back without parameters.

Leaving all of this to you and #1202 — the timestamp-only split is #2594.

src/databricks/labs/lakebridge/reconcile/query_builder/base.py
take60: Following your "timestamps for now" — this hook is not in the timestamp-only split (#2594), which leaves `base.py` untouched entirely. Deferring the scale/precision question to your numeric investigation and #1202.

PR #2594, opened 2026-07-28
Fix BigQuery reconcile timestamp fidelity

take60:
### What does this PR do?

Fixes timestamp handling in the BigQuery reconcile source. Split out of #2592

| # | Defect | Effect |
|---|---|---|
| 1 | Row-hash strings differ from Spark's | `row`/`data`/`all` report mismatches on any table with a `TIMESTAMP` or `DATETIME` column — **even when correctly migrated** |
| 2 | `schema` rejects correct ones | Every `STRUCT` with two or more fields |
| 3 | `schema` accepts lossy targets | `DATETIME`→`timestamp`, `INT64`→`int`, `FLOAT64`→`float` all pass. **Not fixed here** — see below; the affected columns are warned about instead |

None were caught by the existing tests: #2527's end-to-end table was `int64`/`string` only, which are exactly the types that render identically on both engines, and the schema-compare fixture used a single-field `struct<a int64>`.

- Files changed
  - src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
  - src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
  - src/databricks/labs/lakebridge/reconcile/schema_compare.py
  - tests/unit/reconcile/connectors/test_bigquery.py
  - tests/integration/reconcile/test_schema_compare.py
  - docs/lakebridge/docs/reconcile/index.mdx

**Row hash.** `DataType_transform_mapping["bigquery"]` only cast every column to `STRING`, and BigQuery does not render a timestamp the way Spark does:

|---|---|---|
| `TIMESTAMP 2023-12-31 19:00:00` | `2023-12-31 19:00:00+00` | `2023-12-31 19:00:00` |
| `TIMESTAMP …00.100000` | `…00.100+00` | `…00.1` |
| `DATETIME …12:00:00.100000` | `2023-12-31 12:00:00.100` | `…12:00:00.1` |

Both families are now formatted explicitly: `FORMAT_TIMESTAMP('%F %H:%M:%E*S', col, 'UTC')` for `TIMESTAMP` and `FORMAT_DATETIME('%F %H:%M:%E*S', col)` for `DATETIME`. `%E*S` emits only the fractional-second digits that are present, which is what Spark does, and the explicit `'UTC'` drops the offset suffix.

A BigQuery `DATETIME` parses as sqlglot `TIMESTAMP`, while a user-supplied `timestamp_ntz` parses as `TIMESTAMPNTZ` — both spellings reach the mapping, so both are keyed.

**False accepts (now warned about, not fixed).** BigQuery loses the declared width: every integer type is an alias of `INT64` and both float types of `FLOAT64`, and sqlglot collapses them the same way, so a narrower Databricks target compares equal to the source. An earlier revision of this PR pinned `_SCHEMA_QUERY` to report `bigint`/`double`/`timestamp_ntz` to force a mismatch, but that flags correctly-migrated columns — a column *declared* `SMALLINT` in BigQuery is indistinguishable from one declared `INT64` — and one invalid column fails the whole table. It also broke the sampling path: the reported type is the CAST target there, and `CAST(... AS timestamp_ntz)` is not valid BigQuery.

`INT64`, `FLOAT64` and `DATETIME` are therefore reported as-is and added to `_APPROXIMATE_TYPES`, so `get_schema` warns and names the columns instead. This defect is documented, not closed.

**False mismatches.** `_validate_parsed_query` normalized `", "` → `","` on the two *converted* queries but not the two *originals*. BigQuery's `INFORMATION_SCHEMA` reports `STRUCT<a INT64, b STRING>` with that separator, so both equality checks failed for any multi-field `STRUCT`, whatever its element types. Both sides are now normalized, which also lets any source declare a spaced type parameter such as `decimal(38, 0)`.

- **`TIMESTAMP` row hashes assume a UTC Spark session.** The BigQuery side emits `FORMAT_TIMESTAMP(..., 'UTC')`, while the Databricks side stringifies the same instant using `spark.sql.session.timeZone`, which Lakebridge does not pin anywhere in `src/`. On a workspace whose session timezone is not UTC, every `TIMESTAMP` row would still hash differently. Pre-existing — the old `CAST(... AS STRING)` had the same asymmetry plus an offset suffix — and not closed here; the verification below was run on a UTC session. `DATETIME` is unaffected, since `FORMAT_DATETIME` has no timezone component.
- **`schema_compare.py` is shared by all seven sources.** `_normalize` is now scoped to the declared type rather than the whole `create table` string, so it cannot strip `", "` inside a quoted identifier. Regression-checked against every dialect fixture — snowflake 32/33, oracle 31/31, tsql 23/26, redshift 16/16, teradata 16/16 — with zero columns changing verdict versus `main`, including column names containing `", "`. Happy to split this into its own PR if you would prefer.
- **Nested types remain uneven and are out of scope.** `ARRAY<NUMERIC>` is still a false mismatch, and the `_APPROXIMATE_TYPES` regex matches STRUCT *field names* as well as types (`struct<int64 string>` warns). Both pre-date this PR; happy to raise either separately.
- `TIME`, `BIGNUMERIC`, `RANGE<T>` and nested `JSON` remain reported as-is and still surface as schema mismatches; they are covered by `_APPROXIMATE_TYPES` and the docs footnote.

### Linked issues

Progresses #2527
Split from #2592 (kept open for the numeric half)

### Functionality

- [x] added relevant user documentation (BigQuery type footnote in `docs/reconcile`)
- [ ] added new CLI command
- [x] fixed timestamp handling for the `bigquery` reconcile source (`schema` / `row` / `data` / `all`)

### Tests

Verified on live BigQuery and Databricks: ran the hash SQL the query builders actually generate over identical rows containing `TIMESTAMP` and `DATETIME` columns (including fractional seconds). All three row hashes now match byte-for-byte (`d6c50f40…`, `ff75c352…`, `8345e532…`); before this change every row differed.

That check was run on a UTC Spark session; see the first caveat above.

The integration schema-compare counts were computed by driving the real `SchemaCompare` rather than by hand, and the test asserts *which* six columns mismatch, so a wrong count cannot pass silently. A parametrized unit test pins which types warn and which stay silent, asserting on the column name so it cannot pass vacuously.

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
```diff
@@ -41,8 +48,16 @@ class BigQueryDataSource(DataSource):
 
     _LIST_SCHEMAS_QUERY = "select schema_name from `{catalog}`.INFORMATION_SCHEMA.SCHEMATA order by schema_name"
     _LIST_TABLES_QUERY = "select table_name from `{catalog}.{schema}`.INFORMATION_SCHEMA.TABLES order by table_name"
+    # Top-level scalars are pinned to their exact Databricks spelling so that schema compare rejects a
+    # narrower target: sqlglot maps INT64/INT/SMALLINT and FLOAT64/FLOAT to a single category each, so
+    # leaving the BigQuery spelling in place makes INT64 -> int and FLOAT64 -> float compare as equal.
+    # Types nested inside ARRAY<...>/STRUCT<...> keep their BigQuery spelling, because there a match is
+    # only reached by round-tripping the Databricks type back to BigQuery.
```
m-abulazm: I would like to avoid such comments.

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
```diff
@@ -254,8 +254,27 @@ def _get_is_string(column_types_dict: dict[str, DataType], column_name: str) ->
 DataType_transform_mapping: dict[str, dict[str, list[partial[exp.Expression]]]] = {  # pylint: disable=invalid-name
     "universal": {"default": [partial(coalesce, default='_null_recon_', is_string=True), partial(trim)]},
     "bigquery": {
-        # TODO: add timestamps and numbers handling
+        # TODO: add numbers handling
+        # CAST(... AS STRING) does not render a timestamp the way Spark does: BigQuery appends a UTC
```
m-abulazm: same here. if it is necessary then something shorter please

src/databricks/labs/lakebridge/reconcile/schema_compare.py
```diff
@@ -111,17 +111,22 @@ def _validate_parsed_query(cls, source: Dialect, master: SchemaMatchResult) -> N
             * we don't use the databricks column name as it may have been renamed.
             * renaming is checked in the previous step to retrieve the databricks column.
         2. Parse both queries using sqlglot and convert both to the other dialect
-        3. Compare the converted queries with our original queries.
+        3. Compare the converted queries with our original queries. Both sides are normalized the same
```
m-abulazm: same

tests/integration/reconcile/test_schema_compare.py
```diff
@@ -354,19 +354,25 @@ def teradata_databricks_schema():
 
 
 def bigquery_databricks_schema():
+    # Source types are what BigQueryDataSource._SCHEMA_QUERY reports, not the raw INFORMATION_SCHEMA
```
m-abulazm: same

tests/unit/reconcile/connectors/test_bigquery.py
```diff
@@ -82,10 +82,35 @@ def test_get_schema_query_canonicalizes_types_within_family():
...
     # BIGNUMERIC has no exact Databricks equivalent, so it passes through rather than truncating.
     assert "BIGNUMERIC" not in schema_query
 
 
+@pytest.mark.parametrize(
+    "narrower_databricks_type, bigquery_type",
+    [
+        # Rationale for reporting Databricks type names from _SCHEMA_QUERY: a narrower target converts
+        # back to the very BigQuery type the column started as, so a schema compare that reads the
+        # source side as its BigQuery spelling accepts the lossy migration. int and smallint overflow
```
m-abulazm: I dont understand 
> int and smallint overflow INT64's range

m-abulazm (changes requested):
I am not sure about INT64 handling. BigQuery treats all integers as one type and all integer types are aliases to INT64 so I am not sure if scoping INT64 as BIGINT is the right call or not

take60:
Agreed on INT64 — every BigQuery integer type is an alias of it, so the declared width isn't recoverable and pinning it would flag correctly-migrated columns. I've dropped the type pinning entirely; INT64/FLOAT64/DATETIME are now reported as-is and added to `_APPROXIMATE_TYPES`, so `get_schema` names the affected columns instead of failing them.

Dropping the DATETIME pin also fixed a problem I'd introduced: the reported type doubles as the CAST target in the mismatch-sampling query, so `timestamp_ntz` produced `CAST(... AS timestamp_ntz)`, which BigQuery rejects. Constraint is now noted in the docstring.

Row hashes are unchanged — both spellings resolve to the same transform — so this PR is now just the row-hash fix plus the `schema_compare` whitespace fix.

Two things I'm deliberately leaving out of scope, happy to raise either as its own PR:
- `TIMESTAMP` hashes assume a UTC Spark session (the Databricks side follows `spark.sql.session.timeZone`, which we never pin). Pre-existing, not closed here.
- Nested types are still uneven: `ARRAY<NUMERIC>` is a false mismatch, and the warning matches STRUCT field names as well as types.

tests/unit/reconcile/connectors/test_bigquery.py
take60: You're right, that was backwards — `int`/`smallint` are narrower than INT64, not wider. The test carrying that wording is gone with the pinning it justified, and I've fixed the same claim in the docstring.

src/databricks/labs/lakebridge/reconcile/connectors/bigquery.py
take60: Trimmed.

src/databricks/labs/lakebridge/reconcile/query_builder/expression_generator.py
take60: Trimmed.

src/databricks/labs/lakebridge/reconcile/schema_compare.py
take60: Trimmed.

tests/integration/reconcile/test_schema_compare.py
take60: Trimmed.

take60:
@m-abulazm hello, please let me know if you have any other feedback. thanks!

m-abulazm (approved):

m-abulazm:
@take60 looks good. you need a second reviewer.
[#2601](https://github.com/databrickslabs/lakebridge/pull/2601) is the refactor you need to start comparing numeric types. it will also conflict with those changes but one file only

sundarshankar89 (approved):
LGTM

PR #2596, opened 2026-07-28
Add `profiler_run_metadata` to DuckDB extracts

### LINKED ISSUE #2544: [FEATURE]: Include Source System Metadata in Profiler DuckDB Output

Profiler

### Problem statement

We need a programmatic way to identify the source system that was used as input to generate a profiler output (e.g., Redshift, Snowflake, etc.).

Currently, this information has to be inferred either from the file name or by inspecting the DuckDB file for source-specific tables. Both approaches are brittle and error-prone, making them unsuitable for reliable automation.

The profiler output should include explicit source metadata so that consumers can determine the originating source system programmatically without relying on heuristics.

### Proposed Solution

Add a source-independent metadata table to the DuckDB file generated by the profiler.

At a minimum, this table should include the source system used as input (e.g., Redshift, Snowflake, etc.), allowing consumers to determine the originating source programmatically instead of relying on file names or source-specific tables.

Since this table is intended for generic metadata, it can also be extended in the future to capture additional useful information, such as the profiler version, generation timestamp, or other run-level metadata.

### Additional Context

_No response_

dgomez04:
### Summary
- Write a source-independent `profiler_run_metadata` table (wide single row) into every profiler DuckDB extract at the **end** of the run.
- Persist source, pipeline, environment, and outcome fields so consumers can identify the extract and how the run went without filename heuristics or log parsing.

The Profiler owns the write; `PipelineClass.execute()` only runs steps and returns results so the pipeline stays reusable. Failed DDL / `source_ddl` steps abort remaining extracts, but the metadata row is still written (best-effort) before the Profiler raises. `status` is derived from step results (`COMPLETE` / `COMPLETE_WITH_ABSENCES` / `FAILED`); `results` is the per-step JSON.

Closes #2544


m-abulazm (changes requested):
why is source system optional? that complicates downstream processing unnecessarily

tests/integration/assessments/test_pipeline.py
```diff
@@ -18,6 +18,7 @@
 
 _DB_FILE = "test_profiler.db"
 _CREDS_FILE = "test_creds.yml"
+_SOURCE_SYSTEM = "mssql"
```
bishwajit-db: This threads `source_system` through nine call sites but doesn't assert the table they now produce — `_write_run_metadata` is 0% covered, and it's the one consumer-facing contract in this PR.

No sandbox needed: with `steps=[]` and `executor=None`, `execute()` writes the table and returns `[]`, so this fits as a unit test beside `test_duckdb_helpers.py`. Worth asserting column names/order (catches a rename here rather than in someone else's pipeline), the single row, and that a second `execute()` replaces rather than appends.

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -77,6 +83,32 @@ def execute(self) -> list[StepExecutionResult]:
...
+                    source_system VARCHAR,
+                    lakebridge_version VARCHAR,
+                    generated_at TIMESTAMPTZ
+                )
+                """)
+            conn.execute(
+                f"INSERT INTO {PROFILER_RUN_METADATA_TABLE} VALUES (?, ?, ?)",
+                [self._source_system, lakebridge_version, generated_at],
+            )
+            conn.commit()
```
bishwajit-db: `DROP` + `CREATE` + `INSERT` isn't atomic — an interruption after the `DROP` leaves no metadata table at all, worse than the stale row. Also `conn.commit()` with no `conn.begin()`, unlike the other two write paths in this file (`:197`, `:272`).

One statement fixes both.

```suggestion
        with connect_to_profiler_db(self._db_path) as conn:
            conn.execute(
                f"""
                CREATE OR REPLACE TABLE {PROFILER_RUN_METADATA_TABLE} AS
                SELECT ?::VARCHAR AS source_system,
                       ?::VARCHAR AS lakebridge_version,
                       ?::TIMESTAMPTZ AS generated_at
                """,
                [self._source_system, lakebridge_version, generated_at],
            )
```

bishwajit-db (commented):
Add docs. `profiler_run_metadata` is the new contract for extract consumers, and it's source-independent, so `docs/lakebridge/docs/assessment/profiler/index.mdx` seems the right home:

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -18,6 +18,8 @@
 
 logger = logging.getLogger(__name__)
 
+PROFILER_RUN_METADATA_TABLE = "profiler_run_metadata"
```
bishwajit-db: Worth reserving this name now that the pipeline owns it. `Step._validate_name` accepts `profiler_run_metadata`, and a config using it fails mid-run.

Importing the constant into `profiler_config.py` is a circular import (confirmed). Moving it to `_constants.py` as `RESERVED_TABLE_NAMES` and re-exporting from here works; `Step._validate_name` then rejects it with a clear message.

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -44,15 +46,19 @@ def __init__(
         executor: DatabaseConnector | None,
         db_path: Path,
         cred_file_path: Path,
+        *,
+        source_system: str,
```
bishwajit-db: Which vocabulary can consumers rely on here? Via the CLI it's lowercased and checked against `PROFILER_SOURCE_SYSTEM`, but constructed directly it's arbitrary — the tests in this PR pass `redshift_provisioned`, which isn't in that list, so the two already diverge. Either normalize here or document that only CLI runs are canonical.

src/databricks/labs/lakebridge/assessments/profiler.py
```diff
@@ -82,7 +82,13 @@ def _execute(
...
             extractor = Profiler._setup_extractor(source_system, cred_file_path) if connector_required else None
             db_path = output_folder / make_profiler_db_filename(source_system)
-            result = PipelineClass(pipeline_config, extractor, db_path, cred_file_path).execute()
+            result = PipelineClass(
+                pipeline_config,
+                extractor,
+                db_path,
+                cred_file_path,
+                source_system=source_system,
+            ).execute()
```
bishwajit-db: Consider passing `self._variant` through too. For mssql it selects `single_db` vs `multi_db`, which changes which tables are in the extract, so `source_system=mssql` alone doesn't tell a consumer what to expect. `NULL` for sources without variants.

src/databricks/labs/lakebridge/assessments/pipeline.py
m-abulazm: I am also not sure why is this the only positional argument which would have made sense if it was optional

src/databricks/labs/lakebridge/assessments/profiler.py
m-abulazm: this comment got me thinking that we should also add: lakebridge version, auth method, exact pipeline (maybe the table names are just enough), python version maybe, os

tests/integration/assessments/test_pipeline.py
dgomez04: I have removed `_write_run_metadata` and wired it through `save_to_duckdb` - coverage is now as expected through `test_duckdb_helpers` and integration tests.

src/databricks/labs/lakebridge/assessments/pipeline.py
dgomez04: This was a standing problem across all other profilers, I've added transactions to `save_to_duckdb` which should take care of it for this and all other writes.

src/databricks/labs/lakebridge/assessments/pipeline.py
dgomez04: Added `RESERVED_TABLE_NAMES` and `PROFILER_RUN_METADATA_TABLE ` to avoid this from happening.

src/databricks/labs/lakebridge/assessments/pipeline.py
dgomez04: Added `_normalize_source_system ` and fixed the Redshift test. I've removed the `*` and kept it five-argument, keyword-only.

src/databricks/labs/lakebridge/assessments/profiler.py
dgomez04: I have keep the authentication method outside of this pull request as it require an additional lift.

bishwajit-db (approved):
LGTM

src/databricks/labs/lakebridge/assessments/run_metadata.py
```diff
@@ -0,0 +1,28 @@
...
+# from one extract to the next. Column order must match the field order above: the row is
+# inserted positionally.
+PROFILER_RUN_METADATA_SCHEMA = (
+    "source_system VARCHAR, "
+    "variant VARCHAR, "
+    "lakebridge_version VARCHAR, "
+    "python_version VARCHAR, "
+    "operating_system VARCHAR, "
+    "generated_at TIMESTAMPTZ"
+)
```
m-abulazm: as the comment says, the `PROFILER_RUN_METADATA_SCHEMA` column order has to line up with the `ProfilerRunMetadata` field order, because the metadata row gets inserted positionally (`INSERT ... SELECT *` in `_save_overwrite`, not `BY NAME`). 
                                                                                                                                                                                                                                                                                                                            
This can be brittle e.g. `source_system, variant, lakebridge_version`, etc. are all `VARCHAR`, so if someone later reorders the dataclass fields (or the schema string) without touching the other, DuckDB will quietly start writing values into the wrong columns. Or a new column is not added to the schema and the class only                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                            
I looked at flipping the insert to `INSERT ... BY NAME`, but that same code path is shared with the Synapse extracts, so that will be a bigger change. 

Given that, I'd suggest is deriving the schema string from the dataclass itself, so order can only ever come from one place:

```
from dataclasses import dataclass, field, fields                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                            
  @dataclass(frozen=True)                                                                                                                                                                                                                                                                                                   
  class ProfilerRunMetadata:                                                                                                                                                                                                                                                                                                
      source_system: str = field(metadata={"duckdb_type": "VARCHAR"})                                                                                                                                                                                                                                                       
      variant: str | None = field(metadata={"duckdb_type": "VARCHAR"})                                                                                                                                                                                                                                                      
      lakebridge_version: str = field(metadata={"duckdb_type": "VARCHAR"})                                                                                                                                                                                                                                                  
      python_version: str = field(metadata={"duckdb_type": "VARCHAR"})                                                                                                                                                                                                                                                      
      operating_system: str = field(metadata={"duckdb_type": "VARCHAR"})                                                                                                                                                                                                                                                    
      generated_at: datetime = field(metadata={"duckdb_type": "TIMESTAMPTZ"}


  PROFILER_RUN_METADATA_SCHEMA = ", ".join(                                                                                                                                                                                                                                                                                 
      f"{f.name} {f.metadata['duckdb_type']}" for f in fields(ProfilerRunMetadata)                                                                                                                                                                                                                                          
  )  
```

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -81,6 +110,28 @@ def execute(self) -> list[StepExecutionResult]:
 
         return execution_results
 
+    def _write_run_metadata(self) -> None:
```
m-abulazm: this write is unguarded on errors, while the steps themselves degrade to ABSENT/warnings. I don't think this is a real risk in practice: the metadata row is six small, fully-controlled values with an explicit schema, so there's basically no data-driven way for it to fail. The only realistic failure modes are environmental: the DuckDB file being locked, an unwritable path, or a full disk. And those would take down every extraction step too. I would wrap it in a try/except that logs a warning and moves on just to be future safe. 

please see my comment

src/databricks/labs/lakebridge/assessments/run_metadata.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime
+
+
+@dataclass(frozen=True)
+class ProfilerRunMetadata:
```
m-abulazm, line 15: I am still missing some more info: 
```
class ProfilerRunMetadata:
...
      pipeline_name: str                                                                                                                                                                                                                                                                                                    
      pipeline_version: str
      status: str          # COMPLETE | COMPLETE_WITH_ABSENCES | FAILED  (summary, so consumers needn't parse JSON)                                                                                                                                                                                                                        
      results: str         # JSON: [{"step_name", "status", "error_message"}, ...]    
```

Please see my comment

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -48,15 +56,36 @@ def __init__(
...
         self.config = config
         self.executor = executor
         self._db_path = db_path.expanduser()
         self._create_dir(self._db_path.parent)
         self._cred_file_path = cred_file_path
+        self._source_system = self._normalize_source_system(source_system)
+        self._variant = variant
+
+    @staticmethod
+    def _normalize_source_system(source_system: str) -> str:
```
m-abulazm: I am not a big fan of this, we have no direct callers and a bigger concern here is coupling the pipelines with profilers. ideally a pipeline is agnostic of profilers details and can be reused by anything needing a pipeline. This is a regression that we introduced in this PR. 

Please see my comment

m-abulazm (changes requested):
Two changes I'd want. First, the pipeline shouldn't own this write at all; it couples `PipelineClass` to profiler details (source systems, variants, the metadata table) when a pipeline should just run steps and hand back results so anything can reuse it. We have no direct callers other than the Profiler, so moving the write onto the Profiler is mostly a move/rename rather than a rewrite, and the `Profiler` already knows the source system, variant and db path (this is the same point I left on `_normalize_source_system`).                                                                                                             
                                                                                                                                                                                                                                                                                                                            
Second, the row is missing info I'd want when reading an extract — which pipeline produced it and how the run went:                                                                                                                                                                                                       
```                                                                                                                                                                                                                                                                                         
  pipeline_name: str                                                                                                                                                                                                                                                                                                        
  pipeline_version: str                                                                                                                                                                                                                                                                                                     
  status: str          # COMPLETE | COMPLETE_WITH_ABSENCES | FAILED (summary, so consumers needn't parse JSON)                                                                                                                                                                                                              
  results: str         # JSON: [{step_name, status, error_message}, ...]
```                                                                                                                                                                                                                                                    
  status is just derived from the step results (any ERROR -> FAILED, else any ABSENT -> COMPLETE_WITH_ABSENCES, else COMPLETE). So this has to be written at the end of the run, not the start. We also have to stop raising exceptions inside `execute()` and actually gets simpler as both raise blocks disappear:                                                                                                                                                                                                                                                            
```
  def execute(self) -> list[StepExecutionResult]:                                                                                                                                                                                                                                                                           
      results: list[StepExecutionResult] = []                                                                                                                                                                                                                                                                               
      for step in self.config.steps:                                                                                                                                                                                                                                                                                        
          result = self._process_step(step)                                                                                                                                                                                                                                                                                 
          results.append(result)                                                                                                                                                                                                                                                                                            
          self._log_step_result(result)                                                                                                                                                                                                                                                                                     
          if step.type in {"ddl", "source_ddl"} and result.status == StepExecutionStatus.ERROR:                                                                                                                                                                                                                             
              logger.error(f"Aborting run: DDL step {result.step_name} failed")                                                                                                                                                                                                                                             
              break   # remaining steps depend on this DDL; stop, but still report what ran                                                                                                                                                                                                                                 
      return results                                                                                                                                                           
```
  And because `_write_run_metadata` now runs at the end, I'd make the write best-effort, wrap it in a try/except that logs a warning and moves on. 

```
 def _write_run_metadata(self, db_path, pipeline_config, results):                                                                                                                                                                                                                                                         
      try:                                                                                                                                                                                                                                                                                                                  
          metadata = ProfilerRunMetadata(                                                                                                                                                                                                                                                                                   
              source_system=self._source_system,                                                                                                                                                                                                                                                                            
              variant=self._variant,                                                                                                                                                                                                                                                                                        
              pipeline_name=pipeline_config.name,                                                                                                                                                                                                                                                                           
              pipeline_version=pipeline_config.version,                                                                                                                                                                                                                                                                     
              lakebridge_version=lakebridge_version,                                                                                                                                                                                                                                                                        
              status=self._run_status(results),                                                                                                                                                                                                                                                                             
              results=json.dumps([                                                                                                                                                                                                                                                                                          
                  {"step_name": r.step_name, "status": r.status.value, "error_message": r.error_message}                                                                                                                                                                                                                    
                  for r in results                                                                                                                                                                                                                                                                                          
              ]),                                                                                                                                                                                                                                                                                                           
              generated_at=datetime.now(timezone.utc),                                                                                                                                                                                                                                                                      
          )                                                                                                                                                                                                                                                                                                                 
          save_to_duckdb(pd.DataFrame([asdict(metadata)]), PROFILER_RUN_METADATA_TABLE,                                                                                                                                                                                                                                     
                         str(db_path), schema=PROFILER_RUN_METADATA_SCHEMA)                                                                                                                                                                                                                                                 
      except Exception:                                                                                                                                                                                                                                                                                                     
          logger.warning(f"Failed to write {PROFILER_RUN_METADATA_TABLE}", exc_info=True)    
```

and the Profiler can use it and the new non-raising `execute()`:                                                                                                                                                                                                            
```
 results = PipelineClass(pipeline_config, extractor, db_path, cred_file_path).execute()                                                                                                                                                                                                                                    
  self._write_run_metadata(db_path, pipeline_config, results)   # best-effort inside; always has results                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                            
  failed = [r for r in results if r.status == StepExecutionStatus.ERROR]                                                                                                                                                                                                                                                    
  if failed:                                                                                                                                                                                                                                                                                                                
      raise RuntimeError(f"Pipeline failed for {source_system}: {', '.join(r.step_name for r in failed)}") 
```

Together with the ordering point on `PROFILER_RUN_METADATA_SCHEMA`, that's the shape I'd aim for.

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -56,36 +48,21 @@ def __init__(
...
-        """Casefold the label so every extract records one spelling of a given platform.
+    def execute(self) -> list[StepExecutionResult]:
+        """Run every configured step and return per-step outcomes.
 
-        The CLI only offers the canonical names, but a caller constructing the pipeline
-        directly can pass anything. An unknown name is a warning rather than an error:
-        it makes the extract harder to consume, not impossible to produce.
+        Does not raise on step failures: callers (the Profiler) decide how to
+        surface them after recording run metadata. A failed DDL / source_ddl
+        step aborts the remaining steps because later extracts depend on it.
```
```suggestion
        Does not raise on step failures: callers decide how to surface them.
        A failed DDL / source_ddl step aborts the remaining steps because later extracts depend on it.
```
decoupled of profilers. this includes comments

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -144,7 +88,7 @@ def _process_step(self, step: Step) -> StepExecutionResult:
             return StepExecutionResult(step_name=step.name, status=StepExecutionStatus.COMPLETE)
         except (RuntimeError, ConnectionError) as e:
             # Optional: warn + ABSENT (customer isn't failed; maintainers get the cause).
-            # Required: ERROR, which fails the run below.
+            # Required: ERROR — callers (the Profiler) fail the run after recording metadata.
```
```suggestion
            # Required: ERROR
```

src/databricks/labs/lakebridge/assessments/run_metadata.py
```diff
@@ -1,28 +1,32 @@
...
-PROFILER_RUN_METADATA_SCHEMA = (
-    "source_system VARCHAR, "
-    "variant VARCHAR, "
-    "lakebridge_version VARCHAR, "
-    "python_version VARCHAR, "
-    "operating_system VARCHAR, "
-    "generated_at TIMESTAMPTZ"
-)
+# from one extract to the next. Derived from the dataclass so column order can only come
+# from one place — the row is inserted positionally (`INSERT ... SELECT *`).
```
m-abulazm:
```suggestion
# Derived from the dataclass so column order can only come from one place
```

src/databricks/labs/lakebridge/assessments/run_metadata.py
```diff
@@ -0,0 +1,32 @@
...
+    operating_system: str = field(metadata={"duckdb_type": "VARCHAR"})
+    status: str = field(metadata={"duckdb_type": "VARCHAR"})
+    results: str = field(metadata={"duckdb_type": "VARCHAR"})
+    generated_at: datetime = field(metadata={"duckdb_type": "TIMESTAMPTZ"})
+
+
+# Declared rather than inferred: `variant` is NULL for sources without variants, and DuckDB
+# types an all-null column INTEGER, so inference would give the same column a different type
+# from one extract to the next. Derived from the dataclass so column order can only come
+# from one place — the row is inserted positionally (`INSERT ... SELECT *`).
```
m-abulazm:
```suggestion
# Derived from the dataclass so column order can only come from one place
```

tests/unit/assessment/test_run_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+    assert pipeline_name == "warehouse_profiler"
+    assert pipeline_version == "1.2.3"
+    assert recorded_version == lakebridge_version
+    assert status == ProfilerRunStatus.COMPLETE_WITH_ABSENCES.value
+    assert json.loads(results_json) == [
+        {"step_name": "inventory", "status": "COMPLETE", "error_message": None},
+        {"step_name": "optional_metric", "status": "ABSENT", "error_message": "object missing"},
+    ]
+    assert isinstance(generated_at, datetime)
+    assert generated_at.tzinfo is not None or generated_at.replace(tzinfo=timezone.utc)
```
m-abulazm:
```suggestion
    assert generated_at.tzinfo
```
testing timestamps is tricky, but still I am not sure why we need the timezone replace and the `or` clause. 

if you want to test it, here is an idea 
```
assert abs(datetime.now(timezone.utc) - generated_at) < timedelta(minutes=5)
```

m-abulazm (changes requested):
Nice work. Remaining comments are all nits, nothing structural. And please update PR description.

m-abulazm (approved):

PR #2597, opened 2026-07-28
[DO NOT MERGE] Require DuckDB DDL on every SQL profiler step

### LINKED ISSUE #2460: Add optional DDL to SQL pipeline step
The Oracle profiler PR (#2187) introduced a pattern where every logical extractor is split into two YAML steps that share the same `name:` — one with `type: ddl`, one with `type: sql`. This works today because `pipeline.py` dispatches on `step.type` first, but it causes three problems:

1. **Ambiguous failure reporting.** `PipelineClass.execute()` aggregates `StepExecutionResult` by `step.name`, so when the SQL half of `config_containers` fails, users see `Pipeline execution failed due to errors in steps: config_containers` with no signal whether DDL or SQL was the culprit.
2. **Implicit name=table contract.** `_execute_ddl_step` checks `if not self._table_exists(conn, step.name): conn.execute(ddl)`, so the DDL must `CREATE TABLE` whose name is exactly `step.name`. Nothing enforces this; an edit that changes the name in one place but not the other silently no-ops or fails at insert time.
3. **Future fragility.** Anyone refactoring `pipeline.py` to dedupe steps by `name` (perfectly reasonable assumption) breaks the YAML silently.

### Scope / acceptance criteria
- Fold DDL into the SQL step.** Single step with `setup_ddl:` and `extract_sql:` fields. Larger refactor, but conceptually cleaner since DDL+SQL are one extractor.
- Migrate the Oracle pipeline_config (PR #2187).
- Migrate the MSSQL pipeline_config once it converts to `type: sql` (see #2459).
- Document the decision so future profilers follow the same shape.

### References
- PR #2187 review thread on `oracle/pipeline_config.yml:11`.
- Related: #2459 (Convert MSSQL profiler from `type: python` to `type: sql`).

dgomez04:
## Summary
Foundation for required SQL `ddl_source`: DuckDB DDL is applied before inserts, a failed DuckDB DDL / `source_ddl` step aborts the run, and optional source-query failures stay `ABSENT`. The error model is a single `ERROR` status (no separate `ERROR_FATAL`).

This PR is **engine + mechanical YAML folds** for Oracle, MSSQL, and Teradata (existing `*_ddl.sql` already on `main`). It does **not** invent Snowflake, Redshift, or legacy Synapse DuckDB schemas.

## Do not merge to `main` yet
Stacked source PRs (each needs a live-system smoke):
- Snowflake: #2640
- Redshift: #2641
- Legacy Synapse: #2642

Merge those into this branch (`feat/2460-required-sql-ddl-source`) after live tests. The inventory gate (`test_profiler_config_inventory.py`) stays red until all three land. Then merge this PR to `main`.

## Review protocol
Review as engine + mechanical YAML, not as "ship all profilers." Python Synapse and BigQuery are unchanged.

Related: #2460


src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -29,6 +29,10 @@ class StepExecutionStatus(str, Enum):
     ABSENT = "ABSENT"
```
m-abulazm, line 31: lets use ERROR_FATAL and ERROR

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -59,7 +63,10 @@ def execute(self) -> list[StepExecutionResult]:
             execution_results.append(result)
             self._log_step_result(result)
 
-            if step.type in {"ddl", "source_ddl"} and result.status == StepExecutionStatus.ERROR:
+            # source_ddl failures and local DuckDB DDL failures abort immediately.
+            if result.status == StepExecutionStatus.ERROR and (
```
m-abulazm: with thew distinction in error result type, we can simplify this check

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -86,6 +93,9 @@ def _process_step(self, step: Step) -> StepExecutionResult:
         try:
             self._dispatch_step(step)
             return StepExecutionResult(step_name=step.name, status=StepExecutionStatus.COMPLETE)
+        except DuckDBDDLError as e:
+            # Local DuckDB DDL is our schema contract; never tolerate as ABSENT.
+            return StepExecutionResult(step_name=step.name, status=StepExecutionStatus.ERROR, error_message=str(e))
```
```suggestion
            return StepExecutionResult(step_name=step.name, status=StepExecutionStatus.FATAL_ERROR, error_message=str(e))
```

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -126,12 +136,40 @@ def _execute_sql_step(self, step: Step):
 
         logging.info(f"Executing query: {query}")
         result = self.executor.fetch(query)
-        self._save_to_db(result, step.name, str(step.mode))
+        self._save_to_db(result, step.name, step.mode)
+
+    def _apply_duckdb_ddl(self, step: Step) -> None:
+        """Create (or recreate) the DuckDB table declared by ``step.ddl_source``."""
+        if not step.ddl_source:
+            raise DuckDBDDLError(f"DDL execution failed: sql step '{step.name}' is missing ddl_source")
```
```suggestion
            raise DuckDBDDLError(f"DDL missing for sql step '{step.name}')
```

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -126,12 +136,40 @@ def _execute_sql_step(self, step: Step):
...
+                    conn.execute(f"DROP TABLE IF EXISTS {step.name}")
+                    conn.execute(ddl)
+                    logging.debug(f"Recreated table '{step.name}' from DDL")
+                elif not self._table_exists(conn, step.name):
+                    conn.execute(ddl)
+                    logging.debug(f"Created table '{step.name}' from DDL")
+                else:
+                    logging.debug(f"Table '{step.name}' already exists; skipping DDL in append mode")
+                conn.commit()
+        except DuckDBDDLError:
```
m-abulazm: this catch is redundant as the exception would never be raised here.

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -126,12 +136,40 @@ def _execute_sql_step(self, step: Step):
 
         logging.info(f"Executing query: {query}")
         result = self.executor.fetch(query)
-        self._save_to_db(result, step.name, str(step.mode))
+        self._save_to_db(result, step.name, step.mode)
+
+    def _apply_duckdb_ddl(self, step: Step) -> None:
```
m-abulazm: I need a comment on the method what is happening here and the intent behind it. the old execute_ddl was documented quite well

m-abulazm (changes requested):
we should revisit our pipeline flow. when to raise, when to stop execution eagerly and when to tolerate

src/databricks/labs/lakebridge/assessments/pipeline.py
dgomez04: Thanks. Removed

m-abulazm (approved):

src/databricks/labs/lakebridge/resources/assessments/legacy_synapse/requests_ddl.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE requests (
+    request_id VARCHAR,
+    session_id VARCHAR,
+    status VARCHAR,
+    submit_time TIMESTAMP,
+    start_time TIMESTAMP,
+    end_compile_time TIMESTAMP,
+    end_time TIMESTAMP,
+    total_elapsed_time BIGINT,
+    error_id VARCHAR,
```
asnare: Are we missing a column here?

src/databricks/labs/lakebridge/resources/assessments/redshift/serverless/10_cost_incurred_ddl.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE cost_incurred (
+    set_name VARCHAR,
+    seconds_charged DOUBLE,
```
asnare: I think we're missing a column here?

src/databricks/labs/lakebridge/resources/assessments/redshift/serverless/8_chart_concurrent_users_by_hour_ddl.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE chart_concurrent_users_by_hour (
+    set_name VARCHAR,
+    distinct_users VARCHAR,
```
asnare: I think this is the wrong type?

src/databricks/labs/lakebridge/resources/assessments/redshift/provisioned_multi_az/8_chart_concurrent_users_by_hour_ddl.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE chart_concurrent_users_by_hour (
+    set_name VARCHAR,
+    distinct_users VARCHAR,
```
asnare: I think this is the wrong type?

src/databricks/labs/lakebridge/resources/assessments/redshift/provisioned/8_chart_concurrent_users_by_hour_ddl.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE chart_concurrent_users_by_hour (
+    set_name VARCHAR,
+    distinct_users VARCHAR,
```
asnare: I think this is the wrong type?

src/databricks/labs/lakebridge/resources/assessments/snowflake/pipe_usage_ddl.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE pipe_usage (
+    PIPE_NAME VARCHAR,
+    START_TIME TIMESTAMP,
+    END_TIME TIMESTAMP,
+    CREDITS_USED DOUBLE,
+    BYTES_INSERTED DOUBLE,
+    FILES_INSERTED DOUBLE,
```
asnare: The types of these both seem a bit weird, why wouldn't they be an integer type of some sort?

src/databricks/labs/lakebridge/resources/assessments/snowflake/database_objects_ddl.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE database_objects (
+    DATABASE_NAME VARCHAR,
+    DATABASE_OWNER VARCHAR,
+    IS_TRANSIENT VARCHAR,
```
asnare: Might be a (pre-existing bug) here: the corresponding query is a union, on which this column name doesn't match one of the sets being concatenated. Can you check this and log an issue if this is the case? (I don't think the fix is trivial.)

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -126,12 +138,44 @@ def _execute_sql_step(self, step: Step):
...
+        ddl = read_text(Path(step.ddl_source)).strip()
+        logging.info(f"Applying DuckDB DDL for table '{step.name}' (mode={step.mode})")
+
+        try:
+            with duckdb.connect(self._db_path) as conn:
+                conn.begin()
+                if step.mode == "overwrite":
+                    conn.execute(f"DROP TABLE IF EXISTS {step.name}")
+                    conn.execute(ddl)
+                    logging.debug(f"Recreated table '{step.name}' from DDL")
```
asnare: This leads to a change in behaviour:
 * _Before_: a failed query would leave the existing data intact. This was intentional.
 * _After_: a failed query now leads to an empty table being present.

This is a regression against `main`, especially when combined with 'optional' steps which (I believe?) are supposed to leave existing data intact if they fail.

Does this seem accurate?

asnare (changes requested):
I understand the intent of this PR, and agree we should be moving in the direction of defining the schemas of the captured data: this is a contract with downstream tooling, and previously it was implicit in the CTAS nature of how data was saved.

That said, this PR introduces a new sharp edge: the SQL statements need to produce results that are compatible with the DDL defining where the data goes. Essentially each DDL/DQL pair needs to match, and when they don't the step will crash and the run will fail.

At present this failure mode is not detected by any tests: I had a look and by inspection found a few examples where the steps will fail at runtime. Because we don't have any tests covering this, if merged this would not have been caught until someone tried to use the profiler.

With this in mind:
 - This PR cannot be merged until the altered profilers have been tested against the actual target systems and we know they still run correctly.
 - We really need to have some automated tests that verify the queries actually run. At the moment we seem to be missing integration tests for most (all?) of the affected systems. Given this gap, we need a way to test that the ddl/sql pairs appear to match.

Issue #2537, opened 2026-06-26
BigQuery reconcile: handle decimal/timestamp hash parity via DataType_transform_mapping


take60:
Follow-up from review of #2527 (requested by @m-abulazm).

#2527 reverted the special-cased BigQuery decimal transform (`bigquery_decimal_transform` and the
conditional branch in `query_builder/base.py::_get_transform`) to avoid a second place that
conditionally rewrites types. The row/data hash transforms for BigQuery should instead be handled
properly through `DataType_transform_mapping`.

## Problem
For row/data reconciliation, columns are cast to STRING and hashed. BigQuery `CAST(NUMERIC AS STRING)`
strips trailing zeros (`23.7989`) while Spark `CAST(DECIMAL(p,s) AS STRING)` pads to the column scale
(`23.798900000`), so identical decimal values hash differently → false mismatches. TIMESTAMP/FLOAT64
string formatting parity needs similar handling.

## Proposed
Extend the `"bigquery"` entry in `DataType_transform_mapping`
(`reconcile/query_builder/expression_generator.py`) to emit scale-aware decimal formatting (and
timestamp/number handling) — no conditional special-casing in `_get_transform`. @m-abulazm offered to
pair on this.

Marker left in code: `# TODO: add timestamps and numbers handling`.

PR #2601, opened 2026-07-30
Extract column transformations into a ColumnTransformer service

m-abulazm:
### What does this PR do?
The reconcile query builders resolved column transforms inline: each builder reached into DataType_transform_mapping for per-type rendering and carried its own user-transform logic on the QueryBuilder base. That resolution was single-sided (each layer only knew its own schema and dialect), which leaves no seam for cross-engine rendering.

Introduce ColumnTransformer, an abstract service that holds both the source and target sides and renders a column with knowledge of its counterpart. Builders depend only on the interface (transform / transform_user) and receive a RuleBasedColumnTransformer at construction; Reconciliation builds one per table and passes it to the hash, sampling, aggregate, and threshold builders.

No behavior change: query output is identical; this is a structural refactor that gives cross-engine numeric rendering a place to live.
- Move the per-type mapping and default/user transform logic out of QueryBuilder and expression_generator into column_transformer.py.
- transform/transform_user return RenderedColumn, carrying the declared cast_type so the sampler casts reconstructed literal rows without reaching into transformer internals (drops the user_transformed_columns coupling).
- Resolve the Schema.column_name backwards-compat field, repointing reads to ansi_normalized_column_name.
### Linked issues

Progresses #1202, #2537

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


ganeshdogiparthi-db (approved):
LGTM

src/databricks/labs/lakebridge/reconcile/recon_config.py
```diff
@@ -308,10 +308,8 @@ def get_filter(self, layer: str) -> str | None:
 
 @dataclass
```
bishwajit-db, line 309: Schema is now a positional/field-shape break. Dropping column_name and reordering to (ansi_normalized_column_name, data_type, source_normalized_column_name) is intended (the field carried a "TODO remove" note), and all in-repo constructions are updated. Since Schema lives in recon_config, any external/downstream code constructing it positionally will break silently contradicting the stated “no behavior change”. Probably acceptable, but worth a line in the PR description or changelog.

@m-abulazm

bishwajit-db (approved):
LGTM

PR #2602, opened 2026-07-31
[FIX] Redshift concurrent-users metrics collapsing across days

dgomez04:
### Summary
- Fixes `chart_concurrent_users_by_hour` and `rs_avg_concurrent_users` so distinct users are counted within a single calendar day before being reduced to an hour-of-day (`0–23`) profile.
- Chart reports the busiest day for each hour (`max`); the KPI averages those per-day/hour counts.

### Root Cause Analysis
`query_view` exposes the full retained query history (~7 days from `sys_query_history` / previously `stl_query`) with **no `start_time` filter**. Both metrics then grouped only by `date_part(hour, start_time)` and ran `count(distinct user_id)`.

That collapses every day into 24 hour-of-day buckets. Distinct users who ran at the same clock hour on different days are unioned into one count — for example, 5 users at 10:00 on day 1 plus 5 different users at 10:00 on day 2 reported **10**, not a peak of **5**.

The chart name and KPI imply a diurnal concurrency profile (users active in that hour on a typical/busy day), not the union of all users ever seen at that clock hour across the retention window. True session-overlap concurrency is out of scope; this fix preserves the existing “distinct users in an hour” semantics and corrects the cross-day aggregation.


ysmx-github (commented):
Hi Diego, I will not approve this request. we don't need it by day, it is all average by the hour across multiple days, it shows the peak hours and nothing else. Those are unimportant metrics, but let's keep them as they are

dgomez04:
> Hi Diego, I will not approve this request. we don't need it by day, it is all average by the hour across multiple days, it shows the peak hours and nothing else. Those are unimportant metrics, but let's keep them as they are

Hi Yuri, as the pull request description states, the current state is not averaging by the hour across multiple days. 

For example, if there are 5 users at 10:00 on day 1 plus 5 different users at 10:00 on day 2, the queries reported 10, not a peak of 5.

Please re-consider your stance.

ysmx-github (approved):
Diego, I really don't care about this query, so I reconsider my block and approve this PR as long as the final output of the query has the same structure as before

src/databricks/labs/lakebridge/resources/assessments/redshift/sql/4_rs_avg_concurrent_users.sql
```diff
@@ -1,10 +1,12 @@
 --4
-with base as (select count(distinct user_id) as distinct_users
-                    ,date_part(hour,start_time) as hour
-                from query_view 
-                group by 2
-                order by 2)
+-- Distinct users are counted per calendar day and hour so a multi-day query history is not
+-- collapsed into 24 hour-of-day buckets; the average is then taken over those buckets.
+with base as (select date_trunc('day', start_time) as day
```
asnare: Nit:
```suggestion
with base as (select start_time::date as day
```
(Slightly more efficient, and conveys intent better.)

src/databricks/labs/lakebridge/resources/assessments/redshift/sql/8_chart_concurrent_users_by_hour.sql
```diff
@@ -1,8 +1,13 @@
 -- 8 - Chart: concurrent_users_by_hour
+with daily_hourly_users as (select date_trunc('day', start_time) as day
```
asnare: Nit:

```suggestion
with daily_hourly_users as (select start_time::date as day
```
(Likewise)

src/databricks/labs/lakebridge/resources/assessments/redshift/sql/8_chart_concurrent_users_by_hour.sql
```diff
@@ -1,8 +1,13 @@
...
+                                  ,count(distinct user_id) as distinct_users
+                              from query_view
+                             group by 1, 2)
 select 'chart_concurrent_users_by_hour' set_name
-      ,count(distinct user_id) as distinct_users
-      ,date_part(hour,start_time) as hour
-from query_view 
-group by 1,3
-order by 3
+      ,max(distinct_users) as distinct_users
```
asnare, line 11: This makes the query return the _peak_ number of distinct users across the given hour, not the average. Subtle difference.

Given the title doesn't really hint at this, I think a comment is needed above.

src/databricks/labs/lakebridge/resources/assessments/redshift/sql/4_rs_avg_concurrent_users.sql
```diff
@@ -1,10 +1,12 @@
...
+                    ,date_part(hour, start_time) as hour
+                    ,count(distinct user_id) as distinct_users
+                from query_view
+                group by 1, 2)
 select 'rs_avg_concurrent_users' set_name
-      ,round(avg(distinct_users),0)::double precision avg_concurrent_users 
+      ,round(avg(distinct_users),0)::double precision avg_concurrent_users
```
asnare, line 10: Feels weird rounding but leaving the column type as double, but whatever.

asnare (approved):
@dgomez04: Has this been tested against an actual instance?

A few nits on the queries, but they're style rather than correctness.

For the future, we should indeed avoid aggregating in this way during collection: we should just capture the distinct users for each hour, and then both forms of aggregation are post-processing steps on the same data.

dgomez04:
> @dgomez04: Has this been tested against an actual instance?
> 
> A few nits on the queries, but they're style rather than correctness.
> 
> For the future, we should indeed avoid aggregating in this way during collection: we should just capture the distinct users for each hour, and then both forms of aggregation are post-processing steps on the same data.

Yes - this has already been tested manually by Yuri.

src/databricks/labs/lakebridge/resources/assessments/redshift/sql/8_chart_concurrent_users_by_hour.sql
dgomez04: Good catch — added a comment above clarifying that the outer `max()` returns the peak (busiest day) per hour-of-day, not the average across days.

dgomez04: Added

PR #2610, opened 2026-08-06
Stream Snowflake query results as Arrow into DuckDB to reduce out-of-memory errors

### LINKED ISSUE #2607: [BUG]: Snowflake profiler exhausts client memory on large accounts — extract steps buffer the full result set and materialize it three times

### Category of Bug / Issue

Profiler bug

### Current Behavior

On a high-volume Snowflake account, `execute-database-profiler` exhausts client memory and the host process dies. On a VDI this took the whole session host down rather than producing a Python traceback.

The Snowflake extract has no streaming anywhere in its path. A single step's result set is fully buffered and then materialized two more times before it reaches disk:

1. **Unbounded query.** `query_history.sql` selects every query in `SNOWFLAKE.ACCOUNT_USAGE.QUERY_HISTORY` over a 90-day window — ~30 columns, no `LIMIT`, plus an `ORDER BY START_TIME DESC`.

2. **Full buffer into Python objects.** `_BaseConnector.fetch` calls `result.fetchall()` — no `fetchmany`, no cursor iteration, no server-side paging:

   ```python
   # src/databricks/labs/lakebridge/connections/database_manager.py
   with Session(self.engine) as session, session.begin():
       result = session.execute(text(query))
       return FetchResult(list(result.keys()), result.fetchall())
   ```

3. **Second copy into pandas.** `FetchResult.to_df()` copies the whole thing again:

   ```python
   def to_df(self) -> pd.DataFrame:
       return pd.DataFrame.from_records(self.rows, columns=list(self.columns))
   ```

4. **Third materialization into DuckDB.** `PipelineClass._save_to_db` then runs `CREATE TABLE {step} AS SELECT * FROM _result_frame` (or `INSERT INTO ... SELECT *`) against the frame.

Steps 2–4 are all live simultaneously, so peak RSS is roughly 2–3x the decoded result set. `ACCOUNT_USAGE.QUERY_HISTORY` on a busy account is easily tens of millions of rows over 90 days, and as Python row objects that is many GB before pandas takes its copy.

This is not specific to Snowflake — `_BaseConnector.fetch` is shared by the Oracle and Teradata connectors, and the MSSQL and Redshift connectors have their own `fetchall()` implementations with the same shape. Snowflake just hits it first because `ACCOUNT_USAGE` is the largest default extract.

### Expected Behavior

A profiler extract should complete in bounded memory regardless of source-account size. Memory use should be a function of batch size, not result-set size.

### Steps To Reproduce

1. Configure the Snowflake profiler against an account with a high query volume (enough that 90 days of `ACCOUNT_USAGE.QUERY_HISTORY` is in the tens of millions of rows).
2. Run `databricks labs lakebridge execute-database-profiler`.
3. The `query_history` step grows unboundedly in RSS and the process is killed by the OS / the host runs out of memory.

```shell
No Python traceback was captured — the host ran out of memory and the session
terminated before the interpreter could unwind.
```

### Proposed approach

Sketching this rather than prescribing it, since it touches the shared pipeline framework:

- Add a batched fetch path to `DatabaseConnector` (e.g. `fetch_batches(query, size) -> Iterator[FetchResult]`) backed by `cursor.fetchmany()` / SQLAlchemy's `yield_per`, keeping the existing `fetch()` for small steps.
- Have `_execute_sql_step` / `_save_to_db` consume batches and append to DuckDB per batch inside a single transaction, so nothing larger than one batch is resident.
- Prefer `CREATE TABLE ... AS SELECT` from the first batch, then `INSERT` subsequent batches, dropping the intermediate pandas copy where possible — DuckDB can ingest Arrow or record batches directly.
- Consider a configurable batch size with a conservative default.

Related: #2486 (configurable lookback window). That issue would reduce how often this is hit, but not fix it — the failure is a function of rows returned, not of the window, so a large enough account fails at 30 days too. The two are complementary.

### Logs Confirmation

### Sample Query

```shell
-- src/databricks/labs/lakebridge/resources/assessments/snowflake/query_history.sql
SELECT QUERY_ID, QUERY_TYPE, DATABASE_NAME, ... -- ~30 columns
FROM SNOWFLAKE.ACCOUNT_USAGE.QUERY_HISTORY
WHERE START_TIME >= DATEADD('day', -90, CURRENT_TIMESTAMP())
ORDER BY START_TIME DESC;
```

### Operating System

Windows

### Version

latest via Databricks CLI

### LINKED ISSUE #2608: [BUG]: End-user installs resolve dependencies unconstrained — uv.lock does not apply to `databricks labs install`, pyarrow arrives unpinned

### Category of Bug / Issue

Other

### Current Behavior

End-user installs resolve the full dependency tree live from PyPI with no lock file and no constraints, so two users installing the same Lakebridge version on different days can get materially different dependency sets. We hit a transitive-dependency version break on two separate machines during a single customer engagement and had to work around it both times with a hand-written `PIP_CONSTRAINT` file.

The mechanism:

- `databricks labs install lakebridge` unpacks the release zipball to `~/.databricks/labs/lakebridge/lib` and runs `<venv>/python -m pip install .` against it. (Visible in the debug log attached to #2221.)
- That resolves from `pyproject.toml` only. The repo's `uv.lock` is not consulted, and `.build-constraints.txt` covers build-backend packages (hatchling and friends), not runtime dependencies.
- Several runtime dependencies are declared with loose or absent upper bounds, so pip takes whatever is newest at install time.

**pyarrow is the concrete case.** It isn't a declared dependency at all — it arrives transitively and completely unpinned via `db-dtypes>=1.0.0,<2` (pulled in for the BigQuery profiler), which itself requires `pyarrow` with no ceiling. So every install picks up whatever pyarrow released most recently. `uv.lock` does pin pyarrow for development, which means CI is green against a version that no end user is guaranteed to get.

`snowflake-connector-python>=4.5.0` and `snowflake-sqlalchemy>=1.9.0` have the same shape — no upper bound.

The general problem is more important than the specific package: any transitive dependency can ship a breaking release and turn a previously-working `databricks labs install lakebridge` into a broken one, with no change on our side and nothing in the release notes.

### Expected Behavior

Installing a given Lakebridge version should produce a reproducible, tested dependency set. What CI validates should be what users receive.

### Steps To Reproduce

1. `databricks labs install lakebridge` on a clean machine.
2. Inspect the resulting venv: `~/.databricks/labs/lakebridge/state/venv/…/pip list`.
3. Compare the resolved pyarrow version against the one pinned in `uv.lock`. They are only coincidentally the same, and diverge as soon as a new pyarrow ships.

```shell
# From the --debug log in #2221, showing the resolution path:
DEBUG Installing Python dependencies for: C:\Users\...\.databricks\labs\lakebridge\lib
DEBUG running: C:\Users\...\.databricks\labs\lakebridge\state\venv\Scripts\python.exe -m pip install .
```

### Proposed approach

A few options, roughly in increasing order of effort:

- **Add upper bounds** to the loosest runtime dependencies (`db-dtypes`, `snowflake-connector-python`, `snowflake-sqlalchemy`), and declare `pyarrow` explicitly with a tested range rather than inheriting it transitively. Cheapest fix, catches the common case.
- **Ship a constraints file with the release** and have the install path pass `PIP_CONSTRAINT` / `--constraint`, generated from `uv.lock` at release time. Reproducible without over-constraining the declared ranges.
- **Install from the lock** where the labs CLI allows it.

Happy to be told which of these fits the project's release process — the first one alone would have prevented what we hit.

### Logs Confirmation

### Sample Query

```shell

```

### Operating System

Windows

### Version

latest via Databricks CLI

dgomez04:
### Issue

Customers have been hitting overnight hangs with no visbility and **out-of-memory errors** during the `query_history` step, even after Snowflake shows the queries have succeeded. 

After some investigation, we have pointed the root causes to: 
1. **`fetchall()` as a single operation** — the full 90-day result is downloaded and materialized into a list before any `duckdb` write, with no progress boundaries.
2. **Memory duplication** — that list is then copied again via `FetchResult.to_df()` into a full `pandas` DataFrame before insert, so peak memory usage is often tens of gigabytes on high volume accounts. 

### Solution

Stream Snowflake results as **Arrow tables** and write them straight into `duckdb`:

- Use Snowflake’s native `fetch_arrow_batches` so column types come from result metadata, not per-batch `dtype` inference.
- Feed each Arrow batch into `duckdb` via `register(...)` + `SELECT *` on a **single connection** held for the whole step, with a **commit per chunk** to bound uncommitted memory.
- Capability is an explicit typed method (`supports_streaming()`), not `hasattr` reflection. Scope remains Snowflake-only for now.

Other profiler backends stay on `fetch` + `_save_to_db`. Empty-result behavior is unchanged: an all-empty stream writes nothing and leaves any existing table untouched.

### Additions

- `DatabaseConnector.supports_streaming()` / `stream()` — default off; Snowflake overrides and yields `pyarrow.Table` via `cursor.fetch_arrow_batches(force_microsecond_precision=True)`.
- `save_to_duckdb_conn(...)` — write one pandas or Arrow frame on an already-open connection; `save_to_duckdb` becomes the open-connection wrapper for existing callers.
- Profiler `_stream_sql_step` — one DuckDB connection for the step; first non-empty batch uses overwrite when configured, later batches append.
- Direct `pyarrow>=14.0.0` dependency (no longer only transitive).

### Test

- [x] Existing unit tests still pass (`test_snowflake_utils`, `test_database_manager`, `test_duckdb_helpers`) — no Snowflake live connector tests exist today to switch
- [ ] Run `execute-database-profiler` on a high-volume Snowflake account and confirm extract completes without out of memory.


asnare (approved):
I made a few tweaks on top of what was here, and in particular changed the version of pyarrow that we use to something newer with type stubs.

@dgomez04: Have a look at the tweaks to make sure you're happy with them.

src/databricks/labs/lakebridge/assessments/pipeline.py
```diff
@@ -125,9 +129,28 @@ def _execute_sql_step(self, step: Step):
...
+
+        first = True
+        with connect_to_profiler_db(self._db_path) as conn:
+            logging.info(f"Starting query: {query}")
+            for batch in self.executor.stream(query):
+                if batch.num_rows == 0:
+                    continue
+                write_mode: SaveMode = "overwrite" if first and step.mode == "overwrite" else "append"
+                save_to_duckdb_conn(conn, batch, step.name, mode=write_mode)
+                first = False
```
m-abulazm, line 153: Nit: This can be simplified, not blocking though
```suggestion
        write_mode = step.mode
        with connect_to_profiler_db(self._db_path) as conn:
            logging.info(f"Starting query: {query}")
            for batch in self.executor.stream(query):
                if batch.num_rows == 0:
                    continue
                save_to_duckdb_conn(conn, batch, step.name, mode=write_mode)
                write_mode = "append"
```

m-abulazm (approved):
LGTM. not blocking on a nit

asnare (approved):

PR #2618, opened 2026-08-10
Key pair authentication support for the Snowflake Profiler

dgomez04:
## Summary
- Adds key-pair authentication for the Snowflake profiler in response to customer feature requests.
- Declares `cryptography` as an explicit dependency. We import it directly to load PEM PKCS8 keys, so it should not rely only on a transitive install. The floor is set to `>=46.0.5` to match the minimum already required by `snowflake-connector-python`, avoiding a redundant or conflicting lower bound while still allowing newer releases (including 50.x) to resolve.

## Test plan
- [ ] `configure-database-profiler` → snowflake → `pat` still writes and connects as before
- [ ] Same flow with `key_pair` writes `private_key_path` (+ optional passphrase when confirmed)
- [ ] `test-profiler-connection` / `execute-database-profiler` succeeds against a key-pair-enabled Snowflake user
- [ ] Bad key path / wrong passphrase produce clear `ConnectionError`s
- [ ] Existing unit tests for Snowflake configurator / connector still pass


docs/lakebridge/docs/assessment/profiler/snowflake.mdx
```diff
@@ -27,13 +32,27 @@ You'll need:
...
+Existing credentials files without `auth_type` continue to use PAT authentication (`Pat`).
+
 ## Configure and run the profiler
 
 ```bash
 databricks labs lakebridge configure-database-profiler
 ```
 
-Pick **snowflake** when prompted, then paste the connection details and PAT.
+Pick **snowflake** when prompted, then choose **Pat** or **KeyPair** and enter the connection details.
```
asnare: Spelling:
```suggestion
Pick **snowflake** when prompted, then choose **PAT** or **Key-Pair** and enter the connection details.
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -119,19 +120,27 @@ def _connect(self) -> Engine:
...
+        if resolved.password is not None:
+            url_kwargs["password"] = resolved.password
+
+        connect_args: dict[str, Any] = {}
+        if resolved.private_key is not None:
+            connect_args["private_key"] = resolved.private_key
+
+        if connect_args:
+            return create_engine(URL.create(**url_kwargs), connect_args=connect_args)
+        return create_engine(URL.create(**url_kwargs))
```
asnare: Let's see if we can help the linters a bit… maybe:
```suggestion
        snowflake_url = URL.create(
            drivername="snowflake",
            host=account,
            database=f"{database}/{schema}",
            username=user,
            password=resolved.password,
            query={"warehouse": warehouse, "role": role},
        )

        connect_args = {"private_key": resolved.private_key} if resolved.private_key else {}
        return create_engine(snowflake_url, connect_args=connect_args)
```

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -16,6 +20,35 @@ def is_valid_snowflake_account(account_identifier: str) -> bool:
     return bool(account_identifier) and bool(_ACCOUNT_IDENTIFIER.match(account_identifier))
 
 
+def load_snowflake_private_key(key_path: Path, passphrase: str | None = None) -> bytes:
```
asnare, line 23: This needs some unit tests. They should:
 - Generate a key-pair.
 - For the key-pair, parameterize the testing of the password to cover:
    - Unencrypted (`None`)
    - Encrypted (normal password)
    - Edge-case: encrypted with an empty password (`""`)
 - The unhappy paths need to be checked:
    - Wrong password.
    - Missing password.
    - Path missing.
    - Garbage file (not a PKCS#8 file).

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -16,6 +20,35 @@ def is_valid_snowflake_account(account_identifier: str) -> bool:
...
+    password = passphrase.encode() if passphrase else None
+    try:
+        private_key = serialization.load_pem_private_key(
+            key_bytes,
+            password=password,
+            backend=default_backend(),
+        )
+    except (TypeError, ValueError) as e:
+        hint = "check the passphrase" if passphrase else "the key may be encrypted (provide a passphrase)"
+        raise ConnectionError(f"Invalid Snowflake private key at {key_path}: {e} ({hint})") from e
```
asnare: Also need to handle (differently) `UnsupportedAlgorithm` I think?

asnare: The language here isn't quite right: the private key might be valid, but we have the wrong password. These are different problems, and need to be resolved by the user differently.

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -16,6 +20,35 @@ def is_valid_snowflake_account(account_identifier: str) -> bool:
...
+        key_bytes = key_path.read_bytes()
+    except OSError as e:
+        raise ConnectionError(f"Unable to read Snowflake private key at {key_path}: {e}") from e
+
+    password = passphrase.encode() if passphrase else None
+    try:
+        private_key = serialization.load_pem_private_key(
+            key_bytes,
+            password=password,
+            backend=default_backend(),
```
asnare: Deprecated/no-op argument, I think?

src/databricks/labs/lakebridge/connections/snowflake_auth.py
asnare, line 1: We don't seem to have any test coverage of this file?

tests/unit/assessment/test_snowflake_configurator.py
```diff
@@ -47,6 +51,7 @@ def test_local_vault_stores_pat_verbatim(tmp_path):
         "secret_vault_type": "local",
         "snowflake": {
             "connection": {
+                "auth_type": "Pat",
```
asnare: … will need to be:
```suggestion
                "auth_type": "pat",
```

tests/unit/assessment/test_snowflake_configurator.py
asnare, line 1: There need to be more tests in here, we need to have:
 - `auth_type`: `pat`
 - `auth_type`: `key_pair` (no password)
 - `auth_type`: `key_pair` (a password)
 - `auth_type`: `key_pair` (zero-length password `""`)

src/databricks/labs/lakebridge/connections/snowflake_auth.py
```diff
@@ -0,0 +1,86 @@
...
+        if not key_path_value:
+            raise ConnectionError(f"{cls.__name__} requires 'private_key_path' in credentials")
+        passphrase = config.get("private_key_passphrase")
+        passphrase_str = str(passphrase) if passphrase else None
+        private_key = load_snowflake_private_key(Path(str(key_path_value)), passphrase_str)
+        return ResolvedSnowflakeCredentials(private_key=private_key)
+
+
+# User-selectable auth methods, in the order they appear in the configurator prompt.
+AUTH_CHOICES: list[type[SnowflakeAuth]] = [
```
asnare: Please use `Sequence`

src/databricks/labs/lakebridge/connections/snowflake_auth.py
```diff
@@ -0,0 +1,86 @@
...
+    Pat,
+    KeyPair,
+]
+
+_AUTH_REGISTRY: dict[str, type[SnowflakeAuth]] = {cls.__name__: cls for cls in AUTH_CHOICES}
```
asnare: Please use `Mapping`

src/databricks/labs/lakebridge/connections/snowflake_auth.py
```diff
@@ -0,0 +1,86 @@
...
+
+# Accept snake_case values written by the first draft of this feature, plus missing auth_type.
+_AUTH_ALIASES: dict[str, str] = {
```
asnare: Please use `Mapping`.

src/databricks/labs/lakebridge/connections/snowflake_auth.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def resolve_credentials(cls, config: JsonObject) -> ResolvedSnowflakeCredentials:
+        pat = config.get("pat")
+        if not pat:
+            raise ConnectionError(f"{cls.__name__} requires a non-empty 'pat' in credentials")
+        return ResolvedSnowflakeCredentials(password=str(pat))
+
+
+class KeyPair:
```
asnare: These need to have a common base-class, and the base-class should declare two abstract class properties:
 - `auth_type`: `pat` and `key_pair` in the subclasses.
 - `label`: `PAT` and `Key-Pair`

The registry and storage can then use `auth_type` instead of name, and the user-prompts during configuration should use the labels as the choices (but storing auth_type).

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -50,6 +51,54 @@ def _prompt_mssql_auth_credentials(prompts: Prompts, auth_type: str) -> dict[str
     return {}
 
 
+def _prompt_snowflake_auth_credentials(prompts: Prompts, auth_type: str, secret_vault_type: str) -> dict[str, str]:
+    """Prompt for Snowflake auth-specific fields for the chosen strategy.
+
+    Returns a partial dict to merge into the nested ``connection`` block. Shared
+    connection fields (account, user, warehouse, …) are prompted separately.
+    """
+    if auth_type == "Pat":
```
asnare: This (and the check below) will need to be updated to use `.label`.

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -370,22 +419,13 @@ def _configure_credentials(self) -> None:
...
-        )
-
-        # In env mode the stored value is the name of an environment variable that
-        # EnvGetter resolves at runtime, not the token itself, so prompt accordingly.
-        if secret_vault_type == "env":
-            pat = self.prompts.question("Enter the environment variable name holding the PAT")
-        else:
-            pat = self.prompts.password("Enter Programmatic Access Token (PAT)")
+        auth_choices = [cls.__name__ for cls in SNOWFLAKE_AUTH_CHOICES]
+        auth_type = self.prompts.choice("Select authentication method", auth_choices, sort=False)
```
asnare: This will need to be updated to something like:
```suggestion
        auth_choices = {cls.label: cls.auth_type for cls in SNOWFLAKE_AUTH_CHOICES}
        assert len(auth_choices) == len(SNOWFLAKE_AUTH_CHOICES), "Label/auth-type collision."
        auth_label = self.prompts.choice("Select authentication method", list(auth_choices.keys()), sort=True)
        auth_type = auth_choices[auth_label]
```

src/databricks/labs/lakebridge/connections/database_manager.py
asnare: For testing, I think it's worth decomposing this method into two pieces:
1. Build the URL and connect arguments.
2. Connect with them.

This would allow the URL-building and connect-argument logic to be unit-tested. In particular we now have a few branches and it's worth ensuring they're covered. (For example, with key-pair we should get a password-less URL and connect-args containing the decrypted private-key.)

src/databricks/labs/lakebridge/connections/snowflake_auth.py
```diff
@@ -0,0 +1,86 @@
...
+]
+
+_AUTH_REGISTRY: dict[str, type[SnowflakeAuth]] = {cls.__name__: cls for cls in AUTH_CHOICES}
+
+# Accept snake_case values written by the first draft of this feature, plus missing auth_type.
+_AUTH_ALIASES: dict[str, str] = {
+    "pat": Pat.__name__,
+    "key_pair": KeyPair.__name__,
+}
+
```
asnare: Do we need to keep this when we merge to `main`? I'd like to ditch the complexity?

asnare (changes requested):
I know this is a draft, but I've done a review anyway to catch a few things early before we iterate too much further. There are two areas that need to be looked at:
 - Test coverage. I've added inline suggestions.
 - The registry pattern being used. I know this one is based on another, but there are some sharp edges and I think we can do better. In particular we need to distinguish between:
   - The labels we present to users and document (`PAT`, `Key-Pair`.)
   - The low-level identities we store in configuration and pass around. (These are `snake_case` by convention: `pat`, `key_pair`.)
   In dynamic languages like Python it's tempting to tie these to class names, but this leads to problems down the track with maintenance and consistency, plus it's hard to automate.

   I've outlined how this can be done more cleanly. (It can be improved further using a better enum/registration pattern, but that's a heavier lift and can be retrofitted so long as the external contracts are the same. For now updating `mssql` to match can also be deferred, pending this.)

src/databricks/labs/lakebridge/connections/snowflake_utils.py
dgomez04: Covered in `test_snowflake_utils.py`. One caveat on the empt password case (`""`),`cryptography`cannot succesfully decrypt a key encrypted with an empty passphrase. `password=b""` is treated like a missing password on encrypted keys, and `BestAvailableEncryption(b"")` refuses to create such keys.

asnare: Ouch. It's a common thing for people to just hit <enter> as their password… 🤔

docs/lakebridge/docs/assessment/profiler/snowflake.mdx
```diff
@@ -8,13 +8,18 @@ import Admonition from '@theme/Admonition';
 
 ## Prerequisites
 
-- A Snowflake user with permission to create Personal Access Tokens (PATs) and to query the `SNOWFLAKE.ACCOUNT_USAGE` schema (e.g., the `ACCOUNTADMIN` role or a role granted `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`).
+- A Snowflake user that can authenticate with either a **Programmatic Access Token (PAT)** or **key-pair (JWT)** authentication, and that can query the `SNOWFLAKE.ACCOUNT_USAGE` schema (e.g., the `ACCOUNTADMIN` role or a role granted `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`).
```
asnare: Nit, key-pair auth for Snowflake is unrelated to JWT:
```suggestion
- A Snowflake user that can authenticate with either a **Programmatic Access Token (PAT)** or **key-pair** authentication, and that can query the `SNOWFLAKE.ACCOUNT_USAGE` schema (e.g., the `ACCOUNTADMIN` role or a role granted `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`).
```

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -52,6 +59,55 @@ def _prompt_mssql_auth_credentials(prompts: Prompts, auth_type: str) -> dict[str
...
+    ``auth_label`` is the user-facing label from the auth-method prompt (e.g. ``PAT``).
+    Returns a partial dict to merge into the nested ``connection`` block. Shared
+    connection fields (account, user, warehouse, …) are prompted separately.
+    """
+    if auth_label == Pat.label:
+        logger.info("Authentication uses a Programmatic Access Token (PAT). See Snowflake's docs:")
+        logger.info(
+            "  https://docs.snowflake.com/en/user-guide/programmatic-access-tokens"
+            "#generating-a-programmatic-access-token"
+        )
```
asnare: Nit, should be a single log entry:
```suggestion
        logger.info("Authentication uses a Programmatic Access Token (PAT). See Snowflake's docs:   https://docs.snowflake.com/en/user-guide/programmatic-access-tokens#generating-a-programmatic-access-token")
```
(And `make fmt`.)

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -52,6 +59,55 @@ def _prompt_mssql_auth_credentials(prompts: Prompts, auth_type: str) -> dict[str
...
+        # EnvGetter resolves at runtime, not the token itself, so prompt accordingly.
+        # Stored under `pat` (not `password`) to flag this is a rotating
+        # Programmatic Access Token, not a SQL password.
+        if secret_vault_type == "env":
+            return {"pat": prompts.question("Enter the environment variable name holding the PAT")}
+        return {"pat": prompts.password("Enter Programmatic Access Token (PAT)")}
+
+    if auth_label == KeyPair.label:
+        logger.info("Authentication uses key-pair (JWT). See Snowflake's docs:")
+        logger.info("  https://docs.snowflake.com/en/user-guide/key-pair-auth")
```
asnare: Nits, no JWT and single log-entry:
```suggestion
        logger.info("Authentication uses key-pair. See Snowflake's docs: https://docs.snowflake.com/en/user-guide/key-pair-auth")
```

tests/unit/connections/test_snowflake_utils.py
asnare, line 1: Please add type annotations to the `test_*` functions so that mypy can lint the bodies:
 * Types for the fixtures;
 * `-> None` for the return values.

asnare (changes requested):
Very close, only a few minor/mechanical things now, in particular the type annotations.

[I should add that I also tested this manually with a password-protected ECDSA key.]

asnare (approved):

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -52,6 +59,57 @@ def _prompt_mssql_auth_credentials(prompts: Prompts, auth_type: str) -> dict[str
...
+    if auth_label == Pat.label:
+        logger.info(
+            "Authentication uses a Programmatic Access Token (PAT). See Snowflake's docs: "
+            "https://docs.snowflake.com/en/user-guide/programmatic-access-tokens"
+            "#generating-a-programmatic-access-token"
+        )
+        # In env mode the stored value is the name of an environment variable that
+        # EnvGetter resolves at runtime, not the token itself, so prompt accordingly.
+        # Stored under `pat` (not `password`) to flag this is a rotating
+        # Programmatic Access Token, not a SQL password.
```
```suggestion
```

please avoid code comments when the code does not need it

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -52,6 +59,57 @@ def _prompt_mssql_auth_credentials(prompts: Prompts, auth_type: str) -> dict[str
     return {}
 
 
+def _prompt_snowflake_auth_credentials(prompts: Prompts, auth_label: str, secret_vault_type: str) -> dict[str, str]:
+    """Prompt for Snowflake auth-specific fields for the chosen strategy.
+
+    ``auth_label`` is the user-facing label from the auth-method prompt (e.g. ``PAT``).
+    Returns a partial dict to merge into the nested ``connection`` block. Shared
+    connection fields (account, user, warehouse, …) are prompted separately.
```
```suggestion
```

same, one liner is enough

src/databricks/labs/lakebridge/assessments/configure_assessment.py
```diff
@@ -372,22 +430,15 @@ def _configure_credentials(self) -> None:
...
-        )
-
-        # In env mode the stored value is the name of an environment variable that
-        # EnvGetter resolves at runtime, not the token itself, so prompt accordingly.
-        if secret_vault_type == "env":
-            pat = self.prompts.question("Enter the environment variable name holding the PAT")
-        else:
-            pat = self.prompts.password("Enter Programmatic Access Token (PAT)")
+        auth_choices = {cls.label: cls.auth_type for cls in SNOWFLAKE_AUTH_CHOICES}
+        assert len(auth_choices) == len(SNOWFLAKE_AUTH_CHOICES), "Label/auth-type collision."
```
```suggestion
```
lets avoid asserts in production code and with the current choices this can never fire anyway plus unit test `test_auth_choices_expose_snake_case_types_and_labels` already checks this

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -99,13 +100,13 @@ def health_check(self) -> bool:
...
-        # The SDK types JSON values loosely, so narrow to a dict for the accesses below.
-        connection_config = self.config["connection"]
-        if not isinstance(connection_config, dict):
-            raise ConnectionError("Snowflake credentials must be nested under a 'connection' block")
+    @staticmethod
+    def build_engine_args(connection_config: dict[str, Any]) -> tuple[URL, dict[str, Any]]:
+        """Build the SQLAlchemy URL and connect_args for a Snowflake connection config.
 
+        Separated from ``_connect`` so URL / connect-arg construction can be unit-tested
+        without opening a real engine.
```
m-abulazm:
```suggestion
```

src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -117,19 +118,30 @@ def _connect(self) -> Engine:
...
-            password=password,
             host=account,
             database=f"{database}/{schema}",
+            username=user,
+            password=resolved.password,
             query={"warehouse": warehouse, "role": role},
         )
-        return create_engine(snowflake_url)
+        connect_args = {"private_key": resolved.private_key} if resolved.private_key else {}
+        return snowflake_url, connect_args
```
m-abulazm:
```suggestion
        return snowflake_url, resolved.connect_args
```

returning connect_args directly from the strategy is more extendable. `ResolvedSnowflakeCredentials` could carry password + connect_args instead of a private_key field, so future auth methods don't need connector changes.

src/databricks/labs/lakebridge/connections/snowflake_auth.py
```diff
@@ -0,0 +1,96 @@
...
+    @classmethod
+    @abstractmethod
+    def resolve_credentials(cls, config: JsonObject) -> ResolvedSnowflakeCredentials: ...
+
+
+class Pat(SnowflakeAuth):
+    """Programmatic Access Token — stored under `pat` and used as the URL password."""
+
+    auth_type = "pat"
+    label = "PAT"
```
m-abulazm, line 51: I see some earlier threads about this but it is not clear why we need both

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -16,6 +20,52 @@ def is_valid_snowflake_account(account_identifier: str) -> bool:
...
+    Note: ``cryptography`` treats ``b""`` like a missing password when decrypting, so
+    keys encrypted with an empty passphrase cannot be loaded successfully.
+    """
+    try:
+        key_bytes = key_path.read_bytes()
+    except OSError as e:
+        raise ConnectionError(f"Unable to read Snowflake private key at {key_path}: {e}") from e
+
+    # None = unencrypted key; "" = encrypted with an empty passphrase (distinct from None).
+    password = passphrase.encode() if passphrase is not None else None
```
m-abulazm:
```suggestion
    password = passphrase.encode() if passphrase else None
```
we use Prompts that never accept empty strings so the case of "" should not happen and if it happens we need to treat it similar to None.

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -16,6 +20,52 @@ def is_valid_snowflake_account(account_identifier: str) -> bool:
...
+        raise ConnectionError(
+            f"Unable to decrypt Snowflake private key at {key_path}: {e} "
+            "(provide the correct passphrase, or omit it for an unencrypted key)"
+        ) from e
+    except ValueError as e:
+        message = str(e).lower()
+        if "password" in message or "decrypt" in message:
+            raise ConnectionError(
+                f"Unable to decrypt Snowflake private key at {key_path}: {e} " "(check the passphrase)"
+            ) from e
```
m-abulazm:
```suggestion
```
lets keep it simple, one generic ValueError message like "unable to load private key at X: e (check the key file and passphrase)"

tests/unit/assessment/test_snowflake_configurator.py
```diff
@@ -61,7 +88,33 @@ def test_local_vault_stores_pat_verbatim(tmp_path):
...
+
+def test_local_vault_stores_key_pair_with_passphrase(tmp_path):
+    creds = _run(_key_pair_prompts(encrypted=True, passphrase="secret-pass"), tmp_path)
+    connection = creds["snowflake"]["connection"]
+    assert connection["auth_type"] == "key_pair"
+    assert connection["private_key_path"] == "/path/to/rsa_key.p8"
+    assert connection["private_key_passphrase"] == "secret-pass"
+
+
+def test_local_vault_stores_key_pair_with_empty_passphrase(tmp_path):
```
m-abulazm: is this really a valid scenario? given `test_load_private_key_empty_passphrase_cannot_decrypt` this should not happen

src/databricks/labs/lakebridge/connections/snowflake_utils.py
```diff
@@ -16,6 +20,52 @@ def is_valid_snowflake_account(account_identifier: str) -> bool:
...
+def load_snowflake_private_key(key_path: Path, passphrase: str | None = None) -> bytes:
+    """Load a PEM PKCS8 private key and return DER bytes for SQLAlchemy connect_args.
+
+    Snowflake SQLAlchemy expects the private key as unencrypted DER PKCS8 bytes via
+    ``connect_args={"private_key": ...}``. Encrypted ``.p8`` files need ``passphrase``.
+
+    ``passphrase=None`` means the key is expected to be unencrypted. An empty string
+    (``""``) is treated as an empty passphrase (``b""``), distinct from ``None``.
+    Note: ``cryptography`` treats ``b""`` like a missing password when decrypting, so
+    keys encrypted with an empty passphrase cannot be loaded successfully.
```
m-abulazm:
```suggestion
```

m-abulazm (changes requested):

src/databricks/labs/lakebridge/connections/snowflake_auth.py
dgomez04: `auth_type` is what we store and dispatch one, `label` is what we show in prompts and documentation. Andrew asked for in an earlier review.

dgomez04: They serve different audiences:

- `auth_type` (`pat`, `key_pair`) is the stored identity in credentials YAML and the dispatch key at connect time.
- `label` (`PAT`, `Key-Pair`) is only what we show in the configurator prompt and docs.

Keeping them separate means we can reword the UI without breaking existing credential files (and avoids tying the stored value to the class name).

m-abulazm: makes sense

m-abulazm (approved):
LGTM

PR #2623, opened 2026-08-13
analyzer: fix validation of the source path

asnare:
### What does this PR do?

This PR updates the checks we do for the source directory when analysing, it only needs to be readable instead of writable.

At one point the output was written into the source directory, which meant it needed to be writable. However since the analyzer was updated to allow the output location to be specified this has not been necessary.

### Functionality

- modified existing command: `databricks labs lakebridge analyze`

### Tests

- manually tested
- existing unit tests (needed updating)
- existing integration tests


gueniai (approved):
LGTM

sundarshankar89 (approved):
LGTM

PR #2624, opened 2026-08-13
Profilers: eliminate unnecessary `ORDER BY` clauses from queries run on source systems

asnare:
Remove unnecessary ORDER BY queries from SQL that we run on source systems when profiling.

These ORDER BY clauses generate extra work on the sources and are not necessary. Consumers of the DuckDB data that rely on order need to specify that explicitly. (In general the relational model does not guarantee order is preserved across insertions and updates: queries need to specify this if they require it.)

## Changes

This PR removes `ORDER BY` clauses from SQL that we run against source systems when profiling: these are just additional work on the sources and are not necessary. Downstream consumers of the data that rely on the order need to specify that when querying the data. (In general the relation model does not guarantee order is preserved across insertion/update: such queries already need to specify this for correctness.) 

Only outer `ORDER BY` clauses have been removed that weren't part of a top-N construct. (In addition, an inner `ORDER BY` has been removed a CTE where the ordering was discarded anyway.) 

### Functionality

- modified existing command: `databricks labs lakebridge execute-database-profiler`


gueniai (approved):
LGTM

sundarshankar89 (approved):
LGTM

PR #2625, opened 2026-08-13
Reorder Teradata profiler config prompts to match other connectors

dey-abhishek:
The Teradata configurator prompted for the password before host, port, user and database, because the password was read into a local variable ahead of the credential dict literal. This made it look like host/user were never asked when the flow paused at the password prompt.

Prompt for the connection fields first (host, port, database, user, password), matching the Oracle and Redshift configurators. The password is still read last since the env vault stores an env-var name.

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


dgomez04 (approved):
LGTM

gueniai (approved):

PR #2626, opened 2026-08-14
chore: lint/fix imports with ruff

asnare:
## Changes

This is a preliminary change to the repo to ensure that ruff's import-checking linters pass. These checks are enabled in the next version: turning them on here means that the changes needed for ruff can be implemented piecemeal. These supersede the equivalent checks from pylint, which we want to eliminate.

Only imports are affected by this PR; there are no behavioural changes.

### Tests

- existing unit tests
- existing integration tests


gueniai (approved):
LGTM

sundarshankar89 (approved):
This is great finally we will be out of pylint, Thanks for making this happen.

PR #2644, opened 2026-08-25
Configurable DBQL lookback and row cap for the Teradata profiler

dey-abhishek:
### What does this PR do?

Makes the Teradata profiler's look-back windows and the DBQL row cap configurable via execution-time bind parameters, with defaults that preserve (or lightly update) the previous behavior. Previously each of these was hardcoded in SQL:

- **`lookback_days`** (default `30`) — core DBQL extract: `CollectTimeStamp >= CURRENT_DATE - :lookback_days` (raised from the previous hardcoded 7-day window)
- **`pdcr_lookback_days`** (default `180`) — PDCR history extracts (`PDCRINFO.DBQLogTbl_Hst`): `LogDate >= date - :pdcr_lookback_days`
- **`sys_usage_lookback_days`** (default `60`) — system usage aggregation (`DBC.ResUsageSpma`): `TheDate >= date - :sys_usage_lookback_days`
- **`sys_nodes_lookback_days`** (default `30`) — node hardware inventory (`DBC.ResUsageSpma`): `TheDate >= date - :sys_nodes_lookback_days`
- **`max_rows`** (default `100000`) — core DBQL row cap: `QUALIFY ROW_NUMBER() OVER (ORDER BY TotalFirstRespTime DESC) <= :max_rows`

The windows are intentionally independent: `DBC.DBQLogTbl` is the short-lived live table, while `PDCRINFO` retains far more history, so collapsing them onto one value would either shrink the PDCR extract or overrun the live table's retention. All are prompted for by `configure-database-profiler` and stored under the source's `profiler` section in the user-owned credentials file; the packaged defaults keep existing setups working.

> **Scope note:** Teradata result-set streaming was split out of this PR into #2649 so this PR is parameterization-only. On this branch Teradata buffers via `fetch()`, exactly as `main` does today; streaming lands separately.

- **Execution-time bind parameters (not string interpolation).** `DatabaseConnector.fetch()`/`stream()` take a `parameters` mapping; the SQLAlchemy-based connectors bind `:name` placeholders (translated to the driver's paramstyle), so values are never spliced into SQL text. `TOP` can't take a bind parameter, so the row cap is expressed with `QUALIFY ROW_NUMBER()`. Parameters are bound globally to every step, so the core DBQL extract, both PDCR steps, and the two `ResUsageSpma` queries pick up their windows automatically.
- **Resolution precedence:** pipeline defaults < credentials file < explicit overrides. Only parameters the pipeline declares are read from credentials, and the lookup is **scoped to the profiled source** so a shared credentials file with multiple sources can't leak one source's settings into another.
- Configured values are validated `>= 1` so `0`/negative can't silently produce an empty (or future-dated) extract.

- The raw-DBAPI connectors (mssql/redshift) reject bind parameters with `NotImplementedError` until binding is implemented for them; no current pipeline passes them any.
- A single Teradata run extracts the core DBQL workload and additionally attempts the PDCR extraction; the PDCR steps are `optional: true`, so when `PDCRINFO` is absent they are marked `ABSENT` and tolerated and the run still succeeds.

### Functionality

- [x] documented the configurable profiler parameters (Teradata profiler docs)
- [ ] added new CLI command

### Tests

- [x] added/updated unit tests — bind-parameter resolution + source scoping, and the Teradata configurator prompts for all five profiler settings
- [x] `make fmt` clean (black/ruff/mypy, pylint 10.00/10) and `make test` passing
- [x] validated end-to-end against a live Teradata box: all steps complete against the parameterized SQL, with PDCR steps skipped automatically where `PDCRINFO` is absent


asnare (changes requested):
A few immediate topics that we need to address before getting into the finer details.

1. Before we implement or support this, the Teradata profiler needs to be updated to support the streaming interface. Without this we end up with unbounded memory use and allowing users to configure the look-back will only make that more problematic than it already is. (This could be a loose thread that unravels somewhat… we will need to ensure that the streamed batches all have the same schema, and verify that every query has paired DDL in place.)

2. Sorry, but we can't do variable substitution in an adhoc manner using a custom implementation: it's unsafe and will invariably be duplicated down the track into more places. Implementations like this are always brittle: only database-specific drivers (from the vendor) know how to do this safely and properly. (Custom interpolation also prevents us from improving our quality checks on SQL down the track.) The only safe way to allow for interpolation like this is to use prepared statements with execution-time binding. (In some cases this will lead to changes to the queries because variables can't be bound into all locations of the query, but that's a compromise we must accept.)

PR #2654, opened 2026-09-03
Simplify Teradata `pdcr` workload extraction

dgomez04:
### Summary

Modifies `td_pdcr_bi_concurrency_extract.sql‎` workload extract from thirteen dimensions to two (`QueryType` and `WorkloadBucket`) by removing the `PDCRINFO.UserInfo` join and organizational/user/hourly dimensions.

Adds `td_pdcr_bi_concurrency_extract` to track total BI queries and average BI query starts per minute over 180 days. 

### Motivation

The migration model only requires system-level workload proportions.

Grouping across thirteen high-cardinality dimensions (user, department, application, hour) generated massive aggregation/join spool on large `pdcr` repositories, causing spool on production runs. 

### Expected Performance

- Decreasing group keys from 13 to 2 low-cardinality dimensions eliminates join spool and shrinks output cardinality from millions of rows to roughly the number of query types.
- The 180-day source scan remains (plus a scalar, no-GROUP-BY scan for the BI concurrency query). The primary gain is a drastic reduction in peak hash/sort spool rather than guaranteed reductions in total source IO or elapsed time.

### Output Semantics

`td_pdcr_info_agg_extract` maps:

- `ETL` -> `ETL`
- `BI/QUERY` -> `BI`
- `DDL`, `DML`, `SP`, and `OTHER` -> `INTERACTIVE`

`td_pdcr_bi_concurrency_extract.AvgBIQueriesPerMinute` measures BI query starts per minute across the fixed 180-day observation window. It does not claim to measure simultaneous in-flight query execution, which would require interval-overlap analysis.

### Validation

- `201 passed` — complete `tests/unit/assessment` suite
- `git diff --check` passed


sundarshankar89 (approved):
LGTM

gueniai (approved):

PR #2657, opened 2026-09-04
Fix Synapse profiler DuckDB `DECIMAL` overflow on dedicated requests

dgomez04:
## Summary
Prevent DuckDB narrow-`DECIMAL` overflow across **all** profiler sources. When a source query returns Python `Decimal` (SQL `DECIMAL`/`NUMERIC`/`NUMBER` columns and un-cast decimal arithmetic), DuckDB infers a narrow `DECIMAL(p,s)` from a strided ~1000-row sample and overflows on INSERT of a larger value — even when the destination column is `DOUBLE`. Fix: cast to a float type in the query (the DDL type does not protect, because the overflow happens while reading the pandas `Decimal` column).

- **Synapse** (`common/queries.py`): `RESOURCE_ALLOCATION_PERCENTAGE` (the originally-reported path), `table_storage_info` size-MB, and `query_performance_stats` `/ 1000000.0` timings → `FLOAT`.
- **Oracle**: `config_storage` (gb/freegb/maxgb), `perf_heatmap`, `perf_sqltext` → `BINARY_DOUBLE`.
- **Teradata**: `td_sys_usage_agg`, `td_sys_disk_utilization`, `td_dbql_core_info_extract`, `td_pdcr_info_agg_extract`, `td_pdcr_sp_exe_info_agg_extract` aggregates → `FLOAT`.
- Audited, no change needed: Redshift, Snowflake, MSSQL (already cast); BigQuery (`/ power(1024,4)` returns FLOAT64); Synapse `dedicated_storage_info` (integer arithmetic → BIGINT).
- `test_profiler_float_casts.py` guards each cast against regression.

#2578 did not cover these paths.

## Test plan
- [x] `uv run pytest tests/unit/assessment/test_profiler_float_casts.py -q`

### DIFF
--- a/src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
+++ b/src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
@@ -238,7 +238,7 @@ class SynapseQueries:
                 IMPORTANCE,
                 [LABEL],
                 REQUEST_ID,
-                RESOURCE_ALLOCATION_PERCENTAGE,
+                CAST(RESOURCE_ALLOCATION_PERCENTAGE AS FLOAT) AS RESOURCE_ALLOCATION_PERCENTAGE,
                 RESOURCE_CLASS,
                 RESULT_CACHE_HIT,
                 SESSION_ID,
--- a/tests/unit/assessment/test_profiler_float_casts.py
+++ b/tests/unit/assessment/test_profiler_float_casts.py
@@ -14,6 +14,9 @@ _REPO_ROOT = Path(__file__).resolve().parents[3]
 _ASSESSMENTS = _REPO_ROOT / "src/databricks/labs/lakebridge/resources/assessments"
 
 REQUIRED_CASTS: dict[str, tuple[str, ...]] = {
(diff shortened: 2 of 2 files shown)


src/databricks/labs/lakebridge/resources/assessments/synapse/common/queries.py
gueniai, line 385: Should we apply the same fix to this one?

asnare (commented):
Has this change been run against an actual system?

Looking a bit closer, this change is working around a fundamental design issue. Bear in mind that the chain is: SQL -> Pandas -> DuckDB(insert) -> DuckDB(table)

This means that the root cause is that first SQL->Pandas conversion drops the upstream type-information, and DuckDB therefore (when scanning the values to infer the type) guesses wrong. (It has to infer the type before it knows the target type of the table it will land in, which is unfortunate.)

This doesn't seem easy to fix with the current architecture, so a workaround is all we can do although as @gueniai noted it would be best to whack them all. (They're data sensitive, so this is a survey task rather than run/iterate.)

PR #2658, opened 2026-09-09
Fix MSSQL passwords with special characters

surajs-db:
### What does this PR do?
Fixes the Lakebridge MSSQL connection test for passwords containing ODBC-special characters, including ?.

When configuring an on-premises MSSQL source, a password containing ? was inserted directly into the ODBC connection string. The mssql_python parser interpreted the character as connection-string syntax and failed with:

ConnectionStringParseError: Incomplete specification: keyword '?' has no value (missing '=')

This prevented configure-database-profiler from completing the MSSQL connection test.

Formats MSSQL password values with odbc_value before constructing the connection string, so special characters are treated as password data rather than connection-string delimiters.

### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] ... +add your own

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


src/databricks/labs/lakebridge/connections/database_manager.py
```diff
@@ -166,35 +166,39 @@ def stream(self, query: str) -> Iterator[pa.Table]:
 
 
 class MSSQLConnector(DatabaseConnector):
+    @staticmethod
+    def odbc_value(value: str) -> str:
```
m-abulazm: I would rename to something more descriptive e.g. `escape`

tests/unit/connections/test_mssql_auth.py
```diff
@@ -182,7 +182,32 @@ def fake_connect(connection_string, **kwargs):
 
     assert "Authentication=" not in captured["connection_string"]
     assert "UID=alice" in captured["connection_string"]
-    assert "PWD=secret" in captured["connection_string"]
+    assert "PWD={secret}" in captured["connection_string"]
+
+
+def test_mssql_connector_wraps_special_password_in_odbc_braces() -> None:
```
m-abulazm, line 193: since it is a public method, I would test directly
```
assert MSSQLConnector.escape("pa}ss") == "{pa}}ss}"                                                                                                                                                    
assert MSSQLConnector.escape("pa;ss") == "{pa;ss}"
```

m-abulazm (changes requested):

tests/unit/connections/test_mssql_auth.py
surajs-db: Done

